PR #652, opened 2014-02-25
Allowing users to call with single click (skips the dialer)

inssein:
I just tried out TextSecure for the first time today and was annoyed that I had a 2 step process in order to place a call from the conversation.

I am not sure if you guys intentionally didn't include this functionality (maybe you guys didn't want to add an extra permission). I did a quick search of the issues and didn't find anything, so sending a pull request to integrate this functionality.


rhodey:

rhodey:
we added a bunch of integration tests and now run them against PRs automatically. this PR does not contain the build type used for the integration tests so it fails and will keep failing until rebased with master. once it has been rebased I will make sure the tests get re-run so the status icon can be updated.

looks like this is getting held off until we are able to do an update with permission changes, I totally understand if you wanna hold off the rebase work until we decide we're ready for a permission change. thanks for putting all this work in @inssein :)

McLoo:
Permission is in now.
ACTION_DIAL --> ACTION_CALL works as expected. :+1:

inssein:
Thanks for the heads up @McLoo . Sorry @rhodey I totally missed that earlier message, but I will rebase this weekend.

inssein:
@moxie0 should be good to merge now. Sorry my git foo wasn't strong when I started (still really isn't), and this wasn't on a branch. Can I still squash these commits into one easily?

ChaoticMind:
@inssein try `git rebase -i <hash>`

moxie0:
For now, I think I actually like that it takes us to the dialer.  Thanks though!

Issue #859, opened 2014-02-28
[Feature Request] Auto-sync contacts and move UI item


SanderGit:
- Please add a setting to the preferences to refresh the Push Directory every x hours
- The current Settings tab contains both settings and actions.
  Please consider moving the current menu item 'Refresh Push Directory' from Settings to somewhere else, e.g. the 3-dot menu or the action bar at the left side of the screen (not sure why both exist anyway). That will make it easier to find.

Referenced from https://github.com/WhisperSystems/TextSecure/issues/610

PR #890, opened 2014-02-28
Use a SyncAdapter to synchronize the directory "the Android Way"

tinloaf:
This prepares us to do synchronizations on regular intervals (adding this is a one-liner) and also allows the user to enable and disable synchronization via the android settings. This would solve #859. Also, this is the way "all apps do synchronization in android". Or so they say...

Synchronizing the directory from the app could be done by calling this SyncAdapter, but I did not change that yet.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -34,6 +34,10 @@
     <uses-permission android:name="android.permission.READ_CALL_LOG" />
     <uses-permission android:name="android.permission.GET_ACCOUNTS" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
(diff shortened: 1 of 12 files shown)


rhodey:

rhodey:
we added a bunch of integration tests and now run them against PRs automatically. this PR does not contain the build type used for the integration tests so it fails and will keep failing until rebased with master. once it has been rebased I will make sure the tests get re-run so the status icon can be updated.

looks like this is getting held off until we are able to do an update with permission changes, I totally understand if you wanna hold off the rebase work until we decide we're ready for a permission change. thanks for putting all this work in @tinloaf :)

McLoo:
Implemented since 2.23.0
Silently closing?!

Issue #299, opened 2013-08-08
Implement "Quick Responses"


MatejKovacic:
Android has an useful option for rejecting call with SMS. When you receive a call, you can reject it with pre-prepared or custom SMS.

However, this SMS is not sent through TextSecure but through default Android application for sending SMS's.

Can this be changed?

fajabird:
Hi,
seems this issue still persists. What about implementing a small toast-message after someone tries to reject a call with a message that this funtionality is not supported yet because I already had some people asking me (the guy who recommended TS) what happened to their reject-messages. Or is someone already working this out?

fajabird:
Ohh - I just took a look at the code and a "not supported yes" toast-message seems to be already implemented:

Toast.makeText(this,
getString(R.string.QuickResponseService_sorry_quick_response_is_not_yet_supported_by_textsecure),
Toast.LENGTH_LONG).show();

BUT it does not show up for me? Someone else getting that toast?

McLoo:
@fajabird shows up for me. Rejecting call with text ends up in rejecting with 'not supported' toast following :)

merkste:
I'd like to see the quick reply feature working, too. Any news for us on this one?

asafh:
@merkste, unfortunately I don't see myself tackling this in the immediate future - since I don't want to uninstall TextSecure from my daily phone. When I get a secondary phone (not too long I hope, but I can't promise nothing) I'll look into this again.

JavaJens:
@asafh You could:
a) Install two TextSecure versions in parallel and test with the browser extension (very buggy)
b) Specify what needs to be done to continue your development efforts and maybe point others in the right direction

asafh:
@JavaJens what I've done and moxie's input on what needs to be done is described pretty well over PR #962. Basically, what needs to be done is trying to get the key from the key caching service directly, if it is available go ahead with sending the message, otherwise show a toast with the "TextSecure is currently locked. Quick Response messages are disabled." or similar.  

Since the code base has already changed quite a bit, I would the code in that PR's commit as a reference. It will be easier to simply do the similar work on a new branch from main.  

If you decide to tackle this, and have questions you want to consult me about - feel free to send me a message here or at ["hello",String.fromCharCode(31\*2+2)/\*regex scrapper\*/,"ous",".im"].join("")

PR #962, opened 2014-03-03
Implementing Quickreply

asafh:
I've implemented the feature, right now the way it works if TextSecure is locked is it will queue the sending of the message until it is unlocked.
However, if the service is killed before that the message will be lost. We can avoid this using START_REDELIVER_INTENT but this gives some slight chance of resending messages if the service is killed between the sending of the message and the stopSelf call. I opted to avoid this for now.

I tried adhering to your code guidelines but this is my first look into this project, and frankly I'm not a major Android developer though I think I did alright.

I avoided duplicating the "run tasks when you get the master secret" logic in SendReceiveService by extracting it a parent class "MasterSecretTaskService".

Let me know what you think about the way I implemented this.

Issue #299

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -398,7 +398,7 @@
     <string name="SmsReceiver_currently_unable_to_send_your_sms_message">Currently unable to send your SMS message. It will be sent once service becomes available.</string>
 
     <!-- QuickResponseService -->
-    <string name="QuickResponseService_sorry_quick_response_is_not_yet_supported_by_textsecure">Sorry, Quick Response is not yet supported by TextSecure!</string>
+    <string name="QuickResponseService_the_quick_response_will_be_sent_when_textsecure_will_be_unlocked">The Quick Response will be sent when TextSecure will be unlocked!</string>
 
     <!-- auto_initiate_activity -->
     <string name="auto_initiate_activity__you_have_received_a_message_from_someone_who_supports_textsecure_encrypted_sessions_would_you_like_to_initiate_a_secure_session">You have received a message from someone who supports TextSecure encrypted sessions.  Would you like to initiate a secure session?</string>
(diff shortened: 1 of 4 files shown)


McLoo:
@asafh caught up with life again?

asafh:
@McLoo I've went on several vacations since, so not quite. But I'm taking a time out from that now, so you had good timing with your reminder :)

invites18:
This is a really important feature for me because the people closest to me will call me when I can't talk (because I'm in a meeting, having dinner, etc.).  If I could make some simple gestures (taps, swipes) to let them know that I can't talk and will get back  to them, then I can avoid either being hard to get ahold of (and taking a toll on the relationship) or having to stop and create the message the long way.

While I'm not a developer, I appreciate that figuring out the encryption here is hard.  I would be fine with my quickreply messages being unencrypted for three reasons: A) my impression is that if someone can intercept my text messages, they can also intercept my phone calls, so they would know that I did call the person back (even if they don't know what was said), or, if they see that I missed a call and immediately sent a text message, I have to imagine that the bad guys will figure out that the most likely thing that I wrote is "I'll call you back."  B) That I will call someone back is probably the least sensitive message I write.  C) The vast majority of my text messages are to people who don't have textsecure/signal.  I'm trying to win people over, and while I have a few converts, this particular missing feature has been a huge problem.

As it stands, I'm close to dropping textsecure to revert to software that allows a quickreply because my wife wants me to let her know when I can't talk.  (But I really don't want to do this.)

I'd really appreciate it if you could support a quick reply (and allowing me to set a few custom messages would be fantastic).  I'm fine with these messages being unencrypted, or my leaving textsecure unlocked, or even handled by a separate program.  Perhaps quickreplies could be turned on as a setting so users can be told that these are unsecured.


Thank you,
David

invites18:
I just downloaded the latest update and tried out the new feature. It's great - thank you so much! 


David (the same David as the last post from June)

PR #1334, opened 2014-04-10
Disable preview windows to avoid showing the wrong theme

patheticpat:
This fixes the flash of the light theme when you change the theme or
language settings, as @McLoo described in #1322.


DevenSingh27:
but after adding <item name="android:windowDisablePreview">true</item> in theme All the dialog enter animation not work... Is there any workaround for that ???

Issue #103, opened 2012-12-12
Draft of message saved


dolanor:
While typing an SMS, sometimes, you get notifications and switch applications to act with the notification. And after returning to application, sometimes you have come back to home to check agenda or whatever, and the draft of the message has been lost. 
Since every SMS composer on android seems to handle that, it should be great that TextSecure handles it too.

PR #1510, opened 2014-05-18
Improve drafts

AsamK:
The first commit is a small fix to delete a draft when the conversation is deleted. Until now the draft wasn't deleted and could reappear in a new thread. The bug can easily be reproduced by creating a new group, creating a draft, delete the thread and create a new group. The new group now contains the draft from the old group.

The second commit adds the possibility to save drafts for recipients you haven't written to before, i.e. no thread exists. This also fixes the behavior described by @landry314 in #103. It is implemented by creating a new thread (if necessary) when a draft is saved. When the draft is deleted and no other messages exist in the thread, the thread is deleted in ThreadDatabase.update().

### DIFF
--- a/src/org/thoughtcrime/securesms/database/ThreadDatabase.java
+++ b/src/org/thoughtcrime/securesms/database/ThreadDatabase.java
@@ -290,6 +290,7 @@ public class ThreadDatabase extends Database {
   public void deleteConversation(long threadId) {
     DatabaseFactory.getSmsDatabase(context).deleteThread(threadId);
     DatabaseFactory.getMmsDatabase(context).deleteThread(threadId);
+    DatabaseFactory.getDraftDatabase(context).clearDrafts(threadId);
     deleteThread(threadId);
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/database/ThreadDatabase.java
```diff
@@ -398,7 +403,7 @@ public void update(long threadId) {
         else                 timestamp = record.getDateReceived();
 
         updateThread(threadId, count, record.getBody().getBody(), timestamp, record.getType());
-      } else {
+      } else if (!draftExists) {
         deleteThread(threadId);
       }
```
moxie0: Is all this draftExists stuff necessary now? We explicitly set the snippet from the ConversationActivity and don't call update().

AsamK: you're right, it's not necessary, I removed it.

PR #1641, opened 2014-06-24
Replace telephone numbers in a new group chat with contact names

agrajaghh:
I replaced the telephone numers in a new group chat with the corresponding contact names or in case of the own number with "Me". I also made some old strings translatable.

I copied the check for the own number from #1505 and put it in isOwnNumber(). Perhaps we could put this method in a place where it could be used from both PR's?

**before:**
before
**after:**
after

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -176,7 +176,7 @@
     <string name="GroupUtil_group_updated">Group updated.</string>
     <string name="GroupUtil_joined_the_group">joined the group.</string>
     <string name="GroupUtil_title_is_now">Title is now</string>
-    <string name="GroupUtil_me">Me</string>
+    <string name="GroupUtil_you">You</string>
 
     <!-- ImportExportActivity -->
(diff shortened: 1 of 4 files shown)


moxie0:
Hey @agrajaghh I'm going to close this since it's still in progress, but please reopen when you're ready.

agrajaghh:
omg, totally forgot this PR, will have a look soon...

agrajaghh:
hey @moxie0

> Thanks @agrajaghh. Looking at this, it seems that you're doing the contact lookup on the UI thread. What needs to happen is an asynchronous lookup, where the phone number is displayed until the lookup is complete. Similar to how things with Recipients currently work.

I just tried to figure out how the recipients are currently working async but I didn't find the place in the code. In the `ConversationItem` they are just set like this (`ConversationListItem` is working similar):
````java
this.groupStatusText.setText(messageRecord.getIndividualRecipient().toShortString());
````
Could you please point me to the right place where they are retrieved async?

PR #1695, opened 2014-07-13
fixes #1407, sharing contact from the Contacts app

### LINKED ISSUE #1407: textsecure not an option when "sharing" an address book contact
TextSecure is not an option when attempting to share a contact from the address book, in Android 4.4.2 when i select "messaging" it tells me "messaging" is not the default sms app. This would be a nice feature to have added to the app.

payload:
- closes #1407
- TextSecure is listed when you share a contact via Contacts app
- lets you choose a phone number, if there are multiple
- copies just the phone number into a new message
- new message is empty, if there is no phone number

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -19,6 +19,7 @@ package org.thoughtcrime.securesms;
 import android.app.AlertDialog;
 import android.content.ActivityNotFoundException;
 import android.content.BroadcastReceiver;
+import android.content.ContentResolver;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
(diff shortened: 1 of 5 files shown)


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -895,7 +918,10 @@ private void addAttachmentAudio(Uri audioUri) {
   private void addContactInfo(Uri contactUri) {
     ContactAccessor contactDataList = ContactAccessor.getInstance();
     ContactData contactData = contactDataList.getContactData(this, contactUri);
```
McLoo: @moxie0 @payload 
If we change `Contacts._ID` to `OpenableColumns.DISPLAY_NAME` in `contactDataList.getContactData()` everything works fine. 
Contact name will be "SomefileName.vcf" then, but we don't use it anyway right now.
Alternative would be to pass `null` as name instead of `Contacts._ID`

works on CM12 and KK

Mihahn:
@McLoo did you say it works on CM12 and KK with your changes? Could you get this pull request working then?

PR #1929, opened 2014-09-17
MmsCommunication refactor

mcginty:
Introduce MmsConnection class to handle incoming and outgoing network transactions, behaving as a simple abstraction layer over URLConnection objects.

Also another small change for general safety: in retry logic, check protocol before following links and prohibit https->http redirects.

Cuts down on redundant code and reduces a bit of confusion, keeping a majority of the static MMS-related methods to fewer classes.


McLoo:
@mcginty Secure MMS works for me now. Was there a change I missed or is Vodafone Germany behaving well now?
Anyway: Yeay!

mcginty:
Probably a Vodafone change (perhaps they don't require SMIL anymore?), but either way I'm glad to hear it's working now!

PR #1960, opened 2014-09-30
Initial Websockets Implementation

JavaJens:
This PR is to give the maintainers an overview of the status of the websockets implementation.
Most likely things will change (like build variants), but this should be useful as common grounds.


ale5000-git:
@JavaJens: Any update?
There will be another pull request for this?

PR #2279, opened 2014-12-30
rounded ImageView instead of Bitmap, crop-select

mcginty:
// FREEBIE


moxie0:
lgtm, but needs rebase

PR #2284, opened 2015-01-01
Larger max text for PGP key transfer

jamcar23:
Increased the maximum text length so larger text can be transferred (like a PGP key.)


mcginty:
Seems like a bad idea to set to 10,000 when the max GCM push payload size is 4kb. In the long run we'd be better off allowing for text attachments for these kinds of things since convo bubbles aren't optimal for text data transfers like this.

In the short term, we should check performance of conversation bubbles with text > 1000 chars before increasing it further. At first thought, emojifying the text is an O(n) operation, so big blurbs could be a significant tax...

moxie0:
Agreed.

PR #2286, opened 2015-01-01
use PartDatabase to look up thumbnails async

mcginty:
Prevent excessive resource contention from generating two thumbnails for the same piece of media when sending/receiving into an open conversation.


src/org/thoughtcrime/securesms/database/PartDatabase.java
```diff
@@ -85,6 +89,8 @@
     "CREATE INDEX IF NOT EXISTS pending_push_index ON " + TABLE_NAME + " (" + PENDING_PUSH_ATTACHMENT + ");",
   };
 
+  public static final Set<Long> THUMBNAIL_TASKS = new HashSet<>();
```
moxie0: private non-static

src/org/thoughtcrime/securesms/database/PartDatabase.java
```diff
@@ -379,13 +376,59 @@ private InputStream getDataStream(MasterSecret masterSecret, long partId, String
     }
   }
 
+  private static void waitForThumbnail(long partId) {
```
moxie0: non-static

src/org/thoughtcrime/securesms/database/PartDatabase.java
```diff
@@ -379,13 +376,59 @@ private InputStream getDataStream(MasterSecret masterSecret, long partId, String
...
+    Log.w(TAG, "waiting on thumbnail task for part " + partId);
+    synchronized (THUMBNAIL_TASKS) {
+      while (THUMBNAIL_TASKS.contains(partId)) {
+        try {
+          THUMBNAIL_TASKS.wait();
+        } catch (InterruptedException ie) {
+          Thread.currentThread().interrupt();
```
moxie0: We don't (and will never) use interrupts, this should assert.

src/org/thoughtcrime/securesms/database/PartDatabase.java
```diff
@@ -379,13 +376,59 @@ private InputStream getDataStream(MasterSecret masterSecret, long partId, String
...
+        }
+      }
+    }
+  }
+
+  private static void addThumbnailJobIfNecessary(Context context, long partId) {
```
moxie0: This seems a little convoluted.  We're making blocking calls from this thread, why not just generate the thumbnail directly here, rather than kicking off an async job and immediately waiting on it?

src/org/thoughtcrime/securesms/mms/ImageSlide.java
```diff
@@ -81,9 +83,14 @@ public Drawable getThumbnail(int maxWidth, int maxHeight) {
...
       Bitmap thumbnailBitmap;
       long startDecode = System.currentTimeMillis();
-      Log.w(TAG, (part.getThumbnailUri() == null ? "generating" : "fetching pre-generated") + " thumbnail");
-      if (part.getThumbnailUri() != null) thumbnailBitmap = BitmapFactory.decodeStream(PartAuthority.getPartStream(context, masterSecret, part.getThumbnailUri()));
-      else                                thumbnailBitmap = BitmapUtil.createScaledBitmap(context, masterSecret, getUri(), maxWidth, maxHeight);
+
+      if (part.getDataUri() != null && part.getId() > -1) {
+        thumbnailBitmap = BitmapFactory.decodeStream(DatabaseFactory.getPartDatabase(context)
+                                                                    .getThumbnailStream(masterSecret, part.getId()));
+      } else {
```
moxie0: When is this true?

src/org/thoughtcrime/securesms/jobs/ThumbnailGenerateJob.java
```diff
@@ -73,29 +42,36 @@ public void onAdded() { }
...
+
+        database.updatePartThumbnail(masterSecret, partId, part, new ByteArrayInputStream(thumbnailBytes.toByteArray()), aspectRatio);
+      } else {
+        Log.w(TAG, "thumbnail not generated");
+      }
+    } finally {
+      synchronized (PartDatabase.THUMBNAIL_TASKS) {
+        Log.w(TAG, "notifying on task " + partId);
+        PartDatabase.THUMBNAIL_TASKS.remove(partId);
+        PartDatabase.THUMBNAIL_TASKS.notifyAll();
```
moxie0: I don't think we want to expose this so directly.

src/org/thoughtcrime/securesms/mms/ImageSlide.java
mcginty: When you add an attachment and there's no part in the database, but it's represented as a PduPart

mcginty:
comments addressed!

src/org/thoughtcrime/securesms/database/PartDatabase.java
```diff
@@ -85,6 +89,8 @@
     "CREATE INDEX IF NOT EXISTS pending_push_index ON " + TABLE_NAME + " (" + PENDING_PUSH_ATTACHMENT + ");",
   };
 
+  private final Set<Long> THUMBNAIL_TASKS = new HashSet<>();
```
moxie0: thumbnailTasks

mcginty:
Tests added. Mockito is unable to call through to a real method when it's not public, so I inlined the waiting helper method to avoid scoping weirdness.

mcginty:
refactored yesterday (posting this comment for notification purposes)

src/org/thoughtcrime/securesms/database/PartDatabase.java
```diff
@@ -452,6 +550,31 @@ public void updatePartThumbnail(MasterSecret masterSecret, long partId, PduPart
...
+    database.update(TABLE_NAME, values, ID_WHERE, new String[]{partId+""});
+  }
+
+  public void markThumbnailTaskStarted(long partId) {
+    synchronized (thumbnailTasks) {
+      thumbnailTasks.add(partId);
+    }
+  }
+
+  public boolean markThumbnailTaskStartedIfAbsent(long partId) throws FileNotFoundException {
```
moxie0: I don't think you need to double-check here if you're only calling it from the Job, but I think it'd actually be better if you rewrote this to use it in both places.  Have this return an InputStream if it's present, or null *and acquire the lock* if it's not.  Then get rid of isThumbnailInDatabase all together, and refactor the getThumbnailStream to use this method and be super-short.

src/org/thoughtcrime/securesms/jobs/ThumbnailGenerateJob.java
```diff
@@ -72,47 +34,17 @@ public ThumbnailGenerateJob(Context context, long partId) {
...
-  }
-
-  private Bitmap generateImageThumbnail(MasterSecret masterSecret, PduPart part) {
     try {
-      int maxSize = context.getResources().getDimensionPixelSize(R.dimen.thumbnail_max_size);
-      return BitmapUtil.createScaledBitmap(context, masterSecret, part.getDataUri(), maxSize, maxSize);
-    } catch (FileNotFoundException | BitmapDecodingException | OutOfMemoryError e) {
-      Log.w(TAG, e);
-      return null;
+      partDatabase.generatePartThumbnail(masterSecret, partId);
```
moxie0: Doesn't feel to me like you should be able to call a method like this on PartDatabase.  You should be generating the thumbnail somewhere and either loading it or storing it on PartDatabase.  The fact that it happens to occasionally generate one for the caller on load should be opaque to the outside world.

src/org/thoughtcrime/securesms/mms/PartParser.java
```diff
@@ -37,11 +37,11 @@ public static String getMessageText(PduBody body) {
     return bodyText;
   }
 
-  public static PduBody getNonTextParts(PduBody body) {
+  public static PduBody getDisplayableParts(PduBody body) {
```
moxie0: I don't think this is a great name.  Text is a displayable part, in my mind.

src/org/thoughtcrime/securesms/database/PartDatabase.java
```diff
@@ -379,18 +382,106 @@ private InputStream getDataStream(MasterSecret masterSecret, long partId, String
...
+      long    startMillis = System.currentTimeMillis();
+      Bitmap  thumbnail   = MediaUtil.generateThumbnail(context, masterSecret, part.getDataUri(), Util.toIsoString(part.getContentType()));
+
+      if (thumbnail == null) {
+        Log.w(TAG, "media doesn't have thumbnail support, skipping.");
+        return;
+      }
+
+      InputStream jpegStream  = BitmapUtil.toCompressedJpeg(thumbnail);
+      float       aspectRatio = (float) thumbnail.getWidth() / (float) thumbnail.getHeight();
```
moxie0: You do this sequence in two places. Maybe have MediaUtil return a Thumbnail that encapsulates it.

src/org/thoughtcrime/securesms/database/PartDatabase.java
```diff
@@ -197,14 +201,19 @@ public void deleteAllParts() {
     }
   }
 
-  void insertParts(MasterSecret masterSecret, long mmsId, PduBody body) throws MmsException {
+  void insertParts(MasterSecret masterSecret, long mmsId, PduBody body, Map<PduPart, Bitmap> thumbnailMap) throws MmsException {
```
moxie0: This is so crazy that I'm willing to let the thumbnail be in the pdupart.

PR #2288, opened 2015-01-02
Only initialise EmojiGrid when not in edit mode (fixes NPE in edit mode)

samlanning:
Android Studio keeps showing this message:

Suppress initialisation of the emoji_grid_layout.xml when the EmojiDrawer is
being used in a layout where the editor is shown, to prevent an NPE when an
editor is being used.

// FREEBIE


mcginty:
prefer braces on multiline conditionals, but otherwise lgtm!

Issue #2285, opened 2015-01-01
Send button should stick to last method (Signal, insecure SMS) on a per-user basis


deutrino:
What it says on the tin; taken from discussion at the bottom of #845. As @samlanning mentioned and other users indicated, the send method must currently be selected with every message with a chat partner. "Stickiness" would be helpful, as long as the Send button reflected it visually (including for color-impaired users).

Implementing this would also be a workaround to #845 itself.

mcginty:
Right now it's only sticky while the conversation is open in order to make sure users aren't accidentally sending unsecured messages with a user can have secure conversations. We'd need to think of a good way to allow longer-term "stickiness" without users losing awareness of the fact that the conversation *could* be secure.

samlanning:
Hangouts-esque style with the little caret to indicate more options may be appropriate. It would require a re-think of icons on left / right etc... I've been intending on actually implementing exactly this for the last few weeks, but free time has been scarce with all Christmas related activities...

colans:
What about a configuration option in the settings with choices such as the following?

* Not sticky at all (default)
* Sticky per conversation
* Sticky per addressee
* Sticky globally

moxie0:
@colans The answer is never more options.

gordon-morehouse:
@colans, I actually came across the [guidelines for contributing](https://github.com/WhisperSystems/TextSecure/blob/master/contributing.md) last night and tried to craft this feature respect with having no more options in mind. :) (I agree with the sentiment and was taught more than once by people I respect that "every checkbox is a failure." Yeah obviously that's flexible.)

@mcginty I was apparently mistaken in believing that it "un-stuck" between each message based on the recent comments in #845. Somebody was complaining about inconvenience when texting a lot.

samlanning:
@gordon-morehouse the inconvinience will still occur if the activity is closed often (i.e. switching between apps / closing TextSecure)

agrajaghh:
I like the hangout style, its really intuitive...

mgerring:
This is a huge issue for me because several of my friends stopped using Signal, and I didn't find out they haven't been getting my messages for months.

lenisko:
Yeah, should be set as urgent. Saving state shouldn't be so hard, we are already saving color per contact. Also like mentioned before, my friends stopped using it because of this and also lack of some sort notification that message wasn't delivered.

2-4601:
> I didn't find out they haven't been getting my messages for months.

​
[...]

Issue #845, opened 2014-02-27
Unable to send non-encrypted SMS to previous TextSecure user


carpenma:
A friend of mine had TextSecure installed on their device and we exchanged keys.  They have since uninstalled the app.  My device hasn't realized that they no longer use the service so it continues to try to send encrypted push messages to them.  I have tried clearing the app data and refreshing the push directory through the settings to no avail.  I have also tried ending the secure session from the padlock menu in the thread but when I do it just changes back to secure before it sends the next message (presumably because it needs confirmation from the other user?).  I'm not sure if this is a bug or if I am missing something but I would appreciate some guidance either way.  Thank you.

gordon-morehouse:
I opened a feature request for what was discussed in the last few comments (lack of "stickiness" in message transport method when selecting from the Send button) - please see #2285.

wrought:
Having just gone through this issue with a friend, it seems to me that the behavior of the Lock / Unlock icon (button) is not intuitive.

When I select the closed Lock icon, I should be able to clearly end the secure session, even if I am registered to use the Push feature and currently using the Push transport mode. To me, "clearly ending the secure session" would mean that sending new messages should default to the insecure SMS transport mode, I should have an option to re-initiative a secure session, and the "Locked" lock icon should change to the "Unlocked" lock icon. 

If you think that the changes ya'll have specified so far as a result of this thread deliver the experience I've described above (or an equivalent), then I am satisfied. Otherwise, please let me know.

rhodey:
with the dropping of encrypted SMS and MMS and some bug fixes related to transport selection on the send button this should no longer be an issue, closing.

fersingb:
@rhodey Maybe I missed something but I think it's still an issue.

When I try to send a SMS to a friend who used to have TextSecure but uninstalled it, the default behavior in TextSecure is still set to TextSecure. I've to long press on the arrow and select SMS.

Also, my friend is still appearing in the TextSecure users list.

Or have some bug fixes you talked about not been released yet?

agrajaghh:
your friend can unregister here: http://whispersystems.org/textsecure/unregister

fersingb:
That's the info I was missing, thanks !

mgraupner:
[...]

PR #2289, opened 2015-01-03
Updated Transport UI to be more "Hangouts-Like"

samlanning:
Following comments on #2285 #845 #1724.

UI is now suitable to allow for persistent transport selection. 

**Screenshots:**

SMS Only Contact

Other Contacts


samlanning:
@mcginty: I believe this is your domain?

samlanning:
I'm thinking perhaps I need to add a bit of green through the loop of the lock on the SMS icons?

agrajaghh:
Isn't `Secure SMS` missing in the last screenshot?

samlanning:
@agrajaghh it's like that for me with current master too, I haven't changed the behaviour as to what transports are enabled.

samlanning:
I have got contacts where all three are enabled however...

agrajaghh:
okay :)

samlanning:
Rebased ontop of current head, and replaced TS transport icon with latest logo.

@mcginty

moxie0:
Thanks, but we're thinking of adding a tool tip for our current send button selector instead of breaking things out into a separate component.

PR #2290, opened 2015-01-03
[DONT MERGE] Make Transport Selection Persistent (Follows #2289)

samlanning:
A follow-on from #2289, allowing for transport selection to be persistent.

However there's some weird race condition behaviour going on with retrieving the thread from the `ThreadDatabase` to get the transport, and I haven't been able to narrow it down. I've left in some print statements.

**WARNING:** If you run this, i makes changes to DB.


moxie0:
Thanks @samlanning, but please don't submit PRs that aren't ready.

samlanning:
ok no problem.

Issue #2271, opened 2014-12-26
Error on transmission results in large message, crashing TextSecure.


phryneas:
Hi,

the scenario:

* B sends an image to A over encrypted channel.
* There is an error on transmission
* A cannot decrypt the message 
* Textsecure interprets the message as plaintext (in the pane with all chats it is visible in preview as some base64-string)
* On opening of the chat Textsecure tries to render the multi-kilobyte "text" message
* Crash

This is critical as A can not chat with B any more. The only hotfix so far seems that B sends 10-20 messages to A, so the corrupted message is not visible on opening the conversation, but only after scrolling.

It is still impossible to scroll behind the corrupted message this way, so this is just a temporary hotfix.

Suggested fix: all text messages with more than X characters are rendered only partially and have to be clicked on to be shown completely. This will allow for the corruped message to be rendered partially and removed by hand.

(As a side effect this will allow for easier scrolling beyond long messages)

rhodey:
this should now be resolved thanks to #2294, if you're still having this issue I'll reopen this :)

Issue #2252, opened 2014-12-21
Receiving vcards with images crashes Textsecure and makes it unusable.


Richard-W:
A contact of mine sent me a vcard that contained an image via MMS. Textsecure downloaded it and displayed the whole ASCII encoded image as text which slowed down the app to the point where it crashed.

Textsecure was not usable until i managed to delete the MMS (on the 5th try it did not crash).

This could be used as some kind of DOS attack i think. Please include some kind of safeguard.

rhodey:
closing this as resolved by (in crash, maybe not in feature) #2294 and dupe of #499

Issue #2226, opened 2014-12-15
Crash because of big picture sent via SMS


peter-sc:
I received MMS with 2MB picture inside. The thread with this message is opening for about 2mins on 2-core CPU. Then the picture is displayed as encrypted text (no picture). After pressing on the message and trying to delete it, app crashes.
Other MMS messages, which were sent before of smaller size, were received and shown correctly.

Device: Xperia SP, running CM11, nightly 22-11-2014
Textsecure: 2.3.3 from Play store

```
12-15 22:45:40.100 I/PowerManagerService(  775): Waking up from sleep...
12-15 22:45:40.240 E/MP-Decision(  227): Error setting a sleep mode for secondary cores - -38
12-15 22:45:40.240 I/QCOM PowerHAL(  775): Got set_interactive hint
12-15 22:45:40.240 D/SurfaceFlinger(  305): Screen acquired, type=0 flinger=0xb81e54e0
12-15 22:45:40.240 D/qdhwcomposer(  305): hwc_blank: Unblanking display: 0
12-15 22:45:40.291 I/ActivityManager(  775): Process com.android.mms (pid 9535) has died.
12-15 22:45:40.291 W/ActivityManager(  775): Scheduling restart of crashed service com.android.mms/.transaction.TransactionService in 1000ms
12-15 22:45:40.511 D/qdhwcomposer(  305): hwc_blank: Done unblanking display: 0
12-15 22:45:40.521 D/SurfaceControl(  775): Excessive delay in unblankDisplay() while turning screen on: 273ms
12-15 22:45:40.601 W/ActivityManager(  775): getAssistContextExtras failed: no resumed activity
12-15 22:45:40.641 I/WindowManager(  775): Lock screen displayed!
12-15 22:45:40.641 D/DisplayPowerController(  775): Unblocked screen on after 179582903 ms
12-15 22:45:40.751 D/WifiConfigStore(  775):  wifi or gsm to wlan conn type is auto , should auto connect
12-15 22:45:40.831 D/LightsService(  775): Excessive delay setting light: 189ms
12-15 22:45:41.932 D/PhoneStatusBar(20732): disable: < expand* icons alerts ticker* system_info back* home* recent* clock* search >
[...]

PR #2294, opened 2015-01-04
limit displayed message length to 1024 characters.

phryneas:
* limit displayed message length to 1024 characters.
* added a "messageTooLongIndicator" that can be clicked to display the
full content of a truncated message

this should concern issues #2252 #2226 and  #2271 (and possibly more)

even though it does not resolve the problem of corruped messages, corrupted messages won't crash textsecure any more and it is possible to delete them

this also removes a possible DOS on a user's history with another user by sending a very long message that will crash on display and thus prevent scrolling back in history

logcat for the case of a too big message crashing TS:

    01-04 14:18:57.523  14631-14642/org.thoughtcrime.securesms I/art﹕ [ 01-04 14:18:58.045 14631:14642 I/art      ]
        Wrote stack traces to '/data/anr/traces.txt'
    01-04 14:19:00.477  14631-14631/org.thoughtcrime.securesms A/libc﹕ new[] failed to allocate 296136 bytes
    01-04 14:19:00.477  14631-14631/org.thoughtcrime.securesms A/libc﹕ Fatal signal 6 (SIGABRT), code -6 in tid 14631 (crime.securesms)

(content of /data/anr/traces.txt did not seem too interesting to me but if requested, I can provide it)

### DIFF
--- a/res/layout/conversation_item_received.xml
+++ b/res/layout/conversation_item_received.xml
@@ -107,6 +107,23 @@
 
             </LinearLayout>
 
+
+            <!-- displayed if message is longer than ConversationItem.DISPLAY_MAXLENGTH -->
+            <TextView android:id="@+id/conversation_item_message_too_long_indicator"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:textAppearance="?android:attr/textAppearanceSmall"
+                android:layout_gravity="right"
+                android:textColor="?conversation_sent_text_secondary_color"
+                android:visibility="gone"
+                android:layout_marginRight="8dip"
+                android:paddingTop="2dip"
+                android:paddingLeft="8dp"
+                android:paddingRight="5dp"
+                android:paddingBottom="2dp"
+                android:text="@string/ConversationItem_message_too_long_indicator"
+                />
+
             <LinearLayout
                 android:layout_width="wrap_content"
                 android:layout_height="match_parent"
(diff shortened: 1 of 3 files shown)


agrajaghh:
You have to remove the comments to follow the code guidlines

phryneas:
Oh sorry I didn't know there even were such a thing as code guidelines.
I found them now in the wiki and will clean the code up accordingly.

To prevent that in the future: Could a link the code guidelines maybe be added to the README.md under "Contributing Code" or at least the BUILDING.md?

phryneas:
Okay, cleaned up and spacing should be two-spaces now, too.

mcginty:
So, in push messages we limit the length to 1000. If we get a message over that, I think we should just either error out or cut it off, not allowing them to expand the text. I think that's better than having a button that questionably crashes the app :P

phryneas:
I wasn't aware of that - removing that part of the code would be okay with me.

But as the code is already written just as a suggestion: 

Even 1K messages take a while to render on my S4, which is not really a low-spec phone.
So it could be considered to decrease the DISPLAY_MAXLENGTH to something like 500 (or even lower?) to increase overall rendering speed and scrolling smootheness and just remove the warning part from the button text as as you are right, 1K messages will probably crash nobody.

(As a comparison: I have some ~16K messages. Those take quite a long time to render, but still no crashes at that point)

phryneas:
I reduced it to "just show 1000 characters" as suggested by @mcginty .

While I think the other functionality might be useful, I agree that it might take some discussion and I feel it's more important to push this simple fix to the codebase as soon as possible, as it resolves a number of issues. (or at least makes TextSecure still usable for those experiencing these issues)

PR #2301, opened 2015-01-05
fix temp file permission denied

mcginty:
was getting permission denied IOError when downloading attachments on my GB phone


moxie0:
lgtm

PR #2302, opened 2015-01-06
use internal emoji pane for all android versions

### LINKED ISSUE #2183: Smiles choice on participant phone
Hi All,
 I do not know where are smiles in TextSecure coming from , but on Android <4.4 they are brilliant!

screenshot_2014-12-04-09-57-22

 My wife got a new Phone with Android 4.4.x and she was so upset, almost crying when she find out that smiles now are simple rounded as everywhere.
 Is there is a way to collect smile set from the Android <4.4 versions like on a screenshot and add to a newer release as option "Use classic smiles" or similar? This can be also feature request.
Thanks you.

P.S. how it looks now:
screenshot_2014-12-04-14-54-55
screenshot_2014-12-04-14-55-04

### LINKED ISSUE #2251: Enter key does not send
Hitting enter simply makes a carriage return and does not send a text. This is despite the "Enter key sends" setting being checked. 
I'm using version 14.0.9 on Android 5.0.1. Thanks.

### LINKED ISSUE #2300: Emoji-Button
Dear developers,

first of all I would like to say thank you for your hard and great work!

I am using TextSecure for nearly one year now -  with my family, friends and in my circle of acquaintances.

During Christmas time a lot of those guys got new phones or updated to the newest available Android versions, running now mostly Android KitKat.

The problem such "normal" end-users nowadays face, is the missing Emoji-button within the version running on KitKat.
Unfortunately not everyone knows how to install alternative keyboards or is willing to do so. And not every type of keyboard offers the functionality of "Emojis" (e.g. on Huawei Ascend P7).
There are several reasons I somehow could understand those people.
 It's a pity - if such users switch back to commercial messenger apps because of this "problem".

It would be great if you could add the Emoji-button once again in the KitKat-version, so we could spread TextSecure even wider.

Thank you for your time and consideration.

kaz4r

mcginty:
1) Enter key by default (remove as option), still
   keeping "Enter key sends" option

2) New material assets for emoji and keyboard icons,
   inc a xxhdpi version finally

fixes #2300, fixes #2251, fixes #2183


luksen:
Is this still happening? I love using emoji, but sadly am currently stuck with the ugly Samsung ones.

wp9015362:
Hello,

any update on this one?

It would be much appreciated if this change would land in TextSecure.

Regards

moxie0:
I've never really been sold on doing this.  It's built into the Android keyboard, and using our own means everyone unnecessarily loses that much space in the compose edittext.  It makes sense for apps which aren't using the Android emoji set, but we are.

agrajaghh:
seems like some devices come with keyboards which don't work as expected (see #2745), maybe it makes sense to handle this in TS. Even Hangouts is doing it on its own...

mcginty:
Yeah, I no longer believe we can rely on the keyboard to handle emoji.

moxie0:
I can live if you want to do that. However, if you disable the Google keyboard's emoji capability, then our emoji drawer needs to be an exact replica, or at least contain all of the glyphs and functionality that it does.  Currently it looks like our drawer is missing a couple of sections.

Also, this needs to be rebased.

mcginty:
@moxie0 what emoji keyboard are you using? I'm pretty sure our drawer provides the whole set...

mcginty:
rebased

moxie0:
the google keyboard has seven sections, our keyboard only has six. It's definitely missing the ascii faces section, and the section headers are different.

wp9015362:
Hello,

really looking forward to this getting merged.

I am curious:

When this will be merged, what Emojis will TextSecure have in it's Emoji drawer on 4.4. KitKat devices and above?

Will it have the native Emojis, i.e. Samsung Emojis on Samsung devices, LG Emojis on LG devices, HTC Emojis on HTC devices, Google (Noto) Emojis on Nexus devices and so on?

Or will it have a specific set of Emojis that will be built into TextSecure, like WhatsApp and Telegram are doing it for example?

I really would prefer the latter, because the manufacturer specific (native) Emojis all are differening from each other, and that might lead to confusion. I mean if someone sends an Emoji to someone else and the recipient sees a different Emoji because he uses a different device with different Emojis.

Therefore I prefer the way WhatsApp and Telegram are doing it for example. They have one specific set of Emojis which is built into the app, so everyone sees the same Emojis. Much better that way IMHO.

And in case the TextSecure Emoji drawer will have it's own Emoji set: Which set will that be?

Will it be the Emoji One set (http://emojione.com/)?

Or the Twemoji set (https://github.com/twitter/twemoji)?

Or which one?

> @moxie0 wrote:
> 
> However, if you disable the Google keyboard's emoji capability,
> then our emoji drawer needs to be an exact replica,
> or at least contain all of the glyphs and functionality that it does.

I agree that it should not have less functionality. But IMHO it should not be an exact replica.

The Google keyboard Emoji drawer shows the space key and scrolling through the Emojis of each section goes horizontally.

I personally don't like that.

The Emoji drawer in Telegram for example shows no space key, which allows more Emojis to be shown on display and scrolling through the Emojis of each section goes vertically (scrolling/swiping between sections goes horizontally though). I like that much better.

Here's a screenshot of the Telegram Emoji drawer (also shows WhatsApp Emoji drawer next to it):

http://imagenes.es.sftcdn.net/blog/es/2014/02/Whatsapp-and-telegram.jpg

Source: http://features.en.softonic.com/telegram-better-than-whatsapp

Oh, and please do not make the background of the Emoji drawer fully black. Please make it be a little brighter than black. Because using a dark background for the Emoji drawer makes it hard to see some of the darker Emojis. Gray would be better than black for that reason (like WhatsApp is doing it for example).

Regards

moxie0:
in the 2.10.0 branch

Issue #2280, opened 2014-12-30
low resolution for a/v-attachment icons


McLoo:
The icons for audio and video attachments are only available in 72x72 which makes them look pretty ~~ulgy~~ retro in the draft preview. Besides that the media message bubble is too large for them.

Copied both draft preview icons next to each other:

![unbenannt](https://cloud.githubusercontent.com/assets/5326916/5583312/a42679c4-9076-11e4-9f8b-81ca10a43d9f.png)

And somehow they both don't fit into the app design...

Would b more consistent to have ic_attach_audio_holo_light/dark ![ic_attach_audio_holo_light](https://cloud.githubusercontent.com/assets/5326916/5583380/2e17a166-9078-11e4-86b5-c8655d7b46ad.png) and ic_attach_video_holo_light/dark ![ic_attach_video_holo_light](https://cloud.githubusercontent.com/assets/5326916/5583382/384f1ede-9078-11e4-8b51-3afe762358cc.png) indicate the media type.

mcginty:
That 8-bit feel :pager:. Yeah, we can probably yoink some of the latest material icons for the cause.

McLoo:
thanks @sebkur

PR #2304, opened 2015-01-06
Mediaicons

sebkur:
Fix for issue #2280: added high-resolution, themed media icons for audio and video attachments.


McLoo:
I like them. And the generator :)

agrajaghh:
yep, looks good! Did you write the generator?

sebkur:
Yep, I originally wrote a similar version for a different project, but I tend to reuse it a lot :)

sebkur:
By the way, those icons are not the only viable choices for audio/video items, just what I found most appropriate. With the script we can easily pick a different icon from the Material icon set.

manhole11:
+1, those low-res pics are really ugly

mcginty:
Cool script, although I don't know that it belongs in the TextSecure repository and we've moved all source artwork and design guidelines into our pixelapse account at this point. Since the source art you're using is in Google's material-design-icons repository already, we can probably just skip on that as it's versioned elsewhere and we probably want to use the most up-to-date version they have.

src/org/thoughtcrime/securesms/mms/Slide.java
```diff
@@ -125,4 +126,11 @@ protected static void assertMediaSize(Context context, Uri uri)
       if (size > MAX_MESSAGE_SIZE) throw new MediaTooLargeException("Media exceeds maximum message size.");
     }
   }
+
+  public static Drawable resolveIcon(Context c, int iconAttr)
```
mcginty: this looks more fitting for `Util` rather than `Slide`

src/org/thoughtcrime/securesms/mms/Slide.java
```diff
@@ -125,4 +126,11 @@ protected static void assertMediaSize(Context context, Uri uri)
...
+  {
+    TypedValue out = new TypedValue();
```
mcginty: does `out` need to be recycled after resolving the attribute?

sebkur:
My idea was to have the script, and thus the means for creating the PNG icons, in the repo so that it would be easy to update the graphics once for whatever reason we decide to use a different source icon. I think it's good to keep these things together (otherwise how would somebody be able to conveniently create the different scaled PNG files). However, I agree that it does not belong into the repo if there's a different place to store such things. Is it possible to put scripts into that pixelapse project you mentioned?

Concerning the SVG files: I guess I could modifiy the script to download the base files from the material repo into a temporary directory before creating the PNG files, which sound like a good idea to me, indeed.

mcginty:
For the same reason we use a PNG crusher but do not include one in the repo, it's often cleaner to reference dependencies rather than include them in an effort to avoid monolithic codebases. I guess what I'm saying is that I think this script could be cleaned up and used across a *ton* of Android projects, making it a good candidate to be maintained in PyPI and meanwhile added to a readme here as we'd consume it.

agrajaghh:
I used a modified version of this script at #2361 too. I think it could be good to include it here or at pixelapse. Since it specifies suffixes, colors and opacities it's project/file specific.

src/org/thoughtcrime/securesms/mms/AudioSlide.java
```diff
@@ -63,8 +64,8 @@ public SMILMediaElement getMediaElement(SMILDocument document) {
   }
 
   @Override
-  public Drawable getThumbnail(int maxWidth, int maxHeight) {
-    return context.getResources().getDrawable(R.drawable.ic_menu_add_sound);
+  public Drawable getThumbnail(Context context, int maxWidth, int maxHeight) {
```
moxie0: Is there a reason to pass in Context here?  Just looking at the diff, AudioSlide already has a reference.

sebkur: Yes, it already has a context, however it does not represent correctly the current style that is selected by the user, i.e. resolving icons with this instance will still return light icons even if the user selected the dark theme sometime before.

sebkur:
Okay, I generalized the script a bit and moved it to a separate repository here on GitHub (not sure about the process of getting it into PyPI).

Here's the repository:
https://github.com/sebkur/android-res-utils

To generate images now, here's what to do:

Clone the repository

Execute the following:

    /path/to/android-res-utils/pngs_from_svg.py ic_audio.svg 150 /path/to/TextSecure/res/
    /path/to/android-res-utils/pngs_from_svg.py ic_video.svg 150 /path/to/TextSecure/res/

src/org/thoughtcrime/securesms/mms/Slide.java
sebkur: Which `Util` class do you suggest? `util.DynamicTheme` or maybe a new class `util.Resources` or `util.Drawables`?

src/org/thoughtcrime/securesms/mms/Slide.java
sebkur: I do not find any documentation concerning this. For `TypedArray` there is a method `recycle()`, and the docs for `obtainStyledAttributes` state to use it, however for `TypedValue` there seems to be no equivalent counterpart

sebkur:
How do you resolve merge conflicts with pull requests? Should I rebase on the master and `push -f` or do a merge? Or do you resolve merge conflicts yourself?

agrajaghh:
I always rebase on master, I think its the least work for the maintainers...

mcginty:
Yeah, the context passed doesn't have any clue about the theme since it's just whatever context `MmsDatabase` has. Seems reasonable to pass a "target context" that contains the theme you wish to use when styling.

I rebased, squashed your commits, shrunk pngs, put a paragraph in `BUILDING.md`, and moved the icon helper method into a new `ThemeUtil.java`.

Tested on GB and KK for sanity, looks... so, so much better.

McLoo:
sweet!
but now the add attachment dialog icons differ...

sebkur:
Ah, indeed. They should be replaced with Material icons, too...

agrajaghh:
created a PR to fix this #2535

Issue #2296, opened 2015-01-04
Unlock Screen Theme


ghost:
~~don't bother posting issues here, they won't get fixed.. mods don't care~~ :see_no_evil: 
 
something about the theme of the unlock screen.

PR #2307, opened 2015-01-07
Fix theme usage for passphrase activity

danbartram:
### Changes
* Apply the dark theme to the `PassphrasePromptActivity` activity if the setting is enabled.
    * Prior to this commit, this activity is limited to only displaying in the light theme.

### Issues Solved
* [#2296](https://github.com/WhisperSystems/TextSecure/issues/2296)


McLoo:
Awesome. thanks!
The background pattern color doesn't feel 100% right though.

danbartram:
> The background pattern color doesn't feel 100% right though.

Is it too bright or dark?

McLoo:
@danbartram can't tell really. just a feeling :)

But the header text color could be a little lighter brighter.
maybe you can stick to the other colors already used - if you haven't already.

danbartram:
@McLoo I tried to keep the colors similar but that doesn't mean it's the best look.

[light_centered_app_title.xml:13](https://github.com/WhisperSystems/TextSecure/blob/master/res/layout/light_centered_app_title.xml#L13)
```
android:textColor="#55000000"
```

I moved this value into the different themes in `themes.xml`, kept the alpha value the same but inverted the RGB values for the light/dark theme.

**Light Theme**
```
<item name="centered_app_title_color">#55000000</item>
```

**Dark Theme**
```
<item name="centered_app_title_color">#55ffffff</item>
```

----

I think I'd agree that there could be a little bit more contrast in the header title text, but kept the colors close to (what I thought was) consistent.

As for the background pattern, that was pretty much just my personal taste as there wasn't a pre-existing texture to go with. If I made it too dark then the pattern was essentially a solid color.

I'm experimenting with the coloring now to see what looks best.

----

## Changes

### New `background_pattern_dark`
background_pattern_dark

#### Before
screenshot_2015-01-07-23-10-44

#### After
screenshot_2015-01-07-23-12-24

Is this a better color? I've darkened it a bit more and it might look better to you now. I'm not sure if the `EditText` looks too light in comparison now though?

### New `centered_app_title_color` color
I've changed it from the `#55ffffff` value to use the `@color/gray27` value and it seems to look better (compare it to the "after" image of the background pattern comparison).

screenshot_2015-01-07-23-23-11


----

Let me know what you think of these changes.

Also, would you prefer I submit a new PR if you prefer the new changes?

McLoo:
@danbartram last one looks great to me. But it's up to @mcginty what (and if) get's merged.
*Could also be, that they come up with a completely new design after their Winter of Code ;)*

nwdev:
THANK YOU GUYS!

danbartram:
@McLoo - Thank you for the suggestions. If they do like it, I'll open a new PR with the latest changes applied too. We'll just have to wait and see what happens once February hits ;)

@nwdev - You're welcome, enjoy :)

nwdev:
Any idea when this will be merged???

agrajaghh:
since this seemed to be inactive I opened #2627

PR #2325, opened 2015-01-12
Add expiration check on build freshness.

moxie0:


src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
```diff
@@ -48,7 +48,7 @@ public void onAdded() {
   }
 
   @Override
-  public void onRun(MasterSecret masterSecret) throws NoSuchMessageException {
+  public void onRunSend(MasterSecret masterSecret) throws NoSuchMessageException {
```
mcginty: `onSend`?

src/org/thoughtcrime/securesms/util/Util.java
```diff
@@ -288,6 +290,11 @@ public static void fixBackgroundRepeat(Drawable bg) {
     }
   }
 
-
-
+  public static void verifyBuildFreshness() throws TextSecureExpiredException {
```
mcginty: seems like an `isBuildFresh()` fits as a general-purpose Util method and then the caller can throw exceptions as needed?

moxie0:
done

Issue #1963, opened 2014-10-01
Updating directory for new textsecure users only displays after leaving screen


anemes:
When updating directory new users, the screen does not refresh with new text-secure users. Leaving and re-entering the new text scree updates the list of users as expected.

PR #2326, opened 2015-01-12
refresh contacts database on new db query

calvinhu4:
Re: Issue #1963
This should refresh the list of TextSecure users when the refresh button is pressed in the in-app contacts list.


mcginty:
Merged as c89fbabbf3e03292e5cfa6eb4f01a30296966d80

Issue #2245, opened 2014-12-19
Sending SMS (not allowed!)


SourcemakerHD:
I configured TextSecure to not use the SMS feature. So no SMS messages should be sent. Since the last update TextSecure sents SMS messages which are not allowed!

chmac:
I also have this issue. I have all outgoing SMS blocked by  I was trying to send messages to a friend. He unregistered from the PUSH service, and registered with a new number. I went into the old conversation, and I replied to a message. That message was sent by SMS (at significant cost, I'm roaming!). First type of unallowed SMS send.

Second type, I was messaging on the PUSH service, then lost WiFi momentarily, and again, TextSecure sent an SMS without my permission.

In my case, TextSecure is not my default messaging app. I also have some kind of permission manager running, which I've just used to block TextSecure's "Send SMS" permission. Hopefully that avoids this happening again for me, but it appears that this is definitely an issue which can affect other users.

rbu:
There's several possible bugs in this issue.

@SourcemakerHD mentioned in the third comment, part (1) that it may be possible to start a conversation with a non-TS user (where no conversation exists yet). How can you reproduce this? I could not reproduce.

@SourcemakerHD mentioned in (2) that an existing conversation (via PUSH) turns into an SMS conversation when the other person unregisters PUSH. This is a problem I could reproduce and which I would like to raise as well. Sending a large message, a friend of mine encountered the cost of a dozen SMS when someone else unregistered.
When SMS is set to "Nobody" this needs to be respected even in existing conversations!

@chmac raised a third issue which is the app falling back to SMS when PUSH is temporarily unavailable. I cannot reproduce. Are you sure the SMS was actually sent? Or do you have a green error message?

chmac:
In the case where PUSH was temporarily unavailable, the UI shows me the same green message, and when I press and hold on it, then click the info (i), it tells me it was sent via SMS. However, @rbu you're absolutely right, it didn't actually send the message. It also shows me a red triangle on the left, which I guess means "message failed to send".

chmac:
As a point of feedback, showing the same green message for failed messages or for messages sent via the SMS transport is a little confusing. I realise (now) that a small red triangle has been added, but that wasn't clear to me immediately that the message had not been sent. Might be a nice idea to make the background red or grey instead of the green, to show that the message had not been sent at all.

SourcemakerHD:
@rbu
[...]

Issue #2161, opened 2014-11-30
sending sms possible when sms disabled in existing conversation


jagz6262:
In the version 2.3.2, When we choose the settings to allow outgoing SMS to check the options:
Text Secure users or Non-Text Secure Users. I choose only the text secure users and sends message to other party without having the text secure application and the other party was able to receive the message.
Device: Nexus 5 Android Version 5.0 lollipop

agrajaghh:
> Ah yeah. If you have an existing conversation open with a non-TS user, you can currently continue to send SMS with them.

Thats still possible, but intended, right? Closing?

FeuRenard:
This issue refers options that are no more present in Signal. It should be closed, I think.

PR #2350, opened 2015-01-14
prevent SMS/MMS composition in UI when messaging to non-TextSecure users

calvinhu4:
Re: Issue #2161, #2245


rbu:
What is holding back the merge?

moxie0:
@rbu We were waiting for you to write something!

rbu:
@moxie0 I am honestly wondering if there is anything keeping this from being merged. If there is, possibly I could do something to help. I do not understand how sarcasm helps this discussion.

mcginty:
Closing this, since we've changed how SMS is handled after the encrypted SMS change. Also, if we do something here we'd probably want to still allow for the auto-upgrade to a TextSecure message happen if we find out the user has recently registered.

PR #2363, opened 2015-01-18
Add pref to change MMS user agent

### LINKED ISSUE #466: Requesting userAgent mod or option in TextSecure settings
First off let me thank you for a great app.

I am using an unlocked phone on the SoftBank network in Japan.  The SoftBank mms server checks the userAgent field to make sure it matches phones sold by SoftBank.

It would be great if Textsecure could either add the res/xml-mcc440-mnc20/ folder with all files from the xml folder but with mms_config.xml modified as described here:
http://lazydroid.com/2012/04/howto-send-and-receive-mms-on-galaxy-nexus-softbank/
and here
http://forum.xda-developers.com/showthread.php?t=2451413

Or allow users to input userAgent as an option in settings.  I believe GO SMS Pro has this function but I would prefer something with better security.

I have tried to make the edits myself but am getting recompile errors.  I will keep trying but I am sure many more would benefit form such as option.

SamWhited:
Fixes #466


agrajaghh:
I think it would be better to include this setting in the "manual MMS settings" not in the main "SMS and MMS" settings

SamWhited:
I thought about that (see the issue), but since the manual MMS settings have to be turned on and off it felt silly to connect this with all the other manual settings (eg. you turn them on just wanting to set a user agent to get around some block, but now you have to lookup all your providers settings and enter these too). But I agree it would probably be better there; maybe with a second checkbox just for this? I'm hesitant to add too many settings.

mcginty:
The on/off issue with manual MMS is fixed in 534df06794e62a801da618c3efb552bf9c035583

SamWhited:
Moved the pref over to Manual MMS with the new `CustomDefaultPreference`. Don't have my test device on me at the moment to install and check.

Also signed the CLA which, while I understand the purpose, is deeply irritating. Will avert a rant by shutting up now.

res/xml/preferences_manual_mms.xml
```diff
@@ -33,4 +33,11 @@
...
             android:inputType="textVisiblePassword"/>
 
-</PreferenceScreen>
\ No newline at end of file
+    <org.thoughtcrime.securesms.components.CustomDefaultPreference
+            android:key="pref_mms_user_agent"
+            android:title="@string/preferences__mms_user_agent"
+            android:summary="@string/preferences__mms_override_mms_user_agent"
+            android:hint="@string/preferences__default_mms_user_agent"
+            android:inputType="textNoSuggestions"/>
```
moxie0: have you tested this? I don't see how this could work without a custom_pref_toggle specified

SamWhited: No, sorry; I thought I said in my previous comment (but apparently not): I don't have a test device which I can run TS on at the moment.

moxie0:
Oh, I see.  I appreciate the PR, and I think this change is a good idea, but please only submit PRs that you've tested.

SamWhited:
Yah, absolutely, sorry about that (in my defense, the original was tested, but you're correct, I shouldn't have made changes until I could test again).

PR #2369, opened 2015-01-20
Update travis.yml to make builds 'passing' finally

McLoo:
...as soon as Travis gets activated for the TextSecure we no longer see  [Build Status](https://travis-ci.org/WhisperSystems/TextSecure) in readme.md but a neat [Build Status](https://travis-ci.org/WhisperSystems/TextSecure)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,14 @@
 language: android
 android:
-  components:
-    - platform-tools
-    - build-tools-19.1.0
+components:
+    - build-tools-21.1.2
     - android-19
     - extra-android-m2repository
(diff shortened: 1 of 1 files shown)


kaeawc:
I would suggest looking into [Travis CI's container builds](http://docs.travis-ci.com/user/workers/container-based-infrastructure/)

There is a good example of using them in an open source project [here](https://github.com/SecUpwN/Android-IMSI-Catcher-Detector/blob/master/.travis.yml).

McLoo:
rebased. still green :)

@kaeawc thanks for the hint. But TS doesn't feel well inside a container... Last test was at 14 min, which is OK i think.

PR #2375, opened 2015-01-21
material actionbar and fab

mcginty:
* AppCompat v21, new color scheme based on the Signal blue (#2090EA), assets for all actionbar icons
* FAB for creating new conversations instead of an actionbar icon
* New material-er icon
* Move photoshop/AI sources out of artwork folder and into pixelapse repository at https://www.pixelapse.com/openwhispersystems/projects/signal-android/views/assets/

This PR is the most asset-heavy of the material changes. Net loss is pretty nice though ;)

logo-512


src/org/thoughtcrime/securesms/PassphraseCreateActivity.java
```diff
@@ -86,15 +85,6 @@ protected void onPostExecute(Void param) {
     }
   }
 
-  private void mitigateAndroidTilingBug() {
```
agrajaghh: you forgot to delete the call to `mitigateAndroidTilingBug` in line 54, right?

agrajaghh:
I like it! :+1: 

nitpick, but I think it would look better if the keyhole would be a bit higher:

original:
original

higher:
higher

mcginty:
great point @agrajaghh, will fix up that icon

mcginty:
@moxie0 committed the workaround for samsung crashes, testing was done on Samsung devices that were reported in the Android issue to have issues, as well as some LG devices which were also supposedly behaving weirdly. No crashes.

moxie0:
ship it!

agrajaghh:
@mcginty `drawable-xxxhdpi/icon.png` is green not blue, I guess thats a mistake?

Also the old `actionbar_icon_holo_light` is still shipped and referenced in `themes.xml`. But I think you're assigning the icons now in `styles.xml`, right?

midi:
Yay, looking sleek.

Push it to Play please :)

Murazaki:
NEED !

McLoo:
```@drawable/lockscreen_watermark``` is also still more grass than sky

I guess it is not intended, that  the salmon(?) colored warning for submitting debug logs is scrolled out of view once the log is loaded now?

McLoo:
New group from long press FAB would be nice too

McLoo:
The "more menu" (forward) when a single message is selected needs a dark background for the dark theme.
Else it's white on white...

McLoo:
One (last) more thing: I still have difficulties to understand the shaddow dropped by the key*hole*. 
I had expected light falling through...
 
unbenannt

midi:
Another of my most used applications, getting a material face-lift?
Shippitiy-ship it to Play please. Such a nice present.

ship it squirrel

Murazaki:
I've seen @McLoo 's proposal and I thought I could give it a try. I though the long light one was lacking form...

textsecure

Murazaki:
material_icon_gigantic_light

mcginty:
Yeah, I tried a logo with the light coming through and just couldn't get it to a point where I liked it better than the more abstract "floating keyhole." Could very likely just be personal preference, as I've heard others liked the light better. @Murazaki interesting thought with trying to make it have a bit more depth, I might play around with it more later.

@McLoo I have other material reworks in the way, this PR is scoped just for ActionBar and FAB. Reworking group creation is on the chopping block for later :)...

jmue:
From my perspective the shine through effect conflicts with the material launcher spec. To create a shine through effect the light source has to sit below the bubble. But the light source(s) are in front of the bubble like this:
nexus2cee_image_thumb6 1

The Image is taken from here: http://www.androidpolice.com/2014/10/08/googles-new-vision-for-launcher-icons-paper-lighting-and-how-to-get-your-icon-ready/

So - following the spec and take the keyhole as a hole the keyhole itself gets a shadow. But that would not be visible.

Murazaki:
Well specs are always lacking information. This case is really special, and I don't think you could really state that we can't move one light behind the icon to highlight the keyhole and still be doing material design.

mcginty:
Well, it could go both ways. I like the idea of playing off of the material design which dictates shadows being cast on paper and offering an icon that instead provides a source of light. Just don't think we got it quite yet, as it's still a bit flat compared to the shadowy version.

mranta:
My first reaction to seeing @McLoo's version was the same as @jmue's. It contradicts material design and breaks continuity by introducing a new light source. I recommend going through the actual [material design icons style specs](https://www.google.com/design/spec/style/icons.html#icons-product-icons).

The only changes I'd make to @mcginty's [first version](https://cloud.githubusercontent.com/assets/373823/5846102/2f3c3662-a165-11e4-8fc3-a5529463b0bd.png) would be to add tinting and shadowing to the foreground and background layers' edges.

>The top and bottom edges of material elements provide a sense of depth and surface. Material elements have a standard 1dp thickness. All edge distances are measured from an element's interior edge.

>Tint highlights the top edge of all elements. The left, right, and bottom edges do not have a tint applied.

>Shade darkens the bottom edge of all elements. The left, right, and top edges do not have a shade applied.


Tinted edge specs:
>Height: 1dp
>Opacity: 20%
>Color: White (#FFFFFF)

All of the previous variations contradict material design in one way or another.

mcginty:
I refined the icon a bit in https://github.com/WhisperSystems/TextSecure/commit/ba62e018db89793aca4526f241e4b3188cb66b9c.

PR #2380, opened 2015-01-23
Also crop contact photos in ShareList

AsamK:
Commit 62816ee51ac80b446c788ec50bbd94d967ebddbb forgot to switch the ShareListActivity (which appears when sharing from another App) to the RoundedImageView.
Only group icons that were still stored cropped were shown correctly, contact photos were shown as squares. This commit fixes this.


mcginty:
Committed as d586893402b530d4f710b4240fc87c8c3b80201a, thanks @AsamK!

PR #2428, opened 2015-02-08
Workaround for LGE-related NPEs

mcginty:
I didn't see this crash during my most recent AppThwack iterations on LG devices, but it appears it's still in the wild. Below is a proposed fix, limited in scope based on comments in the android project bug. That final comment seems to suggest that while Build.BRAND may not accurately filter buggy devices, Build.MANUFACTURER will.

See: https://code.google.com/p/android/issues/detail?id=78154

### DIFF
--- a/src/org/thoughtcrime/securesms/BaseActionBarActivity.java
+++ b/src/org/thoughtcrime/securesms/BaseActionBarActivity.java
@@ -1,0 +1,22 @@
+package org.thoughtcrime.securesms;
+
+import android.support.annotation.NonNull;
+import android.support.v7.app.ActionBarActivity;
+import android.view.KeyEvent;
+
+
+public abstract class BaseActionBarActivity extends ActionBarActivity {
+  @Override
+  public boolean onKeyDown(int keyCode, KeyEvent event) {
+    return BaseActivity.isKeyCodeWorkaroundRequired(keyCode) || super.onKeyDown(keyCode, event);
+  }
+
+  @Override
+  public boolean onKeyUp(int keyCode, @NonNull KeyEvent event) {
+    if (BaseActivity.isKeyCodeWorkaroundRequired(keyCode)) {
+      openOptionsMenu();
+      return true;
+    }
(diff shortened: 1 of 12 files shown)


src/org/thoughtcrime/securesms/BaseActionBarActivity.java
```diff
@@ -0,0 +1,23 @@
+package org.thoughtcrime.securesms;
+
+import android.support.annotation.NonNull;
+import android.support.v7.app.ActionBarActivity;
+import android.view.KeyEvent;
+
+import static org.thoughtcrime.securesms.BaseActivity.isKeyCodeWorkaroundRequired;
```
moxie0: I'm into static imports for inner classes, but not really for methods.

src/org/thoughtcrime/securesms/BaseActivity.java
```diff
@@ -0,0 +1,28 @@
...
+      openOptionsMenu();
+      return true;
+    }
+    return super.onKeyUp(keyCode, event);
+  }
+
+  public static boolean isKeyCodeWorkaroundRequired(int keyCode) {
+    return (keyCode == KeyEvent.KEYCODE_MENU) &&
+           (Build.VERSION.SDK_INT == 16)      &&
+           ("LGE".equalsIgnoreCase(Build.MANUFACTURER));
```
moxie0: There are some crashes for non-LG devices. Is there a reason to only do this for LG?

mcginty: This is based on the reports from the Google issue - I assumed the device may not be labeled LG, but may still have them as the `MANUFACTURER`. Does the Play console say whether there's another Build.MANUFACTURER affected by this?

We can also just not check. It seemed like a good idea to reduce the number of cases in which we override key presses to a minimum, though. Given LGE has custom code on menu keypresses, we would likely be preventing other non-stock OEM ROMs from running their custom code, and I have no good way to guess what the implications of that are.

PR #2429, opened 2015-02-09
add onboarding "intro"

mcginty:
A pleasant and colorful introduction to TextSecure to give a smoother intro than a flashing progress dialog.


src/org/thoughtcrime/securesms/PassphraseCreateActivity.java
```diff
@@ -37,27 +56,80 @@
 
 public class PassphraseCreateActivity extends PassphraseActivity {
```
moxie0: I think this needs to be renamed if this is what it does now.

src/org/thoughtcrime/securesms/PassphraseCreateActivity.java
```diff
@@ -37,27 +56,80 @@
...
 
   @Override
   public void onCreate(Bundle savedInstanceState) {
+    supportRequestWindowFeature(Window.FEATURE_NO_TITLE);
     super.onCreate(savedInstanceState);
-
     setContentView(R.layout.create_passphrase_activity);
-
     initializeResources();
+    TextSecurePreferences.setOnboardingCompleted(this, false);
```
moxie0: Why do you need a separate flag for this? Never needed one before.

src/org/thoughtcrime/securesms/PassphraseCreateActivity.java
```diff
@@ -37,27 +56,80 @@
...
+
+    getSupportActionBar().hide();
   }
 
   private void initializeResources() {
```
moxie0: This method needs to be heavily refactored, pretty hard to read as is.

src/org/thoughtcrime/securesms/RoutingActivity.java
```diff
@@ -169,7 +169,7 @@ private Intent getPushRegistrationIntent() {
   }
 
   private int getApplicationState() {
-    if (!MasterSecretUtil.isPassphraseInitialized(this))
+    if (!MasterSecretUtil.isPassphraseInitialized(this) || !TextSecurePreferences.isOnboardingCompleted(this))
```
moxie0: You want to make all existing users go through the onboarding or something?

src/org/thoughtcrime/securesms/PassphraseCreateActivity.java
```diff
@@ -37,27 +56,80 @@
...
+
+      @Override
+      public void onPageScrollStateChanged(int state) {
+      }
+    });
+
+    if (!MasterSecretUtil.isPassphraseInitialized(this)) {
+      TextSecurePreferences.setPasswordDisabled(this, true);
+      new SecretGenerator().execute(MasterSecretUtil.UNENCRYPTED_PASSPHRASE);
+    } else {
```
moxie0: I think this should be designed so that you never get here.

moxie0:
In general I don't totally love this idea.  I understand wanting to make the create passphrase process look a little nicer, but turning it into a flow where the user needs to manually intervene seems like a step backwards to me.  My sense is that we want to be moving further in the opposite direction -- less interaction required to get to the app.

mcginty:
I pushed this to v2.7.0, and Before we go further with this screen, I'm going to play around with integrating it into the registration page instead like you were suggesting.

PR #2430, opened 2015-02-09
Conversation thumbnail improvements

mcginty:
* Improve media aesthetics in-conversation with larger and borderless photo thumbnails.
* Preload the thumbnail in the media preview and do scaling animations where supported to make opening images feel snappier (can eventually add in shared element transitions for Lollipop)
* Programmatically manipulate message colors and shapes, so less inflation and less XML, hence this change being a net loss.


agrajaghh:
Good idea! Looks way better than displayed inside the bubble! I like the scaling animation, it really feels snappier :blush: 

I got two remarks:
* The blue box with the time got a white shadow in the bottom. This looks weird for some pictures.
* If you send a picture, it's first displayed normally, after ~0,5s its displayed much brighter (indicating that its not sent yet) and finally displayed normal again. This "flashing" is a bit disturbing. It should be displayed brighter in the beginning I think.

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -258,12 +270,22 @@ private void setSelectionBackgroundDrawables(MessageRecord messageRecord) {
     drawables.recycle();
   }
 
-  private void setBodyText(MessageRecord messageRecord) {
+  private boolean isCaptionlessMms() {
+    return TextUtils.isEmpty(messageRecord.getDisplayBody()) && messageRecord.isMms();
+  }
+
+  private void setBodyText() {
```
moxie0: always pass read-state variables to private methods, we are not heathens

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -271,6 +293,28 @@ private void setBodyText(MessageRecord messageRecord) {
     }
   }
 
+  private void setMedia() {
```
moxie0: same

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -207,38 +214,43 @@ public void setHandler(Handler failedIconHandler) {
...
+        final int color = backgroundDrawables.getColor(background, -1);
+        shadow.setColor(backgroundDrawables.getColor(MESSAGE_SHADOW, -1));
+        message.setColor(color);
+        triangle.setColor(color);
+        media.setColor(color);
+
+        final int topRadius = captionedMms ? 0 : r;
+        message.setCornerRadii(new float[]{r, r, topRadius, topRadius, r, r, r, r});
+        media.setCornerRadii(new float[]{r, r, r, r, topRadius, topRadius, r, r});
       }
```
moxie0: This method needs to be heavily refactored for less side effects and more readability.

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -271,6 +293,28 @@ private void setBodyText(MessageRecord messageRecord) {
...
+    RelativeLayout.LayoutParams parentParams = (RelativeLayout.LayoutParams)conversationParent.getLayoutParams();
+    if (isCaptionlessMms()) {
+      parentParams.addRule(RelativeLayout.BELOW, 0);
+      parentParams.addRule(RelativeLayout.ALIGN_BOTTOM, R.id.mms_view);
+    } else {
+      parentParams.addRule(RelativeLayout.BELOW, R.id.mms_view);
+      parentParams.addRule(RelativeLayout.ALIGN_BOTTOM, 0);
+    }
+    conversationParent.setLayoutParams(parentParams);
+  }
```
moxie0: this needs to be refactored

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -394,14 +442,22 @@ public void onSuccess(final SlideDeck result) {
           public void run() {
             for (Slide slide : result.getSlides()) {
               if (slide.hasImage()) {
-                slide.setThumbnailOn(mmsThumbnail);
+                if (slide.getPart().isPendingPush()) {
+                  mmsDownloading.setVisibility(View.VISIBLE);
+                  mmsContainer.setVisibility(View.VISIBLE);
+                  return;
```
moxie0: I'd prefer it if this were encapsulated in the slide

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -157,13 +158,19 @@ public void onPause() {
...
     imageAttacher = new PhotoViewAttacher(image);
    }
 
   private void displayImage() {
+    try {
+      image.setImageBitmap(BitmapFactory.decodeStream(PartAuthority.getThumbnail(this, masterSecret, mediaUri)));
+      image.setVisibility(View.VISIBLE);
+    } catch (IOException fnfe) {
+      Log.w(TAG, "tried to render thumbnail, but it wasn't found. carrying on.");
+    }
```
moxie0: This seems kinda gross

moxie0:
I'd like it thumbnail/media stuff to be encapsulated so they don't spill out into ConversationItem.  In general I also prefer 1000x XML lines to lines of UI code.

mcginty:
While I'd much prefer to do this entirely in XML and only deal with switching predefined states in the UI, we'd be looking at not only doubling the amount of XML, but also the attributes in the code that need resolving... just for adding variable radius corners into the mix.

When we want to modify the shadows or change their distances... just not going to scale well in XML. I think it'll look better if I refactor it.

manhole11:
How would this be displayed?

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
mcginty: What else should we be doing? If we can't render the thumbnail, we can move on and try to render the fullsize image, so this isn't a critical exception.

mcginty:
This is now updated with a refactor that:

* moves message appearance state management into a new ConversationBubble class
* fixes for the alpha animations that appeared in GB
* re-encapsulates animation management into ImageSlide

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -256,12 +245,26 @@ private void setSelectionBackgroundDrawables(MessageRecord messageRecord) {
...
   }
 
+  private static boolean hasConversationBubble(MessageRecord messageRecord) {
+    return !messageRecord.isGroupAction();
+  }
+
+  private static boolean isCaptionlessMms(MessageRecord messageRecord) {
+    return TextUtils.isEmpty(messageRecord.getDisplayBody()) && messageRecord.isMms();
+  }
+
```
moxie0: static?

src/org/thoughtcrime/securesms/components/OutgoingConversationBubble.java
```diff
@@ -0,0 +1,98 @@
...
+
+  private static final Map<Integer, Integer> TRANSPORT_STYLE_MAP;
+  static {
+    Map<Integer, Integer> styleMap = new HashMap<>(TRANSPORT_STYLE_ATTRIBUTES.length);
+    styleMap.put(TRANSPORT_STATE_PUSH_SENT, 0);
+    styleMap.put(TRANSPORT_STATE_SMS_SENT, 1);
+    styleMap.put(TRANSPORT_STATE_SMS_PENDING, 2);
+    styleMap.put(TRANSPORT_STATE_PUSH_PENDING, 3);
+    TRANSPORT_STYLE_MAP = Collections.unmodifiableMap(styleMap);
+  }
```
moxie0: = Collections.unmodifiableMap(new HashMap<Integer, Integer>() {{ put(..., ...);}}; ?

agrajaghh:
since you're already merging `StyleUtil` and `ThemeUtil`, how about removing `resolveIcon()` from [util/Dialogs.java](https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/util/Dialogs.java#L44) and just use `resolveIcon()` in `ThemeUtil`?

agrajaghh:
I just run your branch again. I really like the design! :smile: 

But unfortunately it's still flickering for me, if I send a picture or even if I send a text but the picture is still visible.

here is a video of sending a text. You can see the image flickering twice. (I had to put the video in a zip file, otherwise the automatic compression would eliminate the flickering...):
https://drive.google.com/open?id=0B6cQGnZWgOWIYTM1WloyQmFYUUk&authuser=0

mcginty:
Moving back to XML got too unwieldy, but agreed that using mutate() to modify whatever drawable was attached to the view was way too rigid.

* I profiled having a Builder class to create bubble shapes as we'd like them, and it's quite performant so I'm going that route instead.
* Migrated all resource-related helper methods into a ResUtil, instead of in ThemeUtil and Dialogs, thanks @agrajaghh for the recommendation.

Next up I'm going to create an ImageView that manages its own entry animation, and will also check out the screen flickr, as unfortunately this change makes it even more obvious (it existed before, just didn't flash to white).

mcginty:
the imageview is now required to handle its own animations.

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -404,14 +403,18 @@ public void onSuccess(final SlideDeck result) {
           public void run() {
             for (Slide slide : result.getSlides()) {
               if (slide.hasImage()) {
-                slide.setThumbnailOn(context, mmsThumbnail);
-                mmsThumbnail.setOnClickListener(new ThumbnailClickListener(slide));
-                mmsThumbnail.setVisibility(View.VISIBLE);
+                if (slide.getPart().isPendingPush()) {
+                  mediaThumbnail.setImageResource(R.drawable.stat_sys_download);
+                  mediaThumbnail.setScaleType(ScaleType.CENTER_INSIDE);
+                  mediaThumbnail.setVisibility(View.VISIBLE);
```
moxie0: this seems like the kind of stuff the mediaThumbnail should encapsulate and be responsible for.  knowing how to display images, as well as knowing how to display progress.  that the caller should have to set the scale time, visibility, etc... feels like a misplacement of responsibility to me.

src/org/thoughtcrime/securesms/ConversationItem.java
moxie0: still static?

src/org/thoughtcrime/securesms/components/OutgoingBubbleContainer.java
```diff
@@ -0,0 +1,107 @@
...
+
+  private static final Map<Integer, Integer> TRANSPORT_STYLE_MAP;
+  static {
+    Map<Integer, Integer> styleMap = new HashMap<>(TRANSPORT_STYLE_ATTRIBUTES.length);
+    styleMap.put(TRANSPORT_STATE_PUSH_SENT, 0);
+    styleMap.put(TRANSPORT_STATE_SMS_SENT, 1);
+    styleMap.put(TRANSPORT_STATE_SMS_PENDING, 2);
+    styleMap.put(TRANSPORT_STATE_PUSH_PENDING, 3);
+    TRANSPORT_STYLE_MAP = Collections.unmodifiableMap(styleMap);
+  }
```
moxie0: = new HashMap<Integer, Integer>() {{ put(..); }};

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -384,42 +384,15 @@ else if (MmsDatabase.Status.DOWNLOAD_APN_UNAVAILABLE == messageRecord.getStatus(
...
-      mmsThumbnail.setVisibility(View.VISIBLE);
-      mmsContainer.setVisibility(View.VISIBLE);
-      mmsThumbnail.setImageDrawable(new ColorDrawable(Color.TRANSPARENT));
+
+      mediaThumbnail.setVisibility(View.INVISIBLE);
+      mediaThumbnail.setImageDrawable(null);
     } else {
-      mmsThumbnail.setVisibility(View.GONE);
-      mmsContainer.setVisibility(View.GONE);
+      mediaThumbnail.setVisibility(View.GONE);
```
moxie0: I still think this should be a complex object capable of managing its own state. having to call setVisibility immediately followed by setImageDrawable(null) feels wrong to me

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -667,4 +635,44 @@ public void onClick(DialogInterface dialogInterface, int i) {
...
+    }
+  }
+
+  private class SlideDeckListener implements FutureTaskListener<SlideDeck> {
+    @Override
+    public void onSuccess(final SlideDeck slideDeck) {
+      if (slideDeck == null) return;
+
+      Slide slide = slideDeck.getThumbnailSlide(context);
```
moxie0: getThumbnailSlide can return null.  It should either return @ NonNull and throw an AssertionError if it's actually impossible to reach that return null statement, or there should be a null check here.

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -667,4 +635,44 @@ public void onClick(DialogInterface dialogInterface, int i) {
...
+      thumbnailFuture = slide.getThumbnail(context);
+      if (thumbnailFuture != null) {
+        thumbnailFuture.addListener(new ThumbnailListener());
```
moxie0: You have an instance variable that's a thumbnailListener, should you have passed that into this class' constructor and used it here instead of creating a new one?

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -667,4 +635,44 @@ public void onClick(DialogInterface dialogInterface, int i) {
...
+  }
+
+  private class SlideDeckListener implements FutureTaskListener<SlideDeck> {
+    @Override
+    public void onSuccess(final SlideDeck slideDeck) {
+      if (slideDeck == null) return;
+
+      Slide slide = slideDeck.getThumbnailSlide(context);
+      thumbnailFuture = slide.getThumbnail(context);
+      if (thumbnailFuture != null) {
```
moxie0: By contrast, it doesn't seem that this can be null? can it be annotated @ NonNull and this null check dropped?

src/org/thoughtcrime/securesms/components/BubbleContainer.java
```diff
@@ -0,0 +1,160 @@
...
+    bodyBubble.setLayoutParams(parentParams);
+  }
+
+  private boolean isMediaPresent(@MediaState int mediaState) {
+    return mediaState != MEDIA_STATE_NO_MEDIA;
+  }
+
+  private boolean isPending(@TransportState int transportState) {
+    return transportState == TRANSPORT_STATE_PUSH_PENDING || transportState == TRANSPORT_STATE_SMS_PENDING;
+  }
```
moxie0: If you used an enum instead of @ intdef, you could put these methods on the enum object itself.

manhole11:
"..larger and borderless photo thumbnails." is there a reason why my thumbnails are not borderless?

Issue #2425, opened 2015-02-08
Crash when trying to attach a picture


gladiac:
Hello,

I'm running version 2.5.0 and it often crashes for me if I try to attach a picture. Here is the backtrace  ...

```
D/Documents(16838): onFinished() [content://com.android.providers.media.documents/document/image%3A2512]
W/ConversationActivity( 5868): onActivityResult called: 1, -1 , Intent { dat=content://com.android.providers.media.documents/document/image:2512 flg=0x43 }
W/ImageSlide( 5868): Got soft reference: null
W/ImageSlide( 5868): generating thumbnail for new part
W/InputMethodManagerService( 3600): Starting input on non-focused client com.android.internal.view.IInputMethodClient$Stub$Proxy@42064820 (uid=10011 pid=5868)
I/dalvikvm-heap( 5868): Clamp target GC heap from 66.112MB to 64.000MB
I/dalvikvm-heap( 5868): Forcing collection of SoftReferences for 1997584-byte allocation
I/dalvikvm-heap( 5868): Clamp target GC heap from 66.112MB to 64.000MB
E/dalvikvm-heap( 5868): Out of memory on a 1997584-byte allocation.
I/dalvikvm( 5868): "AsyncTask #1" prio=5 tid=21 RUNNABLE
I/dalvikvm( 5868):   | group="main" sCount=0 dsCount=0 obj=0x420b9db0 self=0x617469b8
I/dalvikvm( 5868):   | sysTid=6203 nice=10 sched=0/0 cgrp=apps/bg_non_interactive handle=1634789184
I/dalvikvm( 5868):   | state=R schedstat=( 628065124 191858043 491 ) utm=56 stm=6 core=6
I/dalvikvm( 5868):   at android.graphics.BitmapFactory.nativeDecodeStream(Native Method)
I/dalvikvm( 5868):   at android.graphics.BitmapFactory.decodeStreamInternal(BitmapFactory.java:620)
I/dalvikvm( 5868):   at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:596)
I/dalvikvm( 5868):   at org.thoughtcrime.securesms.util.BitmapUtil.createScaledBitmap(BitmapUtil.java:138)
I/dalvikvm( 5868):   at org.thoughtcrime.securesms.util.BitmapUtil.createScaledBitmap(BitmapUtil.java:102)
[...]

Issue #2372, opened 2015-01-20
share / add attachment: picture == > crash


peterpanondrugs:
same problem still existing! .... was closed on thread #2248 
(i didnt know if should open a new one again or post in closed. couldnt find in guidelines)
what happened there on the closed thread? i dont see the progress.
cheers 

Device : Samsung GT-I9305 (m3xx)
Android: omni_i9305-userdebug 4.4.4 KTU84P 440 test-keys
App : TextSecure 2.4.2

 when i send a picture from gallery / or as attachment out of TS or from my camera. it crashes. can someone confirm this? 
log: https://gist.github.com/d0195640abc36b4f8cf5

mcginty:
You did fine opening a new issue :). Sorry you're still hitting this. Pasting the related part (OOM):

```
D/dalvikvm( 4164): GC_BEFORE_OOM freed 79K, 49% free 17815K/34420K, paused 46ms, total 46ms
E/dalvikvm-heap( 4164): Out of memory on a 31961104-byte allocation.
I/dalvikvm( 4164): "JobConsumer-1" prio=5 tid=13 RUNNABLE
I/dalvikvm( 4164):   | group="main" sCount=0 dsCount=0 obj=0x423bb318 self=0x5cd0ba30
I/dalvikvm( 4164):   | sysTid=4203 nice=0 sched=0/0 cgrp=apps handle=1557184136
I/dalvikvm( 4164):   | state=R schedstat=( 6746567746 16561401629 3732 ) utm=631 stm=43 core=3
I/dalvikvm( 4164):   at dalvik.system.VMRuntime.newNonMovableArray(Native Method)
I/dalvikvm( 4164):   at android.graphics.BitmapFactory.nativeDecodeStream(Native Method)
I/dalvikvm( 4164):   at android.graphics.BitmapFactory.decodeStreamInternal(BitmapFactory.java:620)
I/dalvikvm( 4164):   at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:596)
I/dalvikvm( 4164):   at org.thoughtcrime.securesms.util.BitmapUtil.createScaledBitmap(BitmapUtil.java:138)
I/dalvikvm( 4164):   at org.thoughtcrime.securesms.util.BitmapUtil.createScaledBitmap(BitmapUtil.java:102)
I/dalvikvm( 4164):   at org.thoughtcrime.securesms.util.BitmapUtil.createScaledBitmap(BitmapUtil.java:93)
[...]

PR #2438, opened 2015-02-10
memory leak fixes, batch one

mcginty:
Benchmarked with simulated high usage (open many conversations long and small with scrolling and many media previews), forced a GC, then took a `adb shell dumpsys meminfo org.thoughtcrime.securesms`.

Found leaks with heap dumps and MAT.

#### Before (with leaks): 93MB heap allocated
```
Applications Memory Usage (kB):

                   Pss  Private  Private  Swapped     Heap     Heap     Heap
                 Total    Dirty    Clean    Dirty     Size    Alloc     Free
                ------   ------   ------   ------   ------   ------   ------
  Native Heap       16       16        0        0    18356    16939     1416
  Dalvik Heap    69322    68912        0        0    90952    76586    14366
          ...
        TOTAL   110803   101200     6444        0   109308    93525    15782

 Objects
               Views:     4813         ViewRootImpl:        6
         AppContexts:       47           Activities:       44
              Assets:        7        AssetManagers:        7
       Local Binders:       51        Proxy Binders:       68
    Death Recipients:        0
     OpenSSL Sockets:        5
```

#### After: 46MB allocated
```
Applications Memory Usage (kB):

                   Pss  Private  Private  Swapped     Heap     Heap     Heap
                 Total    Dirty    Clean    Dirty     Size    Alloc     Free
                ------   ------   ------   ------   ------   ------   ------
  Native Heap       16       16        0        0    14388    11363     3024
  Dalvik Heap    56341    55924        0        0    77968    34849    43119
          ...
        TOTAL    88908    79892     6284        0    92356    46212    46143

 Objects
               Views:     1260         ViewRootImpl:       12
         AppContexts:       15           Activities:       12
              Assets:        7        AssetManagers:        7
       Local Binders:      141        Proxy Binders:       26
    Death Recipients:        0
     OpenSSL Sockets:        5
```

Should help with #2425 and #2372

### DIFF
--- a/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
+++ b/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
@@ -64,6 +64,7 @@ public class MediaPreviewActivity extends PassphraseRequiredActionBarActivity im
   private final DynamicLanguage dynamicLanguage = new DynamicLanguage();
 
   private MasterSecret masterSecret;
+  private boolean      paused;
 
   private View              loadingView;
   private TextView          errorText;
@@ -114,13 +115,14 @@ public class MediaPreviewActivity extends PassphraseRequiredActionBarActivity im
     } else {
       relativeTimeSpan = null;
     }
-    getSupportActionBar().setTitle(recipient == null ? getString(R.string.MediaPreviewActivity_you) : recipient.getName());
+    getSupportActionBar().setTitle(recipient == null ? getString(R.string.MediaPreviewActivity_you) : recipient.toShortString());
     getSupportActionBar().setSubtitle(relativeTimeSpan);
   }
 
   @Override
   public void onResume() {
     super.onResume();
+    paused = false;
     dynamicLanguage.onResume(this);
     if (recipient != null) recipient.addListener(this);
     initializeMedia();
@@ -129,6 +131,7 @@ public class MediaPreviewActivity extends PassphraseRequiredActionBarActivity im
   @Override
   public void onPause() {
     super.onPause();
+    paused = true;
     if (recipient != null) recipient.removeListener(this);
     cleanupMedia();
   }
@@ -140,7 +143,6 @@ public class MediaPreviewActivity extends PassphraseRequiredActionBarActivity im
     setIntent(intent);
     initializeResources();
     initializeActionBar();
-    initializeMedia();
   }
 
   private void initializeViews() {
@@ -154,9 +156,9 @@ public class MediaPreviewActivity extends PassphraseRequiredActionBarActivity im
     final long recipientId = getIntent().getLongExtra(RECIPIENT_EXTRA, -1);
 
     masterSecret = getIntent().getParcelableExtra(MASTER_SECRET_EXTRA);
-    mediaUri = getIntent().getData();
-    mediaType = getIntent().getType();
-    date = getIntent().getLongExtra(DATE_EXTRA, -1);
+    mediaUri     = getIntent().getData();
+    mediaType    = getIntent().getType();
+    date         = getIntent().getLongExtra(DATE_EXTRA, -1);
 
     if (recipientId > -1) {
       recipient = RecipientFactory.getRecipientForId(this, recipientId, true);
@@ -211,8 +213,12 @@ public class MediaPreviewActivity extends PassphraseRequiredActionBarActivity im
 
       @Override
       protected void onPostExecute(Bitmap bitmap) {
-        loadingView.setVisibility(View.GONE);
+        if (paused) {
+          if (bitmap != null) bitmap.recycle();
+          return;
+        }
 
+        loadingView.setVisibility(View.GONE);
         if (bitmap == null) {
           errorText.setText(R.string.MediaPreviewActivity_cant_display);
           errorText.setVisibility(View.VISIBLE);
--- a/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
+++ b/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
@@ -18,6 +18,7 @@ package org.thoughtcrime.securesms;
 
 import android.annotation.TargetApi;
 import android.content.DialogInterface;
+import android.content.Intent;
 import android.graphics.Bitmap;
 import android.net.Uri;
 import android.opengl.GLES20;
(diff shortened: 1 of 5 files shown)


gladiac:
Thanks for looking into this!

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -99,31 +101,72 @@ private void setFullscreenIfPossible() {
...
+  private void initializeActionBar() {
+    final CharSequence relativeTimeSpan;
+    if (date > 0) {
+      relativeTimeSpan = DateUtils.getRelativeTimeSpanString(date,
+                                                             System.currentTimeMillis(),
+                                                             DateUtils.MINUTE_IN_MILLIS);
+    } else {
+      relativeTimeSpan = null;
+    }
+    getSupportActionBar().setTitle(recipient == null ? getString(R.string.MediaPreviewActivity_you) : recipient.getName());
```
moxie0: If you're listening for resolution, this should probably be recipient.toShortString() rather than recipient.getName().  If it's unresolved, the latter will return null.

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -99,31 +101,72 @@ private void setFullscreenIfPossible() {
...
+  private void initializeResources() {
     final long recipientId = getIntent().getLongExtra(RECIPIENT_EXTRA, -1);
 
     masterSecret = getIntent().getParcelableExtra(MASTER_SECRET_EXTRA);
-    mediaUri     = getIntent().getData();
-    mediaType    = getIntent().getType();
-    date         = getIntent().getLongExtra(DATE_EXTRA, -1);
+    mediaUri = getIntent().getData();
+    mediaType = getIntent().getType();
+    date = getIntent().getLongExtra(DATE_EXTRA, -1);
```
moxie0: alignment

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -137,32 +180,14 @@ public void onModified(Recipient recipient) {
...
-      relativeTimeSpan = DateUtils.getRelativeTimeSpanString(date,
-                                                             System.currentTimeMillis(),
-                                                             DateUtils.MINUTE_IN_MILLIS);
-    } else {
-      relativeTimeSpan = null;
+  private void cleanupMedia() {
+    image.setImageDrawable(null);
+    if (bitmap != null) {
+      bitmap.recycle();
+      bitmap = null;
```
moxie0: This isn't thread safe and will result in crashes.

mcginty: do you mean down the line? because `bitmap` is only set and accessed on the main thread

moxie0: You have an asynctask that's setting the bitmap on an imageview.

mcginty: The bitmap returned back from the AsyncTask won't be recycled afaik, as it's newly allocated from `BitmapUtil.createScaledBitmap`, so its memory is separate from whatever the instance variable `bitmap` contains.

moxie0: Maybe I'm looking at the diff wrong, but I see the lines:

`MediaPreviewActivity.this.bitmap = bitmap;`

Imagine that line is executed, then the thread is preemted.

Now there's the lines:
`````
    if (bitmap != null) {
      bitmap.recycle();
`````

Imagine those execute, then that thread is preempted.

Now back to the first thread:

`image.setImageBitmap(bitmap);`

Comment (mcginty) [inline src/org/thoughtcrime/securesms/MediaPreviewActivity.java]: Yeah, I think the fact that both the onPostExecute argument and MediaPreviewActivity instance variables are called `bitmap` is confusing.

The `image.setImageBitmap(bitmap)` is using the argument passed into `onPostExecute()`, which is nullchecked and will not have been recycled as it's freshly allocated.

`onPostExecute()` only sets `MediaPreviewActivity.this.bitmap`, it doesn't access it. And since `onPostExecute()` and `cleanupMedia()` are both run on the main thread, I can't think of a way that it'd break in the reverse direction either.

Comment (mcginty) [inline src/org/thoughtcrime/securesms/MediaPreviewActivity.java]: Edit: I see what you're saying now, but am I wrong that if both methods are run on the same thread, that kind of pre-emption doesn't happen in Java?

Comment (moxie0) [inline src/org/thoughtcrime/securesms/MediaPreviewActivity.java]: No you're right, that should be fine if it's the same thread.  Only danger is that it doesn't get recycled if the asynctask completes after the activity is closed.

**Comment by gladiac** [comment]
From the discussion about the patches it looks like important comments in the source code are missing to clarify the threading issue ;)

File: src/org/thoughtcrime/securesms/MediaPreviewActivity.java (hunk 3 above)
Comment (mcginty) [inline src/org/thoughtcrime/securesms/MediaPreviewActivity.java]: Good point, fixed!

**Comment by moxie0** [comment]
What are the rules for recycling bitmaps? Right now you're recycling in onPause, which means that the imageview could be (partially) displayed with a recycled bitmap backing it.  Is that a problem?  If not, why can't you immediately recycle a bitmap after setting it on an imageview?

**Comment by mcginty** [comment]
I call `setImageDrawable(null)` right before `recycle()` to no longer display the bitmap that's about to be recycled.

File: src/org/thoughtcrime/securesms/MediaPreviewActivity.java [hunk 4]
```diff
@@ -99,10 +102,58 @@ private void setFullscreenIfPossible() {
...
+  }
+
+  @Override
+  protected void onNewIntent(Intent intent) {
+    super.onNewIntent(intent);
+    if (recipient != null) recipient.removeListener(this);
+    setIntent(intent);
+    initializeResources();
+    initializeActionBar();
+    initializeMedia();
```
moxie0: onResume gets called immediately after onNewIntent, so it looks to me like two async tasks will run.  I think the first task to complete would result in a bitmap that never gets recycled.

mcginty:
(also updated)

moxie0:
lgtm

PR #2472, opened 2015-02-14
add Code style Guidlines to readme

agrajaghh:
I think the style guidlines are too hard to find in the wiki


moxie0:
I think this might be better in contributing.md?

agrajaghh:
yeah, you're right. I moved it

PR #2479, opened 2015-02-16
Prevent NPE when MMS's FROM PduHeader is missing

### LINKED ISSUE #1448: Crash when receiving MMS
This just started in the past couple of days, every time an MMS comes in, TextSecure crashes. My setup is a bit strange since I'm using t-mobile's network with a phone originally designed for AT&T. Recipients of my MMS messages get the "adaptation" error, but this crash when I receive one is new to me.

Here's the debug log that I ran just after the crash/app reboot: http://hastebin.com/xitukakiqa

mcginty:
Fixes #1448


moxie0:
lgtm

PR #2482, opened 2015-02-16
Prevent NPE in PassphraseChangeActivity

mcginty:
The issue's reproducible by tapping OK like a crazed lunatic (or just tapping it twice I guess) when changing your password.

### DIFF
--- a/src/org/thoughtcrime/securesms/PassphraseChangeActivity.java
+++ b/src/org/thoughtcrime/securesms/PassphraseChangeActivity.java
@@ -77,6 +77,8 @@ public class PassphraseChangeActivity extends PassphraseActivity {
   }
 
   private void verifyAndSavePassphrases() {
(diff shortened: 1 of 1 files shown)


moxie0:
Wait, if the handler is running on the UI thread, how can this change make a difference?

mcginty:
I figured that extra clicks are queued and when setMasterSecret is called, in between `bindService` and the `onServiceConnected` async callback happening, the subsequent click events happen, and at least one of them happens after `cleanup()` is called.

moxie0:
If extra clicks are queued, wouldn't there still be a window for a crash to occur before the buttons are turned off?

McLoo:
With sth. like
```
if (hasClicked) return;
hasClicked = true;
```
rather than disabling, queued events should be an issue?!

rhodey:

PR #2483, opened 2015-02-16
Un-jank character count indicators

### LINKED ISSUE #1841: counter of written characters vanishes after "0/63 (1)"
There is a counter for written characters at the bottom of editing screen for new message. After "0/63 (1)" counter vanishes.

Please add counter funcionality continuing after 63 characters. In Germany one text can contain 160 characters. So optimal counter would continue from "0/160 (1)" to "160/160 (2)"
This seems to me to be a bug, because (1) could indicate number of texts.

-> this helps people like me who still have to pay for each text seperately. With text secure its not possible to have control over text message costs. For me this is why I am still not using text secure as standard app. I think the code change will be just a few lines but would be a great improvement.

Thank you m0xie and team for this wonderful software. Thank you for contributing your free time for our privacy.

device info: motorola A953, CM Android 4.4.4, Text secure 2.1.5, I am push registred.

mcginty:
As the final installment of my Winter Cleaning Sunday™

Fixes #1841

### DIFF
--- a/src/org/thoughtcrime/securesms/TransportOption.java
+++ b/src/org/thoughtcrime/securesms/TransportOption.java
@@ -1,24 +1,42 @@
 package org.thoughtcrime.securesms;
 
+import org.thoughtcrime.securesms.util.CharacterCalculator;
+import org.thoughtcrime.securesms.util.CharacterCalculator.CharacterState;
+import org.thoughtcrime.securesms.util.EncryptedSmsCharacterCalculator;
+import org.thoughtcrime.securesms.util.PushCharacterCalculator;
+import org.thoughtcrime.securesms.util.SmsCharacterCalculator;
+
 public class TransportOption {
-  public int    drawable;
-  public String text;
-  public String key;
-  public String composeHint;
+  public int                 drawable;
+  public String              text;
+  public String              key;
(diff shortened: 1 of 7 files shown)


src/org/thoughtcrime/securesms/TransportOption.java
```diff
@@ -20,5 +32,8 @@ public boolean isForcedPlaintext() {
   public boolean isForcedSms() {
     return key.equals("insecure_sms") || key.equals("secure_sms");
   }
-}
 
+  public CharacterState calculateCharacters(int charactersSpent) {
+    return characterCalculator == null ? null : characterCalculator.calculateCharacters(charactersSpent);
```
moxie0: Instead of returning null and having null checks in the calling code, why not have a push character calculator that always returns 2000 or something?

mcginty: Makes twice as much sense now that we have the max limit! Yep, will do.

mcginty:
Added PushCharacterCalculator with 2000 as max size, changed new copyrights to 2015, and renamed `isForcedX` to `isX` since it didn't really make sense.

moxie0:
lgtm

PR #2484, opened 2015-02-16
Add first Espresso interaction test

mcginty:
Put the instrumentation in place to run Espresso 2 tests, plus an example test that a clean start and verifies the Register and Skip buttons are displayed.


moxie0:
lgtm

PR #2489, opened 2015-02-17
Set notification color to colorPrimary on api 21+

Memorion:
This is my first "real" pull request ever, please tell me if I missed something!


agrajaghh:
A picture is worth a thousand words:

after

**before:**
before

Memorion:
oh, sorry, I somehow assumed it would only color the small app icon and just tested with contacts which have their own picture

mcginty:
Closing since it's not quite pull ready (need to change the recipient icon to not have the weird semi-transparent square). Thanks @Memorion!

PR #2490, opened 2015-02-17
Fix icon in GroupMembersDialog

agrajaghh:
The Icon in the `GroupMembersDialog` is too bright in the light theme on Android > 2.3. On Android 2.3 the header of the dialog is black in both themes, so there is no issue...

I got the icons form the [Action Bar Icon Pack](http://developer.android.com/design/downloads/index.html) and used the original names, if you want to keep your filenames just tell me.

**before:**
before

**after:**
after


mcginty:
Now that we depend on `material-dialogs`, shall we just migrate this dialog to it and save the API-specific conditionals?

agrajaghh:
@mcginty yeah, that sounds better, I'll try to do that...

agrajaghh:
@mcginty done in #2492

PR #2492, opened 2015-02-17
migrate GroupMembersDialog to material-dialogs

agrajaghh:
better alternative to #2490 

**Android 5, light theme:**
light

**Android 5, dark theme:**
dark

**Android 2.3, light theme:**
2 3 light

**Android 2.3, dark theme:**
2 3 dark

The title seems too long, what do you think about "Group members"?

PR #2535, opened 2015-02-26
replace icons in the add attachment dialog

agrajaghh:
I kept the original file names from the material icon repo, is that ok? We also have the same licence "problems" as in #2492

should I migrate this (or all AlertDialogs in a seperate PR) to material-dialogs too?

**Android 5.0:**
attachment_5 0

**Android 2.3:**
attachment_2 3


jeremymasters:
Is this identical for the dark theme or inverted?  I know nothing of material design rules. :)

agrajaghh:
@jeremymasters currently all the alert dialogs have a light background...

agrajaghh:
@mcginty I guess you would like the artwork at pixelapse? Now its in the artwork folder...

Issue #2476, opened 2015-02-15
Language selection should ask for confirmation before changing to selected language.


git12a:
Presently, when one selects a language from the list inside Language settings, it just directly switches to that language. Imagine the panic on the mind of normal average user when he sees a language he accidentally tapped on (and don't know it) and wants to revert to previous (or English) language. 

It should ask "do you want to switch to LanguageName?"

DustinLim:
I feel that changing the theme/language is jarring because the app immediately returns you to the main settings screen. Returning to the language picker can be challenging, in case you selected the wrong one.

Asking confirmation might be overkill, however keeping the user on the appearance screen would help a lot.

PR #2552, opened 2015-03-02
Changing appearance settings keeps the user on the same screen

DustinLim:
Hi, this is my first time contributing here. I am part of a student group from the Delft University of Technology, following a course on Software Architecture. We have picked TextSecure as our Open Source project of interest and are at the moment tasked to try to contribute something. Which brings me to this pull request.

Changing the theme or language returns the user to the main settings screen, which can be jarring for the user. If you accidentally tapped on the wrong language this also makes it harder find your way back. This PR keeps the user at the appearance screen by redisplaying that Fragment after the Activity is recreated, when this scenario applies. 

Possibly a fix for #2476.

// FREEBIE


moxie0:
Thanks @DustinLim but this looks hackey.  We'd need you to implement something a little less convoluted in order to achieve the same thing.

PR #2554, opened 2015-03-02
message timestamps update onResume()

### LINKED ISSUE #2519: Time stamps very slow to update

rhodey:
modified conversation view and conversation list view to update message timestamps onResume. Fixes #2519

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -91,6 +91,18 @@ public class ConversationFragment extends ListFragment
     getLoaderManager().restartLoader(0, null, this);
   }
 
+  @Override
+  public void onResume() {
+    super.onResume();
+    getLoaderManager().initLoader(0, null, this);
+  }
+
+  @Override
+  public void onPause() {
+    super.onPause();
+    getLoaderManager().destroyLoader(0);
+  }
+
   private void initializeResources() {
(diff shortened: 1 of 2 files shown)


moxie0:
in an onCreate() situation, isn't this going to result in loading the entire data set, then immediately refreshing it?

rhodey:
@moxie0 good catch, in the `onCreate()` case the `ConversationAdapter.notifyDataSetChanged()` call would happen needlessly in `onResume()`.

Another option is to initialize the list adapter in `onResume()` but on it's own that also results in even more waste.

I think keeping track of `onPause()` and `onResume()` with a boolean `fragmentHasPaused` would be the most efficient and clean way about this. `onPause()` set `fragmentHasPaused = true`, `onResume()` if `fragmentHasPaused` then `ConversationAdapter.notifyDataSetChanged()`, `fragmentHasPaused = false`

rhodey:
the basics of this seem to be that we need to initialize a list once and force refresh a list `onResume()` but not immediately after `onCreate()`-- I can't see another way of doing that without tracking `onPause()`.

rhodey:
modified to include the `onPause()` tracking boolean rebased and pushed.

agrajaghh:
isn't it enough to set `hasPaused = true;` at the end of `onResume()`? So `handleRefreshAdapter()` won't be called the first time, but the second time (and afterwards) and you don't have to override `onPause()`... Perhaps it would be more clear if you rename the variable to something like `isCreated`?

PR #2556, opened 2015-03-03
message timestamps update onResume()

### LINKED ISSUE #2519: Time stamps very slow to update

rhodey:
Modified conversation view and conversation list view to update message timestamps onResume.

The basics of this is that `LoaderManager.initLoader()` is now called `onResume()` instead of `onCreate()` and `LoaderManager.destroyLoader()` is now called `onPause()` so that a new loader will be created `onResume()` which will force the timestamps to update.

The null check and `-1` check on `this.recipients` and `this.threadId` is to handle the case of a new conversation thread where the list of messages would be empty-- in this case no `ListAdapter` is set so these checks prevent an NPE within two of the `LoaderManager` callbacks.

Fixes #2519

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -91,6 +91,18 @@ public class ConversationFragment extends ListFragment
     getLoaderManager().restartLoader(0, null, this);
   }
 
+  @Override
+  public void onResume() {
+    super.onResume();
+    getLoaderManager().initLoader(0, null, this);
+  }
+
+  @Override
+  public void onPause() {
+    super.onPause();
+    getLoaderManager().destroyLoader(0);
+  }
+
   private void initializeResources() {
     this.masterSecret = this.getActivity().getIntent().getParcelableExtra("master_secret");
     this.recipients   = RecipientFactory.getRecipientsForIds(getActivity(), getActivity().getIntent().getLongArrayExtra("recipients"), true);
@@ -103,7 +115,6 @@ public class ConversationFragment extends ListFragment
                                                   (!this.recipients.isSingleRecipient()) || this.recipients.isGroupRecipient(),
                                                   DirectoryHelper.isPushDestination(getActivity(), this.recipients)));
       getListView().setRecyclerListener((ConversationAdapter)getListAdapter());
-      getLoaderManager().initLoader(0, null, this);
     }
   }
 
@@ -159,6 +170,7 @@ public class ConversationFragment extends ListFragment
     this.threadId   = threadId;
 
     initializeListAdapter();
+    getLoaderManager().restartLoader(0, null, this);
   }
 
   public void scrollToBottom() {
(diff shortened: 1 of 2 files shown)


rhodey:
In working on this fix I think I also discovered an extra call to `ConversationListFragment.initializeListAdapter()`

https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/ConversationListFragment.java#L146

`ConversationListFragment.setMasterSecret()` is only called by `ConversationListActivity.initializeResources()` which is called inside of the activity's `onCreate()` method.

https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/ConversationListActivity.java#L256

So `ConversationListFragment.initializeListAdapter()` is called in `onCreate()` of both the activity and the `onActivityCreated()` of the fragment-- we should be able to drop `initializeListAdapter()` from `ConversationListFragment.setMasterSecret()`

rhodey:
dropped extra call to `initializeListAdapter()` within `ConversationListFragment.setMasterSecret()`, tested, and rebased.

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -159,6 +170,7 @@ public void reload(Recipients recipients, long threadId) {
     this.threadId   = threadId;
 
     initializeListAdapter();
+    getLoaderManager().restartLoader(0, null, this);
```
moxie0: why is this necessary to do on every message sent?

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -91,6 +91,18 @@ public void onNewIntent() {
...
+  @Override
+  public void onResume() {
+    super.onResume();
+    getLoaderManager().initLoader(0, null, this);
+  }
+
+  @Override
+  public void onPause() {
+    super.onPause();
+    getLoaderManager().destroyLoader(0);
```
mcginty: Wouldn't we only have to destroyLoader when the threadId changes, so on a new intent or when we send our first message to a user? If not, we should just call restartLoader in onResume instead of both initLoader and destroyLoader.

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -268,12 +280,16 @@ public void onFailure(Throwable error) {
 
   @Override
   public void onLoadFinished(Loader<Cursor> arg0, Cursor cursor) {
-    ((CursorAdapter)getListAdapter()).changeCursor(cursor);
+    if (this.recipients != null && this.threadId != -1) {
```
rhodey: These checks are to prevent an NPE on `getListAdapter()` which is not set in the case that this fragment is for a new message (identifiable by null recipients and threadId of -1). Alternatively could just check that `getListAdapter() != null`, not sure which is better style / more understandable.

src/org/thoughtcrime/securesms/ConversationFragment.java
rhodey: `initLoader()` should be called before any other `LoaderManager` method because it handles some needed initialization but in addition to this it also creates a new loader. So if we placed `restartLoader()` in `onResume()` we'd have to place `initLoader()` in either `onCreate()` or `onResume()` and this would result in one loader being needlessly created.

src/org/thoughtcrime/securesms/ConversationFragment.java
rhodey: thanks, this should be changed to restart the loader only if the `threadId` changes which handles the case where the fragment represented a new conversation and because of this didn't create a loader.

src/org/thoughtcrime/securesms/ConversationFragment.java
rhodey: Also, if we agree that `initLoader()` should happen in `onResume()` to avoid wasting a loader then `destroyLoader()` must happen in `onPause()` as above so that the next call to `initLoader()` in `onResume()` will notice that it's loader is gone, create a new one, and ultimately update the timestamp.

src/org/thoughtcrime/securesms/ConversationFragment.java
rhodey: actually, I'll need to check this. my reasoning for this was that previous to these changes the entire list adapter and loader were reinitialized after every message send.

https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/ConversationFragment.java#L161

PR #2567, opened 2015-03-03
make some strings translateable

agrajaghh:


res/values/strings.xml
```diff
@@ -853,6 +861,11 @@
     <!-- conversation_list -->
     <string name="conversation_list__menu_search">Search</string>
 
+    <!-- conversation_list_item_view -->
+    <string name="conversation_list_item_view__contact_photo_image">Contact Photo Image</string>
+    <string name="conversation_list_item_view__attachment_indicator">Attachment Indicator</string>
+    <string name="conversation_list_item_view__error_alert">Error Alert</string>
```
mcginty: Even though it was like that in the original, want to lower-case these to "Error alert" for example?

mcginty:
This is awesome, thanks so much. To full i18n and beyond!

agrajaghh:
made "Alert" lowercase. All strings except stuff like "+" or "•" should be translateable now :smiley:

PR #2568, opened 2015-03-03
settings preference group summaries

### LINKED ISSUE #2395: Settings menu subheadings copy
This issue arrises from an irc conversation, wherein one of our users failed to find the screen security setting under App Protection due to that section's binary subheading of "On" or "Off".
```
11:53 < redacted> it is extremely non-discoverable, because from the top level settings menu, App Protection looks like a boolean toggle
11:53 < redacted> it says "App protection" (greyed out: "off")
11:53 < redacted> I really didn't expect a sub-menu there!
11:54 < redacted> also, funny that it says "off" when in fact something in that submenu is on
```
Seems like a few of the subheadings in the settings menu could use some polish. Suggestions:
* SMS and MMS
      - Drop the colons, i.e., "Incoming [on|off], outgoing [on|off]"
      - Add delivery reports status to make it more discoverable?
* App Protection
      - Add screen security, i.e., "Passphrase [on|off], screen security [on|off]"
* Appearance
      - Could also drop one or both colons here, i.e., "[Dark|Light] theme, [Language]"

rhodey:
1) created a new string constant for *Passphrase*.
2) modified *App protection* preference group summary to be more verbose.
3) dropped the colons from *Appearance* preference group summary.
4) dropped the colons from and *SMS and MMS* preference group summary.

turns out we didn't have a *Passphrase* string in `strings.xml`, the closest match I found was *Enable passphrase* but the *App protection* preferences group summary doesn't read as nicely, eg...

*Enable passphrase on, Screen security on*
vs
*Passphrase on, Screen security on*

Could revert to using the existing *Enable passphrase* string constant if we really wanna avoid adding a new one, lemme know.

Fixes #2395

### DIFF
--- a/src/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
+++ b/src/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
@@ -153,21 +153,25 @@ public class AppProtectionPreferenceFragment extends PreferenceFragment {
   }
 
   private static CharSequence getPassphraseSummary(Context context) {
-    final int passphraseResId = R.string.preferences__passphrase;
-    final int onResId         = R.string.ApplicationPreferencesActivity_on;
-    final int offResId        = R.string.ApplicationPreferencesActivity_off;
+    final int    passphraseResId = R.string.preferences__passphrase_summary;
+    final String onRes           = context.getString(R.string.ApplicationPreferencesActivity_on);
+    final String offRes          = context.getString(R.string.ApplicationPreferencesActivity_off);
 
-    return context.getString(passphraseResId) + " " +
-           context.getString((TextSecurePreferences.isPasswordDisabled(context) ? offResId : onResId));
+    if (TextSecurePreferences.isPasswordDisabled(context))
+      return context.getString(passphraseResId, offRes);
+
+    return context.getString(passphraseResId, onRes);
   }
 
   private static CharSequence getScreenSecuritySummary(Context context) {
-    final int screenSecurityResId = R.string.preferences__screen_security;
-    final int onResId             = R.string.ApplicationPreferencesActivity_on;
-    final int offResId            = R.string.ApplicationPreferencesActivity_off;
+    final int    screenSecurityResId = R.string.preferences__screen_security_summary;
+    final String onRes               = context.getString(R.string.ApplicationPreferencesActivity_on);
+    final String offRes              = context.getString(R.string.ApplicationPreferencesActivity_off);
 
-    return context.getString(screenSecurityResId) + " " +
-           context.getString((TextSecurePreferences.isScreenSecurityEnabled(context) ? onResId : offResId));
+    if (TextSecurePreferences.isScreenSecurityEnabled(context))
+      return context.getString(screenSecurityResId, onRes);
+
+    return context.getString(screenSecurityResId, offRes);
   }
 
   public static CharSequence getSummary(Context context) {
--- a/src/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
+++ b/src/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
@@ -157,10 +157,11 @@ public class AppProtectionPreferenceFragment extends PreferenceFragment {
     final String onRes           = context.getString(R.string.ApplicationPreferencesActivity_on);
     final String offRes          = context.getString(R.string.ApplicationPreferencesActivity_off);
 
-    if (TextSecurePreferences.isPasswordDisabled(context))
+    if (TextSecurePreferences.isPasswordDisabled(context)) {
       return context.getString(passphraseResId, offRes);
-
-    return context.getString(passphraseResId, onRes);
+    } else {
+      return context.getString(passphraseResId, onRes);
+    }
   }
 
   private static CharSequence getScreenSecuritySummary(Context context) {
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
```diff
@@ -153,10 +152,25 @@ public void onClick(DialogInterface dialog, int which) {
     }
   }
 
-  public static CharSequence getSummary(Context context) {
-    final int onCapsResId  = R.string.ApplicationPreferencesActivity_On;
-    final int offCapsResId = R.string.ApplicationPreferencesActivity_Off;
+  private static CharSequence getPassphraseSummary(Context context) {
+    final int passphraseResId = R.string.preferences__passphrase;
```
mcginty: we probably don't need a variable for this one - with Android Studio now "autocompleting" string resources it's actually more readable to directly reference R.string vars.

agrajaghh: @mcginty just a general question to clarify: should the lines then be wrapped after 100 characters while the line is "collapsed" in Android Studio? I think I put all these strings in variables because the lines would have been ultra long otherwise...

mcginty: Yeah, I can understand that since the ternary operator pushed it to be a pretty long line. Personally, I don't mind it when R.* variables push the line length a bit longer because of the auto-collapsing, but maybe @moxie0 can chime in if he has an opinion on it.

moxie0: i don't collapse them in my environment, so it matters to me

On 03/03/2015 12:20 PM, Jake McGinty wrote:
> In
> src/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
> <https://github.com/WhisperSystems/TextSecure/pull/2568#discussion_r25722771>:
> 
>> @@ -153,10 +152,25 @@ public void onClick(DialogInterface dialog, int which) {
>>      }
>>    }
>>  
>> -  public static CharSequence getSummary(Context context) {
>> -    final int onCapsResId  = R.string.ApplicationPreferencesActivity_On;
>> -    final int offCapsResId = R.string.ApplicationPreferencesActivity_Off;
>> +  private static CharSequence getPassphraseSummary(Context context) {
>> +    final int passphraseResId = R.string.preferences__passphrase;
> 
> Yeah, I can understand that since the ternary operator pushed it to be a
> pretty long line. Personally, I don't mind it when R.* variables push
> the line length a bit longer because of the auto-collapsing, but maybe
> @moxie0 <https://github.com/moxie0> can chime in if he has an opinion on it.
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/WhisperSystems/TextSecure/pull/2568/files#r25722771>.
> 

-- 
http://www.thoughtcrime.org

rhodey: yeah, felt kinda silly about it but just following convention in

https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/preferences/SmsMmsPreferenceFragment.java#L144

and

https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/preferences/AppearancePreferenceFragment.java#L44

etc

moxie0:
will the way you've done this translate to RTL languages? don't you need to use string substitution here?

rhodey:
I don't believe that this or what's on master translates correctly to RTL languages. I can try to fix that in this pull but I think we'd have to do something like [this](http://stackoverflow.com/questions/18996183/identifyng-rtl-language-in-android) string substitution.

For the string substitution method we'd need a set of strings like
`<string name="preferences__passphrase_summary">Passphrase %s</string>`
And the RTL translators would need to swap the location of the `%s`

rhodey:
[this](http://stackoverflow.com/questions/15746091/how-to-handle-rtl-languages-on-pre-4-2-versions-of-android) StackOverflow user seems to suggest that many carriers attempt to handle RTL via custom ROM and that each do this differently, could potentially undo any corrections we attempt via string substitution and can't imagine any easy ways of testing for this.

moxie0:
i think you should at least use substitution instead of appending your internal concats, even for non-rtl language support.

rhodey:
@moxie0 OK, [AppearancePreferenceFragment](https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/preferences/AppearancePreferenceFragment.java#L53) and [SmsMmsPreferenceFragment](https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/preferences/SmsMmsPreferenceFragment.java#L167) do the concat as well, I'll switch both these over since this issue requests that they already be modified to drop the `:`

src/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
```diff
@@ -153,10 +152,29 @@ public void onClick(DialogInterface dialog, int which) {
...
-    final int offCapsResId = R.string.ApplicationPreferencesActivity_Off;
+  private static CharSequence getPassphraseSummary(Context context) {
+    final int    passphraseResId = R.string.preferences__passphrase_summary;
+    final String onRes           = context.getString(R.string.ApplicationPreferencesActivity_on);
+    final String offRes          = context.getString(R.string.ApplicationPreferencesActivity_off);
+
+    if (TextSecurePreferences.isPasswordDisabled(context))
+      return context.getString(passphraseResId, offRes);
+
+    return context.getString(passphraseResId, onRes);
```
moxie0: use if else or the ternary

src/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
```diff
@@ -153,10 +152,29 @@ public void onClick(DialogInterface dialog, int which) {
...
+  private static CharSequence getScreenSecuritySummary(Context context) {
+    final int    screenSecurityResId = R.string.preferences__screen_security_summary;
+    final String onRes               = context.getString(R.string.ApplicationPreferencesActivity_on);
+    final String offRes              = context.getString(R.string.ApplicationPreferencesActivity_off);
 
-    return context.getString(TextSecurePreferences.isPasswordDisabled(context) ? offCapsResId : onCapsResId);
+    if (TextSecurePreferences.isScreenSecurityEnabled(context))
+      return context.getString(screenSecurityResId, onRes);
+
+    return context.getString(screenSecurityResId, offRes);
```
moxie0: use if else or the ternary

PR #2569, opened 2015-03-03
remove ForegroundImageView

agrajaghh:
`ForegroundImageView` got replaced with `com.makeramen.RoundedImageView` in https://github.com/WhisperSystems/TextSecure/commit/62816ee51ac80b446c788ec50bbd94d967ebddbb so its no longer needed


moxie0:
thanks but it looks like @mcginty is actually using this in a change he's working on for thumbnails right now, so let's leave it in

PR #2570, opened 2015-03-03
Update to latest version of libtextsecure for simplified interface.

moxie0:
1) Switch to new TextSecureAddress addressing, rather than mixing
   long-based recipient IDs into libtextsecure.

2) Get rid of RecipientFormattingException throws in calls to
   RecipientFactory.

### DIFF
--- a/src/org/thoughtcrime/securesms/RoutingActivity.java
+++ b/src/org/thoughtcrime/securesms/RoutingActivity.java
@@ -198,12 +198,12 @@ public class RoutingActivity extends PassphraseRequiredActionBarActivity {
     Recipients recipients;
     String body     = getIntent().getStringExtra("sms_body");
     long   threadId = getIntent().getLongExtra("thread_id", -1);
+    String data     = getIntent().getData().getSchemeSpecificPart();
 
-    try {
-      String data = getIntent().getData().getSchemeSpecificPart();
+    if (data != null) {
       recipients = RecipientFactory.getRecipientsFromString(this, data, false);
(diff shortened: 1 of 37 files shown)


src/org/thoughtcrime/securesms/RoutingActivity.java
```diff
@@ -198,12 +198,12 @@ private ConversationParameters getConversationParametersForSendAction() {
     Recipients recipients;
     String body     = getIntent().getStringExtra("sms_body");
     long   threadId = getIntent().getLongExtra("thread_id", -1);
+    String data     = getIntent().getData().getSchemeSpecificPart();
```
mcginty: getData() itself can be null according to android docs, this seems like it could be a NPE opportunity

moxie0: then it is in the current code too, right?  i was just trying to move the equivalent of the recipient == null check out of the getRecipientsFromString method and into here. i can add this null check as well if you'd like

mcginty: Yeah, might as well while we're in the file. I don't think we've seen any NPEs around it, but it seems like it could happen.

mcginty:
lgtm :+1: incredibly satisfying to see all those try-catch blocks disappear.

moxie0:
ok added the check, let me know if that works for you

mcginty:
looks good!

moxie0:
yep, just hope we weren't passing null in there constantly somehow =)

PR #2572, opened 2015-03-04
update message timestamps onResume

### LINKED ISSUE #2519: Time stamps very slow to update

rhodey:
Modified ConversationFragment and ConversationListFragment to update message timestamps on 'onResume'. Improved performance through more effecient use of loaders and 'ConversationAdapter.notifyDataSetChanged()'.

Fixes #2519

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -60,7 +60,6 @@ public class ConversationFragment extends ListFragment
   private Recipients   recipients;
   private long         threadId;
   private ActionMode   actionMode;
-  private boolean      isListRefreshNeeded = false;
 
   @Override
   public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle bundle) {
@@ -95,16 +94,9 @@ public class ConversationFragment extends ListFragment
   public void onResume() {
     super.onResume();
 
-    if (getListAdapter() != null && isListRefreshNeeded) {
+    if (getListAdapter() != null) {
       ((ConversationAdapter) getListAdapter()).notifyDataSetChanged();
     }
-    isListRefreshNeeded = false;
-  }
-
-  @Override
-  public void onPause() {
-    super.onPause();
-    isListRefreshNeeded = true;
   }
 
   private void initializeResources() {
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -60,6 +60,7 @@ public class ConversationFragment extends ListFragment
   private Recipients   recipients;
   private long         threadId;
   private ActionMode   actionMode;
+  private boolean      isListRefreshNeeded = false;
 
   @Override
   public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle bundle) {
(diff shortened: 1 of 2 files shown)


rhodey:
turns out calling `ConversationAdapter.notifyDataSetChanged()` is not equivalent to `LoaderManager.restartLoader()` and is in-fact more efficient. `notifyDataSetChanged()` rebuilds the visible list item views using the data already available to the adapter, where a call to `restartLoader()` creates a new loader and a new cursor, reads from that cursor, and then rebuilds the list item view.

because the data within the cursor isn't changing we can use the `LoaderManager` more efficiently.

ultimately this pull results in TS using one `Loader` per ConversationFragment or ConversationListFragment and rebuilding of the visible list item views every `onResume()` except the one immediately following `onCreate()`.

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -143,7 +155,6 @@ public void onListItemClick(ListView l, View v, int position, long id) {
   public void setMasterSecret(MasterSecret masterSecret) {
     if (this.masterSecret != masterSecret) {
       this.masterSecret = masterSecret;
-      initializeListAdapter();
```
rhodey: not needed, `setMasterSecret()` is only called from within `ConversationListActivity.onCreate()` and `initializeListAdapter()` already happens within `onActivityCreated()`.

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -100,8 +101,6 @@ public void onClick(View v) {
     });
     initializeListAdapter();
     initializeBatchListener();
-
-    getLoaderManager().initLoader(0, null, this);
```
rhodey: already taken care of inside `initializeListAdapter()`

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -113,12 +129,9 @@ private void initializeContextualActionBar() {
   }
 
   private void setCorrectMenuVisibility(Menu menu) {
-    ConversationAdapter adapter        = (ConversationAdapter) getListAdapter();
     List<MessageRecord> messageRecords = getSelectedMessageRecords();
 
     if (actionMode != null && messageRecords.size() == 0) {
-      adapter.getBatchSelected().clear();
-      adapter.notifyDataSetChanged();
       actionMode.finish();
```
rhodey: already taken care of in `onDestroyActionMode()` callback which should come from `actionMode.finish()`

moxie0:
no "isListRefreshNeeded" please

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -155,10 +160,14 @@ private MessageRecord getSelectedMessageRecord() {
...
   public void reload(Recipients recipients, long threadId) {
+    boolean threadIdChanged = this.threadId != threadId;
+
     this.recipients = recipients;
     this.threadId   = threadId;
 
-    initializeListAdapter();
+    if (threadIdChanged) {
+      initializeListAdapter();
+    }
```
moxie0: why do this, instead of leaving restartLoader in onNewIntent and initLoader in initializeListAdapter?  Shouldn't this change just be the new 3 lines in onResume()?

rhodey: In the case that you start a new conversation the list adapter and loader are never initialized because there is no message history so no need for them.

https://github.com/rhodey/TextSecure/blob/issue-2519/src/org/thoughtcrime/securesms/ConversationFragment.java#L109

If you decide to send a message in this new conversation then the adapter and loader are needed, watching for the threadId to change from `-1` is the most straightforward way of doing this.

moxie0: I'm looking for the minimum diff fix for this issue.  If you're making changes to address another issue, please open a separate issue for that to be assigned to another milestone.

PR #2574, opened 2015-03-04
fix icons in import/export dialog

agrajaghh:
There were some issues with the icons in the import/export dialog:
* the `Import plaintext backup" icon had the same pixel size in the mdpi, hdpi, xhdpi and xxhdpi folder
* the dark icons were 100% black

I created the icons again (using the original names from the material repo) with 87% black and made the icons a bit bigger, they seemed a bit too small to me.

screenshot_2015-03-04-12-04-02


mcginty:
lgtm!

PR #2590, opened 2015-03-05
show image preview inside conversation list item views

### LINKED ISSUE #1934: Show image hint instead of (no subject) for media/mms

rhodey:
1) removed two unused ImageViews from conversation_list_item_view
2) created a new view for thread-level media preview (conversation_list_item_with_media_preview.xml)
3) added two new columns ('snippet_part_uri' and 'snippet_part_type') to ThreadDatabase's 'thread' table
4) added appropriate migration to DatabaseFactory
5) modified ThreadRecord to contain a ListenableFutureTask<Slide> and ThreadDatabase to populate it
6) updated ConversationActivity's draft saving logic to work with ThreadDatabase changes
7) modified ConversationListItem, ConversationListAdapter, and ConversationListFragment to work with the new conversation_list_item_with_media_preview.xml view
8) updated ShareListAdapter to work with the new ThreadDatabase.readerFor() method
9) minor convenience changes to ImageSlide and DraftDatabase
10) bumped versionCode to trigger database upgrade

Fixes #1934

@mcginty has some major ImageSlide refactoring in the works, some of which this pull could benefit from, for now I'll just put this up here though.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           xmlns:tools="http://schemas.android.com/tools"
           package="org.thoughtcrime.securesms"
-      android:versionCode="95"
+      android:versionCode="94"
       android:versionName="2.5.3">
 
     <permission android:name="org.thoughtcrime.securesms.ACCESS_SECRETS"
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           xmlns:tools="http://schemas.android.com/tools"
           package="org.thoughtcrime.securesms"
-      android:versionCode="94"
+      android:versionCode="95"
       android:versionName="2.5.3">
 
     <permission android:name="org.thoughtcrime.securesms.ACCESS_SECRETS"
--- a/res/layout/conversation_list_item_view.xml
+++ b/res/layout/conversation_list_item_view.xml
@@ -59,6 +59,22 @@
                       tools:text="Jules Bonnot"
                       android:ellipsize="marquee" />
 
+            <ImageView android:id="@+id/attachment"
+                       android:layout_marginLeft="3dip"
+                       android:layout_height="match_parent"
+                       android:layout_width="20sp"
+                       android:visibility="gone"
+                       android:src="@drawable/ic_attachment_universal_small"
+                       android:contentDescription="Attachment Indicator" />
+
+            <ImageView android:id="@+id/error"
+                       android:layout_marginLeft="3dip"
+                       android:layout_height="match_parent"
+                       android:layout_width="20sp"
+                       android:visibility="gone"
+                       android:src="@drawable/ic_action_warning_red"
+                       android:contentDescription="Error Alert" />
+
             <TextView android:id="@+id/date"
                       android:layout_marginLeft="3dip"
                       android:layout_height="wrap_content"
--- a/res/layout/conversation_list_item_view.xml
+++ b/res/layout/conversation_list_item_view.xml
@@ -59,22 +59,6 @@
                       tools:text="Jules Bonnot"
                       android:ellipsize="marquee" />
 
-            <ImageView android:id="@+id/attachment"
-                       android:layout_marginLeft="3dip"
-                       android:layout_height="match_parent"
-                       android:layout_width="20sp"
-                       android:visibility="gone"
-                       android:src="@drawable/ic_attachment_universal_small"
-                       android:contentDescription="Attachment Indicator" />
-
-            <ImageView android:id="@+id/error"
-                       android:layout_marginLeft="3dip"
-                       android:layout_height="match_parent"
-                       android:layout_width="20sp"
-                       android:visibility="gone"
-                       android:src="@drawable/ic_action_warning_red"
-                       android:contentDescription="Error Alert" />
-
             <TextView android:id="@+id/date"
                       android:layout_marginLeft="3dip"
                       android:layout_height="wrap_content"
--- a/res/layout/conversation_list_item_with_media_preview_view.xml
+++ b/res/layout/conversation_list_item_with_media_preview_view.xml
@@ -1,0 +1,99 @@
+<?xml version="1.0" encoding="utf-8"?>
+<org.thoughtcrime.securesms.ConversationListItem
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="70dp">
+
+    <FrameLayout android:id="@+id/contact_photo_frame"
+                 android:layout_width="wrap_content"
+                 android:layout_height="wrap_content"
+                 android:layout_centerVertical="true"
+                 android:layout_alignParentLeft="true"
+                 android:paddingTop="3dp"
+                 android:paddingBottom="3dp"
+                 android:paddingLeft="10dp"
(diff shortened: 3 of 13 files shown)


AndroidManifest.xml
```diff
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           xmlns:tools="http://schemas.android.com/tools"
           package="org.thoughtcrime.securesms"
-      android:versionCode="94"
+      android:versionCode="95"
```
moxie0: don't do this

res/layout/conversation_list_item_view.xml
```diff
@@ -59,22 +59,6 @@
...
-                       android:contentDescription="Attachment Indicator" />
-
-            <ImageView android:id="@+id/error"
-                       android:layout_marginLeft="3dip"
-                       android:layout_height="match_parent"
-                       android:layout_width="20sp"
-                       android:visibility="gone"
-                       android:src="@drawable/ic_action_warning_red"
-                       android:contentDescription="Error Alert" />
-
```
moxie0: please only include changes relevant to the issue you're addressing

res/layout/conversation_list_item_with_media_preview_view.xml
```diff
@@ -0,0 +1,99 @@
...
+        </LinearLayout>
+
+        <ImageView android:id="@+id/media_preview"
+                   android:layout_width="70dp"
+                   android:layout_height="match_parent"
+                   android:layout_marginLeft="4dp"
+                   android:gravity="end"
+                   android:src="@drawable/ic_attach_picture_holo_light"
+                   android:scaleType="centerCrop"/>
+
```
moxie0: Why is there a whole separate layout for what looks like the addition of a single element?  Is this otherwise subtly different, or just a copy/paste of the standard layout?

src/org/thoughtcrime/securesms/ConversationListAdapter.java
```diff
@@ -41,40 +42,82 @@
...
+        throw new IllegalArgumentException("unsupported item view type given to ConversationListAdapter");
+    }
   }
 
   @Override
   public void bindView(View view, Context context, Cursor cursor) {
-    if (masterCipher != null) {
-      ThreadDatabase.Reader reader = threadDatabase.readerFor(cursor, masterCipher);
+    if (masterSecret != null && masterCipher != null) {
+      ThreadDatabase.Reader reader = threadDatabase.readerFor(cursor, masterSecret, masterCipher);
```
moxie0: intuitively nothing should really need both a mastersecret and a mastercipher, just one or the other

src/org/thoughtcrime/securesms/ConversationListAdapter.java
```diff
@@ -41,40 +42,82 @@
...
+    }
+
+    return VIEW_TYPE_BASE;
+  }
+
+  @Override
+  public int getItemViewType(int position) {
+    Cursor cursor = (Cursor)getItem(position);
+    return getItemViewType(cursor);
+  }
```
moxie0: seems like all of this could be done with a single view, which you selectively show the imageview for in the bind call?

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -189,11 +155,54 @@ private void initializeReminders() {
...
+        adapter.initializeBatchMode(true);
+        adapter.toggleThreadInBatchSet(((ConversationListItem) view).getThreadId());
+        adapter.notifyDataSetChanged();
+        return true;
+      }
+
+      return false;
+    }
+  }
+
```
moxie0: i'm not following, why did all of this need to change?

src/org/thoughtcrime/securesms/ConversationListItem.java
```diff
@@ -161,4 +224,56 @@ public void run() {
...
+        builder.setNegativeButton(R.string.no, null);
+        builder.show();
+      }
+    }
+
+    @Override
+    public boolean onLongClick(View v) {
+      return clickListener.onItemLongClick(null, ConversationListItem.this, -1, -1);
+    }
+  }
```
moxie0: i don't think all this should be in here, showing the thumbnail is probably enough

src/org/thoughtcrime/securesms/database/DatabaseFactory.java
```diff
@@ -717,6 +718,11 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
         db.execSQL("ALTER TABLE mms ADD COLUMN network_failures TEXT");
       }
 
+      if (oldVersion < INTRODUCED_SNIPPET_PART_VERSION) {
+        db.execSQL("ALTER TABLE thread ADD COLUMN snippet_part_uri TEXT");
+        db.execSQL("ALTER TABLE thread ADD COLUMN snippet_part_type TEXT");
```
moxie0: what's the type for? do parts not already contain this information?

src/org/thoughtcrime/securesms/database/DraftDatabase.java
```diff
@@ -153,5 +153,19 @@ public String getSnippet(Context context) {
...
+        return snippetSlideDraft;
+      }
+
+      snippetSlideDraft = getDraftOfType(Draft.VIDEO);
+      if (snippetSlideDraft != null) {
+        return snippetSlideDraft;
+      }
+
+      return getDraftOfType(Draft.AUDIO);
+    }
```
moxie0: maybe just check for != text

res/layout/conversation_list_item_with_media_preview_view.xml
rhodey: this amounts to one additional ImageView but the layout required to make that look nice includes quiet a few changes. also this is fundamentally a different view, right? all the different message types (incomming, outgoing, etc) have separate view layout resources. keeping the two views separate also saves processing during layout inflation in the case where the ImageView is not needed.

moxie0: incoming/outgoing views are different because it is impossible to make them shared.  we don't have a different view for messages with delivery receipt indicators and messages without delivery receipt indicators visible.

src/org/thoughtcrime/securesms/ConversationListItem.java
rhodey: this is required to prevent batch selection mode from acting weird on conversation list items that have the image preview. the changes to `ConversationListFragment` you asked about above are a prerequisite to this.

moxie0: acting weird how? get rid of the tap-to-display-image and it should be unaffected, no?

src/org/thoughtcrime/securesms/database/DatabaseFactory.java
rhodey: media from drafts is not saved into the PartDatabase, it is stored as a URI inside of the DraftDatabase value column. we can't do conversation list media preview for drafts without the `snippet_part_type` column.

src/org/thoughtcrime/securesms/database/DraftDatabase.java
rhodey: this is more future proof though, right? If someone else was to add another draft type and not bother to check this method then it'd possibly break.

src/org/thoughtcrime/securesms/ConversationListFragment.java
rhodey: see below

src/org/thoughtcrime/securesms/ConversationListItem.java
rhodey: oh, gotcha, didn't understand that you wanted to drop tap to display image. I like that idea.

rhodey:
needs some serious refactoring due to recent updates, will open new pull when done.

PR #2591, opened 2015-03-05
update message timestamps onResume

### LINKED ISSUE #2519: Time stamps very slow to update

rhodey:
modified ConversationFragment and ConversationListFragment to update message timesamps on resume.

Fixes #2519.
// FREEBIE


rhodey:
forgive the branch name error, opening new pull

PR #2592, opened 2015-03-06
fix LoaderManager misuse and prevent wasteful view redraws

rhodey:
modified ConversationFragment and ConversationListFragment to use the LoaderManager more efficiently. dropped wasteful call to notifyDataSetChanged() in ConversationFragment.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -103,7 +103,6 @@ public class ConversationFragment extends ListFragment
 
     initializeResources();
     initializeListAdapter();
-    getLoaderManager().restartLoader(0, null, this);
   }
 
   private void initializeResources() {
@@ -117,7 +116,7 @@ public class ConversationFragment extends ListFragment
                                                   (!this.recipients.isSingleRecipient()) || this.recipients.isGroupRecipient(),
                                                   DirectoryHelper.isPushDestination(getActivity(), this.recipients)));
       getListView().setRecyclerListener((ConversationAdapter)getListAdapter());
-      getLoaderManager().initLoader(0, null, this);
+      getLoaderManager().restartLoader(0, null, this);
     }
   }
 
@@ -127,12 +126,9 @@ public class ConversationFragment extends ListFragment
   }
 
   private void setCorrectMenuVisibility(Menu menu) {
-    ConversationAdapter adapter        = (ConversationAdapter) getListAdapter();
     List<MessageRecord> messageRecords = getSelectedMessageRecords();
 
     if (actionMode != null && messageRecords.size() == 0) {
-      adapter.getBatchSelected().clear();
-      adapter.notifyDataSetChanged();
       actionMode.finish();
       return;
     }
@@ -169,10 +165,14 @@ public class ConversationFragment extends ListFragment
   }
 
   public void reload(Recipients recipients, long threadId) {
+    boolean threadIdChanged = this.threadId != threadId;
+
     this.recipients = recipients;
     this.threadId   = threadId;
 
-    initializeListAdapter();
+    if (threadIdChanged) {
+      initializeListAdapter();
+    }
   }
 
   public void scrollToBottom() {
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -165,12 +165,10 @@ public class ConversationFragment extends ListFragment
   }
 
   public void reload(Recipients recipients, long threadId) {
-    boolean threadIdChanged = this.threadId != threadId;
-
     this.recipients = recipients;
-    this.threadId   = threadId;
 
-    if (threadIdChanged) {
+    if (this.threadId != threadId) {
+      this.threadId = threadId;
       initializeListAdapter();
     }
   }
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -96,8 +96,6 @@ public class ConversationListFragment extends ListFragment
     });
     initializeListAdapter();
     initializeBatchListener();
-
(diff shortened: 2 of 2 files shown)


src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -103,7 +102,7 @@ private void initializeListAdapter() {
                                                   (!this.recipients.isSingleRecipient()) || this.recipients.isGroupRecipient(),
                                                   DirectoryHelper.isPushDestination(getActivity(), this.recipients)));
       getListView().setRecyclerListener((ConversationAdapter)getListAdapter());
-      getLoaderManager().initLoader(0, null, this);
+      getLoaderManager().restartLoader(0, null, this);
```
rhodey: @mcginty and I confirmed that `restartLoader` is more efficient and in our use case functionally equivalent.

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -88,7 +88,6 @@ public void onNewIntent() {
 
     initializeResources();
     initializeListAdapter();
-    getLoaderManager().restartLoader(0, null, this);
```
rhodey: no need to call restart loader right after previously `initLoader` or currently `restartLoader`

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -155,10 +151,14 @@ private MessageRecord getSelectedMessageRecord() {
   }
 
   public void reload(Recipients recipients, long threadId) {
+    boolean threadIdChanged = this.threadId != threadId;
+
     this.recipients = recipients;
     this.threadId   = threadId;
 
-    initializeListAdapter();
+    if (threadIdChanged) {
```
rhodey: no need to reinitialize the entire list adapter and loader every time we send a message, only in the case that it is the first message in a thread, identifiable by `threadId` changing from `-1` to something else.

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -100,8 +100,6 @@ public void onClick(View v) {
     });
     initializeListAdapter();
     initializeBatchListener();
-
-    getLoaderManager().initLoader(0, null, this);
```
rhodey: no need to call `initLoader()` right after is has been reset by `initializeListAdapter()`

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -143,7 +141,6 @@ public void onListItemClick(ListView l, View v, int position, long id) {
   public void setMasterSecret(MasterSecret masterSecret) {
     if (this.masterSecret != masterSecret) {
       this.masterSecret = masterSecret;
-      initializeListAdapter();
```
rhodey: no need to reinitialize the adapter and loader on `setMasterSecret()`, this method is only called during create of `ConversationListActivity` and because `initializeListAdapter()` already happens on create of `ConversationListFragment` it is wasteful.

src/org/thoughtcrime/securesms/ConversationFragment.java
mcginty: reads nicer to me if you just have the `this.threadId != threadId` in conditional, not as a variable

mcginty:
the rest looks good to me, haven't tested

src/org/thoughtcrime/securesms/ConversationFragment.java
rhodey: `initializeListAdapter()` uses `this.threadId` for some conditional stuff, so without the boolean the if would have to look something like

```
if (this.threadId != threadId) {
  this.recipients = recipients;
  this.threadId = threadId;
  initializeListAdapter()
} else {
  this.recipients = recipients;
  this.threadId   = threadId;
}

which I thought looked less nice

File: src/org/thoughtcrime/securesms/ConversationListFragment.java (hunk 5 above)
Comment (rhodey) [inline src/org/thoughtcrime/securesms/ConversationListFragment.java]: should be fine dropping this line as long as the fragment never resumes before the parent activity's onCreate is finished

**Comment by mcginty** [comment]
This is probably also a fix for https://github.com/WhisperSystems/TextSecure/issues/2589 if it's still occurring, as my guess would be `onNewIntent()`, and therefore `initLoader`, was being called before `initializeListAdapter()`.

File: src/org/thoughtcrime/securesms/ConversationFragment.java [hunk 6]
```diff
@@ -113,12 +112,9 @@ private void initializeContextualActionBar() {
   }
 
   private void setCorrectMenuVisibility(Menu menu) {
-    ConversationAdapter adapter        = (ConversationAdapter) getListAdapter();
     List<MessageRecord> messageRecords = getSelectedMessageRecords();
 
     if (actionMode != null && messageRecords.size() == 0) {
-      adapter.getBatchSelected().clear();
```
rhodey: already set to happen in the action mode finish callback

rhodey:
@mcginty good catch, I agree. The only thing I'm uncertain of now is dropping the `ConversationListFragment.initializeListAdapter()` call in `ConversationListFragment.setMasterSecret()`. I'm pretty confident this should always be fine and tested it a bit but I think we should get this merged and I'll drop that change if it helps the cause.

rhodey:
rebased with 62181a3d7f59a4b1e2738411291abfbd30d1d794 and tested. FWIW Android Studio memory monitor shows improved performance of ~0.80MB on avg

src/org/thoughtcrime/securesms/ConversationFragment.java
mcginty: then wouldn't this work?

```
this.recipients = recipients;
if (this.threadId != threadId) {
  this.threadId = theradId;
  initializeListAdapter();
}
```

which I think reads a bit nicer

rhodey: yup, gimme a sec

PR #2594, opened 2015-03-06
remove unused resources

agrajaghh:
I saw these two png files in the density-independent res/drawables folder and it seems like they are never displayed. `ImageButton` and `ImageView` both contain `android:visibility="gone"` and are never accessed


mcginty:
good thing they're never displayed if they're density-independent!

agrajaghh:
hehe. What do you mean by "they're safe"?

mcginty:
safe to remove, that is.

agrajaghh:
Ah ok, I see ...

PR #2595, opened 2015-03-06
Update strings.xml

thoughtbox:
Case change.


agrajaghh:
Are you sure? 'Incoming off, outgoing on' seems right, no?

thoughtbox:
Doh.

PR #2604, opened 2015-03-06
prevent ghost thread on batch delete of all messages in a thread

### LINKED ISSUE #2262: hidden conversation when all messages deleted before writing new ones
- enter existing conversation
- delete all messages
- ~~right~~(at least it sounds similar) write a new one 
- exit conversation
--> not in conversation list

rhodey:
when deleting all messages in a thread ConversationFragment now explicitly deletes the thread and resets 'threadId' to -1.

Fixes #2262
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1234,6 +1234,11 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   }
 
   @Override
+  public void setThreadId(long threadId) {
+    this.threadId = threadId;
+  }
+
+  @Override
   public void onAttachmentChanged() {
     initializeSecurity();
   }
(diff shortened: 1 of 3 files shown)


rhodey:
Instead of adding the single purpose `ThreadDatabase.getMessageCountForThreadId()` method I was thinking about adding a `getThreadRecord(long threadId)` method that would simply pull from a `Reader` and then just pulling the `count` off of `ThreadRecord`. Any thoughts on this and the balance of specialized pulls vs setting ourselves up for ease of use in the future?

rhodey:
when deleting all the messages in a thread the thread DB automatically deletes the thread, but ConversationActivity does not understand that the thread has been deleted so any new messages sent are inserted using the stale thread ID. the next time a new thread is started the messages sent post all-message-delete will magically appear there because thread IDs are reused in our DB.

moxie0:
this is too much

moxie0:
Have deleteMessage return whether the thread was deleted instead of adding an additional DB query

PR #2614, opened 2015-03-07
Make function easier, default SMS Provider ever true

TimBo93:


moxie0:
does this correspond to an issue?

TimBo93:
i think that is wrong. Or it is not documented good enough. But if i click at the cancle button the intuition is that TextSecure should not the StandardSmsProvider.

The other part of my Commit is an optimization regarding of simplicity of the code.

moxie0:
The point is that if you dismiss the reminder, we don't want to show it again.

PR #2634, opened 2015-03-09
change FAB icon to ic_create and add more resolutions to ic_add

agrajaghh:
I changed the FAB icon to `ic_create`. I think it makes more sense plus gmail and some other apps use this to write a new message too:
ic_create_black_48dp

I also included more resolutions for `ic_add_white_24dp` since only the xxhdpi version was included.


mcginty:
:+1: lgtm

jeremymasters:
Seems like it's a mixed bag whether to use "+" or the pencil.  I just checked Google Inbox and it uses the "+"...So does Evernote, Google Docs, Google Sheets.  Not that I really care or know what material design condones. :)

mcginty:
I think in Inbox it makes more sense, as the "+" expands to a number of actions that add different things. In our case, as it stands the FAB specifically begins the task of composing a new message. Having the pencil makes that more obvious, which I think is a positive.

PR #2635, opened 2015-03-09
show message timestamp in the notification

### LINKED ISSUE #2632: Notification can show incorrect message receipt time
If I have a notification for two (or more) messages from different senders, the notification correctly displays the time of the most recent message.

However, if I read this second message without reading the first one, the notification time is updated to the time that I *read* the second message, rather than the receipt time of the first message.

agrajaghh:
This fixes #2632 

Currently the notification timestamp is the time of the notification creation. This PR changes the timestamp to the message timestamp.

I think there might be a problem if the messages don't arrive in chronological order since the timestamp of the last received message will be displayed I guess. I couldn't test this but I think its still way better than before ;)


agrajaghh:
@moxie0 am I allowed to bump after one month? ;)

rhodey:
I checked out @agrajaghh's branch, rebased with master and tested. works as expected :)

mcginty:
code lgtm too

moxie0:
in 2.13.0

PR #2636, opened 2015-03-09
lazily initialize emoji drawer

### LINKED ISSUE #2585: don't inflate Emoji pane with the rest of the layout
Migrate the Emoji pane inflation to a `ViewStub` so it's only inflated when set to visible. Will knock off ~100ms when creating a conversation view.

mcginty:
Saves a good smidge of inflation time in onCreate for new conversations.

Closes #2585

### DIFF
--- a/res/layout/conversation_activity.xml
+++ b/res/layout/conversation_activity.xml
@@ -122,11 +122,10 @@
         </LinearLayout>
     </RelativeLayout>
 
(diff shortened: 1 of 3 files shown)


moxie0:
Can this be encapsulated in the view?

mcginty:
That'd mean a placeholder view which is inflated to a ViewStub instead of just a ViewStub which doesn't require any inflation.

PR #2637, opened 2015-03-10
media type hints in message notifications

### LINKED ISSUE #1639: Empty notification when receiving an image without text
Sorry if this was already reported somewhere, didn't find anything using the search.

I just noticed that when you receive an image without any text in a message, the belonging notification just displays the sender but nothing else. WhatsApp solves this by displaying "John Doe - Picture", so that you know what's going on.

This probably concerns the other attachment types as well (didn't test it though).
My suggest would be to follow the example of WhatsApp and just display the type of the received media.

rhodey:
1) changed NotificationItem to provide content type hints for media messages that contain no text
2) changed NotificationState to hold more state so that state can be easily restored on resolve of NotificationItem's SlideDeck future
3) NotificationItem and NotificationState related updates to MessageNotifier

the path to notification media previews is now pretty straightforward.

Fixes #1639
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -46,15 +46,19 @@ import org.thoughtcrime.securesms.database.DatabaseFactory;
 import org.thoughtcrime.securesms.database.MmsSmsDatabase;
 import org.thoughtcrime.securesms.database.PushDatabase;
 import org.thoughtcrime.securesms.database.SmsDatabase;
+import org.thoughtcrime.securesms.database.model.MediaMmsMessageRecord;
 import org.thoughtcrime.securesms.database.model.MessageRecord;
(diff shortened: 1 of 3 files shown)


moxie0:
This diff is pretty big.  Is +100 lines really necessary to display the words "Media Message" or whatever?  I'd like it if this were easier to review.

rhodey:
@moxie0 this change is so large because the only media content type information available from `MediaMmsMessageRecord` comes through a `SlideDeck` future. Our notification setup had to be changed to retain state so that the notification state could be recreated upon resolve of the future.

Another way around this would be to attach part content type information to `MediaMmsMessageRecord` but pulling that from the `PartDatabase` would have to happen synchronously else we'd still need these large notification state changes to rebuild state on future resolve.

moxie0:
just say "media message"  or whatever instead of resolving the type

rhodey:
@moxie0 pshhhh... OK

PR #2643, opened 2015-03-11
fix picture message batch selection behavior

### LINKED ISSUE #2488: Mark/Select picture attachment at conversation - problems.
Hi folks,
when i try to select an attachment (picture) at the conversation while pressing "ON" the picture, it gets marked but as soon i release my finger from the screen the picture automaticly gets unmarked. After repeating the following steps 5-6 times he finally catched my selection.
BUT! 
When i mark next to the picture, on the blank screen, it gets marked immediately.
U need a log for this?

rhodey:
onItemClick and onItemLongClick callbacks were acting up when used alongside onClick and onLongClick callbacks, so, dropped use of the list item specific callbacks.

Fixes #2488
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -22,7 +22,6 @@ import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.AdapterView;
 import android.widget.ListView;
 import android.widget.Toast;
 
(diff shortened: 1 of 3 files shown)


rhodey:
just noticed a regression with the long click animation, previously the entire row would do the pre-long-click animation when building up to the long click event, now only the conversation item bubble thing does this. trying some workarounds :/

moxie0:
Any diff in this area needs to be extremely small.  This is the way it is due to a minefield of android bugs across versions.

PR #2647, opened 2015-03-11
Remove encrypted SMS transport, simplify transport options.

moxie0:


res/layout/registration_activity.xml
```diff
@@ -77,7 +77,8 @@
                       android:orientation="horizontal">
 
             <Button android:id="@+id/skipButton"
-                    android:text="@string/registration_activity__skip"
+                    android:visibility="invisible"
```
mcginty: any reason not to use "gone"?

src/org/thoughtcrime/securesms/jobs/SmsReceiveJob.java
```diff
@@ -67,24 +65,21 @@ public boolean onShouldRetry(Exception exception) {
...
     } else {
-      MessageNotifier.updateNotification(context, masterSecret, messageAndThreadId.second);
+      messageAndThreadId = database.insertMessageInbox(masterSecret, message);
     }
 
     return messageAndThreadId;
   }
 
-  private Optional<IncomingTextMessage> assembleMessageFragments(Object[] pdus) {
+  private Optional<IncomingTextMessage> assembleMessageFragments(Object[] pdus)  {
```
mcginty: " "

src/org/thoughtcrime/securesms/util/MmsCharacterCalculator.java
```diff
@@ -0,0 +1,11 @@
+package org.thoughtcrime.securesms.util;
+
+public class MmsCharacterCalculator extends CharacterCalculator {
+
+  private static final int MAX_SIZE = 5000;
```
mcginty: larger than push character count?

mcginty:
Lots of instances of the word "exchange" in strings.xml still :)

mcginty:
Maybe it'd be more fitting in the character calculation case for MMS, Push, as well as any TextView maximums that we store a value as an R.integer

src/org/thoughtcrime/securesms/jobs/SendJob.java
```diff
@@ -40,6 +40,7 @@ public final void onRun(MasterSecret masterSecret) throws Exception {
 
   protected abstract void onSend(MasterSecret masterSecret) throws Exception;
 
+  // FIXME: This should return a value rather than modifying one.
```
mcginty: agreed, will change

src/org/thoughtcrime/securesms/contacts/ContactsDatabase.java
```diff
@@ -97,7 +97,11 @@ public void close() {
   }
 
   public Cursor query(String filter, boolean pushOnly) {
-    final boolean      includeAndroidContacts = !pushOnly && TextSecurePreferences.isDirectSmsAllowed(context);
+    // FIXME: This doesn't make sense to me.  You pass in pushOnly, but then
+    // conditionally check to see whether other contacts should be included
+    // in the query method itself? I don't think this method should have any
+    // understanding of that stuff.
```
mcginty: Think it's just a bad variable name. That boolean is used when adding push group members to only show push contacts regardless of user preferences.

src/org/thoughtcrime/securesms/util/MmsCharacterCalculator.java
moxie0: The compose box max is like 2k, so it doesn't really matter. I just set it to something bigger than what the max is.

PR #2648, opened 2015-03-12
fix picture message batch selection behavior

### LINKED ISSUE #2488: Mark/Select picture attachment at conversation - problems.

rhodey:
1) discovered that a spurious onItemClick callback occurs when moving left or right across an image before releasing a long click.
2) prevented the false onItemClick from occuring by curating the list view and item view onTouch actions.

nothing else prevents this, I tried everything. thoroughly tested.

Fixes #2488
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -20,7 +20,9 @@ import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
+import android.view.MotionEvent;
 import android.view.View;
+import android.view.View.OnTouchListener;
 import android.view.ViewGroup;
 import android.widget.AdapterView;
 import android.widget.ListView;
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -119,6 +127,28 @@ private void initializeListAdapter() {
   private void initializeContextualActionBar() {
     getListView().setOnItemClickListener(selectionClickListener);
     getListView().setOnItemLongClickListener(selectionClickListener);
+    getListView().setOnTouchListener(new OnTouchListener() {
+      @Override
+      public boolean onTouch(View v, MotionEvent event) {
+        if (event.getAction() == MotionEvent.ACTION_UP) {
```
rhodey: `ListView` should only ever experience an `ACTION_UP` after `ConversationItem` gets an `ACTION_CANCEL` in the following situations:

1) after an `ACTION_DOWN` on `ListView`
2) after an `ACTION_MOVE` on `ConversationItem`, caused by `ACTION_DOWN` on `ConversationItem` canceling into a gesture via one or more `ACTION_MOVE`'s on `ConversationItem`.

So... when we see an `ACTION_UP` on `ListView` and it is not the result of an `ACTION_DOWN` on the `ListView` or gesture on the `ConversationItem` we ignore it-- thus preventing the spurious `onListItemClick` callback from surfacing and undoing our media message batch selection.

moxie0:
This seems like something that needs to be encapsulated in the view rather than the controller.

moxie0:
Please reopen when encapsulated

PR #2649, opened 2015-03-12
fix empty mms notifications

### LINKED ISSUE #1639: Empty notification when receiving an image without text

rhodey:
notifications for media messages that contain no text now read 'Media message' in italics.

Fixes #1639
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -51,6 +51,7 @@ import org.thoughtcrime.securesms.recipients.Recipient;
 import org.thoughtcrime.securesms.recipients.RecipientFactory;
 import org.thoughtcrime.securesms.recipients.Recipients;
 import org.thoughtcrime.securesms.service.KeyCachingService;
+import org.thoughtcrime.securesms.util.SpanUtil;
 import org.thoughtcrime.securesms.util.TextSecurePreferences;
 import org.whispersystems.textsecure.api.messages.TextSecureEnvelope;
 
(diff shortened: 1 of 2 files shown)


agrajaghh:
I'm not sure, but I think it could be usefull to display "Media message" regardless if there is a text set. Imagine you receive a picture with a really short text. If you see the notification, you already see the text and might think it was just a textmessage. If you don't want to reply you might just clear the notification without seeing the picture.

So how about displaying "Media message: text"?

rhodey:
I like the idea @agrajaghh, I'll try it out :)

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -352,7 +352,7 @@ private static NotificationState constructNotificationState(Context context,
         body.setSpan(new StyleSpan(android.graphics.Typeface.ITALIC), 0, body.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
       }
```
moxie0: Looks to me like this whole diff should be a 3-line change right here.

agrajaghh:
@rhodey looks good! :smile: 

What do you think about adding the same to the ConversationListActivity?

PR #2665, opened 2015-03-13
Create NameSpace Conversation and move codefiles into that

TimBo93:


mcginty:
Please read the code guidelines before submitting pull requests. I don't think this really helps the organization of the code. Thanks.

TimBo93:
do you mean this ?
https://github.com/WhisperSystems/RedPhone/wiki/Code-style-Guidelines
Which policy is violated here ?
I honestly do not think the current code very structured. This should be rapidly improved.

mcginty:
Yes, the indentation change here violates the style guidelines: https://github.com/WhisperSystems/TextSecure/pull/2665/files#diff-d9440c6e717759c11a7320a50b294379R440

Also, unless there's a conflict of two of the same class name being used in a file, one shouldn't use the full package name as seen in https://github.com/WhisperSystems/TextSecure/pull/2665/files#diff-23139aee74479a72a1ad628505568a01R68. Also we tend to avoid creating variables when it's just as short to use the getter (and there is no performance impact, like in this case).

That's a strong opinion, but we need a compelling alternative.

TimBo93:
Thank you for your response. (: What if the 2 issues you marked would be fixed? What about a discussion or plan about namespace configurations? I think textsecure can profit from a better structure. Is there a way to communicate the ideas (etherpad for example, dev wiki)? I thought it could help you if i spend my time with structure work or refactoring and you can focus to some more complicated work. Sorry for my lazy english - i hope you can follow me. Thank you for your time to show me my failures.

mcginty:
Right now all our Activity classes are in the base package, which makes sense to me at least. While it may not be perfect, I don't think it's particularly awful to see a list of activities that make up how a user interacts with an app.

TimBo93:
I know a functionality in maven which allows you to generate uml package or uml class diagrams. For me as a newcomer in that project it is not simple to include myself efficient into that. Maybe there is an option for gradle too. To see where the leaknesses in coupling (or cohesion) are.

PR #2671, opened 2015-03-13
Fix deprecation warnings

tinloaf:
//FREEBIE


moxie0:
This isn't our code. It's up to @mcginty, but I think we probably want to keep the diff as minimal as possible to make any updates smoother?

moxie0:
Thanks but let's add suppresswarnings instead, we don't want to touch that junk

Issue #2672, opened 2015-03-14
Fix post-encrypted SMS settings strings


jtracey:
In 2.7.0+, now that encrypted SMS is no longer a thing, some of the strings for settings are a bit confusing. In particular, there should no longer be a distinction between "incoming" and "outgoing" SMS settings. I'm also not sure about the description of the Push message setting, given that it's the primary feature of the app. Maybe change it to "Disable TextSecure messages" or something?
Screenshots:
![incoming on](https://cloud.githubusercontent.com/assets/2998852/6649507/f521c19c-c9a8-11e4-96f3-a77affbbb2fb.png)
![incoming sms enabled](https://cloud.githubusercontent.com/assets/2998852/6649509/f93e5510-c9a8-11e4-89a8-2efabe327c14.png)

moxie0:
yeah we should touch these up

moxie0:
@agrajaghh can i assign this to you? =)

agrajaghh:
@moxie0 sure, will have a look

PR #2682, opened 2015-03-14
adapt SMS strings in preferences

agrajaghh:
Fix for #2672 

* changed "Push messages" to "TextSecure messages"
* changed sms summary to: "On", "Off", "SMS on, MMS off", "SMS off, MMS on"
* changed SMS pref on >= KITKAT to: "SMS Disabled" and "SMS Enabled"
* removed unused strings

Please comment if you have better suggestions for the strings :blush: 

**Android 2.3 settings:**
2 3_settings

**Android 2.3 SMS and MMS:**
2 3_sms

**Android >= KITKAT settings:**
5 0_settings

**Android >= KITKAT SMS and MMS (disabled):**
5 0_sms_disabled

**Android >= KITKAT SMS and MMS (enabled):**
5 0_sms_enabled

**Advanced:**
5 0_advanced


Geblaat:
Imho it would be good to mention encryption or security, because since encrypted SMS is gone, you now need to register for TextSecure messages if you want encryption. How about "Use encrypted messaging and avoid SMS fees by using the data channel for secure communication with other TextSecure users"?

agrajaghh:
what about just adding "encrypted"?
> Increase privacy and avoid SMS fees by using the data channel for encrypted communication with other TextSecure users

Geblaat:
Yes, that would work as well I guess.

jtracey:
IMO, if push settings are neccessary to keep, they should be burried in Advanced Settings. It made sense to put it where it is when it was an experimental way of doing things, but now, I can't immagine very many circumstances where someone would actually even need to do this. Putting an option to disable the entire point of the app as the first thing you see in settings just seems somewhat dangerous to me.

agrajaghh:
@jtracey good point since there is the non-dismissable banner to register for push messages anyway. I'll move it into the advanced options ...

agrajaghh:
moved the "TextSecure messages" option to advanced and updated the screnshots

moxie0:
@mcginty You cool with these changes?

mcginty:
@moxie0 @agrajaghh looks good. I think the subtext under "TextSecure messages" is too long. Maybe something more to the point like "Free private messaging to TextSecure and Signal users." The registration page will have more opportunity for details.

mcginty:
On that note we can probably remove "viewing and storing" from the incoming SMS and MMS descriptions.

agrajaghh:
@mcginty changed and updated the screenshots

PR #2712, opened 2015-03-17
Fix for inconsistent thumbnail aspect ratio

### LINKED ISSUE #2676: Image bubble size differs in detail, outgoing/incoming messages
Seems as if the max width is different - didn't code check... 
Conversation 
screenshot_2015-03-14-03-31-05 01 01

Detail 

screenshot_2015-03-14-03-29-52 01

mcginty:
Problems arose from any app that targets below API 19 and uses
views that rely on RelativeLayouts giving correct measurement specs to
their onMeasure().

Resolves #2676

### DIFF
--- a/res/layout/conversation_bubble_incoming.xml
+++ b/res/layout/conversation_bubble_incoming.xml
@@ -12,28 +12,34 @@
           android:layout_alignParentLeft="true"
           android:layout_alignParentStart="true" />
 
-    <org.thoughtcrime.securesms.components.ForegroundImageView
-            android:id="@+id/image_view"
-            android:layout_width="wrap_content"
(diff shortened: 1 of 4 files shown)


moxie0:
i live in ignorance, so lgtm, but that means very little

moxie0:
But is there a reason this wrapper layout isn't just a part of the ForegroundImageView?

mcginty:
@moxie0 I tried that at first, but ended up having other layout problems so I kept it simple instead. Can revisit later.

moxie0:
When is this mythical "later?"  =)

mcginty:
defining later is on the roadmap

PR #2717, opened 2015-03-18
Acquire wakelocks for incoming messages.

moxie0:
Please also see the accompanying change:
https://github.com/WhisperSystems/jobmanager/commit/c7a1ac531018646557bb11d0830c9c7a136a7999

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,7 @@ dependencies {
     compile 'com.madgag.spongycastle:prov:1.51.0.0'
     provided 'com.squareup.dagger:dagger-compiler:1.2.2'
 
-    compile 'org.whispersystems:jobmanager:0.10.0'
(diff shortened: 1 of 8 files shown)


mcginty:
the `withWakeLock(boolean)` still reads strange to me but code lgtm!

agrajaghh:
Couldn't it be problematic to set a wakelock without a timeout? For example in `PushReceiveJob`: What would happen when the job gets added (and so the wakelock gets acquired) but the requirement isn't present (no internet connection). Wouln't the phone be awake until the internet connection is back or did I miss something?

moxie0:
@agrajaghh They're only set without a timeout if there are no requirements.

agrajaghh:
@moxie0 shame on me, you're right....

PR #2719, opened 2015-03-19
fix initialization of emoji cache

### LINKED ISSUE #2700: Emoji pane losing history
Hi!

Since version 2.6, TextSecure loses the emoji history (last used emojis) from time to time.
Never happened before with older version.

Thanks for your great work!

rhodey:
Fix initialization of emoji cache.

Could someone test this on a pre-kitkat phone? I have none. Intuition is that if you force-stop TextSecure on 2.6.4 you should be able to reproduce #2700, might need to clear app cache too but admittedly I have no idea what that button actually does.

Fixes #2700
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/util/Emoji.java
+++ b/src/org/thoughtcrime/securesms/util/Emoji.java
@@ -311,8 +311,6 @@ public class Emoji {
         Log.w(TAG, e);
         recentlyUsed = new LinkedHashSet<>();
       }
-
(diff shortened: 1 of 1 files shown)


moxie0:
All PRs need to be tested on GB. If you don't have a GB device, please use an emulator.

moxie0:
Please reopen when tested.

rhodey:
tested on Nexus One running Android 2.3.6, solves issue.

PR #2720, opened 2015-03-19
use dynamic language setting for message time deltas

### LINKED ISSUE #2684: Dates are displayed in system locale regardless of language setting
Currently all the dates and times are diplayed regarding the android system locale using `DateUtils.formatDateTime`. 

Sometimes its even mixed because the plurals `hours_ago` and `minutes_ago` are translated.

My system language is set to german but I'm using TextSecure in english:
date

It would be more consistent if the dates are formatted regarding the textsecure language setting I guess.

rhodey:
modified DynamicLanguage to set and restore system default Locale.

Fixes #2684

### DIFF
--- a/src/org/thoughtcrime/securesms/util/DynamicLanguage.java
+++ b/src/org/thoughtcrime/securesms/util/DynamicLanguage.java
@@ -43,6 +43,9 @@ public class DynamicLanguage {
       context.getResources().updateConfiguration(configuration,
                                                   context.getResources().getDisplayMetrics());
     }
+    if (!Locale.getDefault().equals(selectedLocale)) {
+      Locale.setDefault(selectedLocale);
+    }
   }
 
   private static Locale getActivityLocale(Activity activity) {
@@ -53,7 +56,11 @@ public class DynamicLanguage {
     String language[] = TextUtils.split(TextSecurePreferences.getLanguage(context), "_");
 
     if (language[0].equals(DEFAULT)) {
-      return Locale.getDefault();
+      String sysLanguage = System.getProperty("user.language", "en");
+      String sysRegion   = System.getProperty("user.region",   "US");
+      String sysVariant  = System.getProperty("user.variant",  "");
+
+      return new Locale(sysLanguage, sysRegion, sysVariant);
     } else if (language.length == 2) {
       return new Locale(language[0], language[1]);
     } else {


src/org/thoughtcrime/securesms/util/DynamicLanguage.java
```diff
@@ -53,7 +56,11 @@ private static Locale getSelectedLocale(Context context) {
     String language[] = TextUtils.split(TextSecurePreferences.getLanguage(context), "_");
 
     if (language[0].equals(DEFAULT)) {
-      return Locale.getDefault();
+      String sysLanguage = System.getProperty("user.language", "en");
```
rhodey: defaults from Locale.java

agrajaghh:
working for me :+1:

moxie0:
1. No side effects from private methods please.
2. If you have to comment on your diff, that means the diff isn't clear.
3. Just pass a locale into simpledateformat instead of this.

rhodey:
@moxie0 `3` is the first thing I tried, modified 'DynamicLanguage.getSelectedLocale()` to be public and passed that into `SimpleDateFormat`. However, this method is only used within `MessageDetailsActivity`. Then noticed that `android.text.format.DateUtils.formatDateTime()` ends up using `Local.getDefault()` even though it is passed in a reference to context, so I replaced all calls to `android.text.format.DateUtils.formatDateTime()` with a new method `org.thoughtcrime.securesms.util.DateUtils.getFormattedDateTime()`

```java
private static String getFormattedDateTime(Context context, long millis, int flags) {
  Formatter f = new Formatter(new StringBuilder(50), DynamicLanguage.getSelectedLocale());
  return formatDateRange(context, f, millis, millis, flags).toString();
}
```

This also didn't work, in the end the situation is that lots of the date utilities in Android use `Locale.getDefault()` even with a reference to context. So, instead of replacing all of `android.text.format.DateUtils` it made more sense to manipulate the default locale.

My comments on the diff were only to excuse the choice of default Locale as US. If [setContextLocale](https://github.com/WhisperSystems/TextSecure/blob/d68047ce724b9cd71fa89667b668143c0daaba54/src/org/thoughtcrime/securesms/util/DynamicLanguage.java#L38) was renamed `setAppLocale()` would you agree that no side effects exist?

moxie0:
there was a reason we didn't just reset the default locale to begin with

rhodey:
the only use of `java.util.Locale` in all of the codebase is in `DateUtils` and `DynamicLanguage`, but sure, I'll try a different work around.

PR #2725, opened 2015-03-19
use dynamic language setting for message time deltas

### LINKED ISSUE #2684: Dates are displayed in system locale regardless of language setting

rhodey:
modified DateUtils to use SimpleDateFormat for everything because unlike formatDateTime() it respects Locale.

Fixes #2684

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -173,7 +173,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
 
     setContentView(R.layout.conversation_activity);
     getSupportActionBar().setDisplayHomeAsUpEnabled(true);
-    fragment = initFragment(R.id.fragment_content, new ConversationFragment(), masterSecret);
+    fragment = initFragment(R.id.fragment_content, new ConversationFragment(), masterSecret, dynamicLanguage.getCurrentLocale());
 
     initializeReceivers();
     initializeViews();
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -67,7 +67,7 @@ public class ConversationFragment extends ListFragment
   public void onCreate(Bundle icicle) {
     super.onCreate(icicle);
     this.masterSecret = getArguments().getParcelable("master_secret");
-    this.locale       = (Locale) getArguments().getSerializable(PassphraseRequiredActionBarActivity.ARG_LOCALE);
+    this.locale       = (Locale) getArguments().getSerializable(PassphraseRequiredActionBarActivity.LOCALE_EXTRA);
   }
 
   @Override
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -45,6 +45,7 @@ import org.thoughtcrime.securesms.util.SaveAttachmentTask.Attachment;
 
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Locale;
 
 public class ConversationFragment extends ListFragment
   implements LoaderManager.LoaderCallbacks<Cursor>
@@ -60,11 +61,13 @@ public class ConversationFragment extends ListFragment
   private Recipients   recipients;
   private long         threadId;
   private ActionMode   actionMode;
+  private Locale       locale;
 
   @Override
   public void onCreate(Bundle icicle) {
     super.onCreate(icicle);
     this.masterSecret = getArguments().getParcelable("master_secret");
+    this.locale       = (Locale) getArguments().getSerializable(PassphraseRequiredActionBarActivity.ARG_LOCALE);
   }
 
   @Override
@@ -116,7 +119,7 @@ public class ConversationFragment extends ListFragment
 
   private void initializeListAdapter() {
     if (this.recipients != null && this.threadId != -1) {
-      this.setListAdapter(new ConversationAdapter(getActivity(), masterSecret, selectionClickListener,
+      this.setListAdapter(new ConversationAdapter(getActivity(), masterSecret, locale, selectionClickListener,
                                                   (!this.recipients.isSingleRecipient()) || this.recipients.isGroupRecipient(),
                                                   DirectoryHelper.isPushDestination(getActivity(), this.recipients)));
       getListView().setRecyclerListener((ConversationAdapter)getListAdapter());
--- a/src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
+++ b/src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
@@ -19,8 +19,12 @@ import org.thoughtcrime.securesms.service.KeyCachingService;
 import org.thoughtcrime.securesms.service.MessageRetrievalService;
 import org.thoughtcrime.securesms.util.TextSecurePreferences;
 
+import java.util.Locale;
+
 public abstract class PassphraseRequiredActionBarActivity extends BaseActionBarActivity implements MasterSecretListener {
   private static final String TAG = PassphraseRequiredActionBarActivity.class.getSimpleName();
+
+  public static final String ARG_LOCALE = "arg_locale";
 
   private static final int STATE_NORMAL                   = 0;
   private static final int STATE_CREATE_PASSPHRASE        = 1;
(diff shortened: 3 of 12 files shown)


src/org/thoughtcrime/securesms/util/DynamicLanguage.java
```diff
@@ -49,7 +49,7 @@ private static Locale getActivityLocale(Activity activity) {
     return activity.getResources().getConfiguration().locale;
   }
 
-  private static Locale getSelectedLocale(Context context) {
+  public static Locale getSelectedLocale(Context context) {
```
moxie0: return currentLocale instead

src/org/thoughtcrime/securesms/util/DateUtils.java
```diff
@@ -39,6 +40,13 @@ private static int convertDelta(final long millis, TimeUnit to) {
     return (int) to.convert(System.currentTimeMillis() - millis, TimeUnit.MILLISECONDS);
   }
 
+  public static String getFormattedDateTime(Context context, long time, String template) {
+    final Locale locale           = DynamicLanguage.getSelectedLocale(context);
```
moxie0: Pass the locale in, rather than pulling it off a static DynamicLanguage method

moxie0:
@mcginty will have to comment on the rest

src/org/thoughtcrime/securesms/util/DynamicLanguage.java
rhodey: can't modify this method like that, need to add a new non-static `getCurrentLocale()` method and then change the `DateUtils` method signatures to take a `DynamicLanguage` instance. doing this now but definite larger diff.

moxie0: exactly. larger diff but not disgusting.

moxie0:
whoa, this is way too much

rhodey:
@moxie0 

+ `DateUtils` should use `DynamicLanguage`
+ `DynamicLanguage` is not static
+ `ConversationItem` and ConversationListItem` need to use `DateUtils`
+ `ConversationItem` and ConversationListItem` should probably use the same `DynamicLanguage` as their parent activities

So, lots of method signatures had to be changed to pass `DynamicLanguage` down. This diff can be drastically shrunk if `ConversationItem` and ConversationListItem` were allowed to create their own `DynamicLanguage` instance, else I don't feel this can be shrunk without refactoring `DynamicLanguage` to be static-- how about that?

moxie0:
why not just pass a locale?

rhodey:
@moxie0 ok, I agree that it makes more sense to pass in `Locale` instead of `DynamicLanguage`, but adding that change would not reduce the size of this diff because we'd just be passing down `DynamicLanguage.getCurrentLocale()` instead of `DynamicLanguage`.

moxie0:
Why do you need a provider interface? Just pass the locale instead of the interface.

rhodey:
@moxie0 `ConversationFragment` and `ConversationListFragment` both need the locale and it is my understanding that using constructor arguments for fragments is poor form. The interface exists so that the fragments can get a reference to Locale `onAttach()`.

A protected `getLocale()` method could be added to both the activities and then their fragments could grab it `onAttach()` but casting activities like that also seemed kinda poor form.

moxie0:
locale is serializable

McLoo:
for the record:
 In 2.9.0 times (minutes, hours ) change between TS and system locale. Dates still show in system locale. 
Didn't figure out the when they switch. Will recheck, ones this is merged.

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -725,6 +727,8 @@ private void initializeResources() {
     masterSecret     = getIntent().getParcelableExtra(MASTER_SECRET_EXTRA);
 
     recipients.addListener(this);
+
+    getIntent().putExtra(DYNAMIC_LOCALE_EXTRA, dynamicLanguage.getCurrentLocale());
```
moxie0: look at mcginty's routingactivity pr and follow his lead

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -61,6 +61,7 @@
   private Recipients   recipients;
   private long         threadId;
   private ActionMode   actionMode;
+  private Locale       dynamicLocale;
```
moxie0: what's dynamic about it?

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -102,14 +103,17 @@ public void onNewIntent() {
   }
 
   private void initializeResources() {
-    this.masterSecret = this.getActivity().getIntent().getParcelableExtra("master_secret");
-    this.recipients   = RecipientFactory.getRecipientsForIds(getActivity(), getActivity().getIntent().getLongArrayExtra("recipients"), true);
-    this.threadId     = this.getActivity().getIntent().getLongExtra("thread_id", -1);
+    final Intent parentIntent = this.getActivity().getIntent();
```
moxie0: unnecessary

src/org/thoughtcrime/securesms/util/DynamicLanguage.java
```diff
@@ -35,6 +35,10 @@ public void updateServiceLocale(Service service) {
     setContextLocale(service, currentLocale);
   }
 
+  public Locale getCurrentLocale() {
+    return (currentLocale != null) ? currentLocale : Locale.getDefault();
```
moxie0: when would you want this to be true?

src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
```diff
@@ -78,6 +82,20 @@ public void onMasterSecretCleared() {
...
+                                                @NonNull Locale locale) {
+    Bundle args = new Bundle();
+    args.putParcelable("master_secret", masterSecret);
+    args.putSerializable(ARG_LOCALE, locale);
+    fragment.setArguments(args);
+    getSupportFragmentManager().beginTransaction()
+                               .replace(target, fragment)
+                               .commit();
+    return fragment;
+  }
```
moxie0: one of these methods should call the other

src/org/thoughtcrime/securesms/util/DateUtils.java
```diff
@@ -49,34 +55,34 @@ public static String getBriefRelativeTimeSpanString(final Context c, final long
...
-      if (isWithin(timestamp, 6, TimeUnit.DAYS)) {
-        formatFlags |= DateUtils.FORMAT_SHOW_WEEKDAY | DateUtils.FORMAT_ABBREV_WEEKDAY;
-      } else if (isWithin(timestamp, 365, TimeUnit.DAYS)) {
-        formatFlags |= DateUtils.FORMAT_SHOW_DATE | DateUtils.FORMAT_NO_YEAR | DateUtils.FORMAT_ABBREV_ALL;
-      } else {
-        formatFlags |= DateUtils.FORMAT_SHOW_DATE | DateUtils.FORMAT_ABBREV_ALL;
-      }
-      return DateUtils.formatDateTime(c, timestamp, formatFlags);
+      String format = "";
+      if      (isWithin(timestamp,   6, TimeUnit.DAYS)) format += "EEE ";
```
mcginty: the += doesn't make sense here since format is always empty

rhodey: good catch, I'll try the StringBuilder pattern and see how that works out.

src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
```diff
@@ -78,9 +82,12 @@ public void onMasterSecretCleared() {
 
   protected <T extends Fragment> T initFragment(@IdRes int target,
                                                 @NonNull T fragment,
-                                                @NonNull MasterSecret masterSecret) {
+                                                @NonNull MasterSecret masterSecret,
+                                                Locale locale) {
```
mcginty: `@Nullable`

src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
```diff
@@ -78,9 +82,12 @@ public void onMasterSecretCleared() {
...
     Bundle args = new Bundle();
     args.putParcelable("master_secret", masterSecret);
+    args.putSerializable(ARG_LOCALE, locale);
```
mcginty: `LOCALE_EXTRA` i think is the canonical form for bundle constants

PR #2727, opened 2015-03-20
Use new MMS APIs available from Lollipop and Lollipop MR1

### LINKED ISSUE #1937: MMS sending (and receiving) fails if on WiFi and APN is set to IPv6
Hello,

I am having trouble sending and receiving MMS on my T-Mobile USA Samsung galaxy S4 (Android 4.4.4) while connected to WiFi. I am using TextSecure 2.1.8.

It looks like the problems have something to do with the switch from IPv4 (on WiFi) to IPv6 on the cellular network.

If I manually disable WiFi and send a MMS through TextSecure on the IPv6 only APN, the MMS is correctly sent and received.

If I manually switch the APN to IPv4 only, then TextSecure correctly sends and receives MMS while connected to WiFi (by transitioning to the cellular network to send the MMS).

However, if I use the default APN of the phone (IPv6 only) and I am connected to the IPv4 only WiFi, sending fails. The logs show that TextSecure attempts to connect to the cellular, but then times out connecting to the MMSC (2607:7700:0:5::ab8:4b81) because of Network unreachable.

mcginty:
Fixes #1937


mcginty:
Diff updated, tested MMS to/from lollipop and non-lollipop device with good results :).

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -314,8 +314,10 @@ private void setSentStatusIcons() {
 
   private void setFailedStatusIcons() {
     dateText.setText(R.string.ConversationItem_error_not_delivered);
-    indicatorText.setText(R.string.ConversationItem_click_for_details);
-    indicatorText.setVisibility(View.VISIBLE);
+    if (indicatorText != null) {
+      indicatorText.setText(R.string.ConversationItem_click_for_details);
+      indicatorText.setVisibility(View.VISIBLE);
+    }
```
moxie0: why is this in this diff?

src/org/thoughtcrime/securesms/mms/IncomingLollipopMmsConnection.java
```diff
@@ -0,0 +1,105 @@
...
+    if (!ACTION.equals(intent.getAction())) {
+      Log.w(TAG, "received broadcast with unexpected action " + intent.getAction());
+      return;
+    }
+    if (VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP_MR1) {
+      Log.w(TAG, "HTTP status: " + intent.getIntExtra(SmsManager.EXTRA_MMS_HTTP_STATUS, -1));
+    }
+    Log.w(TAG, "code: " + getResultCode() + ", result string: " + getResultData());
+
+    finished = true;
```
moxie0: this has to be in the synchronized block

src/org/thoughtcrime/securesms/mms/OutgoingLollipopMmsConnection.java
```diff
@@ -0,0 +1,101 @@
...
+  private static final String TAG = OutgoingLollipopMmsConnection.class.getSimpleName();
+  private static final String ACTION = OutgoingLollipopMmsConnection.class.getCanonicalName() + "MMS_SENT_ACTION";
+
+  private Context context;
+  private byte[] response;
+  private boolean finished;
+  private long messageId;
+  private byte[] pduBytes;
+
+  public OutgoingLollipopMmsConnection(Context context, byte[] pduBytes, long messageId) {
```
moxie0: pdubytes should be a paramter of send, and i don't think this class should know anything about message ids

src/org/thoughtcrime/securesms/mms/OutgoingLegacyMmsConnection.java
```diff
@@ -0,0 +1,164 @@
...
+import ws.com.google.android.mms.pdu.PduParser;
+import ws.com.google.android.mms.pdu.SendConf;
+
+@SuppressWarnings("deprecation")
+public class OutgoingLegacyMmsConnection extends LegacyMmsConnection implements OutgoingMmsConnection {
+  private final static String TAG = OutgoingLegacyMmsConnection.class.getSimpleName();
+
+  private byte[] mms;
+
+  public OutgoingLegacyMmsConnection(Context context, byte[] mms) throws ApnUnavailableException {
```
moxie0: mms should be a parameter of send

src/org/thoughtcrime/securesms/mms/IncomingLollipopMmsConnection.java
```diff
@@ -0,0 +1,105 @@
...
+      PendingIntent pendingIntent = PendingIntent.getBroadcast(context, 1, new Intent(ACTION), PendingIntent.FLAG_ONE_SHOT);
+      Uri contentUri = ContentUris.withAppendedId(MmsBodyProvider.CONTENT_URI, messageId);
+      Log.w(TAG, "downloading multimedia from " + contentLocation + " to " + contentUri);
+      SmsManager.getDefault().downloadMultimediaMessage(context, contentLocation, contentUri, null, pendingIntent);
+
+      synchronized (this) {
+        while (!finished) Util.wait(this, 30000);
+      }
+
+      context.getApplicationContext().unregisterReceiver(this);
```
moxie0: should probably be finally

src/org/thoughtcrime/securesms/mms/IncomingLollipopMmsConnection.java
```diff
@@ -0,0 +1,105 @@
...
+public class IncomingLollipopMmsConnection extends BroadcastReceiver implements IncomingMmsConnection {
+  public static final String ACTION = IncomingLollipopMmsConnection.class.getCanonicalName() + "MMS_DOWNLOADED_ACTION";
+  private static final String TAG = IncomingLollipopMmsConnection.class.getSimpleName();
+
+  private Context context;
+  private String  contentLocation;
+  private long    messageId;
+  private boolean finished;
+
+  public IncomingLollipopMmsConnection(Context context, String contentLocation, long messageId) {
```
moxie0: contentlocation should be a parameter of receive, and i don't think this class should know anything about message ids

src/org/thoughtcrime/securesms/mms/IncomingLegacyMmsConnection.java
```diff
@@ -0,0 +1,148 @@
...
+                                            boolean usingRadio,
+                                            boolean useProxy)
+      throws ApnUnavailableException
+  {
+    try {
+      NotifyRespInd notifyResponse = new NotifyRespInd(PduHeaders.CURRENT_MMS_VERSION,
+                                                       transactionId,
+                                                       PduHeaders.STATUS_RETRIEVED);
+
+      OutgoingLegacyMmsConnection connection = new OutgoingLegacyMmsConnection(context, new PduComposer(context, notifyResponse).make());
```
moxie0: these classes should probably be peers, not dependencies

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -666,7 +666,7 @@ private void initializeMmsEnabledCheck() {
     new AsyncTask<Void, Void, Boolean>() {
       @Override
       protected Boolean doInBackground(Void... params) {
-        return OutgoingMmsConnection.isConnectionPossible(ConversationActivity.this);
+        return OutgoingLegacyMmsConnection.isConnectionPossible(ConversationActivity.this);
```
moxie0: shouldn't there be a lollipop check somewhere around here now?

src/org/thoughtcrime/securesms/ConversationItem.java
mcginty: i can throw it into another PR, but causes NPEs if there's an incoming failure

src/org/thoughtcrime/securesms/ConversationActivity.java
mcginty: that's what I forgot to do! Yeah, I was looking into how to do that check on Lollipop and obviously got distracted. I added a util method to check for both Lollipop and legacy mms capabilities. Right now we don't have any code in to override the APN settings in Lollipop, so I should disable the settings or add in overrides in this PR too.

src/org/thoughtcrime/securesms/ConversationItem.java
moxie0: is the npe specific to this change?

src/org/thoughtcrime/securesms/util/Util.java
```diff
@@ -279,4 +284,11 @@ public static SecureRandom getSecureRandom() {
   public static boolean isBuildFresh() {
     return BuildConfig.BUILD_TIMESTAMP + TimeUnit.DAYS.toMillis(180) > System.currentTimeMillis();
   }
+
+  @TargetApi(VERSION_CODES.LOLLIPOP)
+  public static boolean isMmsCapable(Context context) {
+    return (VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP)
+           ? SmsManager.getDefault().getCarrierConfigValues().getBoolean(SmsManager.MMS_CONFIG_MMS_ENABLED)
+           : OutgoingLegacyMmsConnection.isConnectionPossible(context);
```
moxie0: I think this method is written too generically. IIRC, what we really want to know is whether we need to prompt the user to enter their custom MMSC information into TextSecure in the case where we can't access it for some reason.  For lollipop, that should never be true, since we're delegating everything to the system.

mcginty: I'm modifying that as we speak :)

src/org/thoughtcrime/securesms/ConversationItem.java
mcginty: No, just came across it during testing. Busting it out into another PR.

mcginty:
Now displays error if MMS isn't enabled by carrier in lollipop and higher, otherwise will go the normal manual MMS route.

Disabled manual MMS controls on lollipop as well since they're ignored. A TODO could be to integrate the new SocketFactory API in the legacy sending logic to support manual overrides, but that's for another PR.

src/org/thoughtcrime/securesms/mms/IncomingLegacyMmsConnection.java
```diff
@@ -0,0 +1,146 @@
...
+                                            boolean usingRadio,
+                                            boolean useProxy)
+      throws ApnUnavailableException
+  {
+    try {
+      NotifyRespInd notifyResponse = new NotifyRespInd(PduHeaders.CURRENT_MMS_VERSION,
+                                                       transactionId,
+                                                       PduHeaders.STATUS_RETRIEVED);
+
+      OutgoingLegacyMmsConnection connection = new OutgoingLegacyMmsConnection(context);
```
moxie0: I still think these classes should be peers and not dependencies

mcginty: And have another wrapper class that implements the OutgoingMmsConnection interface and uses both Incoming/OutgoingLegacyMmsConnection?

moxie0: Some might say MmsDownloadJob is that interface. You might even call it a "controller." =)

mcginty: The outgoing lollipop API appears to send the confirmation automatically, so I felt it was most symmetric/nice to have our legacy implementation do the same thing.

moxie0: I think it's structurally weird, and could result in some functional weirdness (what happens with a failure on just the ack part?), but I can live if it's what you want to do.

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -540,11 +541,19 @@ private void handleAddAttachment() {
...
-    startActivity(intent);
+      Intent intent = new Intent(this, PromptMmsActivity.class);
+      intent.putExtras(getIntent().getExtras());
+      startActivity(intent);
+    } else {
+      new MaterialDialog.Builder(this).title(R.string.ConversationActivity_mms_not_supported_title)
+                                      .content(R.string.ConversationActivity_mms_not_supported_message)
+                                      .iconRes(R.drawable.ic_error_red_24dp)
+                                      .neutralText(android.R.string.ok)
+                                      .show();
```
moxie0: Should we even do this?  We could just have the message fail as usual.

Maybe it's nice, but it could also cause a wrinkle in the "auto-upgrade" path when sending to recipients who are TS users but aren't yet in the synchronized directory.

src/org/thoughtcrime/securesms/jobs/MmsDownloadJob.java
```diff
@@ -86,71 +101,32 @@ public void onRun(MasterSecret masterSecret)  {
 
     database.markDownloadState(messageId, MmsDatabase.Status.DOWNLOAD_CONNECTING);
 
-    String   contentLocation = new String(notification.get().getContentLocation());
-    byte[]   transactionId   = notification.get().getTransactionId();
-    MmsRadio radio           = MmsRadio.getInstance(context);
+    String contentLocation = new String(notification.get().getContentLocation());
+    byte[] transactionId = notification.get().getTransactionId();
```
moxie0: alignment

src/org/thoughtcrime/securesms/jobs/MmsDownloadJob.java
```diff
@@ -86,71 +101,32 @@ public void onRun(MasterSecret masterSecret)  {
...
 
-    Log.w(TAG, "About to parse URL...");
+    Log.w(TAG, "Downloading mms at " + Uri.parse(contentLocation).getHost());
 
-    Log.w(TAG, "Downloading mms at " +  Uri.parse(contentLocation).getHost());
+    download(masterSecret, contentLocation, transactionId);
+  }
 
+  private void download(MasterSecret masterSecret, String contentLocation, byte[] transactionId) {
```
moxie0: I think it makes sense to merge these methods now.

src/org/thoughtcrime/securesms/jobs/MmsSendJob.java
```diff
@@ -90,59 +89,36 @@ public void onCanceled() {
     notifyMediaMessageDeliveryFailed(context, messageId);
   }
 
-  public MmsSendResult deliver(MasterSecret masterSecret, SendReq message)
+  private MmsSendResult deliver(MasterSecret masterSecret, SendReq message)
```
moxie0: I don't think this needs to be a separate method anymore.

src/org/thoughtcrime/securesms/mms/IncomingLegacyMmsConnection.java
```diff
@@ -0,0 +1,146 @@
...
+import ws.com.google.android.mms.pdu.PduParser;
+import ws.com.google.android.mms.pdu.RetrieveConf;
+
+@SuppressWarnings("deprecation")
+public class IncomingLegacyMmsConnection extends LegacyMmsConnection implements IncomingMmsConnection {
+  private static final String TAG = IncomingLegacyMmsConnection.class.getSimpleName();
+
+  private byte[] transactionId;
+
+  public IncomingLegacyMmsConnection(Context context, Apn apn, byte[] transactionId) {
```
moxie0: i still think the latter parameters should be params of retrieve

src/org/thoughtcrime/securesms/mms/IncomingLollipopMmsConnection.java
```diff
@@ -0,0 +1,100 @@
...
+
+public class IncomingLollipopMmsConnection extends BroadcastReceiver implements IncomingMmsConnection {
+  public static final String ACTION = IncomingLollipopMmsConnection.class.getCanonicalName() + "MMS_DOWNLOADED_ACTION";
+  private static final String TAG = IncomingLollipopMmsConnection.class.getSimpleName();
+
+  private Context context;
+  private String  contentLocation;
+  private boolean finished;
+
+  public IncomingLollipopMmsConnection(Context context, String contentLocation) {
```
moxie0: latter param should be of retrieve

src/org/thoughtcrime/securesms/mms/IncomingLegacyMmsConnection.java
mcginty: I think the Lollipop API was kind of poorly designed for that reason (if your receiver fails for some reason, you'd like to be able to fetch it again).

src/org/thoughtcrime/securesms/mms/IncomingLegacyMmsConnection.java
mcginty: lol. just looked at how these notifications are sent back if we don't pass the transactionId to the lollipop API. It's not even serialized by the PduComposer. I'm not sure what it's there for.

src/org/thoughtcrime/securesms/ConversationActivity.java
mcginty: you're totally right. letting it fail.

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -540,11 +541,13 @@ private void handleAddAttachment() {
   }
 
   private void handleManualMmsRequired() {
-    Toast.makeText(this, R.string.MmsDownloader_error_reading_mms_settings, Toast.LENGTH_LONG).show();
+    if (Util.canOverrideMms()) {
```
moxie0: Is this check necessary?

src/org/thoughtcrime/securesms/mms/OutgoingLollipopMmsConnection.java
```diff
@@ -0,0 +1,94 @@
...
+    context.getApplicationContext().registerReceiver(this, new IntentFilter(ACTION));
+    long nonce = System.currentTimeMillis();
+    try {
+      Uri contentUri = ContentUris.withAppendedId(MmsBodyProvider.CONTENT_URI, nonce);
+      Util.copy(new ByteArrayInputStream(pduBytes), context.getContentResolver().openOutputStream(contentUri, "w"));
+
+      SmsManager.getDefault().sendMultimediaMessage(context, contentUri, null, null,
+                                                    PendingIntent.getBroadcast(context, 1, new Intent(ACTION), PendingIntent.FLAG_ONE_SHOT));
+
+      while (!finished) Util.wait(this, 30000);
```
moxie0: Why does this have a timeout if you loop forever?

mcginty: yeah, that logic doesn't make sense (I had the timeout before I did the looped conditional). I changed it so we actually timeout after 30s of waiting.

src/org/thoughtcrime/securesms/mms/IncomingLollipopMmsConnection.java
```diff
@@ -0,0 +1,105 @@
...
+      Log.w(TAG, "downloading multimedia from " + contentLocation + " to " + contentUri);
+      SmsManager.getDefault().downloadMultimediaMessage(context, contentLocation, contentUri, null, pendingIntent);
+
+      long timeoutExpiration = System.currentTimeMillis() + 30000;
+      while (!finished) {
+        Util.wait(this, Math.max(1, timeoutExpiration - System.currentTimeMillis()));
+        if (System.currentTimeMillis() >= timeoutExpiration) {
+          throw new TimeoutException("timeout when waiting for MMS");
+        }
+      }
```
moxie0: You've written this so that it's only possible for a single thread to use this at a time, and such that an instance can only be used once before it needs to be destroyed.  Personally I don't think that's great, but given that it's the case, this logic doesn't really make sense.  You call notify in exactly one place, immediately after setting finished to true.  Just do if(!finished) Util.wait(this, 30000)

If you want this to be thread-safe and reusable, that's fine with me too, but I'd like you to pick a paradigm.

mcginty: Java's own documentation instructs you to always wrap an Object.wait() with a conditional, since "interrupts and spurious wakeups are possible." This logic's not about thread safety, it's about implementing waits as correctly as possible.

moxie0: Spurious wakeups should only happen if the process receives a signal.  I guess it is entirely possible that Android uses signals in some way I don't know about.

src/org/thoughtcrime/securesms/mms/IncomingLollipopMmsConnection.java
```diff
@@ -0,0 +1,105 @@
...
+
+import ws.com.google.android.mms.MmsException;
+import ws.com.google.android.mms.pdu.PduParser;
+import ws.com.google.android.mms.pdu.RetrieveConf;
+
+public class IncomingLollipopMmsConnection extends BroadcastReceiver implements IncomingMmsConnection {
+  public static final String ACTION = IncomingLollipopMmsConnection.class.getCanonicalName() + "MMS_DOWNLOADED_ACTION";
+  private static final String TAG = IncomingLollipopMmsConnection.class.getSimpleName();
+
+  private Context context;
```
moxie0: final

src/org/thoughtcrime/securesms/mms/IncomingLollipopMmsConnection.java
```diff
@@ -0,0 +1,105 @@
...
+
+    finished = true;
+    notifyAll();
+  }
+
+  @Override
+  @TargetApi(VERSION_CODES.LOLLIPOP)
+  public synchronized RetrieveConf retrieve(String contentLocation, byte[] transactionId) throws MmsException {
+    context.getApplicationContext().registerReceiver(this, new IntentFilter(ACTION));
+    long nonce = System.currentTimeMillis();
```
moxie0: inlign

src/org/thoughtcrime/securesms/mms/IncomingLollipopMmsConnection.java
```diff
@@ -0,0 +1,105 @@
...
+    try {
+      PendingIntent pendingIntent = PendingIntent.getBroadcast(context, 1, new Intent(ACTION), PendingIntent.FLAG_ONE_SHOT);
+      Uri contentUri = ContentUris.withAppendedId(MmsBodyProvider.CONTENT_URI, nonce);
```
moxie0: align

src/org/thoughtcrime/securesms/mms/OutgoingLollipopMmsConnection.java
```diff
@@ -0,0 +1,103 @@
...
+    response = intent.getByteArrayExtra(SmsManager.EXTRA_MMS_DATA);
+    finished = true;
+    notifyAll();
+  }
+
+  @Override
+  @TargetApi(VERSION_CODES.LOLLIPOP)
+  public synchronized SendConf send(byte[] pduBytes) throws UndeliverableMessageException {
+    context.getApplicationContext().registerReceiver(this, new IntentFilter(ACTION));
+    long nonce = System.currentTimeMillis();
```
moxie0: inlign

src/org/thoughtcrime/securesms/mms/OutgoingLollipopMmsConnection.java
```diff
@@ -0,0 +1,103 @@
...
+      SmsManager.getDefault().sendMultimediaMessage(context, contentUri, null, null,
+                                                    PendingIntent.getBroadcast(context, 1, new Intent(ACTION), PendingIntent.FLAG_ONE_SHOT));
+
+      long timeoutExpiration = System.currentTimeMillis() + 30000;
+      while (!finished) {
+        Util.wait(this, Math.max(1, timeoutExpiration - System.currentTimeMillis()));
+        if (System.currentTimeMillis() >= timeoutExpiration) {
+          throw new TimeoutException("timeout when waiting for MMS");
+        }
+      }
```
moxie0: this looks familiar...

src/org/thoughtcrime/securesms/mms/IncomingLollipopMmsConnection.java
mcginty: I don't know either, but figured better safe to do it this way. Let me know if you want to just go with the simple version instead.

mcginty:
Addressed comments. Decided to create an abstract LollipopMmsConnection class that manages most of the synchronization logic and wraps it in the form of a transaction. This way the class doesn't render itself totally useless after the first transaction.

Also decided to abstract out all the content provider stuff and just have a MMS body "pointer" for temporary receptacles needed to transact with the Android API.

src/org/thoughtcrime/securesms/jobs/MmsDownloadJob.java
```diff
@@ -222,26 +164,8 @@ private void storeRetrievedMms(MasterSecret masterSecret, String contentLocation
...
-      OutgoingMmsConnection connection = new OutgoingMmsConnection(context, radio.getApnInformation(), new PduComposer(context, notifyResponse).make());
-      connection.sendNotificationReceived(usingRadio, useProxy);
-    } catch (InvalidHeaderValueException | IOException e) {
-      Log.w(TAG, e);
-    }
-  }
-
-  private void handleDownloadError(MasterSecret masterSecret, long messageId, long threadId,
-                                   int downloadStatus, String error, boolean automatic)
+  private static void handleDownloadError(Context context, MasterSecret masterSecret, long messageId, long threadId,
```
moxie0: static?

PR #2733, opened 2015-03-20
created a set of basic espresso tests.

rhodey:
Created a set of basic espresso tests.
// FREEBIE

All these espresso tests pass on my Nexus One running Android 2.3.6 and Nexus 5 running Android 5.1 (with animations disabled). Some notes from my experience with the espresso testing framework:

## out-of-app activities
The display of any activities or prompts that come from outside the app being tested trigger a failure of the current test. This means that through espresso it is not possible to set TextSecure as the default SMS app (post Kit-Kat) and depending on device, opening the notification sound preference also results in test failure. Choosing media for attachments and group images also currently shouldn't work, however, I expect that the future in-app media capture will be testable by espresso.

## Espresso.closeSoftKeyboard()
At the moment this method doesn't block, this sometimes leads to tests failing because the keyboard is covering a view under test. On emulators this reportedly isn't a problem but on real devices it is a definite confirmed issue. Two workarounds exist:

1) Thread.sleep()
2) install any 3rd party soft keyboard

Both are gross, but I think `1` is less gross. To avoid sporadic test failues use the `actuallyCloseSoftKeyboard()` method in place of `closeSoftKeyboard()` and `typeTextAndClose()` in place of `typeText()`.

## RoutingActivity makes stuff weird
Espresso kinda has this 'one activity one test' idea built in where you parameterize `ActivityInstrumentationTestCase2` with the activity you plan on testing. Then you can use this instance method `getActivity()` that returns an instance of the activity under test **and** creates it if necessary. The bummer of `RoutingActivity` is that when `getActivity()` is called you get a handle to a newly created `RoutingActivity` and not the activity you've been routed to, also if you've been routed to a new activity that activity will be destroyed. Having a handle to the activity under test is cool because it allows you to manipulate and test the activity lifecycle and get references to actual views, so, in conclusion, we can't do that stuff.

## options menus
All options menu items have to be matched using `withText()`, the more concrete `withId()` method simply doesn't work here.

## probably shouldn't fake push registration
An espresso test could be written to go through a valid registration flow but many test devices don't have SIM cards and I'm not sure we should count on internet access for anything either. Additionally I don't think we should attempt to fake a push registration state because keeping up with exactly what that state should look like with every new release is complicated and I also think it is outside the intent of espresso tests. I could be wrong about faking this state though, if it's trivial then it would definitely increase the # of things we could test, maybe this will get easier with the end of RoutingActivity.

### DIFF
--- a/androidTest/java/org/thoughtcrime/securesms/ApplicationPreferencesActivityTest.java
+++ b/androidTest/java/org/thoughtcrime/securesms/ApplicationPreferencesActivityTest.java
@@ -1,0 +1,59 @@
+package org.thoughtcrime.securesms;
+
+import android.content.Context;
+import android.test.suitebuilder.annotation.LargeTest;
+
+import org.hamcrest.Matchers;
+
+import static android.support.test.espresso.Espresso.onData;
+import static android.support.test.espresso.action.ViewActions.click;
+import static android.support.test.espresso.assertion.ViewAssertions.matches;
+import static android.support.test.espresso.matcher.PreferenceMatchers.withKey;
+import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
+
+/**
+ * rhodey
+ */
+@LargeTest
+public class ApplicationPreferencesActivityTest extends SkipRegistrationInstrumentationTestCase {
+
+  public ApplicationPreferencesActivityTest() {
+    super();
+  }
+
+  public ApplicationPreferencesActivityTest(Context context) {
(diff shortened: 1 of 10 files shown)


rhodey:
also, there is some risk here in testing too specifically and setting ourselves up for super frequent test touch-ups. interested to hear opinions on this.

rhodey:
needs major update due to routing activity changes, will reopen when fixed.

PR #2743, opened 2015-03-21
migrate to material-dialogs

agrajaghh:
This migrates all `AlertDialogs` to material-dialogs using `AlertDialogWrapper` which also improves the UI in the dark theme.
I also updated the used icons to the material ones and added dark/light versions.

Unfortunately the installation on gingerbread device was failing with `E/dalvikvm LinearAlloc exceeded capacity` after updating the material-dialogs dependency to the newest version, so I had to add some proguard rules.


mcginty:
That's really nice that their wrapper keeps this diff so small! Proguard makes me nervous. This will be a persistent problem for every nontrivial addition to the codebase until we work something out that scales better...

agrajaghh:
Yeah, I'm not happy with these small proguard changes as well... I didn't understand this error completly: I thougt the LinearAlloc limit was 5MB for all android < 4. Why is TS 2.7.1 working in my android 2.3.3 emulator but not on the android 2.3.5 device at #2753?

moxie0:
Looks like this dependency upgrades appcompat?  Didn't you explicitly not want to do that?

agrajaghh:
@moxie0 just material-dialogs is using v22, the rest should continue using v21, or did I misunderstand the system? Otherwise I'll downgrad material-dialogs to 0.6.3.5 (last version using appcompat v21)

moxie0:
@agrajaghh It doesn't work that way, there can only be one version of a class in your class loader's classpath.

agrajaghh:
ah okay, sry...

I downgraded to 0.6.3.5 which makes it possible to add less proguard rules as well. But since the ability to pass a `OnClickListener` for a custom adapter with `AlertDialogWrapper` was just added in 0.6.4.0 I couldn't use this wrapper and had to use `dialog.getListView().setOnItemClickListener()`.

moxie0:
The other thing you can do is to exclude the appcompat stuff in your dependency declaration.  That allows you to use the newer version of the dependency, but with our version of the transitive dependency.  It obviously won't work if the dependency is expecting functions or classes with signatures that don't exist in the old transitive dependency, but often times that's not the case.

agrajaghh:
@moxie0 I did that, and it seems to work! :blush: Now using the wrapper again for the attachment dialog...

What do you think about adding `failOnVersionConflict()` to the gradle file, so we don't have version confilcts by accident?

moxie0:
I'm fine with failOnVersionConflict, although the gradle-witness stuff should already prevent this from happening accidentally.

agrajaghh:
I guess you're right. I just tried it out, and it seems a bit complicated to maintain...

PR #2746, opened 2015-03-21
display the phone number in the confirm-AlertDialog during registration

agrajaghh:
The confirmation dialog in the `RegistrationActivity` currently doesn't show the phonenumber although `String.format` is used. The `%s` was missing in the string...


mcginty:
How about we set the phone number as the title so it's bold and highlighted? Then we don't have to deal with string variables either.

agrajaghh:
good idea, done. less work for the translators ;)

PR #2757, opened 2015-03-23
fix send button assets

mcginty:
1) dark-themed SMS button was used in light theme, making the "unlocked" indicator invisible.
1) removed unused resources and attribues


moxie0:
lgtm

Issue #2751, opened 2015-03-22
Zero character passphrases?


NorthwestBased:
Textsecure 2.7.1
Samsung SGH-T399
Android 4.2.2

When setting up a passphrase, textsecure accepts a passphrase that is zero characters long. From a usability standpoint and to avoid confusion, it seems like we should require passphrases to be at least one character long. What do people think?

neuhaus:
Zero characters may be OK for certain scenarios, like when you're on a solo sail tour. TextSecure could display a warning if you pick a passphrase shorter than X characters.

mcginty:
Yeah, zero character passwords shouldn't be a thing. If you're on a solo sail tour, you can just disable your password :).

PR #2762, opened 2015-03-24
check for and deny the creation of zero-length passphrases

NorthwestBased:
fixes issue #2751.


src/org/thoughtcrime/securesms/PassphraseChangeActivity.java
```diff
@@ -96,6 +96,12 @@ private void verifyAndSavePassphrases() {
                        Toast.LENGTH_SHORT).show();
         this.newPassphrase.setText("");
         this.repeatPassphrase.setText("");
+      } else if (passphrase.equals("")) {
+        Toast.makeText(getApplicationContext(),
+                       R.string.PassphraseChangeActivity_need_new_passphrase_exclamation,
+                       Toast.LENGTH_SHORT).show();
+        this.newPassphrase.setText("");
```
rhodey: should be able to drop both of the calls to `setText()` here because the if block confirmed that they're empty.

NorthwestBased:
@rhodey good catch. I removed both of the redundant calls.

PR #2782, opened 2015-03-25
Changing app appearance fixes user jumping to root settings

DustinLim:
Changing the theme or language returns the user to the root settings screen, which can be jarring. If you accidentally tapped on the wrong language this also makes it harder find your way back. 

For Android 3.0 Honeycomb and up, this can be prevented by calling `recreate()` on the Activity instead of using `finish()` and manually starting a new instance.

Possibly a fix for #2476.

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
+++ b/src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
@@ -18,6 +18,8 @@ package org.thoughtcrime.securesms;
 
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.os.Build.VERSION;
+import android.os.Build.VERSION_CODES;
 import android.os.Bundle;
 import android.preference.Preference;
 import android.support.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
```diff
@@ -113,9 +117,17 @@ public void onDestroy() {
   @Override
   public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
     if (key.equals(TextSecurePreferences.THEME_PREF)) {
-      dynamicTheme.onResume(this);
+      if (VERSION.SDK_INT >= VERSION_CODES.HONEYCOMB) {
+        recreate();
+      } else {
+        dynamicTheme.onResume(this);
+      }
```
mcginty: is there a reason not to do this conditional within DynamicTheme and DynamicLanguage?

DustinLim: For ApplicationPreferencesActivity, I had to add a if-condition in `onCreate()` for this to work. Without it, `onCreate()` would do a new FragmentTransaction on top of the restored fragment state when recreated.

In theory, putting this conditional within DynamicTheme and DynamicLanguage may require checking all other Activities and their `onCreate()` for compatibility.

rhodey:

rhodey:
hey @DustinLim can you rebase this with master? we added a bunch of integration tests and now run them against PRs automatically. this PR does not contain the build type used for the integration tests so it fails and will keep failing until rebased with master. once it has been rebased I will make sure the tests get re-run so the status icon can be updated.

thanks for helping with the project :D

DustinLim:
@rhodey I have done the rebase with master.

mcginty:
Rebased and small aesthetic tweak - in 3.5.0. Thanks! I agree it's safer to just do this in the preference instead of worrying about the different edge cases across all activities for the time being.

PR #2801, opened 2015-03-26
fix ConversationItem manipulating view in wrong thread

### LINKED ISSUE #2800: ConversationItem manipulating view in wrong thread
If there is ever an error while resolving a conversation item thumbnail then `ConversationItem.ThumbnailListener` will get a callback to `onFailure()` which will in turn trigger an uncaught `android.view.ViewRoot$CalledFromWrongThreadException` exception.

https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/ConversationItem.java#L615

call should be moved into `handler.post(...)`

rhodey:
moved manipulation of media thumbnail view into UI handler thread.

Fixes #2800
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -615,7 +615,7 @@ public class ConversationItem extends LinearLayout {
       handler.post(new Runnable() {
         @Override
         public void run() {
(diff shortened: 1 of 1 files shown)


mcginty:
even though it currently has the same effect, `.hide()` should probably be used on that image view.

PR #2810, opened 2015-03-27
Set target SDK to 22 to enable Lollipop statusbar color

axelbd:


mcginty:
It's not that easy, unfortunately. We need to fix the notification color/appearance first since bumping up the targetSdk removes that compatibility mode.

axelbd:
Ok i understand. I didn't saw that issue: i only tested it on lollipop. Thanks for your explanation.

PR #2812, opened 2015-03-27
close cursors used in getMessageRecord

mcginty:
Noticed it while skimming MmsDatabase. Should stop a cursor leak.


moxie0:
cursorloader is supposed to manage that, no?

mcginty:
Ah, it does it does. Oops.

PR #2813, opened 2015-03-27
Fix sendMessage manual MMS dialog logic

mcginty:
Previous logic

1. Didn't request manual MMS settings when there was a single recipient but an attachment was present
2. Did request manual MMS settings before the chance of an upgrade to data was possible.

New logic only requests manual MMS settings when MMS is forced so we know an upgrade isn't possible.


moxie0:
...but what happens when there is no upgrade?

mcginty:
Message will fail, and we might want to create a new failure state that requests manual APN settings.

rhodey:

PR #2817, opened 2015-03-28
Fix reacting to messages when not the default SMS provider

pR0Ps:
Since TextSecure no longer supports encrypted SMS, it shouldn't grab all incoming SMS messages and check them for encryption unless it's the default SMS provider.

This commit changes the behaviour of the app on KitKat and up to not intercept encrypted messages or key exchange messages unless it's set as the default SMS provider.

The number verification SMS still goes through in all cases.

This would allow other applications that use the TextSecure encryption protocol over SMS to coexist on the same phone as TextSecure without having TextSecure showing it's "Received a message encrypted with an old version..." every time an encrypted message is received.

### DIFF
--- a/src/org/thoughtcrime/securesms/service/SmsListener.java
+++ b/src/org/thoughtcrime/securesms/service/SmsListener.java
@@ -119,7 +119,7 @@ public class SmsListener extends BroadcastReceiver {
       return true;
     }
 
-    return WirePrefix.isEncryptedMessage(messageBody) || WirePrefix.isKeyExchange(messageBody);
+    return false;
   }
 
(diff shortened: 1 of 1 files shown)


moxie0:
We're intentionally doing this until everyone has had time to upgrade.

sarciszewski:
Not to be a bother, but have you any estimates on when that could be?

moxie0:
6 months is the upgrade window

taziden:
Why on earth would TextSecure catch SMS if it's not the default sms app ???

agrajaghh:
afaik to receive encrypted sms from old TextSecure versions...

PR #2831, opened 2015-03-30
fix notification sound silent on install

### LINKED ISSUE #2793: Notification sound doesn't work upon installation
I noticed after installation that when I receive messages, it didn't make a sound, it only flashed the notification light.
I go to Settings -> Notifications, and I can see "Sound = Argon" is displayed. I press on 'Sound', which takes me to the sound selection list, but here there is none selected. I now select Argon from the list, back in settings it says "Sound = Argon" as before, but from now the app is making notification sounds.

My colleague also had the same problem.

rhodey:
1) removed unused import.
2) corrected default value of 'pref_key_ringtone'.

Fixes #2793
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/util/TextSecurePreferences.java
+++ b/src/org/thoughtcrime/securesms/util/TextSecurePreferences.java
@@ -5,8 +5,6 @@ import android.os.Build;
 import android.preference.PreferenceManager;
 import android.util.Log;
 
-import org.thoughtcrime.securesms.R;
-
 import java.io.IOException;
 
 public class TextSecurePreferences {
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/util/TextSecurePreferences.java
```diff
@@ -378,7 +376,7 @@ public static boolean isNotificationsEnabled(Context context) {
   }
 
   public static String getNotificationRingtone(Context context) {
-    return getStringPreference(context, RINGTONE_PREF, null);
+    return getStringPreference(context, RINGTONE_PREF, "content://settings/system/notification_sound");
```
mcginty: Better to use an Android-provided Uri instead of hardcoded. Apparently there are 3 sources of this Uri, so I guess pick your favorite: http://stackoverflow.com/a/29250664/4188946

rhodey:
thanks @mcginty, done and tested.

PR #2861, opened 2015-04-02
Fix re-registering missing MasterSecret

mcginty:
There shouldn't be a scenario where one registers and the MasterSecret isn't available, so it seems best to just make it a PassphraseRequiredActionBarActivity.


moxie0:
same should be true for databaseupgradeactivity right?

PR #2864, opened 2015-04-02
remove old, unused, broken tests

rhodey:
1) removed old, unused, and broken MmsReceiverTest.
2) removed RegistrationActivityTest which got entirely broken when 8573316fdff8e978f3c7ff2a5b2cb9cf22c330a7 removed the class it inherited from (RoutedInstrumentationTestCase).

all tests in connectedCheck target now passing on Nexus One GB test device.

// FREEBIE


moxie0:
git commit messages need to be wrapped at 80 chars, and the first line should be <=76

PR #2879, opened 2015-04-03
fix message details recipient list for incoming messages

### LINKED ISSUE #2874: 'Message Details' of a received media lists your own phone number as a sender
1. Open an existing conversation (regular or group, doesn't matter)
2. Find a message with media that **someone has sent to you**
3. Tap and hold the message for 'Message Details'
4. On 'Message Details' your own phone number is listed as a sender below the text 'From:'

Expected behavior is that your own number should not be there (just like it is not with plain text messages).

Affects all supported media: picture, audio and video (ok, ok, I failed at testing with video because recording a <280kb video turned out to be too cumbersome – even with the shittiest resolution a one second video was twice as large as TextSecure's size limit :)

Device info: Nexus 4, stock Android 5.0.1, TextSecure 2.9.3
(Also tested with another device that runs on 4.0.4 and has 2.9.3)

An example of an image message:
textsecure message details sender phone

rhodey:
check that recipient list does not contain self for incoming messages.

Fixes #2874
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/MessageDetailsActivity.java
+++ b/src/org/thoughtcrime/securesms/MessageDetailsActivity.java
@@ -47,6 +47,9 @@ import org.thoughtcrime.securesms.util.DirectoryHelper;
 import org.thoughtcrime.securesms.util.DynamicLanguage;
 import org.thoughtcrime.securesms.util.DynamicTheme;
 import org.thoughtcrime.securesms.util.GroupUtil;
+import org.thoughtcrime.securesms.util.TextSecurePreferences;
+import org.thoughtcrime.securesms.util.Util;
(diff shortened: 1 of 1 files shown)


agrajaghh:
total nitpick, but the blank lines at 258 at 263 look weird to me ;)

rhodey:
@agrajaghh :P my personal coding style for try {} blocks, I'll clean it up

moxie0:
immutability should be your mantra

moxie0:
seems like this doesn't resolve the underlying issue. the sender is right on the messagerecord, but the code is looking up all the addresses in the db. just filtering out the client's address from the result feels like a hack.

rhodey:
gotta look into this more, will reopen when figured out.

PR #2885, opened 2015-04-03
Use Glide for loading part thumbnails

mcginty:
Since Glide manages its memory cache relative to the available heap and reuses Bitmaps whenever possible, we should see an improvement in memory management meaning fewer dreaded OOMEs.

As for fresco, this kind of sums it up: 
cbhuqy1wgaavhb2

Glide is less of a behemoth, more mature, and doesn't have any native libs to screw us over in ways yet unimagined. Perhaps as Fresco works out its kinks, we can re-evaluate it as an option.


mcginty:
Found an issue with loading audio/video thumbnail resources, will fix in a few.

AndroidManifest.xml
```diff
@@ -45,6 +45,9 @@
     <meta-data android:name="com.google.android.gms.version"
                android:value="@integer/google_play_services_version" />
 
+    <meta-data android:name="org.thoughtcrime.securesms.mms.TextSecureGlideModule"
+               android:value="GlideModule" />
```
moxie0: what is this?

moxie0:
Looks like the logic is still pretty spread out, and the glide stuff feels pretty unencapsulated to me.  Can't this be as simple as having a GlideImageView or whatever, that you call .setImageResource(slideDeckFuture) on or something?

AndroidManifest.xml
mcginty: When the Glide singleton initializes, it looks in the manifest metadata for `GlideModule`s, which is the config telling it not to have a disk cache.

mcginty:
In that case, we'd probably want to have a ThumbnailView that also manages opening the full media when clicked.

mcginty:
@moxie0 If we move the ThumbnailClickListener and SlideDeckListener logic into a ThumbnailView, things get kind of hairy with dealing with batch selection...

moxie0:
@mcginty I believe in you.  I don't see why ThumbnailClickListener has to move exactly, but if you want to move it you can still have ConversationItem override it just to handle the batch selection part.

mcginty:
ForegroundImageView now accepts a SlideDeck ListenableFutureTask.

moxie0:
why is there still a slidedecklistener in conversationitem?

mcginty:
@moxie0 updated, the ConversationItem doesn't do anything with SlideDeck futures any more. Added a new ThumbnailClickListener interface so the imageview can pass back the slide contents of the clicked thumbnail.

moxie0:
doesn't look like there have been any commits since our last review?

mcginty:
Updated, apologies for assigning too early!

src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
```diff
@@ -38,6 +44,7 @@ protected final void onCreate(Bundle savedInstanceState) {
     routeApplicationState(masterSecret);
     super.onCreate(savedInstanceState);
     if (!isFinishing()) {
+      Glide.get(this).register(Uri.class, InputStream.class, new EncryptedStreamUriLoader.Factory(masterSecret));
```
moxie0: this is crazy gross, there's no way to make this explicit at load time?

src/org/thoughtcrime/securesms/components/ThumbnailView.java
```diff
@@ -0,0 +1,172 @@
...
+        builder.override(thumbDimens.first, thumbDimens.second);
+      }
+    } else {
+      builder = Glide.with(getContext()).load(slide.getPlaceholderRes(getContext().getTheme()))
+                                        .fitCenter()
+                                        .crossFade();
+    }
+
+    return builder.error(R.drawable.ic_missing_thumbnail_picture);
+  }
```
moxie0: this looks like it needs to be broken up

src/ws/com/google/android/mms/pdu/PduPart.java
```diff
@@ -136,6 +127,7 @@
      private boolean isPendingPush;
      private long    dataSize;
      private Bitmap  thumbnail;
+     private float   aspectRatio;
```
moxie0: why is this here?

src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
mcginty: I agree, but it needs to be available when the Glide operations occur. At first I was doing it in each ThumbnailView instance, passing it along and re-registering this loader factory each time which seemed horribly inefficient.

Right now, there's no way to pass in a Uri loader factory explicitly to a request.

moxie0: I'm not familiar with the API, but I don't think this is an acceptable approach.  I'd even prefer encoding it in the URL.

src/ws/com/google/android/mms/pdu/PduPart.java
mcginty: I use it now to instruct glide on what the final image size should be.

mcginty:
MasterSecret is now passed in via a custom Glide loader model, which is much cleaner. Didn't know it was possible until I dug a bit deeper.

src/ws/com/google/android/mms/pdu/PduPart.java
moxie0: an aspect ratio doesn't strike me as something a PduPart has

mcginty: well now it doesn't :) found out I can make a custom transformation to do the same job

mcginty:
reverted changes to PartDatabase and PduPart, removed debug statements that were too verbose

moxie0:
in 2.13.0

rhodey:
@mcginty what of #3028? should this also be resolved now?

PR #2887, opened 2015-04-04
Preserve shortcuts to old launcher activity

### LINKED ISSUE #2878: Update to 2.10.0 deletes Icon from Android favs
Not a really huge bug, but I'll report anyway: 
The most recent update deleted the TextSecure icon from my app bar. First time something like this happened. Re-adding succesful

Running on Cyanogenmod 11.

veeti:
Add an activity alias for RoutingActivity that points to the new launcher activity. Fixes #2878.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -95,14 +95,19 @@
     <activity android:name=".ConversationListActivity"
           android:label="@string/app_name"
           android:launchMode="singleTask"
-          android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize">
+          android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"
+          android:exported="true" />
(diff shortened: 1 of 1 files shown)


mcginty:
Whoa, that's a thing? Thanks so much for posting this.

moxie0:
@mcginty You're good with this?

Exporting an activity doesn't have any side effects we care about?

veeti:
FWIW, an activity with filters is supposed to be implicitly exported:

> [The default value depends on whether the activity contains intent filters. The absence of any filters means that the activity can be invoked only by specifying its exact class name. This implies that the activity is intended only for application-internal use (since others would not know the class name). So in this case, the default value is "false". On the other hand, the presence of at least one filter implies that the activity is intended for external use, so the default value is "true".](https://developer.android.com/guide/topics/manifest/activity-element.html)

I don't think this will have any side effects.

PR #2895, opened 2015-04-04
Updated gradle dependencies versions, hashes, targetSdkVersion and some files

adrianchifor:
- Updated build.gradle dependencies library versions where possible
- Re-ran calculateChecksums and updated dependencyVerification hashes ([terminal screenshot](https://cloud.githubusercontent.com/assets/6370064/6993312/f5388d12-daf8-11e4-9f18-02bed16e7b45.png))
- Bump targetSdkVersion from 19 to 22
- Updated some project files with new path for new version of com.makeramen.roundedimageview library

Did a ./gradlew build and the build was successful.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,14 +31,14 @@ repositories {
 }
 
 dependencies {
-    compile 'me.leolin:ShortcutBadger:1.0.2-WS2'
-    compile 'se.emilsjolander:stickylistheaders:2.2.0'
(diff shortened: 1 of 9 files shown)


moxie0:
Thanks but we can't update the target SDK until we address the things that the compatibility layer is giving us right now, we can't update shotcutbadger until we update our fork and publish our own artifact, and in general we don't want to update dependencies unless we have a reason to -- and then have thoroughly tested each change.

PR #2896, opened 2015-04-04
Updated some exception object identifiers

adrianchifor:
Updated "_" (underscore) exception object identifiers to something more conventional (‘e’) as the underscore identifier might not be supported in releases after Java 8.

This will also fix some of the warnings given when building TextSecure. ([terminal screenshot](https://cloud.githubusercontent.com/assets/6370064/6993521/a396369c-daff-11e4-8ad1-ca33690454bb.png))


moxie0:
This isn't our code. We don't want to modify it in case we need to pull in changes.

PR #2906, opened 2015-04-06
Fix identity verification screen design issues

### LINKED ISSUE #2904: Text in "Verify identity" screen gets cut off when system font size is set to maximum
Hello,

in TextSecure 2.9.3, the text in the "Verify identity" screen gets cut off when the system font size is set to maximum.

The text goes "behind" the blue bar at the top of the screen (the blue bar which has the "Verify identity" and the barcode logo on it).

Could you please fix it?

Regards

veeti:
- Corrected improper usage of ScrollView that would cut off contents when taller than the layout.

- Adjusted margins and paddings to match Material specs (20dp from sides).

- Fixed monospaced font being overridden on Lollipop.

Fixes #2904. FREEBIE.


mcginty:
:+1: lgtm on both lollipop and below

PR #2918, opened 2015-04-08
fix proguardDebugTest build target

### LINKED ISSUE #2871: Unable to run instrumentation tests due to proguard
Running `./gradlew proguardDebugTest` now fails (probably an issue with the [refactor](https://github.com/WhisperSystems/TextSecure/commit/597f7bbafd2ff7ef2779ca5292a2fc8ef568d489)).

The log is huge so I've put it on pastebin: http://pastebin.com/rkQ0BxUZ

rhodey:
added a new set of proguard rules to fix the 'proguardDebugTest' target.

Fixes #2871
// FREEBIE

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -167,6 +167,7 @@ android {
                           'proguard-square-okio.pro',
                           'proguard-spongycastle.pro',
                           'proguard-rounded-image-view.pro',
+                          'proguard-testing.pro',
                           'proguard.cfg'
         }
         release {
--- a/proguard-testing.pro
+++ b/proguard-testing.pro
@@ -1,0 +1,10 @@
+-dontobfuscate
+-dontskipnonpubliclibraryclassmembers
+
+-dontwarn android.test.**
+-dontwarn com.android.support.test.**
+-dontwarn sun.reflect.**
+-dontwarn org.assertj.**
+-dontwarn org.hamcrest.**
+-dontwarn org.mockito.**
+-dontwarn com.squareup.**


mcginty:
Do the tests actually run for you?

mcginty:
I believe `-dontobfuscate` will break the LGE menu bug since we rely on renaming that conflicting class.

rhodey:
All tests run and pass on my GB Nexus One with these changes. I wasn't aware of that LGE bug and unfortunately already spent **a long** time trying to do without `-dontobfuscate`-- but alas, seems I must keep at it.

mcginty:
@rhodey another option which I haven't tried and don't know if it will work is to keep `-dontobfuscate` but add a rule specifically for that LGE menu class that explicitly allows obfuscation (I think it proguard it looks like `-keep,allowobfuscation`)

rhodey:
@mcginty I'll give that a shot, interestingly enough you can drop the `-dontobfuscate` line from this diff and the `proguardDebugTest` build target still passes fine, however `connectedCheck` fails ceaselessly with

```
com.android.builder.testing.ConnectedDevice > hasTests[...device name...] FAILED 
No tests found. 
```

`connectedCheck` is all I really care about since I'm trying to revive my espresso branch :|

rhodey:
@mcginty no luck with the `allowobfuscation` flag, seems to have no affect when used in combination with `-dontobfuscate`, as verified by diffing `mapping.txt`.

mcginty:
laaame

rhodey:
This commit could still go through and not break release, but the downside of this is that we'd have two different sets of proguard rules, and only one of these (debug) would ever be tested.

I've spent all day debugging and reading up on this and [this bug report](https://code.google.com/p/android/issues/detail?id=77475) finally has me convinced that what we want to do is not currently possible in gradle, where what we want to do is run instrumentation tests on a proguard build of TS where the proguard rules match what is used for the release target.

rhodey:
fixes `proguardDebugTest` but not `connectedCheck`, will reopen when works with the latter.

PR #2922, opened 2015-04-08
remove overly-verbose sqlite query logging

mcginty:
It's always in the logs, and I've never found it useful for debugging. Opening as a PR if I'm the only one :).


moxie0:
I kinda like it just because it gives me some perspective on what the user is doing in the app at those points in the log. I agree we could probably accomplish that differently.

Issue #2740, opened 2015-03-21
Crash on message receive


merkste:
I was reading a conversation as two messages arrived and TS crashed instantly. One of them was for the conversation I was looking into. Crash log is here: https://gist.github.com/dacb1a8a3ef662a2921f

rhodey:
resolved by moving view operations to the UI thread in some commit I cannot find reference to.

PR #2931, opened 2015-04-09
moved more view manipulation into UI thread

rhodey:
issue #2740 contains a strange java.lang.IllegalStateException, this exception resulted in a crash because SlideDeckListener was attempting to change the visibility of a view while outside the UI thread.

// FREEBIE

see debug log [here](https://gist.github.com/anonymous/dacb1a8a3ef662a2921f)
references crash at [this line](https://github.com/WhisperSystems/TextSecure/blob/e0737451db1cb52209c5ae6a00521551be5b728d/src/org/thoughtcrime/securesms/ConversationItem.java#L637)

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -638,13 +638,18 @@ public class ConversationItem extends LinearLayout {
           return;
         }
       }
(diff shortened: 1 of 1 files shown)


moxie0:
this should be obviated by @mcginty's pending glide changes

PR #2946, opened 2015-04-11
fix emoji sort order to be last in first out

### LINKED ISSUE #2967: Emoji drawer recent category is reversed
Hello,

the recent category in TextSecure for Android is reversed.

Recently used Emojis are being added to the bottom right (the end) of the list.

This is counter intuitive and seems illogical, since you need to scroll down to the bottom of the list to pick a recent Emoji.

Every other keyboard or built-in Emoji drawer I know does it the other way around.

Other keyboards and built-in Emoji drawers add recent Emojis to the top left (the beginning) of the list.

So, could you please fix it in the TextSecure Emoji drawer?

Regards

rhodey:
EDIT: the recent emoji tab was sorting emoji as last in last out instead of last in first out. also, if an emoji was already in the recent emoji cache its order would not be updated on reinsert-- for example, [A, B, C, B] resulted in the ordered set [A, B, C] instead of [A, C, B].

1) reverse array index used in recent emoji getter.
2) remove duplicate entries to the recent emoji cache before inserting.

Fixes #2967
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/util/Emoji.java
+++ b/src/org/thoughtcrime/securesms/util/Emoji.java
@@ -330,9 +330,7 @@ public class Emoji {
         prefs = PreferenceManager.getDefaultSharedPreferences(context);
       }
 
-      if (recentlyUsed.contains(asset)) {
-        recentlyUsed.remove(asset);
-      }
+      recentlyUsed.remove(asset);
       recentlyUsed.add(asset);
 
       if (recentlyUsed.size() > EMOJI_LRU_SIZE) {
--- a/src/org/thoughtcrime/securesms/util/Emoji.java
+++ b/src/org/thoughtcrime/securesms/util/Emoji.java
@@ -263,7 +263,8 @@ public class Emoji {
   }
 
   public Pair<Integer, Drawable> getRecentlyUsed(int position, double size, PageLoadedListener pageLoadedListener) {
-    String code = EmojiLRU.getRecentlyUsed(context)[position];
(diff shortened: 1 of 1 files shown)


moxie0:
This is already a mess, let's not pile another hack on top of it. Converting the list to an array for every index is o(n), just iterating to the position in the list would actually be faster. If you want to reverse it, just invert the index.

rhodey:
yup :| reversing the index is a much better idea. fixed that, rebased, and corrected my commit message.

src/org/thoughtcrime/securesms/util/Emoji.java
```diff
@@ -329,6 +330,9 @@ public static void putRecentlyUsed(Context context, String asset) {
         prefs = PreferenceManager.getDefaultSharedPreferences(context);
       }
 
+      if (recentlyUsed.contains(asset)) {
+        recentlyUsed.remove(asset);
+      }
```
moxie0: this is necessary?

rhodey: yeah, it's gross, especially on a set, but here is the example:

emoji seq >> :), :(, :D, :)
without that conditional remove, the ordered set is >> :), :(, :D
with the conditional remove, the ordered set it >> :(, :D, :)

so, in summary, this implementation of ordered set does not update order for items already inserted.

moxie0:
then just call remove, the conditional makes performance worse not better

rhodey:
yup -__- done, gotta work on my O

moxie0:
In 2.13.0

Issue #2769, opened 2015-03-24
Font size in message input box is larger than font size in messages


wp9015362:
Hello,

in TextSecure-Android, the font size in the message input box is different than the font size used for messages.

Could you please make the message input box font size be equal to the message font size?

IMHO the font size for what I am typing (message input box) should not differ from what I am reading (messages).

Regards

jeremymasters:
For the record, mine looks like the same size in both messages and the input box....galaxy s4, stock 4.4.2 using swiftkey.

wp9015362:
On LG G2 with stock KitKat 4.4.2 ROM for example, the message input box font size always is larger than the message font size. Regardless of the keyboard and regardless of the system font size setting.

wp9015362:
@mcginty :

I see you closed a pull request which was about to fix this issue a few months ago:

https://github.com/WhisperSystems/TextSecure/issues/2969

Since you have fixed https://github.com/WhisperSystems/TextSecure/issues/3317 now (thanks for that), could you now also please have a look at this issue?

I still think the font size in the message input box should not be larger than the font size in the messages. The font size should be equal for both.

When the font size in the message input box is larger than the font size in the messages, then the eyes always have to re-adjust between writing a message and reading a message.

And I also noticed something else:

When the message input box is empty, then there is a text in there in light gray which says "Send TextSecure message".

The font size of this text also is different.

Could you please fix that and make the message input box always have the same font size as the messages?

It would be much appreciated.

Regards

andrewsawczyn:
Should this not be an easy fix?  Still unresolved 2 years later. Problem still exists with 3.29.6 and Android 7.1.

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #2969, opened 2015-04-13
Changed bubble message font size to match the input text message.

killickjoe:
This should fix the issue #2769, where the input text message size was different to the actual message (sent and received). Tested on several different devices.

### DIFF
--- a/res/layout/conversation_bubble_incoming.xml
+++ b/res/layout/conversation_bubble_incoming.xml
@@ -50,7 +50,7 @@
                   android:paddingRight="10dp"
                   android:textAppearance="?android:attr/textAppearanceSmall"
                   android:textColor="?conversation_received_text_primary_color"
-                  android:textSize="16sp"
+                  android:textSize="18sp"
                   android:autoLink="all"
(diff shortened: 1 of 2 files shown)


mcginty:
that looks pretty massive to me

mcginty:
I think we should just make the composition text smaller.

agrajaghh:
actually, I don't mind it a bit bigger. Whatsapp has it one or two sizes bigger as the bubbles as well..

mcginty:
WhatsApp has it a bit bigger, Messenger has it about the same size. I don't have strong opinions either way on composition font size, other than I don't want to make the clickzone of the composition any smaller.

killickjoe:
If it'd help, we can set a default size, then use volume button listener to grow or shrink size as user desires.

mcginty:
I don't think we should override controls in a way that violate users' assumptions. Personally, I like having the ability to control my music volume while still in a TextSecure conversation :). Going to close this issue for now since I don't think it's quite what we want.

Thanks!

PR #2884, opened 2015-04-03
fix forward message to contact for which no conversation exists

rhodey:
set list adapter to null when (recipients or threadId == null) to prevent old messages from being displayed when forwarding a message to a recipient for which there is no existing conversation.

Fixes #2883
// FREEBIE

PR #2977, opened 2015-04-13
don't change cursor if adapter is null

mcginty:
Addresses #2970

I wasn't able to find a way to reproduce this. This fix is just a nullcheck, but there's still the problem of this loader being called when the list adapter wasn't initialized. @rhodey do you know why `ConversationFragment`'s `onNewIntent` doesn't check for recipients being null, while it does in `initializeListAdapter`?

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -283,12 +283,16 @@ public class ConversationFragment extends ListFragment
 
   @Override
   public void onLoadFinished(Loader<Cursor> arg0, Cursor cursor) {
-    ((CursorAdapter)getListAdapter()).changeCursor(cursor);
+    if (getListAdapter() != null) {
+      ((CursorAdapter) getListAdapter()).changeCursor(cursor);
+    }
   }
 
   @Override
   public void onLoaderReset(Loader<Cursor> arg0) {
-    ((CursorAdapter)getListAdapter()).changeCursor(null);
+    if (getListAdapter() != null) {
+      ((CursorAdapter) getListAdapter()).changeCursor(null);
+    }
   }
 
   public interface ConversationFragmentListener {


rhodey:
@mcginty I agree that adding the null recipient check to the `onNewIntent()` method may resolve this bug. I'm gonna spend some time trying to reproduce this also.

mcginty:
Cool, will assign to you since you're checking that out

rhodey:
not so sure the loader call in `onNewIntent` is the cause of this anymore. the `-1` check was added in #2884 and just remembered that I felt good about adding that restart loader call there because we used to do it unconditionally [like so](https://github.com/rhodey/TextSecure/blob/ca2c8b961e9b756be6c3478f423a8f77eb102f1c/src/org/thoughtcrime/securesms/ConversationFragment.java#L106)

rhodey:
Some reading...

[LoaderManager.onLoadFinished()](http://developer.android.com/reference/android/app/LoaderManager.LoaderCallbacks.html#onLoadFinished(android.content.Loader<D>, D))]
```
... Note that normally an application is not allowed to commit fragment transactions while in this call, since it can happen after an activity's state is saved ...
```

So, my interpretation of this is that it's possible for the LoaderManager to start a load and not finish until after `onStop()`-- where the ListView and it's adapter may have been released by the system but not the loader. Reading this I understand why we should add a null check to `onLoadFinished()`, nothing I've read here hints that we should need one in `onLoaderReset()` but might as well?

@mcginty looking good imo

rhodey:
It is also possible that 1f9a0fab3ce66ad6ddf4de97e657a2c3027a8714 resolves the issue. I can imagine a situation where `sendComplete()` is called by an AsyncTask after leaving `ConversationActivity`-- this could then lead to `ConversationFragment.initializeListAdapter()` being called which would restart the loader and eventually result in a callback to `onLoadFinished()`

rhodey:
Some notes on getting to the bottom of this bug, all observations based on b692d48b98eabd7f7c8c04880b977fb7b049e9a2

1) loader callbacks such as `onLoadFinished` could only ever result from a call to `LoaderManager.restartLoader()`
2) this only happens in [onNewIntent()](https://github.com/WhisperSystems/TextSecure/blob/b692d48b98eabd7f7c8c04880b977fb7b049e9a2/src/org/thoughtcrime/securesms/ConversationFragment.java#L108) and [initializeListAdapter()](https://github.com/WhisperSystems/TextSecure/blob/b692d48b98eabd7f7c8c04880b977fb7b049e9a2/src/org/thoughtcrime/securesms/ConversationFragment.java#L123)

## onNewIntent
if `onNewIntent()` is called before the list adapter is initialized **and** `threadId == -1` I might expect this bug to surface. I'm skeptical that this is the cause of this bug only because not too long ago we used to reset the loader in `onNewIntent()` unconditionally [like so](https://github.com/rhodey/TextSecure/blob/ca2c8b961e9b756be6c3478f423a8f77eb102f1c/src/org/thoughtcrime/securesms/ConversationFragment.java#L106). So if anything I'd expect this bug to surface less often now in this case because there is an added conditional.


## initializeListAdapter
`initializeListAdapter()` is called from `onActivityCreated()`, `onNewIntent()`, and conditionally in `reload()` **only if 

start new conversation
go home
receive notification from diff conv
open notification

rhodey:
Some notes on getting to the bottom of this bug, all observations based on b692d48b98eabd7f7c8c04880b977fb7b049e9a2

1) loader callbacks such as `onLoadFinished()` could only ever result from a call to `LoaderManager.restartLoader()`
2) this only happens in [onNewIntent()](https://github.com/WhisperSystems/TextSecure/blob/b692d48b98eabd7f7c8c04880b977fb7b049e9a2/src/org/thoughtcrime/securesms/ConversationFragment.java#L108) and [initializeListAdapter()](https://github.com/WhisperSystems/TextSecure/blob/b692d48b98eabd7f7c8c04880b977fb7b049e9a2/src/org/thoughtcrime/securesms/ConversationFragment.java#L123)

## onNewIntent
if `onNewIntent()` is called before the list adapter is initialized **and** `threadId == -1` I would expect this bug to surface. I'm skeptical that this is the cause of this bug because not too long ago we used to restart the loader in `onNewIntent()` unconditionally [like so](https://github.com/rhodey/TextSecure/blob/ca2c8b961e9b756be6c3478f423a8f77eb102f1c/src/org/thoughtcrime/securesms/ConversationFragment.java#L106). Also I cannot think of a case where these conditions would be met.

## initializeListAdapter
`initializeListAdapter()` is called from `onActivityCreated()`, `onNewIntent()`, and conditionally in `reload()` **only if `threadId` has changed**. I can't imagine that any of these calls should cause a problem because `initializeListAdapter()` inits the ListAdapter and touches ListView before restarting the loader so there definitely shouldn't be an NPE in loader callbacks.

So, whats left?

Well I believe this can still be explained by a super long load completing only after the activity has stopped, or an activity stopping super quick before a whatever-length load completes. [LoaderManager documentation](http://developer.android.com/reference/android/app/LoaderManager.LoaderCallbacks.html#onLoadFinished(android.content.Loader%3CD%3E,%20D)) seems to claim that this is possible.

Also I've heard mention of some weird behavior where `onNewIntent()` is called after `onFinish()`-- only if this was to happen could I imagine a scenario where the conditions described in `onNewIntent()` would possibly surface. Imaginary scenario:

1) start conversation with recipient for whom no thread exists
2) `threadId == -1` so `initializeListadapter()` skips actual initialization
3) activity is finished somehow
4) start conversation with a different recipient for whom no thread exists
5) onNewIntent()
6) `threadId == -1` so restart loader --> NPE

FWIW... :|

PR #2982, opened 2015-04-14
fix the connectedCheck build target

### LINKED ISSUE #2871: Unable to run instrumentation tests due to proguard

rhodey:
1) only obfuscate the package android.support.v7.internal.view.menu to prevent LGE ROM bug.
2) '-keepattributes Exceptions' to allow for throwing from mocks
3) -dontskipnonpubliclibraryclassmembers and -dontwarn for everything else

Fixes #2871
// FREEBIE

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -167,6 +167,7 @@ android {
                           'proguard-square-okio.pro',
                           'proguard-spongycastle.pro',
                           'proguard-rounded-image-view.pro',
+                          'proguard-testing.pro',
                           'proguard.cfg'
         }
         release {
--- a/proguard-appcompat-v7.pro
+++ b/proguard-appcompat-v7.pro
@@ -1,7 +1,8 @@
+# only obfuscate this one pagacke because LGE ROM bug
+-keepnames class !android.support.v7.internal.view.menu.**, ** { *; }
+
 -keep public class android.support.v7.widget.** { *; }
 -keep public class android.support.v7.internal.widget.** { *; }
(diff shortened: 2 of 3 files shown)


rhodey:
It should be noted that this pull moves us into a non-ideal situation, where, though our tests may pass we now have a different set of proguard rules for our debug VS our release target-- at the time of writing this is unavoidable because of how gradle applies proguard rules to instrumentation tests.

Instrumentation tests actually result in two output APKs, one is the debug APK (assembleDebug), the other is a test runner APK. Ideally the proguard changes made by this build could be applied exclusively to the test runner APK **but** this is not currently possible. gradle will always apply the debug proguard rules to the test runner APK, therefor if you want to apply rules to the test runner they will also be applied to the debug APK.

Still beat no tests though, IMO :)

rhodey:
also, you can check this out, run gradle build, and check the various mappings.txt files output by proguard into your build dirs. they contain the mapping of obfuscated classes, so I've checked here to verify that we have no regressions related to the LGE ROM bug.

moxie0:
i'd still like you to do a *different* variant that's not debug for this, debug and release need to be the same

rhodey:
oops, forgot about that-- will do.

PR #2988, opened 2015-04-14
support repeating keyevents for emoji backspace

### LINKED ISSUE #2945: Long-pressing emoji backspace deletes only one character
**Another lovely emoji keyboard related issue:**
If you want to remove ten characters with the emoji drawer's backspace you have to tap it ten times. Long press does not work.
Compare to the regular keyboards' backspace which you can tap and hold to remove everything at once.

Tested with: TS 2.11.0, Android 5.0.1, Android 4.0.4

mcginty:
Fixes #2945


McLoo:
@moxie0 Not sure about how binding the milestones are, but the according issue is set the 2.12.0

moxie0:
this needs to be encapsulated

mcginty:
encapsulated into a reusable KeyRepeater

src/org/thoughtcrime/securesms/components/EmojiDrawer.java
```diff
@@ -86,7 +90,7 @@ private void initializeResources() {
     this.backspace = (ImageButton              ) findViewById(R.id.backspace);
     this.emoji     = Emoji.getInstance(getContext());
 
-    this.backspace.setOnClickListener(new BackspaceClickListener());
+    KeyRepeater.attach(backspace, composeText, KeyRepeater.DELETE_KEY_EVENT);
```
moxie0: the view's logic needs to be encapsulated in the view.  this is the emojidrawer, which shouldn't know anything about backspace repeating.  it should have a reference to an EmojiBackspace or whatever, which allows you to set an onClickListener that fires whenever there's a backspace event.  It should be responsible for making sure those events happen when appropriate (repeating, or on a single click).

mcginty:
done

moxie0:
lgtm

moxie0:
in 2.14.0

PR #2994, opened 2015-04-16
Use REST interface instead of WebSockets on GCM events.

moxie0:
// FREEBIE


src/org/thoughtcrime/securesms/jobs/PushReceivedJob.java
```diff
@@ -105,4 +64,5 @@ private boolean isActiveNumber(Context context, String e164number) {
     return isActiveNumber;
   }
 
+
```
mcginty: straggling space

mcginty:
lgtm, haven't run it

PR #3005, opened 2015-04-16
prevent NPE from GooglePlayServicesUtil

### LINKED ISSUE #2505: Crash Browser - NPE
`````
java.lang.RuntimeException: Unable to start activity ComponentInfo{org.thoughtcrime.securesms/org.thoughtcrime.securesms.PlayServicesProblemActivity}: java.lang.NullPointerException
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2215)
at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2265)
at android.app.ActivityThread.access$800(ActivityThread.java:145)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1206)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:136)
at android.app.ActivityThread.main(ActivityThread.java:5116)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:515)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:781)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:597)
at dalvik.system.NativeStart.main(Native Method)
Caused by: java.lang.NullPointerException
at android.support.v4.app.DialogFragment.onActivityCreated(DialogFragment.java:381)
at android.support.v4.app.Fragment.performActivityCreated(Fragment.java:1794)
at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:967)
at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1126)
at android.support.v4.app.BackStackRecord.run(BackStackRecord.java:739)
at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1489)
at android.support.v4.app.FragmentActivity.onStart(FragmentActivity.java:548)
at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1171)
at android.app.Activity.performStart(Activity.java:5241)
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2178)
... 11 more
`````

@mcginty You know what this could be?

### PR DESCRIPTION
**Pull request by rhodey**
GooglePlayServicesUtil.getErrorDialog() returns null when passed the error code ConnectionResult.SERVICE_INVALID so we have to provide our own error dialog for this. Error message taken directly from android documentation on this code.

1) correct placement of @NonNull annotation
2) create new string resource for service invalid error message
3) return custom dialog for ConnectionResult.SERVICE_INVALID

Fixes #2505
// FREEBIE

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -672,7 +672,7 @@
     <string name="message_details_header__with">With:</string>
 
     <!-- PlayServicesProblemFragment -->
-    <string name="PlayServicesProblemFragment__google_play_services_is_not_authentic">The version of the Google Play services installed on this device is not authentic.</string>
+    <string name="PlayServicesProblemFragment__please_install_an_authentic_version_of_google_play">Please install an authentic version of Google Play Services and try again.</string>
 
     <!-- AndroidManifest.xml -->
     <string name="AndroidManifest__create_passphrase">Create passphrase</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -671,6 +671,9 @@
     <string name="message_details_header__from">From:</string>
     <string name="message_details_header__with">With:</string>
 
+    <!-- PlayServicesProblemFragment -->
+    <string name="PlayServicesProblemFragment__google_play_services_is_not_authentic">The version of the Google Play services installed on this device is not authentic.</string>
+
     <!-- AndroidManifest.xml -->
     <string name="AndroidManifest__create_passphrase">Create passphrase</string>
     <string name="AndroidManifest__enter_passphrase">Enter passphrase</string>
--- a/src/org/thoughtcrime/securesms/PlayServicesProblemFragment.java
+++ b/src/org/thoughtcrime/securesms/PlayServicesProblemFragment.java
@@ -18,13 +18,13 @@
 package org.thoughtcrime.securesms;
 
 import android.app.Dialog;
+import android.content.DialogInterface;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.v4.app.DialogFragment;
 import android.util.Log;
 
 import com.afollestad.materialdialogs.MaterialDialog;
-import com.afollestad.materialdialogs.MaterialDialog.ButtonCallback;
 import com.google.android.gms.common.ConnectionResult;
 import com.google.android.gms.common.GooglePlayServicesUtil;
 
@@ -33,15 +33,7 @@ public class PlayServicesProblemFragment extends DialogFragment {
   private Dialog getPlayServicesInvalidDialog() {
     return new MaterialDialog.Builder(getActivity())
                .content(R.string.PlayServicesProblemFragment__please_install_an_authentic_version_of_google_play)
-               .cancelable(false)
-               .positiveText(android.R.string.ok)
-               .callback(new ButtonCallback() {
-                 @Override
-                 public void onPositive(MaterialDialog dialog) {
-                   super.onPositive(dialog);
-                   getActivity().finish();
-                 }
-               })
+               .negativeText(android.R.string.ok)
                .build();
   }
 
@@ -60,6 +52,18 @@ public class PlayServicesProblemFragment extends DialogFragment {
         Log.w(getClass().getName(), "received error code " + code);
         return getPlayServicesInvalidDialog();
     }
+  }
+
+  @Override
+  public void onCancel(DialogInterface dialog) {
+    super.onCancel(dialog);
+    getActivity().finish();
+  }
+
+  @Override
+  public void onDismiss(DialogInterface dialog) {
+    super.onDismiss(dialog);
+    getActivity().finish();
   }
 
 }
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by rhodey** [comment]
tested by starting `PlayServicesProblemActivity` onCreate of `NewConversationActivity` and overriding error code to always equal `ConnectionResult.SERVICE_INVALID` in `PlayServicesProblemFragment`

**Comment by moxie0** [comment]
never display an error message to the user without telling them how to resolve it. also consider the language you use and how people will interpret it.

**Comment by rhodey** [comment]
This error occurs because a signature check on google play services fails, which means:

1) user has flashed custom junk onto their device
2) device comes from a manufacturer who did not pay licensing fees

But most likely case 1, lots of searching around the net supports this, I found no documented cases of the later. To resolve this issue the user needs to uninstall the junk GPS from their device and install a valid copy. The steps to resolve are difficult to explain non-technically but I am certain this only shows up in the above situations so we shouldn't be confusing anyone who didn't bring this on themselves.

`Please install an authentic version of Google Play Services and try again.` ?

File: src/org/thoughtcrime/securesms/PlayServicesProblemFragment.java [hunk 1]
```diff
@@ -21,15 +21,45 @@
...
+               })
+               .build();
+  }
+
   @Override
-  public Dialog onCreateDialog(@NonNull Bundle bundle) {
+  @NonNull
+  public Dialog onCreateDialog(Bundle bundle) {
     int code = GooglePlayServicesUtil.isGooglePlayServicesAvailable(getActivity());
-    return GooglePlayServicesUtil.getErrorDialog(code, getActivity(), 9111);
```
moxie0: Leave this here, if it's null, then build your own dialog.

src/org/thoughtcrime/securesms/PlayServicesProblemFragment.java
```diff
@@ -21,15 +21,45 @@
...
+import com.afollestad.materialdialogs.MaterialDialog.ButtonCallback;
+import com.google.android.gms.common.ConnectionResult;
 import com.google.android.gms.common.GooglePlayServicesUtil;
 
 public class PlayServicesProblemFragment extends DialogFragment {
 
+  private Dialog getPlayServicesInvalidDialog() {
+    return new MaterialDialog.Builder(getActivity())
+               .content(R.string.PlayServicesProblemFragment__please_install_an_authentic_version_of_google_play)
+               .cancelable(false)
```
moxie0: why?

src/org/thoughtcrime/securesms/PlayServicesProblemFragment.java
```diff
@@ -21,15 +21,45 @@
...
+               .positiveText(android.R.string.ok)
+               .callback(new ButtonCallback() {
+                 @Override
+                 public void onPositive(MaterialDialog dialog) {
+                   super.onPositive(dialog);
+                   getActivity().finish();
```
moxie0: Is this necessary? It's not going to be present in the dialog returned by GoolgePlayServicesUtil, right?

res/values/strings.xml
```diff
@@ -671,6 +671,9 @@
     <string name="message_details_header__from">From:</string>
     <string name="message_details_header__with">With:</string>
 
+    <!-- PlayServicesProblemFragment -->
+    <string name="PlayServicesProblemFragment__please_install_an_authentic_version_of_google_play">Please install an authentic version of Google Play Services and try again.</string>
```
moxie0: try not using the word authentic

src/org/thoughtcrime/securesms/PlayServicesProblemFragment.java
rhodey: I discovered that the `finish()` was necessary through testing, else you're just left with a transparent activity when closing the dialog and a tap of the back button is required.

"It's not going to be present in the dialog returned by GoolgePlayServicesUtil, right?" not exactly sure what you mean by this but it gives me the idea of testing if the user is left with a transparent activity after closing the google dialog-- will check.

src/org/thoughtcrime/securesms/PlayServicesProblemFragment.java
rhodey: Android's own documentation says that `GooglePlayServicesUtil.getErrorDialog()` should only ever be called with `SERVICE_MISSING`, `SERVICE_VERSION_UPDATE_REQUIRED`, or `SERVICE_DISABLED`. Additionally all public knowledge on this subject says that null will always be returned for `SERVICE_INVALID`. I don't think that's a good idea.

res/values/strings.xml
rhodey: OK, I'd like to discuss this here then commit something that we agree upon instead of guessing in commits.

`Please install an authentic version of Google Play Services and try again.`
`Please install a valid version of Google Play Services and try again.`
`Please install an valid build of Google Play Services and try again.`

The first is still my favorite, the most factually correct, and the most easily google-able. Any suggestions?

moxie0:
in 2.13.0

Issue #1699, opened 2014-07-13
Add Android Wear support


cypressious:
Hangouts has the possibility to answer with predefined answers or voice. Something similar would be awesome for TextSecure. Since all this works over notifications, this should be reasonably easy to intergrate.

kristovatlas:
Wear support would be unbelievably cool...

vinilox:
I agree, Android wear support would definitely be cool !

espeed:
+1

midi:
This was merged a while back: https://github.com/WhisperSystems/TextSecure/commit/424a463b21869b9afa674430709a4dc3e03be31c

But sadly has no option to reply.
Please consider adding that, as it's super useful.

joeykrim:
Should this be closed, as https://github.com/WhisperSystems/Signal-Android/pull/3244 was merged allowing replies from Android Wear?

lightmaster:
This Feature Request is several years old now, are there any plans to add a native Android Wear app? I would really like to see an app where you can view your messages, reply to a chat that's not in your notifications, start a new chat, etc. It's very annoying how, currently, you can reply to a notification on Android Wear, but if you want to send a second message, you have to either open up Signal on your phone, or wait until they reply back.

darkpixel:
I'm a bit late to the Android Wear game, but I'd love to see support too.  I occasionally get messages while out for a walk or run and would love to tape on my watch and send a voice reply.

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #3010, opened 2015-04-17
Added Wearable-Actions to Notifications

chr1shaefn3r:
To better support wearable devices, I also added the notification actions via the "WearableExtender()" to the NotificationCompat.Builder.

This should give all the Android Wear-compatible device owner the possibility to mark incoming messages as read without the need to pull out their phone, aka. enable maximum laziness (but why else would you own a smart watch, right?).

Tested it with my own Pebble and worked great.

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -249,8 +249,11 @@ public class MessageNotifier {
     builder.setDeleteIntent(PendingIntent.getBroadcast(context, 0, new Intent(DeleteReceiver.DELETE_REMINDER_ACTION), 0));
 
     if (masterSecret != null) {
-      builder.addAction(R.drawable.check, context.getString(R.string.MessageNotifier_mark_all_as_read),
-                        notificationState.getMarkAsReadIntent(context, masterSecret));
+       NotificationCompat.Action markAllAsReadAction =
+         new NotificationCompat.Action(R.drawable.check, context.getString(R.string.MessageNotifier_mark_all_as_read),
+                                       notificationState.getMarkAsReadIntent(context, masterSecret));
+       builder.addAction(markAllAsReadAction);
+       builder.extend(new NotificationCompat.WearableExtender().addAction(markAllAsReadAction));
     }
 
     InboxStyle style = new InboxStyle();
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -196,8 +196,11 @@ public class MessageNotifier {
     if (recipient.getContactUri() != null) builder.addPerson(recipient.getContactUri().toString());
 
     if (masterSecret != null) {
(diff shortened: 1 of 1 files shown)


mcginty:
Can you confirm that action is really missing on `HEAD` of master? I just tried it with a Wear emulator and it looks like it's already available even without this change.

chr1shaefn3r:
Yes, I can confirm that the actions is really missing on HEAD of master on my Pebble before this change.
Maybe this is more of an "Added Pebble support" PR then.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -196,8 +196,11 @@ private static void sendSingleThreadNotification(Context context,
     if (recipient.getContactUri() != null) builder.addPerson(recipient.getContactUri().toString());
 
     if (masterSecret != null) {
-      builder.addAction(R.drawable.check, context.getString(R.string.MessageNotifier_mark_as_read),
-                        notificationState.getMarkAsReadIntent(context, masterSecret));
+      NotificationCompat.Action markAsReadAction =
+        new NotificationCompat.Action.Builder(R.drawable.check, context.getString(R.string.MessageNotifier_mark_as_read),
+                                              notificationState.getMarkAsReadIntent(context, masterSecret)).build();
```
moxie0: looks to me like there's no reason to use a builder here

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -246,8 +249,11 @@ private static void sendMultipleThreadNotification(Context context,
     builder.setDeleteIntent(PendingIntent.getBroadcast(context, 0, new Intent(DeleteReceiver.DELETE_REMINDER_ACTION), 0));
 
     if (masterSecret != null) {
-      builder.addAction(R.drawable.check, context.getString(R.string.MessageNotifier_mark_all_as_read),
-                        notificationState.getMarkAsReadIntent(context, masterSecret));
+       NotificationCompat.Action markAllAsReadAction =
+         new NotificationCompat.Action.Builder(R.drawable.check, context.getString(R.string.MessageNotifier_mark_all_as_read),
+                                               notificationState.getMarkAsReadIntent(context, masterSecret)).build();
```
moxie0: same

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
chr1shaefn3r: true that, will fix.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
chr1shaefn3r: here too

midi:
Thank you @chr1shaefn3r  
I noticed that the reply action is missing, could you add that?
It is super useful.

Another issue: #1699

chr1shaefn3r:
@midi you are welcome.

I'm working on porting @Karalix's PR (SMSSecure#135) to the TextSecure code base.
If I don't run in any major problem, expect a PR  around next week.

Issue #1363, opened 2014-04-14
Indicating in the conservation overview whether the last message is yours or from your conversation partner


bvorak:
Description given.

franciscod:
this has been solved here https://github.com/WhisperSystems/Signal-Android/commit/7c95adc7e61e8157a686caa19ac8b6c8cf593eae right?

2-4601:
Closed too hastily. I'm not sure if it's planned to add "You" or contact names in the conversation list? Or is the current situation good enough?

moxie0:
I think it's good as is

PR #3012, opened 2015-04-18
Indicate whether or not last message is yours

ghost:
This should be a fix for #1363 (plus I fixed a typo)
Sorry if the PR looks like a mess, it's my first attempt to use github. Hope it's fine

### DIFF
--- a/src/org/thoughtcrime/securesms/database/model/ThreadRecord.java
+++ b/src/org/thoughtcrime/securesms/database/model/ThreadRecord.java
@@ -69,7 +69,7 @@ public class ThreadRecord extends DisplayRecord {
     } else if (!getBody().isPlaintext()) {
       return emphasisAdded(context.getString(R.string.MessageNotifier_encrypted_message));
     } else if (SmsDatabase.Types.isEndSessionType(type)) {
-      return emphasisAdded(context.getString(R.string.TheadRecord_secure_session_ended));
+      return emphasisAdded(context.getString(R.string.ThreadRecord_secure_session_ended));
     } else if (MmsSmsColumns.Types.isLegacyType(type)) {
       return emphasisAdded(context.getString(R.string.MessageRecord_message_encrypted_with_a_legacy_protocol_version_that_is_no_longer_supported));
     } else if (MmsSmsColumns.Types.isDraftMessageType(type)) {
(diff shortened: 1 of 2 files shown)


ChaoticMind:
Try out `git rebase -i 586b743`, it lets you squash your commits and the PR will have a cleaner history.

volcano99:
Thanks! I think it worked.

src/org/thoughtcrime/securesms/database/model/ThreadRecord.java
```diff
@@ -78,9 +78,11 @@ public SpannableString getDisplayBody() {
     } else {
       if (TextUtils.isEmpty(getBody().getBody())) {
         return new SpannableString(context.getString(R.string.MessageNotifier_no_subject));
+      } else if (SmsDatabase.Types.isOutgoingMessageType(type)) {
+        return new SpannableString(context.getString(R.string.ThreadRecord_sent) + " " + getBody().getBody());
       } else {
-        return new SpannableString(getBody().getBody());
-      }
+            return new SpannableString(getBody().getBody());
```
chr1shaefn3r: Wrong intentation level.

chr1shaefn3r:
I would favour a solution where I would also have the "delivered" information in the contact list. So additionally to the information that I send the last message, I would also know that it was delivered to the recipient.

agrajaghh:
I agree, delived would be helpfull. What about some icons, maybe on the right side below the date?
checkmark for deliverd, arrow (to the right?) for sent messages and perhaps arrow to the left for received messages?

volcano99:
Looks like that would require a little more time to develop, and considering I don't have the icons... anyone is welcome to trash my PR and open another one :-)

agrajaghh:
@volcano99 I could assist you with the icon's if thats helping you?

volcano99:
My main problem is that I don't have a ton of experience coding, so I'm afraid I'd end up slowing things down :/
Plus, first I want to hear from the official developers if they'd like the idea. Do you already know of any icons?

agrajaghh:
I don't think its a really urgent issue, and the best way of gaining experience is just to try ;)

But you're right lets wait what the main devs will say...

Just made a quick mockup, but I'm not totally pleased with it though....
I tried the TS checkmark and a normal one .. I think the simple one looks better.

volcano99:
The normal one looks SO much better, why is it not being used in conversations too?
(Just a thought... we need a pair of icons for each type, because they will change depending on the theme...)

rhodey:

mcginty:
Closing since this account was deleted. I'm in favor of displaying the delivery status of a message instead of a "You: " in the summary, since it provides even more information and still lets the user know it was a message sent from them.

Issue #2930, opened 2015-04-09
ConversationActivity has incorrect ActionBar size and font size after orientation change


turtlekiosk:
The actionbar size and font size of the text in the actionbar are incorrect if you enter the ConversationActivity and then rotate the device (specifically it retains the correct sizes for the opposite orientation).

Entering from portrait:
![portrait_correct](https://cloud.githubusercontent.com/assets/1096636/7078767/5981b344-deea-11e4-836c-3f5405b75b9e.png)

Entering from landscape and rotating to portrait:
![portrait_incorrect](https://cloud.githubusercontent.com/assets/1096636/7078775/6759ded8-deea-11e4-9259-1b47164823d8.png)

Entering from landscape:
![landscape_correct](https://cloud.githubusercontent.com/assets/1096636/7078786/7ec9fc88-deea-11e4-8508-d06077ffad63.png)

Entering from portrait and rotating to landscape:
![landscape_incorrect](https://cloud.githubusercontent.com/assets/1096636/7078797/9ad7c3c4-deea-11e4-9e0f-6400ec8bf0ac.png)

Device info: Moto X 2013, Android 4.4.4, TextSecure 2.10.2

turtlekiosk:
I think the views are not being redrawn because the the orientation config change is ignored for ConversationActivity in the AndroidManifest.

riyapenn:
Is this still an issue?

turtlekiosk:
Yep, still an issue as of 2.28.1

merkste:
It still persist in 3.6.1

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

Issue #520, opened 2014-01-01
Feature: Direct Photo/Video/Audio attachments while staying in app (API calls to camera, sound recorder, video camera) instead of multistep separate apps


travisnj:
I love the app... I want to replace it full time from Textra, BUT there is a big draw back, why don't you allow for direct photos, video and sounds to be taken. So you don't have to use two apps to get the file to send. 

Example: My wife takes a photo of something she wants my opinion on buying, (in textra or native) presses add, camera comes up, takes photo and sends

Example: Same situation, but with Text Secure you have to take the photo first and save image, then goto app and add via the menu of "where" the documents are to attach and send

So, as you can imagine carrying on a conversation and needing to send several photos during that conversation it would be a bit frustrating and very time consuming using Text Secure. 

Native, Textra and about all others I have used use API's and directly use the photo/audio to send.

Mihahn:
+1 I think this would be a great improvement!

ole-tange:
+1 with #345

andunix:
+1

streblo:
+1

Here's a SO post on how to use contentproviders to do this securely: https://stackoverflow.com/questions/20058793/android-how-to-save-the-camera-result-to-a-private-file

GeniusKratos:
+1

mcginty:
Closed in favor of #3563 and #3564 since all these different types of media captures will be separate PRs, making it easier to track progress.

PR #3018, opened 2015-04-20
Add camera preview to message composition

turtlekiosk:
Long overdue work from WBoC15, this adds a Facebook Messenger/Google Messenger style camera preview that lets you quickly take photos without exiting the ConversationActivity view. (part of #520) ~~This does not currently encrypt the files because I didn't want the work to conflict with the changes from the Glide PR.~~ (EDIT: Since the Glide PR got merged in, I updated the branch.) Also, this probably fixes and definitely affects #2930 because the new view cannot handle not being redrawn on rotation. (QuickCamera.java requires knowledge of the device orientation to rotate the camera preview display and image properly).

screenshot_2015-04-20-12-46-32 portrait
landscape


turtlekiosk:
Note: I made the decision to not support half-expanded camera preview/fling on Android versions < 4.0 and only show a full screen preview. I looked at other apps with a similar UI element, Google Messenger, Textra, and Facebook Messenger, and saw that the first two did not support < 4.0 and that Facebook Messenger had made the same decision regarding full screen camera previews on those devices. Let me know if this is an important issue.

mcginty:
The experience is pretty choppy on my Gingerbread phone. I'm wondering if we should only offer this live preview to Android > 4.0 at least for the first release of it.

turtlekiosk:
That'd be fine with me if you think that's the best way to go

bungabunga:
hey, @turtlekiosk, thank you for implementing a long awaited TextSecure feature!

your implementaton looks very nice but what i see here, is a usability problem. i think you can't make decent photos without a full screen preview. i know Google Messenger, Textra and Facebook Messenger are doing the same, but they don't really give a fuck about taking decent photos. they're looking at it strictly from a "gadget perspective" not from usability stand point. if you want to make a decent photo of someone (or something) you really got to have the bigest preview screen possible.

so i'm asking you if it's possible to have at least an option to switch to a full screen preview in your implementation?

thanks!

McLoo:
@bungabunga if I got it right, you can tap the lower right icon to expand the photo view

bungabunga:
@McLoo: that's great! in this case it would be cool to change the icon to be more explanatory. the icon currently in use can have different meanings if used in a context of a camera viewfinder. 
fullscreen-icon-logo-black
full screen1
fullscreen_alt-512

turtlekiosk:
the icon I use is p standard:

Google Messenger:&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Textra:
screenshot_2015-04-30-10-00-04 screenshot_2015-04-30-10-03-26

Also, I updated the original comment to include a gif of the view in action.

bungabunga:
sorry, then my comments are pointless...

agrajaghh:
@turtlekiosk nice work! Really like the UX :blush: 

But unfortunately it's still buggy, while extensive testing my phone crashed and restarted several time times (Nexus 4, Stock 5.1). Unfortunately logcat is empty after restarting... I'll try to reproduce this.

Besides that, the aspect ratios don't match the native camera one's:
Portrait small: ~1,44 ( I guess thats okay since the image is cropped anyway)
Portrait fullscreen: 0,56 (1,78)
Landscape fullscreen: 1,77

The native camera is taking pictures with 1,33 or 0,75

Would it also be possible to keep the camera view open while rotating?

turtlekiosk:
Thanks for checking the view on your device, I'll def follow up on that. Regarding keeping the camera view open, yeah, if we agree that's the ux we want, I can make it happen.

turtlekiosk:
Also I want to clarify that the images are taken from the preview frame, not from converting the raw image sensor data. The reasons for this are:

1. It's much faster to grab a frame and makes the interface much snapper.
2. It takes much less memory, particularly handy when rotating the picture. (Some camera apps don't rotate images, they just store exif data detailing the proper orientation.) Rotating a full resolution image can easily cause an OOM exception.

The cons are:
1. Lower resolution images.
2. Images sizes are dependent on the preview sizes returned by the Camera API rather than native image format size.

But if anyone has any answers to resolve these issues, please lmk!

mcginty:
@turtlekiosk did you look into just using the camera logic from https://github.com/commonsguy/cwac-camera? Seems like from their repo that there are many device-specific strangenesses that we might be able to avoid by using their code.

turtlekiosk:
I hadn't considered it, but sounds like a good idea, I don't think it'd be too much work to plug that in either.

rhodey:

mcginty:
Going to close this in favor of the CWAC rework.

PR #3029, opened 2015-04-21
Replace rather than insert into push db on duplicate incoming.

### LINKED ISSUE #2287: notification unread message when locking textsecure
Evertytime I lock TextSecure I get a notification about an unread message from a certain contact (always the same contact), but it is wrong as I already have read all messages.

moxie0:
Combined with the switch to server acked messages, this will
prevent the race condition that occurred when an incoming message
showed up at exactly the moment the app updated.

Fixes #2287

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/DatabaseUpgradeActivity.java
+++ b/src/org/thoughtcrime/securesms/DatabaseUpgradeActivity.java
@@ -58,7 +58,7 @@ public class DatabaseUpgradeActivity extends BaseActivity {
   public static final int ASYMMETRIC_MASTER_SECRET_FIX_VERSION = 73;
   public static final int NO_V1_VERSION                        = 83;
   public static final int SIGNED_PREKEY_VERSION                = 83;
-  public static final int NO_DECRYPT_QUEUE_VERSION             = 84;
+  public static final int NO_DECRYPT_QUEUE_VERSION             = 113;
 
   private static final SortedSet<Integer> UPGRADE_VERSIONS = new TreeSet<Integer>() {{
     add(NO_MORE_KEY_EXCHANGE_PREFIX_VERSION);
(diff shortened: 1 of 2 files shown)


McLoo:
what about a wiping old pending messages?

moxie0:
@McLoo That should do the trick

src/org/thoughtcrime/securesms/DatabaseUpgradeActivity.java
```diff
@@ -58,7 +58,7 @@
   public static final int ASYMMETRIC_MASTER_SECRET_FIX_VERSION = 73;
   public static final int NO_V1_VERSION                        = 83;
   public static final int SIGNED_PREKEY_VERSION                = 83;
-  public static final int NO_DECRYPT_QUEUE_VERSION             = 84;
+  public static final int NO_DECRYPT_QUEUE_VERSION             = 113;
```
mcginty: why 113?

src/org/thoughtcrime/securesms/database/PushDatabase.java
```diff
@@ -80,6 +88,31 @@ public Reader readerFor(Cursor cursor) {
     return new Reader(cursor);
   }
 
+  private Optional<Long> contains(TextSecureEnvelope envelope) {
```
mcginty: the name "contains" suggests a boolean return to me. more like `find`?

src/org/thoughtcrime/securesms/database/PushDatabase.java
```diff
@@ -31,14 +33,20 @@ public PushDatabase(Context context, SQLiteOpenHelper databaseHelper) {
   }
 
   public long insert(TextSecureEnvelope envelope) {
```
mcginty: non-null?

PR #3030, opened 2015-04-21
fix handling of sms:, smsto:, mms:, mmsto: URIs

### LINKED ISSUE #2578: Add sms: uri schema handler
If I click on an sms hyperlink in a web browser (such as
```html
<a href='sms:555-555-5555'>Call me, maybe?</a>
```
textsecure does not come up as an option.
Since you can only send sms from your default sms app, this makes it so these links essentially are unusable when textsecure is your default sms app.

Note: smsto: does not work either.

rhodey:
1) added VIEW action and BROWSABLE category to the intent filter of SmsSendtoActivity
2) created two helper methods to make up for bugs in the processing of RFC5724 SMS URIs by android.net.Uri

Fixes #2578
// FREEBIE

tested on my Nexus 5 (A 5.1) & LG Optimus 2X (A 2.3.4).

### DIFF
--- a/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
+++ b/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
@@ -59,7 +59,7 @@ public class Rfc5724Uri {
     for (String keyValue : uri.split("\\?")[1].split("&")) {
       String[] parts = keyValue.split("=");
 
-      if(parts.length == 1) queryParams.put(parts[0], "");
+      if (parts.length == 1) queryParams.put(parts[0], "");
       else                  queryParams.put(parts[0], URLDecoder.decode(parts[1]));
     }
 
--- a/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
+++ b/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
@@ -19,57 +19,63 @@ package org.thoughtcrime.securesms.util;
 
 import java.net.URISyntaxException;
 import java.net.URLDecoder;
+import java.util.HashMap;
+import java.util.Map;
 
 public class Rfc5724Uri {
 
-  private static final String SCHEMA_SMS   = "sms";
-  private static final String SCHEMA_SMSTO = "smsto";
-  private static final String SCHEMA_MMS   = "mms";
-  private static final String SCHEMA_MMSTO = "mmsto";
-
-  private final String uri;
-  private final String schema;
-  private final String recipients;
+  private final String              uri;
+  private final String              schema;
+  private final String              path;
+  private final Map<String, String> queryParams;
 
   public Rfc5724Uri(String uri) throws URISyntaxException {
-    this.uri        = uri;
-    this.schema     = parseSchema(uri);
-    this.recipients = parseRecipients(uri);
+    this.uri         = uri;
+    this.schema      = parseSchema();
+    this.path        = parsePath();
+    this.queryParams = parseQueryParams();
   }
 
-  private static String parseSchema(String uri) throws URISyntaxException {
-    String schema      = uri.split(":")[0];
-    String schemaLower = schema.toLowerCase();
+  private String parseSchema() throws URISyntaxException {
+    String[] parts = uri.split(":");
 
-    if (!schemaLower.equals(SCHEMA_SMS) && !schemaLower.equals(SCHEMA_SMSTO) &&
-        !schemaLower.equals(SCHEMA_MMS) && !schemaLower.equals(SCHEMA_MMSTO))
-    {
-      throw new URISyntaxException(uri, "supplied URI does not contain a valid schema");
+    if (parts.length < 1 || parts[0].length() < 1) throw new URISyntaxException(uri, "invalid schema");
+    else                                           return parts[0];
+  }
+
+  private String parsePath() throws URISyntaxException {
+    String[] parts = uri.split("\\?")[0].split(":", 2);
+
+    if (parts.length < 2 || parts[1].length() < 1) throw new URISyntaxException(uri, "invalid path");
+    else                                           return parts[1];
+  }
+
+  private Map<String, String> parseQueryParams() throws URISyntaxException {
+    Map<String, String> queryParams = new HashMap<>();
+    if (!uri.contains("?") || uri.split("\\?").length < 2) {
+      return queryParams;
     }
 
-    return schema;
-  }
+    for (String keyValue : uri.split("\\?")[1].split("&")) {
+      if (keyValue.contains("=")) {
+        String[] parts = keyValue.split("=");
+        String   value = (parts.length == 1) ? "" : parts[1];
+        queryParams.put(parts[0], URLDecoder.decode(value));
+      }
+    }
 
-  private static String parseRecipients(String uri) throws URISyntaxException {
-    String[] parts = uri.split("\\?")[0].split(":");
-
-    if (parts.length < 2) throw new URISyntaxException(uri, "supplied URI contains no recipients");
-    else                  return parts[1];
+    return queryParams;
   }
 
   public String getSchema() {
     return schema;
   }
 
-  public String getRecipients() {
-    return recipients;
+  public String getPath() {
+    return path;
   }
 
-  public String getQueryParam(String key) {
-    String[] parts = uri.split(key + "=");
-
-    if      (!uri.contains(key + "=")) return null;
-    else if (parts.length == 1)        return "";
-    else                               return URLDecoder.decode(parts[1].split("&")[0]);
+  public Map<String, String> getQueryParams() {
+    return queryParams;
   }
 }
--- a/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
+++ b/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
@@ -39,20 +39,20 @@ public class Rfc5724Uri {
   private String parseSchema() throws URISyntaxException {
     String[] parts = uri.split(":");
 
-    if (parts.length < 1 || parts[0].length() < 1) throw new URISyntaxException(uri, "invalid schema");
-    else                                           return parts[0];
+    if (parts.length < 1 || parts[0].isEmpty()) throw new URISyntaxException(uri, "invalid schema");
+    else                                        return parts[0];
   }
 
   private String parsePath() throws URISyntaxException {
     String[] parts = uri.split("\\?")[0].split(":", 2);
 
-    if (parts.length < 2 || parts[1].length() < 1) throw new URISyntaxException(uri, "invalid path");
-    else                                           return parts[1];
+    if (parts.length < 2 || parts[1].isEmpty()) throw new URISyntaxException(uri, "invalid path");
+    else                                        return parts[1];
   }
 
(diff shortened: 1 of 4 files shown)


moxie0:
this is a lot of code, what is the bug you're trying to fix?

rhodey:
@moxie0 please see my most recent comment in issue #2578, included in this pull by reference.

moxie0:
why are you parsing this uri by hand instead of using Uri or URI?

rhodey:
@moxie0 calling `Uri.getQueryParameter()` on `sms:+437722735932?body=hello%20there` throws an exception stating `this is not a hierarchical url`, none of the Uri query parameter methods work with RFC5724 uris.

Let me try creating a URI from it and using their methods...

rhodey:
URI is also unable to cope with RFC5724, throwing `java.net.URISyntaxException: Scheme-specific part expected at index 4: sms:` from its constructor.

I don't believe Uri or URI can be used in this situation unless we do something weird and abuse the `sms:, smsto:, mms:` etc uris to play nicely with the Uri or URI expectations-- which itself would require some parsing by hand.

Would throwing these methods into util and adding unit tests help the cause, @moxie0?

moxie0:
yes you'll have to write a uri class for this if there's nothing available, along with tests for the pathological cases to ensure you don't end up with underflows or overflows

moxie0:
Validate as you parse, and parse everything on construction time and then have getters that return the parsed parts.  Use split instead of indexof and substring (there should be three logical splits, ":", "?", and "&"), don't worry about validating anything else.

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,75 @@
...
+
+  private static String parseSchema(String uri) throws URISyntaxException {
+    String schema      = uri.split(":")[0];
+    String schemaLower = schema.toLowerCase();
+
+    if (!schemaLower.equals(SCHEMA_SMS) && !schemaLower.equals(SCHEMA_SMSTO) &&
+        !schemaLower.equals(SCHEMA_MMS) && !schemaLower.equals(SCHEMA_MMSTO))
+    {
+      throw new URISyntaxException(uri, "supplied URI does not contain a valid schema");
+    }
```
moxie0: don't bother validating this

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,75 @@
...
+
+    if (parts.length < 2) throw new URISyntaxException(uri, "supplied URI contains no recipients");
+    else                  return parts[1];
+  }
+
+  public String getSchema() {
+    return schema;
+  }
+
+  public String getRecipients() {
```
moxie0: getPath

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,75 @@
...
+    return recipients;
+  }
+
+  public String getQueryParam(String key) {
+    String[] parts = uri.split(key + "=");
+
+    if      (!uri.contains(key + "=")) return null;
+    else if (parts.length == 1)        return "";
+    else                               return URLDecoder.decode(parts[1].split("&")[0]);
```
moxie0: put the query params in a map at construction time

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,75 @@
...
+  private final String schema;
+  private final String recipients;
+
+  public Rfc5724Uri(String uri) throws URISyntaxException {
+    this.uri        = uri;
+    this.schema     = parseSchema(uri);
+    this.recipients = parseRecipients(uri);
+  }
+
+  private static String parseSchema(String uri) throws URISyntaxException {
```
moxie0: lose the static

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,75 @@
...
+    if (!schemaLower.equals(SCHEMA_SMS) && !schemaLower.equals(SCHEMA_SMSTO) &&
+        !schemaLower.equals(SCHEMA_MMS) && !schemaLower.equals(SCHEMA_MMSTO))
+    {
+      throw new URISyntaxException(uri, "supplied URI does not contain a valid schema");
+    }
+
+    return schema;
+  }
+
+  private static String parseRecipients(String uri) throws URISyntaxException {
```
moxie0: lose the static

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,75 @@
...
+    String[] parts = uri.split("\\?")[0].split(":");
```
moxie0: limit 2

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.net.URISyntaxException;
+import java.net.URLDecoder;
+
+public class Rfc5724Uri {
+
+  private static final String SCHEMA_SMS   = "sms";
+  private static final String SCHEMA_SMSTO = "smsto";
+  private static final String SCHEMA_MMS   = "mms";
+  private static final String SCHEMA_MMSTO = "mmsto";
```
moxie0: you shouldn't need these

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,75 @@
...
+
+  private final String uri;
+  private final String schema;
+  private final String recipients;
```
moxie0: path

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
rhodey: I don't understand this comment.

moxie0: split(":", 2)

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,81 @@
...
+    this.uri         = uri;
+    this.schema      = parseSchema();
+    this.path        = parsePath();
+    this.queryParams = parseQueryParams();
+  }
+
+  private String parseSchema() throws URISyntaxException {
+    String[] parts = uri.split(":");
+
+    if (parts.length < 1 || parts[0].length() < 1) throw new URISyntaxException(uri, "invalid schema");
```
moxie0: parts.length can never be less than 1

moxie0: parts[0].isEmpty

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,81 @@
...
+    else                                           return parts[0];
+  }
+
+  private String parsePath() throws URISyntaxException {
+    String[] parts = uri.split("\\?")[0].split(":", 2);
+
+    if (parts.length < 2 || parts[1].length() < 1) throw new URISyntaxException(uri, "invalid path");
```
moxie0: parts[1].isEmpty

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,81 @@
...
+    else                                           return parts[1];
+  }
+
+  private Map<String, String> parseQueryParams() throws URISyntaxException {
+    Map<String, String> queryParams = new HashMap<>();
+    if (!uri.contains("?") || uri.split("\\?").length < 2) {
```
moxie0: contains should not appear in this entire class

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private Map<String, String> parseQueryParams() throws URISyntaxException {
+    Map<String, String> queryParams = new HashMap<>();
+    if (uri.split("\\?").length < 2) {
+      return queryParams;
+    }
+
+    for (String keyValue : uri.split("\\?")[1].split("&")) {
+      if (keyValue.contains("=")) {
+        String[] parts = keyValue.split("=");
```
rhodey: `contains()` necessary for distinction between `sms:555?body, body == null` & `sms:555?body=, body == ""` because `"sms:555?body".split("=").length == 1` & `"sms:555?body=".split("=").length == 1`

moxie0:
doesn't look like this has been updated from the last round of feedback

rhodey:
the only remaining discrepancy that I'm aware of is the use of `contains()` [here](https://github.com/WhisperSystems/TextSecure/pull/3030/files#diff-05daefa13a1a1537982fa06e7d37151dR62), `split()` alone is not enough, see:

```
String[] parts = "body=".split("=");
parts.length == 1
...
String[] parts = "body".split("=");
parts.length == 1
```

without `contains("=")`, `sms:555?body=` (legal, `body == ""`) is indistinguishable from `sms:555?body` (illegal). I can't do anything to remove the `contains()` unless we wanna be weird and say `sms:555?body=` is also illegal, then a check for `parts.length == 2` could work. split isn't all that great >:|

moxie0:
it's fine without the contains

rhodey:
ok, then

```
sms:555?body=, body == ""
&&
sms:555?body, body == ""
&&
sms:555?a=b, body == null
```

done.

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,80 @@
...
+  private Map<String, String> parseQueryParams() throws URISyntaxException {
+    Map<String, String> queryParams = new HashMap<>();
+    if (uri.split("\\?").length < 2) {
+      return queryParams;
+    }
+
+    for (String keyValue : uri.split("\\?")[1].split("&")) {
+      String[] parts = keyValue.split("=");
+
+      if(parts.length == 1) queryParams.put(parts[0], "");
```
agrajaghh: space :P

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -0,0 +1,80 @@
...
+    String[] parts = uri.split(":");
+
+    if (parts.length < 1 || parts[0].isEmpty()) throw new URISyntaxException(uri, "invalid schema");
+    else                                        return parts[0];
+  }
+
+  private String parsePath() throws URISyntaxException {
+    String[] parts = uri.split("\\?")[0].split(":", 2);
+
+    if (parts.length < 2 || parts[1].isEmpty()) throw new URISyntaxException(uri, "invalid path");
```
jpsingleton: @rhodey @moxie0 I believe not having a path (phone number) is valid. Please see my comment on the related issue (https://github.com/WhisperSystems/TextSecure/issues/2578#issuecomment-120620069). Thanks!

PR #3032, opened 2015-04-22
improved challenge sms verification in SmsListener

### LINKED ISSUE #2919: Allow carriers add suffix in the verification code sms
I'd received a message like:
```
Your TextSecure verification code: ***-***[CEG]
```
I think the regular expression here should be changed to ignore any suffix just like you've ignored the prefix in the message:
https://github.com/WhisperSystems/TextSecure/blob/4397b55ceb8b82de1e8545e01dc7b6864fa85846/src/org/thoughtcrime/securesms/service/SmsListener.java#L131

rhodey:
some carriers prepend or append arbitrary text to sms messages enroute, SmsListener failed to handle the append case.

Fixes #2919
// FREEBIE

tested on Android 4.4 & 2.3.4

### DIFF
--- a/src/org/thoughtcrime/securesms/service/SmsListener.java
+++ b/src/org/thoughtcrime/securesms/service/SmsListener.java
@@ -16,6 +16,7 @@
  */
 package org.thoughtcrime.securesms.service;
 
+import android.annotation.SuppressLint;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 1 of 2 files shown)


moxie0:
this should be a one line change

rhodey:
last commit brings it closer to a one line change. changing the regex in `isChallenge()` isn't enough because `parseChallenge('...code: xxx1337-1337xxx')` uses primitive `split(':'), split('-')` and would return `13371337xxx`, at this point makes most sense and smallest diff to switch to the `Matcher` pattern.

src/org/thoughtcrime/securesms/service/SmsListener.java
```diff
@@ -128,7 +130,7 @@ private boolean isChallenge(Context context, Intent intent) {
     if (messageBody == null)
       return false;
 
-    if (messageBody.matches(".*Your TextSecure verification code: [0-9]{3,4}-[0-9]{3,4}") &&
+    if (CHALLENGE_PATTERN.matcher(messageBody).find() &&
```
moxie0: matches()

src/org/thoughtcrime/securesms/service/SmsListener.java
```diff
@@ -138,11 +140,14 @@ private boolean isChallenge(Context context, Intent intent) {
...
-    String[] messageParts = messageBody.split(":");
-    String[] codeParts    = messageParts[1].trim().split("-");
+    String  messageBody      = getSmsMessageBodyFromIntent(intent);
+    Matcher challengeMatcher = CHALLENGE_PATTERN.matcher(messageBody);
 
-    return codeParts[0] + codeParts[1];
+    if (challengeMatcher.find()) {
+      return challengeMatcher.group(1) + challengeMatcher.group(2);
+    } else {
+      throw new IllegalArgumentException("call to parseChallenge() before isChallenge()");
```
moxie0: which argument is illegal here?

rhodey: `intent` is illegal because it doesn't contain a challenge. Does `IllegalStateException` make more sense? The code on master throws an `IndexOutOfBoundsException` in this case.

moxie0:
````
if (!challengeMatcher.matches()) {
  throw new AssertionError("Expression should match.")
}

return challengeMatcher.group(1) + challengeMatcher.group(2)
````

PR #3034, opened 2015-04-22
fix for issue #3004

mderosa:
This exception happens when one thread is accessing the database helper through a reference after it has been closed by another thread. I have tried to do a minimal fix that will prevent this error. As a note both the original and modified versions invoke destroyInstance() which effectively destroys the in memory database with each call of loadInBackground (can be verified by adding a log statement / break point in DatabaseOpenHelper::ContactsDatabase.onCreate() ) that methodology seems like it could made more performant but since it has been performant enough so far, I am not going to be intrusive and do more. Also included is a test for this issue.  -- Best, Marc

### DIFF
--- a/androidTest/java/org/thoughtcrime/securesms/contacts/ContactsCursorLoaderTest.java
+++ b/androidTest/java/org/thoughtcrime/securesms/contacts/ContactsCursorLoaderTest.java
@@ -1,0 +1,52 @@
+package org.thoughtcrime.securesms.contacts;
+
+import android.database.Cursor;
+import android.os.Looper;
+
+import org.thoughtcrime.securesms.TextSecureTestCase;
+
+import java.util.concurrent.Executor;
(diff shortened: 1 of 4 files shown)


moxie0:
@mcginty What's the logic behind making this a singleton that it appears you destroy per-use?

mcginty:
It was originally a singleton to ensure that only one instance of the database ever existed, and that it would get killed when the database was no longer needed in memory. The destroy per-use was added when the contact list wasn't refreshing after added the refresh button to the new conversation activity.

Now I'm thinking the least ugly way to do this would be to just keep an instance of ContactsDatabase in the loader that's closed on its onReset without the singleton stuff. That we can be sure there's a valid instance when it's needed whose state isn't changing from other threads...

moxie0:
@mderosa thanks a lot for tracking this down

PR #3041, opened 2015-04-23
maintain one ContactsDatabase instance per Loader

mcginty:
A proposed alternative fix for #3004 as suggested in #3034 that removes the unnecessary singleton logic and enforces thread safety when doing database operations on the contacts db.

### DIFF
--- a/src/org/thoughtcrime/securesms/contacts/ContactsCursorLoader.java
+++ b/src/org/thoughtcrime/securesms/contacts/ContactsCursorLoader.java
@@ -30,6 +30,7 @@ public class ContactsCursorLoader extends CursorLoader {
   private final Context          context;
   private final String           filter;
   private final boolean          pushOnly;
+  private final Object           dbLock = new Object();
   private       ContactsDatabase db;
 
   public ContactsCursorLoader(Context context, String filter, boolean pushOnly) {
@@ -37,17 +38,22 @@ public class ContactsCursorLoader extends CursorLoader {
     this.context  = context;
     this.filter   = filter;
     this.pushOnly = pushOnly;
+    this.db       = new ContactsDatabase(context);
   }
 
   @Override
   public Cursor loadInBackground() {
-    ContactsDatabase.destroyInstance();
-    db = ContactsDatabase.getInstance(context);
-    return db.query(filter, pushOnly);
+    synchronized (dbLock) {
+      return db.query(filter, pushOnly);
+    }
   }
 
   @Override
   public void onReset() {
+    synchronized (dbLock) {
+      db.close();
+      db = new ContactsDatabase(context);
+    }
     super.onReset();
   }
 }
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/contacts/ContactsCursorLoader.java
```diff
@@ -41,13 +42,18 @@ public ContactsCursorLoader(Context context, String filter, boolean pushOnly) {
 
   @Override
   public Cursor loadInBackground() {
-    ContactsDatabase.destroyInstance();
-    db = ContactsDatabase.getInstance(context);
-    return db.query(filter, pushOnly);
+    synchronized (dbLock) {
+      if (db == null) db = new ContactsDatabase(context);
+      return db.query(filter, pushOnly);
+    }
```
moxie0: Can't you just create ContactsDatabase in the constructor? Why is this lazy initialization necessary?

mcginty: I had incorrectly thought that onCreate in the database helper was called during construction which would be a big main-thread hit. Just checked the android code and I think it only happens when you getReadableDatabase.

mcginty: Either way we can't create the ContactsDatabase in the constructor since there's no destructor that I know of. There's only `onReset` which only tells the loader to reset its data with the possibility that it may be reused with `startLoading`.

moxie0: Not sure I understand. Create it in the constructor and re-create it in onReset?

mcginty: I guess that could work since the database would never be opened and `close()` would be unnecessary. Seems kind of weird though.

moxie0: Weirder than using a synchronized block to lazily initialize it? =)

mcginty: good point

mcginty:
updated

src/org/thoughtcrime/securesms/contacts/ContactsCursorLoader.java
```diff
@@ -30,6 +30,7 @@
   private final Context          context;
   private final String           filter;
   private final boolean          pushOnly;
+  private final Object           dbLock = new Object();
```
moxie0: let's synchronize either on the db itself, or on ContactsCursorLoader

moxie0:
...and this won't break the list update on refresh, right? because that's when onRestart gets called?

src/org/thoughtcrime/securesms/contacts/ContactsCursorLoader.java
moxie0: actually, i don't know how many times loadInBackground is called concurrently, but it might make sense to use a Semaphore of it is.  You could have a semaphore with 100 free permits, and then call acquire() in loadInBackground and acquire(100) in onReset.  That way only onReset would block.

mcginty:
updated to use semaphores to avoid unnecessary blocking (although I think the case where there are concurrent loadInBackgrounds should be an edge case).

from the logs, can verify that onReset is called and ContactsDatabase is repopulating its push contacts.

PR #2398, opened 2015-01-30
Add "do not warn again" checkbox to media preview dialog.

softwareanimal:
This is for bug #1915.

PR #3056, opened 2015-04-25
add "don't ask again" option to media warning

### LINKED ISSUE #1915: Give option for user to suppress warning on opening attachments
Friends who are using TextSecure have found it frustrating that it alerts them every time they want to view an image.

alert

Even if they have the passphrase option disabled it still prompts them.

My "I'm totally an artist" mockup:
yibfmub

But I'd recommend the box being **unchecked** by default and an option in settings to reenable the alert.

rhodey:
1) merge media preview and saving warnings into one dialog
2) remove old string resources

Fixes #1915
// FREEBIE

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -167,10 +167,10 @@
     <string name="ExportFragment_success">Success!</string>
 
     <!-- ExternalMediaWarningDialog -->
-    <string name="ExternalMediaWarningDialog_expose_secure_media">Expose secure media?</string>
+    <string name="ExternalMediaWarningDialog_reveal_secure_media">Reveal secure media?</string>
     <string name="ExternalMediaWarningDialog_previewing_or_saving_media_outside_of">Previewing or saving media outside of TextSecure allows other apps to see it.</string>
     <string name="ExternalMediaWarningDialog_dont_ask_again">Don\'t ask again</string>
-    <string name="ExternalMediaWarningDialog_continue">Continue</string>
+    <string name="ExternalMediaWarningDialog_reveal">Reveal</string>
     <string name="ExternalMediaWarningDialog_cancel">Cancel</string>
 
     <!-- GcmRefreshJob -->
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -168,7 +168,7 @@
 
     <!-- ExternalMediaWarningDialog -->
     <string name="ExternalMediaWarningDialog_expose_secure_media">Expose secure media?</string>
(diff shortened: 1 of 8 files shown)


rhodey:
@mcginty what do you think of the string resources? I'm not in love with them but tried to use your feedback from #2398

preview

McLoo:
Never more settings/options, I know.
But what about re-enable warnings from advanced settings?

res/values/strings.xml
```diff
@@ -170,6 +166,13 @@
     <string name="ExportFragment_error_while_writing_to_sd_card">Error while writing to SD card.</string>
     <string name="ExportFragment_success">Success!</string>
 
+    <!-- ExternalMediaWarningDialog -->
+    <string name="ExternalMediaWarningDialog_expose_secure_media">Expose secure media?</string>
+    <string name="ExternalMediaWarningDialog_previewing_or_saving_media_outside_of">Previewing or saving media outside of TextSecure allows other apps to see the media.</string>
```
mcginty: probably just "see it" instead of "see the media"

rhodey:
@McLoo If we added this "don't ask again" option to more dialogs in the future I agree that an advanced "start asking again" (vocabulary to be determined :P) setting would be nice.

rhodey:
@mcginty any further suggestions related to design or vocabulary? @moxie0 this diff can't get any smaller: new view, new strings, new preference, and swap out of dialog callbacks. in need of some feedback please.

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -224,9 +224,9 @@ protected void onPostExecute(Bitmap bitmap) {
   }
 
   private void saveToDisk() {
-    SaveAttachmentTask.showWarningDialog(this, new DialogInterface.OnClickListener() {
+    ExternalMediaWarningDialog.showIfNecessary(this, new WarningListener() {
```
mcginty: why not just have the user pass in a ButtonCallback directly? Seems like the same amount of code for the caller and simplified code in the dialog class.

mcginty:
I think for visual consistency the checkbox text should be styled the same as the main dialog content text.

mcginty:
Also, to be consistent with material dialogs, the "CONTINUE" button should be "EXPOSE" in this case. Could consider using "reveal" instead of "expose".

rhodey:
done, new screenshot screen

rhodey:
the theme I was using to style the dialog text just got wiped out by 2d8de5291ef5b2867d38a33ac7afb2c4b4098b06, will reopen after switching to new text appearance style and testing

PR #3059, opened 2015-04-25
fix race conditions around sending attachments.

### LINKED ISSUE #3051: Thumbnails not rendering
Built a RC and tried sending a few images.  This happens every so often, where the entire message bubble (except for the arrow) fails to render:

glide-sucks

mcginty:
There appeared to be multiple race conditions leading to the issues in #3051.

First, we needed to wait until the thumbnail for the draft attachment is generated before sending so it can be passed through to be inserted into the PartDatabase properly. Otherwise, we get a race that generates two thumbnails and takes *way* more memory.

Second, it seems like there's a race between the message being inserted into MmsDatabase and the parts being inserted into the part database. The CursorLoader would get notified that the dataset changed underneath it, and if the parts weren't inserted into the database by the time the listview refreshed, there would be no media slides available. To fix this, I just threw the message insertion into a transaction so we can still get the messageId properly, but the data set change event isn't triggered until the parts also exist.

As sugar on top, I think it will be better to show an error indicator in a message instead of just a little triangle tick, so did that too.

### DIFF
--- a/src/org/thoughtcrime/securesms/database/MmsDatabase.java
+++ b/src/org/thoughtcrime/securesms/database/MmsDatabase.java
@@ -749,15 +749,21 @@ public class MmsDatabase extends MessagingDatabase {
 
     contentValues.put(PART_COUNT, PartParser.getSupportedMediaPartCount(body));
 
-    long messageId = db.insert(TABLE_NAME, null, contentValues);
+    db.beginTransaction();
+    try {
+      long messageId = db.insert(TABLE_NAME, null, contentValues);
 
-    addressDatabase.insertAddressesForId(messageId, headers);
-    partsDatabase.insertParts(masterSecret, messageId, body);
+      addressDatabase.insertAddressesForId(messageId, headers);
+      partsDatabase.insertParts(masterSecret, messageId, body);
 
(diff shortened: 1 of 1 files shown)


moxie0:
The data set change shouldn't be triggered until we manually call notifyDataSetChanged.  I don't see how a transaction changes that?

Also, wasn't the part database thumbnail storage explicitly about doing lazy thumbnail generation in a memory safe way? In any case, seems very wrong for the thumbnail *view* to be the thing that populates a message.

mcginty:
commenting out the notifyConversationListeners in `insertMediaMessage` still leads to a refresh in the conversation list, so there's still a dataset change event that seems to be causing this race case. I think I'm going to break this PR into the race case and then a proposed cleaning up of the thumbnail stuff separately.

moxie0:
Are you sure that's the source of the bug then? My understanding is that a DB insert shouldn't have any effect on content observers unless they're explicitly notified. If that's incorrect, we should stop notifying them manually.

mcginty:
Well, I guess the general point is that it doesn't matter where or when the content observer notification happens, anything like a delivery notification could cause it. When it does happen, the list will refresh regardless of any ongoing DB operations, and if it refreshes at just the right time we'll get a state where there's an MMS message in the DB without its parts.

PR #3081, opened 2015-04-29
fix: issue 3027 (incomming bubble end at screen right border)

mderosa:
I just added a margin to the incoming bubble on its right side (50dp) . That may not be the right size for your grid(?) but I left it at that as it seems to accord with your intent  (assumed from the conversation _bubble_incomming.xml file) and just eye-ing the screen seems to balance the left hand image.
I also added an edit mode check to the ConversationItem view class so that it will be viewable in the designer. Being able to view layouts in the designer helps me to debug  faster -- so I left that check code in. You may feel the same way; otherwise just delete the check code.


rhodey:

mcginty:
already fixed, thanks!

PR #3033, opened 2015-04-22
QuickContact in GroupMembersDialog

chr1shaefn3r:
Added onClickListener to the GroupMembersDialog which gives the same functionality as taps on avatars.

Fixes #2837

PR #3088, opened 2015-04-29
fix click listener NPE on group members dialog

mcginty:
found while testing the appcompat v22, but has existed probably forever?


McLoo:
Or maybe #3033 for a more useful tap handling 😉

mcginty:
Ah! I missed that PR. Let me close this one and add a milestone to #3033

Issue #2550, opened 2015-03-01
HTC bug - "System UI has stopped unexpectedly. Would you like to tell HTC?"


krasnoj:
Hi!
Cell phone Android 4.0.3, HTC Rhyme
When I start TextSecure the System first crashes with "Default Lock Screen" and then with SystemUI closes unexpectedly (German: "SystemUI wurde unerwarted beendet"). Then the header line including the clock disappears and the SystemUI is started again with the clock. 
This happens every time I start TextSecure. I already tried to reinstall and reflash the cell phone. But the problem still occurs.
If you need any additional information, just tell me what you need.

Cheers.

mcginty:
@krasnoj please include a debug log

krasnoj:
The build in log functionality is attached to https://gist.github.com/anonymous/024d4c3efd302572cbc8

But there is nothing usefull in this logfile. And yes I posted the whole logfile (using the   and TextSecure was open while the SystemUI crashed.

I posted under <http://debianforum.de/forum/pastebin.php?mode=view&s=38354> a larger logfile made with a third party app. The log file should be complete. Just the entries with "WifiStateMachine" have been filtered by me.

moxie0:
looks like a ROM bug

On 03/02/2015 03:18 PM, krasnoj wrote:
> The build in log functionality is attached to
> https://gist.github.com/anonymous/024d4c3efd302572cbc8
> 
> But there is nothing usefull in this logfile. And yes I posted the whole
> logfile (using the and TextSecure was open while the SystemUI crashed.
> 
> I posted under
> http://debianforum.de/forum/pastebin.php?mode=view&s=38354 a larger
> logfile made with a third party app. The log file should be complete.
> Just the entries with "WifiStateMachine" have been filtered by me.
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/WhisperSystems/TextSecure/issues/2550#issuecomment-76848979>.
> 

-- 
http://www.thoughtcrime.org

rhodey:
I'm unable to reproduce this and agree that this seems like a ROM bug. If you're running into this issue on a very popular non-stock Android ROM then it might be worth looking into a fix but in general we want to remain focused on compatibility with stock Android.

krasnoj:
Hi!

I just want to mention that the Bug occurs with the official upgrade for the HTC Rhyme cell phone. So there is no third party software involved.

Steps I did:
[...]

PR #3089, opened 2015-04-29
Upgrade to AppCompat v22.1

mcginty:
Had to expedite this transition because a clean build no longer works if you don't have a previous cached version of `material-dialogs`.

1) update material-dialogs to fix build
2) change to AppCompat autocomplete and fix visual bug in GB

Manually tested on GB and Lollipop, going through the settings, conversation, and group interactions, and @rhodey has been testing AppCompat v22 on TestObject with no crashes.

AppCompat v22 will allow for the possibility of dropping our `material-dialogs` dependency, and normalizing the EditText and Button views across platforms thanks to their new widgets.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,14 +40,14 @@ dependencies {
     compile 'com.github.chrisbanes.photoview:library:1.2.3'
     compile 'com.github.bumptech.glide:glide:3.5.2'
     compile 'com.makeramen:roundedimageview:1.5.0'
-    compile ('com.afollestad:material-dialogs:0.6.4.7') {
+    compile ('com.afollestad:material-dialogs:0.7.3.1') {
         exclude module: 'appcompat-v7'
         exclude module: 'recyclerview-v7'
         exclude module: 'support-annotations'
     }
 
     compile 'com.soundcloud.android:android-crop:0.9.10@aar'
-    compile 'com.android.support:appcompat-v7:21.0.3'
+    compile 'com.android.support:appcompat-v7:22.1.1'
     compile 'com.android.support:recyclerview-v7:21.0.3'
     compile 'com.melnykov:floatingactionbutton:1.1.0'
     compile 'com.google.zxing:android-integration:3.1.0'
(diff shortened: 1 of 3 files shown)


rhodey:
give me ~30 minutes and I'll merge this into my espresso-tests branch and run it though a dozen devices on TestObject so we can feel better about no crashes. as I understand it, extra attention should be given to LG and Samsung devices, right @mcginty? Any other especially crashy devices?

rhodey:
up and running, this could take awhile though >> https://app.testobject.com/#/share/78dfa82d-f608-4267-8970-acf83e1ac5a1/suite-report

agrajaghh:
@mcginty I could replace all the material-dialogs with [android.support.v7.app.AlertDialog](https://developer.android.com/reference/android/support/v7/app/AlertDialog.html) except the `ProgressDialog`s,  are you okay with it?

rhodey:
I ran a bunch of espresso tests on 11 devices and found no (new) crash bugs, I say new because `HTC EVO 3D` still crashes its lock screen process as documented in #2550, also HTC still fails a bunch of tests related to contact number comparison.

One new thing I noticed from these tests is that the contact filter in `NewConversationActivity` seems to behave differently on different devices. I need to add 2 more tests to be certain of this but for the contact `agrajaghh` all devices I've ever used will match on `jag`, but many of the devices I tested on don't. Anyways, almost certainly unrelated to the v22 update but I can't be certain because I don't have the cached dependencies to build master.

My only concern with merging this is that 2 out of ~500 tests failed with the error message `Instrumentation run failed due to 'Process crashed.'`, both these tests were on a `Fairphone` running 4.2.2 and for some reason the adb logs for this test are empty. I sent a support request to TestObject to see if they have any idea why the logs are empty. The tests that fail both involve enabling app protection and then setting a passphrase. Strangely, test `testChangePassphrase` succeeds which itself involves first enabling app protection, setting a passphrase, and then changing the passphrase, so this is probably just some bug with the testing infrastructure or the device.

agrajaghh:
@rhodey You could use `localMaven()` and build material-dialogs 0.6.4.7 yourself (or I could provide the aar/pom?) if you want to be sure and compare against master?

rhodey:
I updated my test suite and ran another set of tests in attempt to confirm the strange `NewConversationActivity` contact filter behavior, this went well. Additionally the two tests that before failed on `Fairphone` passed this time, no crashes :).

Unfortunately the Espresso test framework simply doesn't work for newer devices so we have no tests for that beyond our own personal devices, however I now feel that we can safely merge this without introducing new crash bugs. So, LGTM.

mcginty:
Merged to `2.14.0` branch.

Issue #3096, opened 2015-05-01
Text message not delivered for certain phone number (unsecured SMS)


brenns10:
For one particular contact, I attempt to send an unsecured SMS and immediately the app displays it with a red exclamation mark, saying "Not delivered" and "Tap for details".  I can send unsecured SMS to other contacts just fine, but this one reliably fails.  I plugged my phone into ADB and reproduced the issue while running logcat.  The wall of text below is what looks to be the relevant section from logcat (I expunged the phone number):

```
W/MmsSmsDatabase( 6309): Executing query: SELECT _id, body, type, thread_id, address, address_device_id, subject, date_sent, date_received, m_type, msg_box, status, part_count, ct_l, tr_id, m_size, exp, st, delivery_receipt_count, mismatched_identities, network_failures, transport_type FROM (SELECT DISTINCT date_sent * 1 AS date_sent, date * 1 AS date_received, _id, body, read, thread_id, type, address, address_device_id, subject, NULL AS m_type, NULL AS msg_box, status, NULL AS part_count, NULL AS ct_l, NULL AS tr_id, NULL AS m_size, NULL AS exp, NULL AS st, delivery_receipt_count, mismatched_identities, NULL AS network_failures, 'sms' AS transport_type FROM sms WHERE (thread_id = 15) UNION ALL SELECT DISTINCT date * 1000 AS date_sent, date_received * 1000 AS date_received, _id, body, read, thread_id, NULL AS type, address, address_device_id, NULL AS subject, m_type, msg_box, NULL AS status, part_count, ct_l, tr_id, m_size, exp, st, delivery_receipt_count, mismatched_identities, network_failures, 'mms' AS transport_type FROM mms WHERE (thread_id = 15) ORDER BY date_received DESC) LIMIT 1
I/LatinIME:LogUtils( 1405): Dictionary info: dictionary = ContextualDictionary.en_US ; version = 1430519766 ; date = ?
D/BubblePopupHelper( 1405): BubblePopupHelper has no instance. Hence we cannot call setShowingBubblePopup : false
D/KeyguardUpdateMonitor( 1210): received broadcast com.lge.softkeypad.intent.SHOW
[...]

Issue #1516, opened 2014-05-20
Can't send SMS to numbers with MS Outlook phone number format


liayn:
I'm trying out TextSecure for the first time and wanted to send SMS to friends, who's contacts are stored in my exchange account.
The numbers were entered in MS Outlook and are stored in this (year long default) format:
```
+xx (xxx) xxxxxxx
```

The sending of the message fails without giving a reason. (I can text these people with the on-board messaging app and I can call them, so the number is correct.)

After playing around a bit, I noticed that the parentheses seem to be the culprit. Any other Android app silently discards them and dials the number.
I added a new number manually (without parentheses) and things worked out just fine.

Then I changed the number of the aforementioned friend to not contain the parentheses anymore in my address book and tried to re-add him to TextSecure. Unfortunately TextSecure somehow caches the old/other number somewhere and I'm not able to enter his number in any way. Neither by selecting it from the (now updated) address book, nor by entering it manually without any special characters. I guess #791 is related here.

Both problems are real show stoppers for using TextSecure. (I'm not able to update all phone numbers in the corporate phone book ;-))

Device-Specs:
Alcatel One Touch Idol X+ Dual SIM (but just using 1 SIM currently)
Android: 4.2.2

rrthomas:
I have the same problem, specifically with numbers containing parens in my address book (numbers starting with a + work fine, and spaces are OK too).

I have a Fairphone 1U running its latest Android 4.2.2, I'm in the UK on the T-Mobile network. In this case I simply deleted the parens from the relevant contact's number, and that fixed the problem.

liayn:
Closing this issue now, due to lack of response of the dev-team and serious attempts to get this fixed. I lost interest into TextSecure meanwhile anyway.

rrthomas:
@liayn, please can you reopen the issue, and let the developers close it if they decide they definitely won't fix it?

liayn:
1 year is enough grace period. Feel free to open a new issue and reference this one. I want that one off my issue list here in github. thx for understanding.

rrthomas:
@liayn feel free to unsubscribe.

liayn:
Doesn't remove it from my open issue list.

rrthomas:
[...]

PR #3099, opened 2015-05-02
Fix for recipients with unusual characters in their phone number.

brenns10:
I did some debugging and found that the reason for my NullPointerException in #3096 was because there was a no-break space (Unicode code point 0xA0) in between the area code and the phone number.  After quite a bit of trial and error, I was able to reproduce this with other contacts by adding/removing non-breaking spaces.  So I just used `PhoneNumberUtils.stripSeparators()` to remove those characters so that `SmsManager` wouldn't throw any exceptions, and that solves the issue.

This might be the reason for the issue described in the note in the code.

If I had to guess where the non-breaking space came from in my contact, I would say that it had to do with the web interface of Google Contacts (but that's just speculation).

### DIFF
--- a/src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
+++ b/src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
@@ -4,6 +4,7 @@ import android.app.PendingIntent;
 import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
+import android.telephony.PhoneNumberUtils;
 import android.telephony.SmsManager;
 import android.util.Log;
 
@@ -90,6 +91,10 @@ public class SmsSendJob extends SendJob {
     ArrayList<PendingIntent> sentIntents      = constructSentIntents(message.getId(), message.getType(), messages, false);
     ArrayList<PendingIntent> deliveredIntents = constructDeliveredIntents(message.getId(), message.getType(), messages);
     String recipient                          = message.getIndividualRecipient().getNumber();
+
+    // Remove all the separators from the phone number, since some (e.g. no-break space) can
+    // cause NPEs in SmsManager's sendTextMessage() function.
+    recipient = PhoneNumberUtils.stripSeparators(recipient);
 
     // NOTE 11/04/14 -- There's apparently a bug where for some unknown recipients
     // and messages, this will throw an NPE.  We have no idea why, so we're just


brenns10:
Ok, I would love to sign this Contributor License Agreement, but the [link](https://open-whisper-cla.appspot.com/cla-server/validate?redirect_url=https%3A%2F%2Fapi.github.com%2Frepos%2FWhisperSystems%2FTextSecure%2Fpulls%2F3099) takes me to a page that I can't fill out.  It looks like the Javascript for form submission expects to have jQuery, but doesn't.  So, the Submit button is broken.

moxie0:
sounds like a rom bug, what are you running?

McLoo:
Maybe #1516 is related here. Sounds similar

brenns10:
I'm running [xdabbeb's VS980](http://forum.xda-developers.com/verizon-g2/development/xdabbeb-s-vs982-5-v1-0-0-t2975551), which is just a modified stock 4.4.2 ROM.  I don't know enough to say whether my ROM has anything to do with it, but I'm sure it's a possibility.  I could potentially come up with a minimal working example APK that tries to send an SMS with a no-break space in it.  If it fails on other people's phones, it's probably not a ROM bug.

brenns10:
Minimum working example for the non-breaking space bug is [here](https://github.com/brenns10/NoBreakSpaceMWE).  The "NBSP" button should crash the app, while the "No Spaces" button will send a text message.  Works (i.e. crashes) on my device like a charm.  Doesn't crash on the emulator, but that could be because the emulator can't send text messages.

If the MWE crashes on other devices, that would indicate that it's not an issue with my ROM.

brenns10:
My MWE crashes on a different device: HTC One M8 GPE 4.4 ROM.  This suggests it is not a ROM issue specific to my phone.

2-4601:
Tested your example app on my U8880 with 4.0.4.
It does not crash when I press "NBSP"

I have an outdated SIM-card in the phone. I cannot send any SMS with it anymore. Don't know if this matters.

brenns10:
Thanks for testing.  It may make a difference that your phone can't do SMS -- the emulators don't crash while running it, and they can't do SMS either.  In any case, more data points are always helpful.

McLoo:
no crash on 4.4.4 Sony Z1 compact
used 11 and 13 digits

brenns10:
My code was written quickly, so I didn't make it robust.  It only attempts to send an SMS if it receives a 10 digit phone number (which is normal in USA).  Anything else and it just does nothing.  If you need a phone number to test it on, I can figure out a way to give you mine privately.

agrajaghh:
I had no problems too (12 digits, removed the check) on a Nexus 4 with 5.1.
I guess we need to test it on another LG G2 running unmodified stock 4.4.2...

brenns10:
OK, this is definitely not a ROM issue.  But, it looks like it is CDMA specific.  I traced all the way into the Android source (got it from Android SDK, api level 19 to match my phone).  

*Warning,* really long analysis incoming:  Let's start with the stack trace I got in #3096:

```
W/SmsSendJob( 6309): java.lang.NullPointerException
W/SmsSendJob( 6309):    at android.os.Parcel.readException(Parcel.java:1478)
W/SmsSendJob( 6309):    at android.os.Parcel.readException(Parcel.java:1426)
W/SmsSendJob( 6309):    at com.android.internal.telephony.ISms$Stub$Proxy.sendText(ISms.java:909)
W/SmsSendJob( 6309):    at android.telephony.SmsManager.sendTextMessage(SmsManager.java:194)
W/SmsSendJob( 6309):    at android.telephony.SmsManager.sendMultipartTextMessage(SmsManager.java:399)
W/SmsSendJob( 6309):    at org.thoughtcrime.securesms.jobs.SmsSendJob.deliver(SmsSendJob.java:99)
W/SmsSendJob( 6309):    at org.thoughtcrime.securesms.jobs.SmsSendJob.onSend(SmsSendJob.java:55)
W/SmsSendJob( 6309):    at org.thoughtcrime.securesms.jobs.SendJob.onRun(SendJob.java:39)
W/SmsSendJob( 6309):    at org.thoughtcrime.securesms.jobs.MasterSecretJob.onRun(MasterSecretJob.java:18)
W/SmsSendJob( 6309):    at org.whispersystems.jobqueue.JobConsumer.runJob(JobConsumer.java:76)
W/SmsSendJob( 6309):    at org.whispersystems.jobqueue.JobConsumer.run(JobConsumer.java:46)
W/SmsSendJob( 6309): org.thoughtcrime.securesms.transport.UndeliverableMessageException: java.lang.NullPointerException
W/SmsSendJob( 6309):    at org.thoughtcrime.securesms.jobs.SmsSendJob.deliver(SmsSendJob.java:113)
W/SmsSendJob( 6309):    at org.thoughtcrime.securesms.jobs.SmsSendJob.onSend(SmsSendJob.java:55)
W/SmsSendJob( 6309):    at org.thoughtcrime.securesms.jobs.SendJob.onRun(SendJob.java:39)
W/SmsSendJob( 6309):    at org.thoughtcrime.securesms.jobs.MasterSecretJob.onRun(MasterSecretJob.java:18)
W/SmsSendJob( 6309):    at org.whispersystems.jobqueue.JobConsumer.runJob(JobConsumer.java:76)
W/SmsSendJob( 6309):    at org.whispersystems.jobqueue.JobConsumer.run(JobConsumer.java:46)
W/SmsSendJob( 6309): Caused by: java.lang.NullPointerException
W/SmsSendJob( 6309):    at android.os.Parcel.readException(Parcel.java:1478)
W/SmsSendJob( 6309):    at android.os.Parcel.readException(Parcel.java:1426)
W/SmsSendJob( 6309):    at com.android.internal.telephony.ISms$Stub$Proxy.sendText(ISms.java:909)
W/SmsSendJob( 6309):    at android.telephony.SmsManager.sendTextMessage(SmsManager.java:194)
W/SmsSendJob( 6309):    at org.thoughtcrime.securesms.jobs.SmsSendJob.deliver(SmsSendJob.java:107)
```

It starts in `Parcel.readException()`, because the exception occurred in a different Service ("isms"), and is being read in.  The cause of the exception was in `SmsManager.java` (`android.telephony.SmsManager`).  Here is the code:

```java
    public void sendTextMessage(
            String destinationAddress, String scAddress, String text,
            PendingIntent sentIntent, PendingIntent deliveryIntent) {
        if (TextUtils.isEmpty(destinationAddress)) {
            throw new IllegalArgumentException("Invalid destinationAddress");
        }

        if (TextUtils.isEmpty(text)) {
            throw new IllegalArgumentException("Invalid message body");
        }

        try {
            ISms iccISms = ISms.Stub.asInterface(ServiceManager.getService("isms"));
            if (iccISms != null) {
                iccISms.sendText(ActivityThread.currentPackageName(), destinationAddress,
                        scAddress, text, sentIntent, deliveryIntent);
            }
        } catch (RemoteException ex) {
            // ignore it
        }
    }
```

The exception occurs on the `icclSms.sendText()` call.  It's not caught, since it's an NPE.  The implementation of the `ISms` service is in `com.android.internal.telephony.IccSmsInterfaceManager`.  Of note, their implementation of `sendText`:

```java
    public void sendText(String callingPackage, String destAddr, String scAddr,
            String text, PendingIntent sentIntent, PendingIntent deliveryIntent) {
        mPhone.getContext().enforceCallingPermission(
                Manifest.permission.SEND_SMS,
                "Sending SMS message");
        if (Rlog.isLoggable("SMS", Log.VERBOSE)) {
            log("sendText: destAddr=" + destAddr + " scAddr=" + scAddr +
                " text='"+ text + "' sentIntent=" +
                sentIntent + " deliveryIntent=" + deliveryIntent);
        }
        if (mAppOps.noteOp(AppOpsManager.OP_SEND_SMS, Binder.getCallingUid(),
                callingPackage) != AppOpsManager.MODE_ALLOWED) {
            return;
        }
        mDispatcher.sendText(destAddr, scAddr, text, sentIntent, deliveryIntent);
    }
```

The dispatcher is an instance of `com.android.internal.telephony.ImsSMSDispatcher`, so we go there to see the next level.  Here we go:

```java
    protected void sendText(String destAddr, String scAddr, String text,
            PendingIntent sentIntent, PendingIntent deliveryIntent) {
        Rlog.d(TAG, "sendText");
        if (isCdmaMo()) {
            mCdmaDispatcher.sendText(destAddr, scAddr,
                    text, sentIntent, deliveryIntent);
        } else {
            mGsmDispatcher.sendText(destAddr, scAddr,
                    text, sentIntent, deliveryIntent);
        }
    }
```

Since I have Verizon, I use the CDMA dispatcher, which is an instance of `com.android.internal.telephony.cdma.CdmaSMSDispatcher`.  Their implementation of `sendText()`:

```java
    protected void sendText(String destAddr, String scAddr, String text,
            PendingIntent sentIntent, PendingIntent deliveryIntent) {
        SmsMessage.SubmitPdu pdu = SmsMessage.getSubmitPdu(
                scAddr, destAddr, text, (deliveryIntent != null), null);
        HashMap map = getSmsTrackerMap(destAddr, scAddr, text, pdu);
        SmsTracker tracker = getSmsTracker(map, sentIntent,
                deliveryIntent, getFormat());
        sendSubmitPdu(tracker);
    }
```

So, `SmsMessage.getSubmitPdu()` seems to be what's constructing a CDMA SMS PDU (protocol data unit).  Looking in the implementation of that, we see:

```java
    public static SubmitPdu getSubmitPdu(String scAddr, String destAddr, String message,
            boolean statusReportRequested, SmsHeader smsHeader) {

        /**
         * TODO(cleanup): Do we really want silent failure like this?
         * Would it not be much more reasonable to make sure we don't
         * call this function if we really want nothing done?
         */
        if (message == null || destAddr == null) {
            return null;
        }

        UserData uData = new UserData();
        uData.payloadStr = message;
        uData.userDataHeader = smsHeader;
        return privateGetSubmitPdu(destAddr, statusReportRequested, uData);
    }
```

They delegate the main responsibility to `privateGetSubmitPdu`, which starts like this:

```java
    private static SubmitPdu privateGetSubmitPdu(String destAddrStr, boolean statusReportRequested,
            UserData userData) {

        /**
         * TODO(cleanup): give this function a more meaningful name.
         */

        /**
         * TODO(cleanup): Make returning null from the getSubmitPdu
         * variations meaningful -- clean up the error feedback
         * mechanism, and avoid null pointer exceptions.
         */

        /**
         * North America Plus Code :
         * Convert + code to 011 and dial out for international SMS
         */
        CdmaSmsAddress destAddr = CdmaSmsAddress.parse(
                PhoneNumberUtils.cdmaCheckAndProcessPlusCodeForSms(destAddrStr));
        if (destAddr == null) return null;

    //...
}
```

We can see that this will return null if `CdmaSmsAddress.parse()` returns null.  What happens if `privateGetSubmitPdu()` returns null?  Looking back, we see that `CdmaSMSDispatcher.sendText()` eventually calls its `getSmsTrackerMap()` function with `pdu` (which would be null).  This is inherited from `com.android.internal.telephony.SMSDispatcher`, and here is its code:

```java
    protected HashMap<String, Object> getSmsTrackerMap(String destAddr, String scAddr,
            int destPort, byte[] data, SmsMessageBase.SubmitPduBase pdu) {
        HashMap<String, Object> map = new HashMap<String, Object>();
        map.put("destAddr", destAddr);
        map.put("scAddr", scAddr);
        map.put("destPort", destPort);
        map.put("data", data);
        map.put("smsc", pdu.encodedScAddress);
        map.put("pdu", pdu.encodedMessage);
        return map;
    }
```

You can see that it does `pdu.encodedScAddress`.  If `pdu` is null, then we have an NPE.  So, if we can get `CdmaSMSAddress.parse()` to return null, we have the exception.  Here is that method:

```java
    public static CdmaSmsAddress parse(String address) {
        CdmaSmsAddress addr = new CdmaSmsAddress();
        addr.address = address;
        addr.ton = CdmaSmsAddress.TON_UNKNOWN;
        byte[] origBytes = null;
        String filteredAddr = filterNumericSugar(address);
        if (filteredAddr != null) {
            origBytes = parseToDtmf(filteredAddr);
        }
        if (origBytes != null) {
            addr.digitMode = DIGIT_MODE_4BIT_DTMF;
            addr.numberMode = NUMBER_MODE_NOT_DATA_NETWORK;
            if (address.indexOf('+') != -1) {
                addr.ton = TON_INTERNATIONAL_OR_IP;
            }
        } else {
            filteredAddr = filterWhitespace(address);
            origBytes = UserData.stringToAscii(filteredAddr);
            if (origBytes == null) {
                return null;
            }
            addr.digitMode = DIGIT_MODE_8BIT_CHAR;
            addr.numberMode = NUMBER_MODE_DATA_NETWORK;
            if (address.indexOf('@') != -1) {
                addr.ton = TON_NATIONAL_OR_EMAIL;
            }
        }
        addr.origBytes = origBytes;
        addr.numberOfDigits = origBytes.length;
        return addr;
    }
```

So, the first thing that happens to the address is that `filterNumericSugar()` gets called on it.

```java
    // Class fields and private initializer (relevant to code)
    private static final char[] numericCharsDialable = {
        '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '*', '#'
    };

    private static final char[] numericCharsSugar = {
        '(', ')', ' ', '-', '+', '.', '/', '\\'
    };

    private static final SparseBooleanArray numericCharDialableMap = new SparseBooleanArray (
            numericCharsDialable.length + numericCharsSugar.length);
    static {
        for (int i = 0; i < numericCharsDialable.length; i++) {
            numericCharDialableMap.put(numericCharsDialable[i], true);
        }
        for (int i = 0; i < numericCharsSugar.length; i++) {
            numericCharDialableMap.put(numericCharsSugar[i], false);
        }
    }



    private static String filterNumericSugar(String address) {
        StringBuilder builder = new StringBuilder();
        int len = address.length();
        for (int i = 0; i < len; i++) {
            char c = address.charAt(i);
            int mapIndex = numericCharDialableMap.indexOfKey(c);
            if (mapIndex < 0) return null;
            if (! numericCharDialableMap.valueAt(mapIndex)) continue;
            builder.append(c);
        }
        return builder.toString();
    }
```

For each character, `filterNumericSugar()` looks it up in the Map, which is initialized above.  Since NBSP is not in that map, `indexOfKey()` returns -1, so `filterNumericSugar()` returns null.  Going back to the `parse` function above, this means that the `else` block gets executed.  So, `filterWhitespace()` gets called, which looks like this:

```java
    private static String filterWhitespace(String address) {
        StringBuilder builder = new StringBuilder();
        int len = address.length();
        for (int i = 0; i < len; i++) {
            char c = address.charAt(i);
            if ((c == ' ') || (c == '\r') || (c == '\n') || (c == '\t')) continue;
            builder.append(c);
        }
        return builder.toString();
    }
```

Since this code hardcodes the allowable whitespace (i.e. doesn't use Java's character methods), the returned string would still contain a NBSP, if it were present.

The next call in `parse()` is `UserData.stringToAscii()`.  Here is the relevant code in that class:

```java
    /**
     * US ASCII character mapping table.
     *
     * This table contains only the printable ASCII characters, with a
     * 0x20 offset, meaning that the ASCII SPACE character is at index
     * 0, with the resulting code of 0x20.
     *
     * Note this mapping is also equivalent to that used by both the
     * IA5 and the IS-91 encodings.  For the former this is defined
     * using CCITT Rec. T.50 Tables 1 and 3.  For the latter IS 637 B,
     * Table 4.3.1.4.1-1 -- and note the encoding uses only 6 bits,
     * and hence only maps entries up to the '_' character.
     *
     */
    public static final char[] ASCII_MAP = {
        ' ', '!', '"', '#', '$', '%', '&', '\'', '(', ')', '*', '+', ',', '-', '.', '/',
        '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', ':', ';', '<', '=', '>', '?',
        '@', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O',
        'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', '[', '\\', ']', '^', '_',
        '`', 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o',
        'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', '{', '|', '}', '~'};

    /**
     * Character to use when forced to encode otherwise unencodable
     * characters, meaning those not in the respective ASCII or GSM
     * 7-bit encoding tables.  Current choice is SPACE, which is 0x20
     * in both the GSM-7bit and ASCII-7bit encodings.
     */
    static final byte UNENCODABLE_7_BIT_CHAR = 0x20;

    /**
     * Only elements between these indices in the ASCII table are printable.
     */
    public static final int PRINTABLE_ASCII_MIN_INDEX = 0x20;
    public static final int ASCII_NL_INDEX = 0x0A;
    public static final int ASCII_CR_INDEX = 0x0D;
    public static final SparseIntArray charToAscii = new SparseIntArray();
    static {
        for (int i = 0; i < ASCII_MAP.length; i++) {
            charToAscii.put(ASCII_MAP[i], PRINTABLE_ASCII_MIN_INDEX + i);
        }
        charToAscii.put('\n', ASCII_NL_INDEX);
        charToAscii.put('\r', ASCII_CR_INDEX);
    }

    /*
     * TODO(cleanup): Move this very generic functionality somewhere
     * more general.
     */
    /**
     * Given a string generate a corresponding ASCII-encoded byte
     * array, but limited to printable characters.  If the input
     * contains unprintable characters, return null.
     */
    public static byte[] stringToAscii(String str) {
        int len = str.length();
        byte[] result = new byte[len];
        for (int i = 0; i < len; i++) {
            int charCode = charToAscii.get(str.charAt(i), -1);
            if (charCode == -1) return null;
            result[i] = (byte)charCode;
        }
        return result;
    }
```

The `charToAscii` map used above doesn't contain NBSP, so when `stringToAscii()` gets to NBSP, it returns null.  Looking back at the `parse` function, since this returned null, `parse` also returns null.  And as we concluded earlier, when `parse` returns null, a NPE gets thrown.  And there we have it!  NPE is thrown when you attempt to send an SMS to an address containing an NBSP (or really anything that isn't in that `charToAscii` table) on CDMA.  I haven't checked GSM, and I really don't intend to (that analysis took all morning).

I don't know if this really qualifies as a bug in AOSP, given that there is no real reason to support a stupid character like NBSP in an SMS address.

brenns10:
I guess the TL;DR for the above long-winded debug session is:

* Found source of NPE in Android source.
* Not a ROM issue.
* Not sure if this is something to be fixed in TextSecure, or reported upstream?

agrajaghh:
Nice Analysis! Sounds like a bug in AOSP to me... But I guess it takes a *while* until its fixed and everybody got an update or a new phone ;) It seems like they knew about the NPEs already:
```` 
        /**
         * TODO(cleanup): Make returning null from the getSubmitPdu
         * variations meaningful -- clean up the error feedback
         * mechanism, and avoid null pointer exceptions.
         */
````

**Comment by brenns10** [comment]
Thanks!  I saw a few of those comments about returning null, so you would think they'd have thrown in a null check higher up the call stack as a catch-all.  Also, you'd think that there might be some more in-depth coverage in the Android documentation about what is and is not a valid SMS address (so you could have at least some clue why you were getting an NPE).  To be honest, I'm not sure if my fix (using `PhoneNumberUtils.stripSeparators()`) will screw up other phone numbers (like international ones), because I haven't found any good documentation about it.

**Comment by moxie0** [comment]
Yeah, problem is that I'm pretty sure a blanket stripSeparators call with break something somewhere.

**Comment by agrajaghh** [comment]
might be a bit hacky, but what about removing everything that is not in `ASCII_MAP`?

**Comment by moxie0** [comment]
looks like that's only for cdma, and it's not guaranteed to be the same on every device in every region

**Comment by brenns10** [comment]
As far as I can tell, the biggest problem with `stripSeparators()` would be sending text messages to email addresses.  I don't think it will change the meaning of any phone numbers.

**Comment by brenns10** [comment]
To elaborate on my previous comment, `stripSeparators()` does pretty much the same thing as your [NumberUtil.filterNumber()](https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/util/NumberUtil.java#L43) function.  Unfortunately, Android Studio says that that function is unused, so I can't tell whether anyone considers that safe to use.  But I imagine that it would be safe, so long as I did this:

```java
if (!NumberUtil.isValidEmail(recipient)) {
    recipient = PhoneNumberUtil.stripSeparators(recipient);
}
```

moxie0:
Yep, there's email, but also who knows what else. In general, I think we should be delivering messages to whatever people type in. If you think this nbsp thing is common, I might be fine with stripping those out, but I'm wary of trying to figure out the exact set of "correct" characters.

brenns10:
I'll admit that I don't think the NBSP issue is common.  I think NPE's from this particular area of Android might be common and should be dealt with upstream.  But when you think of the set of people likely to experience this (TextSecure user, CDMA, *and* happens to have NBSP in a contact, *and* texts them), it's not surprising that I'm the first (or one of very few) TextSecure user who noticed it.  I wouldn't expect you as responsible maintainers to make any changes that have a chance to hurt other users, in order to fix a bug that might occur in a handful of users.

Besides, filtering non-breaking spaces seems more like a hack than a fix.  Thinking of the code path I traced earlier, there are probably a lot of characters that would cause this sort of problem.  Filtering one doesn't prevent the others.  The best I could hope for is a more informative error message that gives users a place to start.  The reason I went so far down the debugging rabbit trail on this one was that I was given no feedback on what was wrong with my message or how I could fix it.

moxie0:
I'm fine with a PR for a better error message.

moxie0:
Closing this in anticipation of a PR that clarifies the error to the user, thanks for all the research.

PR #3104, opened 2015-05-04
Initial avatars

moxie0:


McLoo:
guess you want to remove 
https://github.com/WhisperSystems/TextSecure/blob/4397b55ceb8b82de1e8545e01dc7b6864fa85846/src/org/thoughtcrime/securesms/recipients/Recipient.java#L150
also

res/layout/conversation_list_item_view.xml
```diff
@@ -17,12 +17,12 @@
...
-        <com.makeramen.RoundedImageView
+        <org.thoughtcrime.securesms.components.AvatarImageView
             android:id="@+id/contact_photo_image"
             android:foreground="@drawable/contact_photo_background"
             android:layout_width="50dp"
             android:layout_height="50dp"
-            app:riv_oval="true"
+            app:circle="true"
             android:cropToPadding="true"
             android:scaleType="centerCrop"
```
mcginty: why not just have AvatarImageView hardcode the scaletype to `centerCrop` and always make it a circle since those values don't change and we can enforce consistency?

moxie0: Sure I can do that, I was just preserving what's there.

mcginty: Yeah, I feel like if we're going to name it `AvatarImageView` now it can dictate how it should look even more specifically.

moxie0:
updated

volcano99:
Does "initial avatars" means this is initial work on avatars, or does it mean the avatars will only be initials? ^^"

mcginty:
@volcano99 this change is to move the default placeholder avatar to being a color + initials instead of the boring gray one.

volcano99:
Thanks for clarifying! Are profile pictures on the road map though?

moxie0:
updated

PR #3105, opened 2015-05-04
smarter MMS logic, more chance of success

mcginty:
Instead of using only the Lollipoop API on Android > v21 (typo, but I will not change it given our success rate with that API), we'll use the legacy API with a couple changes after looking 

1. Utilize the hidden API `requestRouteToHostAddress` that takes an `InetAddress` (IPv6-capable) instead of a forced IPv4 integer-encoded address. Will fallback to the IPv4 one if reflection fails for whatever reason.
2. If on Lollipop and our manual MMS code doesn't work, will try to use the Lollipop API and give it 60 seconds instead of 30, since I did run into the timeout not being long enough in certain conditions and I'm thinking maybe it just wasn't long enough for some carriers.


moxie0:
looks reasonable to me

mcginty:
pushed to 2.14.0 branch

PR #3107, opened 2015-05-05
don't update the recent emoji tab while it is open

### LINKED ISSUE #3101: update emoji order after closing
With the last update, the emoji drawer changes the order of emojis in the recently used every time you use one. So it is kind of difficult to use the same emoji twice (you can't just tap twice as the emoji moves after the first tap). It would be more intuitive if the order was updated after the drawer is closed and reopened, and not "live" as it's being done now.
Did I explain it well?

rhodey:
don't update the recent emoji tab while it is open

Fixes #3101
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -324,7 +324,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   @Override
   public void onBackPressed() {
     if (emojiDrawer.isPresent() && emojiDrawer.get().getVisibility() == View.VISIBLE) {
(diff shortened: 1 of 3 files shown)


moxie0:
It's hard for me to follow what's going on here, but this seems like a tipping point to me. I think the whole thing needs to be refactored into a more sane MVC approach.

rhodey:
agreed, I wanted to bulldoze all this but instead tried to keep with the existing style. I would like if this was at least left open until the refactoring can be done and if for some reason that takes awhile we could merge this in the meantime.

moxie0:
i think this is a little too hacktastic to merge, and @mcginty is working on emoji this week, so i'm going to close this

PR #3112, opened 2015-05-05
consider search query when selecting all threads

### LINKED ISSUE #2996: "There are four lights!" or How conversation list's 'select all' disregards search results
1. Go to your conversation list
2. Search for some subset of your conversations
3. Tap and hold to select at least one of the conversations from the search results
4. Now tap 'select all' from the top right corner
5. TS will select **all of your conversations** instead of just all the search results

(6. If you want to live dangerously try to delete the selected messages by tapping the trash bin. It will remove all conversations. This proves that it really selects all conversations and not just misrepresent the number of selected messages.)

A .gif demonstration (does not strictly follow the above steps):
i see four lights

TS 2.11.1, Android 5.0.1 and also 4.0.4

rhodey:
consider search query when selecting all threads

Fixes #2996
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListAdapter.java
+++ b/src/org/thoughtcrime/securesms/ConversationListAdapter.java
@@ -131,17 +131,9 @@ public class ConversationListAdapter extends CursorRecyclerViewAdapter<Conversat
   }
 
   public void selectAllThreads() {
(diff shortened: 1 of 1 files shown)


moxie0:
in 2.14.0

PR #3116, opened 2015-05-07
switch default led color to blue

### LINKED ISSUE #3115: Update the default led notification color to match the current UI
This is such a minor issue that it feels almost a sin reporting this but sometimes details are important.

Currently the default led notification color is green. However the only green left in TextSecure are the message bubbles when using unencrypted SMS. Otherwise TextSecure is all about blue now (icon, UI, push messages). Therefore the default led color should follow _suit_.

rhodey:
switch default led color to blue

Fixes #3115
// FREEBIE

### DIFF
--- a/res/xml/preferences_notifications.xml
+++ b/res/xml/preferences_notifications.xml
@@ -21,7 +21,7 @@
 
     <ListPreference
         android:key="pref_led_color"
-        android:defaultValue="green"
+        android:defaultValue="blue"
         android:title="@string/preferences__led_color"
         android:dependency="pref_key_enable_notifications"
         android:entries="@array/pref_led_color_entries"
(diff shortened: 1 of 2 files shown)


moxie0:
iirc not all devices have blue

rhodey:
`MessageNotifier` is using `Color.parseColor(String)` on `NotificationCompat.Builder.setLights()`, the string "blue" is resolved using a static map, it resolves to `0xFF0000FF`.

I spent ~8 minutes searching the web for statistics on devices that don't support blue LED colors, *lots* of reports of LEDs not working at all or working sporadically, but I found nothing that related this behavior to a specific LED color. Some devices even disable LED notifications in power saving mode, like, this is something ROMs mess up a lot, but I don't think changing the color will make us any more or less vulnerable to their issues.

PR #3141, opened 2015-05-08
Fix for occasional generated avatar mis-sizing in conversation.

moxie0:
Drawables are (strangely) mutable objects.  We reuse a single
drawable for each recipient, but some avatar views (the
conversation list -- 40dp) are larger than others (the
conversation -- 30dp).

This results in a situation where TextDrawable doesn't render
itself appropriately, because the bounds are modified by a larger
view.

Giving the Drawable an intrinsic width and height resolves this
conflict.

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/contacts/ContactPhotoFactory.java
+++ b/src/org/thoughtcrime/securesms/contacts/ContactPhotoFactory.java
@@ -54,18 +54,24 @@ public class ContactPhotoFactory {
   }
 
   public static Drawable getDefaultContactPhoto(Context context, @Nullable String name) {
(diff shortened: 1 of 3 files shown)


McLoo:
for the record:
Lust noticed a pulsing (normal <> oversize) bubble in the conversation *list*, when upgrading an older version on a 5.0.2 device.
There where some unread messages which seemed to cause that, while the Lollipoop notifier was open.

moxie0:
@mcginty I'd like to put together 2.14.2 if you're cool with this

mcginty:
yeah, code looks reasonable

mcginty:
merged into master, corrected the whitespace at `src/org/thoughtcrime/securesms/contacts/ContactPhotoFactory.java:65`

PR #3153, opened 2015-05-12
create emoji package and better compartmentalized views

### LINKED ISSUE #2936: native system emoji overriding TextSecure-provided ones
Hello,

the Emoji drawer in TextSecure [currently is using the Google "Noto Color Emoji" Emoji set](https://github.com/WhisperSystems/TextSecure/issues/2894).

A lot of Android phones do not have the Google "Noto Color Emoji" Emoji set as their native system Emoji set though.

Samsung uses their own Emoji set, LG uses their own Emoji set, HTC uses their own Emoji set and so on...

You can see some of those Emoji sets over there:

http://emoji.codes/family

Unfortunately though, TextSecure does not replace the native system Emojis with the Google "Noto Color Emoji" Emojis, when the native system Emojis are not the Google "Noto Color Emoji" Emojis, like on Samsung, LG and HTC devices for example.

So, when Samsung/LG/HTC users tap on one of the Google "Noto Color Emoji" Emojis in the TextSecure Emoji drawer, a Samsung/LG/HTC Emoji appears in the message input box and in the messages themselves, and not the Google "Noto Color Emoji" Emojis the user actually selected.

So, could you please replace/overwrite the Emojis in the message input box and in the messages (and everywhere else where it is possible) with the Emojis that are being used in the Emoji drawer?

[Telegram for Android](https://github.com/DrKLO/Telegram) also has a built-in Emoji drawer and also has built-in Emojis.

And Telegram for Android does replace/overwrite the Emojis in the message input box and in the messages (and everywhere else where it is possible) with the Emojis that are being used in the Emoji drawer. So it is possible to do it.

Could you please fix this and do it like Telegram for Android is doing it?

Regards

### LINKED ISSUE #2938: Emoji category icons in drawer are not highlighted when selected
Hello,

the Emoji catergory icons in the TextSecure Emoji drawer are not being highlighted when they are selected.

They simply remain gray when they are selected.

With "Emoji category icons" I mean the clock, smiley, flower, bell, car and symbol icons at the top of the Emoji drawer, which allow to toggle between the different Emoji categories.

They are all gray and remain gray when selected.

But they should turn white when being selected so that they are visibly highlighted.

They also turn white on selection on the Google (AOSP) keyboard, on the Telegram for Android Emoji drawer, on the WhatsApp for Android Emoji drawer and so on...

TextSecure should do the same.

In the Telegram for Android and WhatsApp for Android Emoji drawers, the scrollbar below the Emoji categories also has a different color than grey, it's blue/green there.

It would be nice if that scrollbar in TextSecure would also have a special color (white would be okay as well I think) to make it stick out a little more.

Could you please do that?

Regards

mcginty:
Pages and category icons are now expressed in XML, and the general goal is to make it so the only classes necessary for interaction outside of the `emoji` package are views: `EmojiTextView`, `EmojiEditText`, and `EmojiDrawer`.

This change will make it very easy to break into its own project/repository.

fixes #2938
fixes #2936

##### notes
* `com.astuetz.pagerslidingtabstrip` has a min SDK of 10, but doesn't use any 10-only APIs.
* diff is +978/-773 lines if you don't count the emoji code XML :)


AndroidManifest.xml
```diff
@@ -5,7 +5,7 @@
       android:versionCode="115"
       android:versionName="2.14.1">
 
-    <uses-sdk tools:overrideLibrary="com.amulyakhare.textdrawable"/>
+    <uses-sdk tools:overrideLibrary="com.amulyakhare.textdrawable,com.astuetz.pagerslidingtabstrip"/>
```
moxie0: looks like you removed the dependency though?

src/org/thoughtcrime/securesms/components/emoji/Emoji.java
```diff
@@ -0,0 +1,257 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.ref.SoftReference;
+import java.util.Arrays;
+import java.util.concurrent.ExecutorService;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class Emoji {
```
moxie0: In an MVC world, it's not clear to me why this class still needs to exist or what its role is.

src/org/thoughtcrime/securesms/components/emoji/EmojiDrawer.java
```diff
@@ -0,0 +1,118 @@
...
+  public void show() {
+    int keyboardHeight = getKeyboardHeight();
+    Log.w("EmojiDrawer", "setting emoji drawer to height " + keyboardHeight);
+    setLayoutParams(new LinearLayout.LayoutParams(RelativeLayout.LayoutParams.MATCH_PARENT, keyboardHeight));
+    requestLayout();
+    setVisibility(View.VISIBLE);
+  }
+
+  private void initializeEmojiGrid() {
+    if (RecentEmoji.getInstance(getContext()).getCount() <= 0) {
```
moxie0: This seems strange to me.  The drawer is the controller, and should only be coordinating between events from the views (the strip and the page views). RecentEmoji as a singleton that it references directly seems kind of strange.

src/org/thoughtcrime/securesms/components/emoji/EmojiDrawer.java
```diff
@@ -0,0 +1,118 @@
...
+    strip.setViewPager(pager);
+  }
+
+  private void insertEmoji(Integer unicodePoint) {
+    final char[] chars = Character.toChars(unicodePoint);
+    String characters = new String(chars);
+    int start = composeText.getSelectionStart();
+    int end   = composeText.getSelectionEnd();
+
+    CharSequence text = Emoji.getInstance(getContext()).emojify(characters, new Emoji.InvalidatingPageLoadedListener(composeText));
```
moxie0: Why does the controller do this? Isn't that the job of the view?

src/org/thoughtcrime/securesms/components/emoji/RecentEmoji.java
```diff
@@ -0,0 +1,98 @@
...
+
+import com.fasterxml.jackson.databind.type.TypeFactory;
+
+import org.thoughtcrime.securesms.util.JsonUtils;
+
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+
+public class RecentEmoji {
```
moxie0: I think this should just be the implementation of a model interface, not something special.

moxie0:
Basically, I still think there should be four classes here: EmojiDrawer (controller), EmojiPageModel (model), EmojiPageView (view), EmojiStripView (view).

AndroidManifest.xml
mcginty: The fork I moved to still uses the same package name as the original no-longer-maintained one.

src/org/thoughtcrime/securesms/components/emoji/Emoji.java
mcginty: There needs to be some container that manages the emoji bitmaps that are shared in both the drawer and EmojiTextViews. An alternative would be to manage this as statics within EmojiDrawable, but that sounds like a strange place to put it.

src/org/thoughtcrime/securesms/components/emoji/EmojiDrawer.java
mcginty: Sure, we can move this into the pager.

src/org/thoughtcrime/securesms/components/emoji/EmojiDrawer.java
mcginty: I missed this when I was removing emojify calls, fixing.

src/org/thoughtcrime/securesms/components/emoji/RecentEmoji.java
mcginty: And add an onEmojiSelected to the EmojiPageModel interface to alert models when an item has been chosen?

src/org/thoughtcrime/securesms/components/emoji/Emoji.java
moxie0: Why statics? It's logic for the view, the view seems like the best place to put it.

src/org/thoughtcrime/securesms/components/emoji/EmojiPager.java
```diff
@@ -0,0 +1,72 @@
...
+    attrSet.recycle();
+
+    final int[] icons = ResUtil.getResourceIds(getContext(), iconsRes);
+    final int[] pages = ResUtil.getResourceIds(getContext(), pagesRes);
+    final List<EmojiPageModel> models = new LinkedList<>();
+    models.add(new RecentEmojiPageModel(getContext()));
+    for (int i = 0; i < icons.length; i++) {
+      models.add(new StaticEmojiPageModel(icons[i], getResources().getIntArray(pages[i])));
+    }
+    return models;
```
moxie0: It should be the job of the controller to set models on the view, and it should be the job of the model to be parsing the data.

AndroidManifest.xml
moxie0: Somewhat suspicious that someone would fork a library and then update the minSDK.  Seems like they'd be most likely to do that if they were actually using APIs introduced at higher levels?

src/org/thoughtcrime/securesms/components/emoji/EmojiPagerAdapter.java
```diff
@@ -0,0 +1,51 @@
...
+import android.view.ViewGroup;
+import android.widget.ImageView;
+
+import com.astuetz.PagerSlidingTabStrip;
+
+import org.thoughtcrime.securesms.components.emoji.EmojiPageFragment.EmojiSelectionListener;
+
+import java.util.List;
+
+public class EmojiPagerAdapter extends FragmentStatePagerAdapter
```
moxie0: feel like this should be an inner class of EmojiDrawer

moxie0:
can you rebase this?

mcginty:
done

Issue #1628, opened 2014-06-22
Force close after reopen unsaved image draft


Deccoy:
Steps to Reproduce
- open conversation
- add image attachment, but no text
- go back to contact list
- reopen conversation

--> Error, Force Close

Issue #1556, opened 2014-06-05
Crash on reading drafts with attachments


wickedshimmy:
I'll take a look at this tomorrow, but in case anyone else has seen it:
- Added attachment (image) to a message but did not send.
- Hit back, closing the message. TextSecure notified that it was saving draft.
- Go back into that conversation, crash.

On reopening TextSecure the draft is also gone.

Reproducible with different images, with and without an accompanying message, and with both an encrypted SMS session and a regular SMS conversation (I'm not able to test with push contacts).

```
--------- beginning of /dev/log/system
E/AndroidRuntime( 1342): FATAL EXCEPTION: main
E/AndroidRuntime( 1342): Process: org.thoughtcrime.securesms, PID: 1342
E/AndroidRuntime( 1342): java.lang.SecurityException: Permission Denial: opening provider com.android.providers.media.MediaDocumentsProvider from ProcessRecord{434c3718 1342:org.thoughtcrime.securesms/u0a146} (pid=1342, uid=10146) requires android.permission.MANAGE_DOCUMENTS or android.permission.MANAGE_DOCUMENTS
E/AndroidRuntime( 1342):    at android.os.Parcel.readException(Parcel.java:1465)
E/AndroidRuntime( 1342):    at android.os.Parcel.readException(Parcel.java:1419)
E/AndroidRuntime( 1342):    at android.app.ActivityManagerProxy.getContentProvider(ActivityManagerNative.java:2887)
E/AndroidRuntime( 1342):    at android.app.ActivityThread.acquireProvider(ActivityThread.java:4496)
E/AndroidRuntime( 1342):    at android.app.ContextImpl$ApplicationContentResolver.acquireUnstableProvider(ContextImpl.java:2328)
E/AndroidRuntime( 1342):    at android.content.ContentResolver.acquireUnstableProvider(ContentResolver.java:1425)
E/AndroidRuntime( 1342):    at android.content.ContentResolver.openTypedAssetFileDescriptor(ContentResolver.java:1047)
E/AndroidRuntime( 1342):    at android.content.ContentResolver.openAssetFileDescriptor(ContentResolver.java:904)
E/AndroidRuntime( 1342):    at android.content.ContentResolver.openInputStream(ContentResolver.java:629)
E/AndroidRuntime( 1342):    at org.thoughtcrime.securesms.util.BitmapUtil.createScaledBytes(BitmapUtil.java:29)
[...]

Issue #1544, opened 2014-05-31
Force close when pressing back after adding attachment


nh2:
To reproduce:
- Start TextSecure
- Go into first conversation
- Click menu (`...`)
- `Add attachment` -> `Picture`
- Click first picture
- Do not send picture, press back button
- You are now in the conversation list and a Toast appears saying `Saved draft`
- Press the first contact again
- `Unfortunately, TextSecure has stopped`

Tested on Moto X, Android 4.4.2.

Can you reproduce?

Issue #926, opened 2014-03-02
Crashing after saving screenshot as draft


phime42:
After I composed a message with a screenshot in it and went back, TextSecure showed "Saving draft.." and crashed afterwards. Passphrase had to be entered to start it again.

PR #3154, opened 2015-05-12
Remove ACTION_OPEN_DOCUMENT from AttachmentManager

### LINKED ISSUE #1744: Attachments cannot be sent from gallery - on certain phones sometimes
When trying to add an attachment, e.g. a picture, the dialog "Open" will only show the folder "Recent" and "Downloads". Sending pictures or videos that are stored in the gallery would be really nice.

unrulygnu:
Fixes #1744

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
+++ b/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
@@ -183,7 +183,7 @@ public class AttachmentManager {
                                             @NonNull final MediaType mediaType,
                                             @NonNull final MediaConstraints constraints)
   {
-    new AsyncTask<Void, Void, Uri>() {
+    new AsyncTask<Void, Void, InputStream>() {
       @Override
       protected void onPreExecute() {
         thumbnail.clear();
@@ -192,13 +192,9 @@ public class AttachmentManager {
       }
 
       @Override
-      protected @Nullable Uri doInBackground(Void... params) {
+      protected InputStream doInBackground(Void... params) {
         try {
-          final InputStream input = context.getContentResolver().openInputStream(uri);
-          if (input == null) return null;
-          // TODO: handle null, wrap in getCorrectedMimeType() if needed
-          final String mimeType = MediaUtil.getMimeType(context, uri);
-          return PersistentBlobProvider.getInstance(context).create(masterSecret, input, mimeType);
+          return context.getContentResolver().openInputStream(uri);
         } catch (IOException ioe) {
           Log.w(TAG, ioe);
           return null;
@@ -206,12 +202,33 @@ public class AttachmentManager {
       }
 
       @Override
-      protected void onPostExecute(@Nullable final Uri persistentUri) {
-        if (persistentUri == null) {
-          abortAttachment(context.getString(R.string.ConversationActivity_sorry_there_was_an_error_setting_your_attachment));
-        } else {
-          setMediaWithPersistentUri(masterSecret, persistentUri, mediaType, constraints, false);
+      protected void onPostExecute(InputStream result) {
+        final String mimeType = MediaUtil.getMimeType(context, uri);
+        if (result == null || mimeType == null) {
+          abort();
+          return;
         }
+
+        PersistentBlobProvider.getInstance(context).createFuture(masterSecret, result, mimeType)
+            .addListener(new ListenableFuture.Listener<Uri>() {
+              @Override
+              public void onSuccess(Uri result) {
+                if (result == null) {
+                  abort();
+                } else {
+                  setMediaWithPersistentUri(masterSecret, result, mediaType, constraints, false);
+                }
+              }
+
+              @Override
+              public void onFailure(ExecutionException e) {
+                abort();
+              }
+            });
+      }
+
+      private void abort() {
+        abortAttachment(context.getString(R.string.ConversationActivity_sorry_there_was_an_error_setting_your_attachment));
       }
     }.execute();
   }
--- a/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
+++ b/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
@@ -23,7 +23,6 @@ import android.content.Intent;
 import android.graphics.Bitmap;
 import android.net.Uri;
 import android.os.AsyncTask;
-import android.os.Build;
 import android.provider.ContactsContract;
 import android.provider.MediaStore;
 import android.support.annotation.NonNull;
@@ -55,6 +54,7 @@ import org.thoughtcrime.securesms.util.concurrent.ListenableFuture.Listener;
 import org.whispersystems.libaxolotl.util.guava.Optional;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -174,12 +174,62 @@ public class AttachmentManager {
                        @NonNull final MediaType mediaType,
                        @NonNull final MediaConstraints constraints)
   {
-    new AsyncTask<Void, Void, Slide>() {
+    if (MediaUtil.isPersistent(uri)) setMediaWithPersistentUri(masterSecret, uri, mediaType, constraints, true);
+    else                             setMediaWithNonPersistentUri(masterSecret, uri, mediaType, constraints);
+  }
+
+  private void setMediaWithNonPersistentUri(@NonNull final MasterSecret masterSecret,
+                                            @NonNull final Uri uri,
+                                            @NonNull final MediaType mediaType,
+                                            @NonNull final MediaConstraints constraints)
+  {
+    new AsyncTask<Void, Void, Uri>() {
       @Override
       protected void onPreExecute() {
         thumbnail.clear();
         thumbnail.showProgressSpinner();
         attachmentView.setVisibility(View.VISIBLE);
+      }
+
+      @Override
+      protected @Nullable Uri doInBackground(Void... params) {
+        try {
+          final InputStream input = context.getContentResolver().openInputStream(uri);
+          if (input == null) return null;
+          // TODO: handle null, wrap in getCorrectedMimeType() if needed
+          final String mimeType = MediaUtil.getMimeType(context, uri);
+          return PersistentBlobProvider.getInstance(context).create(masterSecret, input, mimeType);
+        } catch (IOException ioe) {
+          Log.w(TAG, ioe);
+          return null;
+        }
+      }
+
+      @Override
+      protected void onPostExecute(@Nullable final Uri persistentUri) {
+        if (persistentUri == null) {
+          abortAttachment(context.getString(R.string.ConversationActivity_sorry_there_was_an_error_setting_your_attachment));
+        } else {
+          setMediaWithPersistentUri(masterSecret, persistentUri, mediaType, constraints, false);
+        }
+      }
+    }.execute();
+  }
+
+  private void setMediaWithPersistentUri(@NonNull final MasterSecret masterSecret,
+                                         @NonNull final Uri uri,
+                                         @NonNull final MediaType mediaType,
+                                         @NonNull final MediaConstraints constraints,
+                                                  final boolean initThumbnail)
+  {
+    new AsyncTask<Void, Void, Slide>() {
+      @Override
+      protected void onPreExecute() {
+        if (initThumbnail) {
+          thumbnail.clear();
+          thumbnail.showProgressSpinner();
+          attachmentView.setVisibility(View.VISIBLE);
+        }
       }
 
       @Override
@@ -199,15 +249,9 @@ public class AttachmentManager {
       @Override
       protected void onPostExecute(@Nullable final Slide slide) {
         if (slide == null) {
-          attachmentView.setVisibility(View.GONE);
-          Toast.makeText(context,
-                         R.string.ConversationActivity_sorry_there_was_an_error_setting_your_attachment,
-                         Toast.LENGTH_SHORT).show();
+          abortAttachment(context.getString(R.string.ConversationActivity_sorry_there_was_an_error_setting_your_attachment));
         } else if (!areConstraintsSatisfied(context, masterSecret, slide, constraints)) {
-          attachmentView.setVisibility(View.GONE);
-          Toast.makeText(context,
-                         R.string.ConversationActivity_attachment_exceeds_size_limits,
-                         Toast.LENGTH_SHORT).show();
+          abortAttachment(context.getString(R.string.ConversationActivity_attachment_exceeds_size_limits));
         } else {
           setSlide(slide);
           attachmentView.setVisibility(View.VISIBLE);
@@ -224,6 +268,11 @@ public class AttachmentManager {
         }
       }
     }.execute();
+  }
+
+  private void abortAttachment(String message) {
+    attachmentView.setVisibility(View.GONE);
+    Toast.makeText(context, message, Toast.LENGTH_SHORT).show();
   }
 
   public boolean isAttachmentPresent() {
@@ -289,17 +338,6 @@ public class AttachmentManager {
   private static void selectMediaType(Activity activity, String type, int requestCode) {
     final Intent intent = new Intent();
     intent.setType(type);
-
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
-      intent.setAction(Intent.ACTION_OPEN_DOCUMENT);
-      try {
-        activity.startActivityForResult(intent, requestCode);
-        return;
-      } catch (ActivityNotFoundException anfe) {
-        Log.w(TAG, "couldn't complete ACTION_OPEN_DOCUMENT, no activity found. falling back.");
-      }
-    }
-
     intent.setAction(Intent.ACTION_GET_CONTENT);
     try {
       activity.startActivityForResult(intent, requestCode);
--- a/src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
+++ b/src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
@@ -55,6 +55,7 @@ public class PersistentBlobProvider {
   }
 
   private final Context           context;
+
   @SuppressLint("UseSparseArrays")
   private final Map<Long, byte[]> cache    = Collections.synchronizedMap(new HashMap<Long, byte[]>());
   private final ExecutorService   executor = Executors.newCachedThreadPool();
--- a/src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
+++ b/src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
@@ -14,6 +14,8 @@ import org.thoughtcrime.securesms.crypto.DecryptingPartInputStream;
 import org.thoughtcrime.securesms.crypto.EncryptingPartOutputStream;
 import org.thoughtcrime.securesms.crypto.MasterSecret;
 import org.thoughtcrime.securesms.util.Util;
+import org.thoughtcrime.securesms.util.concurrent.ListenableFuture;
+import org.thoughtcrime.securesms.util.concurrent.SettableFuture;
 
 import java.io.ByteArrayInputStream;
 import java.io.File;
@@ -55,7 +57,6 @@ public class PersistentBlobProvider {
   }
 
   private final Context           context;
-
   @SuppressLint("UseSparseArrays")
   private final Map<Long, byte[]> cache    = Collections.synchronizedMap(new HashMap<Long, byte[]>());
   private final ExecutorService   executor = Executors.newCachedThreadPool();
@@ -68,7 +69,7 @@ public class PersistentBlobProvider {
                     @NonNull byte[] blobBytes,
                     @NonNull String mimeType)
   {
-    final long id = System.currentTimeMillis();
+    final long id = newId();
     cache.put(id, blobBytes);
     return create(masterSecret, new ByteArrayInputStream(blobBytes), id, mimeType);
   }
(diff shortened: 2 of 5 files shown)


rhodey:
hey @unrulygnu, thanks for looking into this. can you clarify which device(s) you used to reproduce this bug and your PR? I'd like to gather a list of affected devices and use [TestObject](https://testobject.com) to make sure this causes no regressions or other weird behavior. Intent actions are often super buggy or device specific on Android and usually no fun :|

unrulygnu:
@rhodey , thanks for the quick response.

* Google Nexus 5 D820
* Android 5.1 LMY47I stock; no superuser
* TextSecure 2.9.3
* AOSP Gallery2 1.1.40030 (com.android.gallery3d)
* Google Photos DISABLED

Please see <a href="https://github.com/WhisperSystems/TextSecure/issues/1744#issuecomment-101235654">my comment on the issue</a> for a full report.

Also, I can't sign the CLA because the <a href="http://support.whispersystems.org/customer/portal/questions/11686974-can-t-use-the-contributor-license-agreement">CLA page is apparently broken.</a>  Love the project, and would love to contribute more; do you know whom to contact regarding fixing the website?

rhodey:
Sure, thanks for the info and PR :) I think @moxie0 has noticed that the CLA is no longer working (if not, well, here ya go) and I'm sure he'll get around to is when possible. I'll check back in after running the tests.

rhodey:
just adding some context in here as I was searching about...

Samsung devices have the most trouble with this: http://stackoverflow.com/questions/24616788/kitkat-action-open-document-does-not-show-documents-on-samsung-devices

Google contradicts themselves [here](https://developer.android.com/guide/topics/providers/document-provider.html#client) in first stating:

> ACTION_OPEN_DOCUMENT is not intended to be a replacement for ACTION_GET_CONTENT. The one you should use depends on the needs of your app:

> Use ACTION_GET_CONTENT if you want your app to simply read/import data. With this approach, the app imports a copy of the data, such as an image file.
> Use ACTION_OPEN_DOCUMENT if you want your app to have long term, persistent access to documents owned by a document provider. An example would be a photo-editing app that lets users edit images stored in a document provider.

but then later and on the same page [here](https://developer.android.com/guide/topics/providers/document-provider.html#custom) they state:

> The ACTION_OPEN_DOCUMENT intent is only available on devices running Android 4.4 and higher. If you want your application to support ACTION_GET_CONTENT to accommodate devices that are running Android 4.3 and lower, you should disable the ACTION_GET_CONTENT intent filter in your manifest for devices running Android 4.4 or higher. A document provider and ACTION_GET_CONTENT should be considered mutually exclusive. If you support both of them simultaneously, your app will appear twice in the system picker UI, offering two different ways of accessing your stored data. This would be confusing for users.

the use of `ACTION_OPEN_DOCUMENT` BY TextSecure was first introduced by fa3cb871d05417e286511c72e40f44e74947886c, the commit message is rather detailed but doesn't actually make much sense to me and the link provided for context is the same from which I grabbed the contradiction and confusing text above.

This is actually a difficult thing to test for because it's not so much device specific as it is app specific. I think ideally we would have some way to figure out which action would yield the most results and then go with that, short of some magic that gives us that insight I think we should revert fa3cb871d05417e286511c72e40f44e74947886c by merging this commit. This pull even works better for my device (stock N5) which is not evenly reportedly a problem device.

check out the photos in [this imgur album](http://imgur.com/a/1Cg7X), the first is from this PR, the second from master. in conclusion, looks good to me and would like to see this merged :)

unrulygnu:
Awesome, thanks for the additional info!  Wasn't aware of the Samsung issue at all.

I think that Google wasn't contradicting themselves, but rather talking about two different things.  The first quote is in under the "Writing a Client App" heading (using `startActivityForResult()` to access data from other apps), whereas the second quote is under the "Writing a Custom Document Provider" heading (providing data to other apps).

In that context, it makes sense, although the second quote could use some clarification.  I'm figuring that if an app has a Document Provider, that provider is going to show up in the `DocumentsActivity` (I'm assuming above the line, as Drive is in your screenshots) for both `ACTION_OPEN_DOCUMENT` and `ACTION_GET_CONTENT`. So if the same app also as an `intent-filter` for `ACTION_GET_CONTENT`, then the app will also show up below the line in that case.  Conditionally filtering 'ACTION_GET_CONTENT' only for pre-KitKat devices prevents this duplication, hence the mutual exclusivity.

Glad the PR seems to make an improvement.  Looking forward to more substantial contributions soon! Next stop, completing the Korean translation. ^^

rhodey:
@unrulygnu I think the CLA page should be fixed now, FYI.

moxie0:
Yeah iirc @tinloaf put this in, and there was some subtle reason why it needed to be there.  @tinloaf what do you think about this change?

tinloaf:
See #928 - the problem is that on KitKat, without ACTION_OPEN_DOCUMENT, you only get a read-once handle of the file. If you safe a draft and later return to that draft, ye olde TextSecure tried to open the file again (for showing the draft) and crashed. Not sure if that is still the behavior today, though.

See https://developer.android.com/guide/topics/providers/document-provider.html#client for documentation on the difference between ACTION_OPEN_DOCUMENT and ACTION_GET_CONTENT.

unrulygnu:
@rhodey Many thanks, CLA signed.

@tinloaf Further down on the same page is <a href="https://developer.android.com/guide/topics/providers/document-provider.html#permissions">a discussion about maintaining persistent read/write access</a>:

> Persist permissions
> 
> When your app opens a file for reading or writing, the system gives your app a URI permission grant for that file. It lasts until the user's device restarts. But suppose your app is an image-editing app, and you want users to be able to access the last 5 images they edited, directly from your app. If the user's device has restarted, you'd have to send the user back to the system picker to find the files, which is obviously not ideal.
> 
> To prevent this from happening, you can persist the permissions the system gives your app. Effectively, your app "takes" the persistable URI permission grant that the system is offering. This gives the user continued access to the files through your app, even if the device has been restarted...

    final int takeFlags = intent.getFlags()
                & (Intent.FLAG_GRANT_READ_URI_PERMISSION
                | Intent.FLAG_GRANT_WRITE_URI_PERMISSION);
    // Check for the freshest data.
    getContentResolver().takePersistableUriPermission(uri, takeFlags);

When checking TextSecure for specific reasons for implementing `ACTION_OPEN_DOCUMENT`, I also looked for <a href="https://developer.android.com/reference/android/content/ContentResolver.html#takePersistableUriPermission%28android.net.Uri,%20int%29">`ContentResolver#takePersistableUriPermission`</a>, but came up dry.  Without anything that stood out as obvious, I submitted the PR.

Thanks, everyone, for the followup.

rhodey:
I just tested this PR vs master (4c89b242be51afee7e28861912c9e0a2b7e62ebd) on a Samsung Galaxy S4 running KitKat 4.4.2.

On master the devices behaves much like every other device I've used, when adding attachments you get this UI that allows you to chose from recent or google drive photos and the gallery.

With this PR I get a dialog asking which app I want to use to complete the action, I'm given 3 options-- gallery, photos, and some samsumg cloud junk. I took a few pictures and tested adding them from gallery or photos, saving them as drafts, and restarting the phone, ie everything I could to produce this error-on-second-read thing we're attempting to avoid with ACTION_OPEN_DOCUMENT. I never ran into any crashes or unexpected behavior.

This PR has improved UX on every platform I've tested, so definitely still feeling good about merging it.

rhodey:
so I'm seeing all these old crash reports that are seemingly related to KitKat and use of `ACTION_GET_CONTENT`, for example #926, #1628, #1556, and #1544 but none of the steps to reproduce in any of those issues result in a crash or unexpected behavior when tested on this PR.

looking back it appears that the use of `ACTION_OPEN_DOCUMENT` did stop the crashes from happening but it is also clear that for our use case `ACTION_GET_CONTENT` is actually what we want (RE: all the issues reported in #1744).

So, I think that `ACTION_GET_CONTENT` is definitely the correct action to use across all API levels *but maybe* we should add some permission magic to this URI conditionally for API levels above KitKat? I don't know and we can't understand any of this until we can crash a device using `ACTION_GET_CONTENT` again :'(

maybe `AttachmentManager` has changed a lot since then? maybe glide fixed this? @unrulygnu sorry for all the hold ups but it is difficult to merge this until we understand why `ACTION_GET_CONTENT` used to cause so many crashes on Kitkat and now does not.

rhodey:
OK, I was finally able to reproduce some behavior reminiscent of all those earlier crash bugs.

1) open conversation
2) tap add attachment
3) from "recent" category select an image that you haven't ever sent before
4) notice that the image preview is now shown as a draft
5) back all the way out of the app
6) force close the app by swiping it out of the recent activities UI thing
7) open app, return to the conversation
8) notice that the draft preview is broken, log says file not found

note: image definitely has to be from the recent category and must not have ever been opened by TS before.

the above does not happen on master (4c89b242be51afee7e28861912c9e0a2b7e62ebd), only when testing this PR. I think we still need to stop using `ACTION_OPEN_DOCUMENT` to resolve #1744 but we probably need to add some URI permission magic similar to what's described [in this writeup](http://www.andreamaglie.com/blog/2015/01/09/access-storage-framework-uri-permission/)

unrulygnu:
@rhodey , nothing to apologize for; if anything, your thorough investigation, documentation, and communication with a new contributor on related history is encouraging.  Keep up the awesome!  :)

On a Nexus 5 running stock Android 5.1 LMY47I, I followed your procedure exactly and was unable to reproduce the issue on master (7ecc580); however, I *did* get the `FileNotFoundException` in the logs for both `ACTION_OPEN_DOCUMENT` and `ACTION_GET_CONTENT`:

**`ACTION_OPEN_DOCUMENT`**

    05-21 12:12:52.575      745-761/? I/ActivityManager﹕ START u0 {act=android.intent.action.OPEN_DOCUMENT typ=image/* cmp=com.android.documentsui/.DocumentsActivity} from uid 10104 on display 0
    05-21 12:12:52.578      745-761/? V/WindowManager﹕ addAppToken: AppWindowToken{2fab8e46 token=Token{2b570a21 ActivityRecord{3bf3f088 u0 com.android.documentsui/.DocumentsActivity t1672}}} to stack=1 task=1672 at 2
    ...
    05-21 12:13:58.868  10628-10628/org.thoughtcrime.securesms E/BitmapFactory﹕ Unable to decode stream: java.io.FileNotFoundException: : open failed: ENOENT (No such file or directory)
    05-21 12:13:58.868  10628-10628/org.thoughtcrime.securesms W/BitmapDrawable﹕ BitmapDrawable cannot decode

**`ACTION_GET_CONTENT`**

    05-21 12:18:44.264     745-1418/? I/ActivityManager﹕ START u0 {act=android.intent.action.GET_CONTENT typ=image/* cmp=com.android.documentsui/.DocumentsActivity} from uid 10104 on display 0
    05-21 12:18:44.267     745-1418/? V/WindowManager﹕ addAppToken: AppWindowToken{39edca00 token=Token{19b65883 ActivityRecord{553d332 u0 com.android.documentsui/.DocumentsActivity t1674}}} to stack=1 task=1674 at 2
    ...
    05-21 12:19:48.279  19477-19477/org.thoughtcrime.securesms E/BitmapFactory﹕ Unable to decode stream: java.io.FileNotFoundException: : open failed: ENOENT (No such file or directory)
    05-21 12:19:48.279  19477-19477/org.thoughtcrime.securesms W/BitmapDrawable﹕ BitmapDrawable cannot decode

Again, in both cases, the draft preview image thumbnail displayed correctly (Nexus 5).

After lunch, I'll test on an emulator running KitKat, then see if I can get my hands on a Samsung Galaxy S4.  Shouldn't be too hard in Seoul, right?

Once I reproduce, I'll investigate further (thanks for the link you provided!) and update the PR.

rhodey:
I tried to find another way to reproduce this kinda behavior using less steps but came up with nothing less complex. However, I did discover that in addition to swiping TextSecure out of the recent activities UI, [Settings > Applications > TextSecure > Force stop] also results in the same behavior once revisiting the draft.

UI as viewed from step 4
screen

UI as viewed from step 8
screen

rhodey:

unrulygnu:
@rhodey , thanks for the update, and I apologize for not having followed up in the last week.

I did manage to look into the <a href="http://www.andreamaglie.com/blog/2015/01/09/access-storage-framework-uri-permission/">permissions solution</a> you had linked earlier, but it seems to be specific to `ACTION_OPEN_DOCUMENT` and possibly a content provider in the same app.  Just to confirm, I tried out some code, called in `onActivityResult()`:

      private void takeAttachmentPermission(Intent data) {
        if (VERSION.SDK_INT >= VERSION_CODES.KITKAT) {
          //grantUriPermission(getPackageName(), data.getData(), Intent.FLAG_GRANT_READ_URI_PERMISSION);
          final int takeFlags = data.getFlags() & Intent.FLAG_GRANT_READ_URI_PERMISSION;
          //noinspection ResourceType
          getContentResolver().takePersistableUriPermission(data.getData(), takeFlags);
        }
      }

With `ACTION_GET_CONTENT`, the last line throws a `java.lang.SecurityException: No persistable permission grants found`.  With `ACTION_OPEN_DOCUMENT`, it runs fine.

I finally found a victim with a Samsung phone running Kitkat, and I hope to get my hands on it on Monday or Tuesday to try to reproduce the issue.  Sorry again for the delay.

rhodey:
no worries :) glad to see you're making progress on this and other PR's, the issue this PR addresses is especially strange.

unrulygnu:
@rhodey , sorry for the further delay; was focusing my spare time on finishing the Korean translation for TextSecure before coming back to this PR.

Tonight, I tested on a Japanese Samsung device running KitKat:

* Samsung Galaxy J SC-02F
* Android 4.4.2
* Kernel: 3.4.0-2862522
* Build KOT49H.SC02FOMUFNI3

Unfortunately, with `ACTION_GET_CONTENT`, the system launches the pre-KitKat chooser dialog rather than the `DocumentsActivity`.  I confirmed this behavior on another app, as well. With `ACTION_OPEN_DOCUMENT`, however, the `DocumentsActivity` is launched as expected in KitKat and later.

Because of this behavior, which may be due to a special OS build, I cannot reproduce the issue you're seeing on this device; it looks like I may specifically need an S4 running KitKat to continue following up.  I'll keep harassing people I know.  :)

rhodey:
awesome updates @unrulygnu, keep with it :+1: :D

unrulygnu:
It seems that even when using `ACTION_GET_CONTENT`, any URI's with the `com.android.providers.media.documents` authority (i.e., media picked in `DocumentsActivity`) are given the you-are-only-allowed-to-see-this-once-bro permissions treatment from the `MediaDocumentsProvider`.  This results in the issue @rhodey indicated when selecting an image in a draft message, killing the app, then being greeted by an unavailable image in the draft preview.

As a workaround, we can convert the `com.android.providers.media.documents` URI into a good old fashioned `MediaStore` URI, which doesn't include any of that permissions stuff. I have implemented this in the most recent commit to the PR.

This workaround does violate the abstraction that the Storage Access Framework is supposed to implement, and it depends upon an implementation detail of `MediaDocumentsProvider`, where it uses the same ID for media as the `MediaStore`.  Although highly unlikely, if this were to change, media attachments will break.  

I can add validation to help preclude that scenario, if needed.  Please let me know what you think.

unrulygnu:
@rhodey :

I rebased on latest master, looking forward to your feedback.  I've tested on several devices over the past three weeks with zero issues. :)

rhodey:
tested extensively with: photo, audio, and video coming from: in app, and sharing from 3rd party apps after: for stopping, and full device restart on: Gingerbread and KitKat. works as expected :)

only glitch I ran into was gingerbread is slow to init all the media related things so once I restarted the GB device and then opened TextSecure very soon after boot and received a file not found exception. issue is non-reproducible if you wait awhile for GB to warm up and other apps displayed similar behavior in this situation so no problem but good to have a non-perfect thing else I'd suspect I missed something.

lgtm.

unrulygnu:
Seems like there's some demand for this patch: https://github.com/WhisperSystems/TextSecure/issues/1744#issuecomment-122345219

@rhodey : is there anything else I need to do to see about getting this PR merged?

unrulygnu:
@moxie0 : Checking in to see if this PR can be merged.  Please see the following:

Latest update summary: https://github.com/WhisperSystems/TextSecure/pull/3154#issuecomment-113183808 
@rhodey 's review: https://github.com/WhisperSystems/TextSecure/pull/3154#issuecomment-121088777 
Recent issue hate: https://github.com/WhisperSystems/TextSecure/issues/1744#issuecomment-122345219

unrulygnu:
Rebased on current master (eedbc66).  Tested, works.

moxie0:
Hey @unrulygnu thanks for all the hard work on this.  I'm extremely hesitant to break the media framework's abstraction layer by hand-crafting URIs.  If the permission bug is KitKat-only (?), maybe we should just use ACTION_OPEN_DOCUMENT for <= KK, and ACTION_GET_CONTENT for LP+?

unrulygnu:
@moxie0 : Unfortunately, in both KitKat and Lollipop, any media picked in `DocumentsActivity` -- even via `ACTION_GET_CONTENT` -- return URI's with the `com.android.providers.media.documents` authority and the resulting view-only-once permission.  I have not yet tested in Marshmallow preview, however.

I share your reluctance to violate the abstraction layer.  I'll give it a fresh look when I get the chance to see if there are better alternatives.

Thanks for the follow up! ^^

unrulygnu:
@mcginty: are you going to continue on and eventually commit "unjankify incoming media sharing" to master?  What you're doing there may be a good solution to our dilemma here.

moxie0:
@unrulygnu That got merged...

unrulygnu:
Thanks, on it. :+1:

unrulygnu:
59f2446 's `PersistentBlobProvider` implementation is awesome!

I pushed a new commit that checks if the attachment URI has the `com.android.providers.media.documents` authority, in which case it implements  `PersistentBlobProvider` to cache a local copy that will persist through app restarts while the attachment is still a draft. 

This [tests well] (https://github.com/WhisperSystems/TextSecure/pull/3154#issuecomment-104069377) so far, but I'd like to give it one or two days of further testing before considering it ready to merge. Any feedback before then is welcome. ^^

unrulygnu:
@moxie0: Tested, rebased, good to go. :+1:

unrulygnu:
New issue comment requesting what this PR offers: https://github.com/WhisperSystems/Signal-Android/issues/1744#issuecomment-160559544

@moxie0: Do you need anything else from me on this one?  It's still labeled "Waiting on Contributor".

src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -138,6 +139,44 @@ public void setMedia(@NonNull final MasterSecret masterSecret,
                        @NonNull final MediaType mediaType,
                        @NonNull final MediaConstraints constraints)
   {
+    if ("com.android.providers.media.documents".equals(uri.getAuthority())) {
```
moxie0: There's no public constant for this? That makes me nervous

unrulygnu: Previously and again at the time of this PR, I couldn't find a publicly accessible constant (like `MediaStore.AUTHORITY`), and this hack seems to be the industry standard for identifying URI's from the document provider.  

My hands got full yesterday and today, but I'll do more research in the next few days to see if there's any safer method.

unrulygnu:
@@ -138,6 +139,44 @@ public void setMedia(@NonNull final MasterSecret masterSecret,
                            @NonNull final MediaType mediaType,
                            @NonNull final MediaConstraints constraints)
       {
    +    if ("com.android.providers.media.documents".equals(uri.getAuthority())) {

> There's no public constant for this? That makes me nervous

Changed the check from the only condition *of which we know* under which the URI is **non-persistent**, to the only conditions under which the URI can be safely assumed is a **persistent** URI.  

This not only keeps us in happy constant heaven, but it also is a better and safer implementation overall, now that we treat third-party URI's as non-persistent by default. 

Thanks, @moxie0.  Tested and updated.

src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -138,6 +139,47 @@ public void setMedia(@NonNull final MasterSecret masterSecret,
...
+           PersistentBlobProvider.AUTHORITY.equals(uri.getAuthority()) ||
+           ContentResolver.SCHEME_FILE     .equals(uri.getScheme());
+  }
+
+  private void setMediaWithNonPersistentUri(@NonNull final MasterSecret masterSecret,
+                                            @NonNull final Uri uri,
+                                            @NonNull final MediaType mediaType,
+                                            @NonNull final MediaConstraints constraints)
+  {
+    new AsyncTask<Void, Void, Uri>() {
```
moxie0: We don't want to display anything here? It's a full file copy, could be kinda slow right?

unrulygnu:
Sorry for the delay, juggling went full circus this week.  I encountered an attachment error several times while running this branch this week, and I'm not yet sure whether it's related to this PR.  I'll follow up this weekend.

unrulygnu:
@moxie0 :
> We don't want to display anything here? It's a full file copy, could be kinda slow right?

Fixed.

Also the attachment error in my previous comment was due to the lack of memory cache on one of the `PersistentBlobProvider#create` overloads.  All public `PersistentBlobProvider#create` methods now cache the blob in memory before writing, with a `PersistentBlobProvider#createWithoutCache` method available for special cases.

Refactored `Util#readFully` to use `Util#copy`, as it was redundant.

Rebased, tested.

src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
```diff
@@ -60,15 +61,26 @@ private PersistentBlobProvider(Context context) {
...
-    cache.put(id, imageBytes);
-    return create(masterSecret, new ByteArrayInputStream(imageBytes), id);
+    cache.put(id, blobBytes);
+    return create(masterSecret, new ByteArrayInputStream(blobBytes), id);
   }
 
-  public Uri create(@NonNull MasterSecret masterSecret,
-                    @NonNull InputStream input)
+  public @Nullable Uri create(@NonNull MasterSecret masterSecret,
+                              @NonNull InputStream input)
```
moxie0: I don't understand the rationale for this. The point of using a stream is explicitly to avoid holding an in-memory copy.  I can't think of any situation where we'd want to load an entire stream into memory when we don't have to.

unrulygnu: Do you have any objection to making `PersistentBlobProvider#persistToDisk` synchronous, placing the burden on callers of `PersistentBlobProvider#create` to do so off of the main thread?  This would, for instance, defer the `AttachmentManager#setSlide` call until after the persistent file is written, during which there will be a progress spinner and an unblocked UI thread anyway.

I can explore other alternatives, but this seems like the option of lowest impact so far.

moxie0: I'm not sure what would work, given some of the callers. Might make more sense to have it return a ListenableFuture<Uri>...

unrulygnu: Thanks for the feedback.  I'll revisit it this weekend.

unrulygnu:
@moxie0: 

> Might make more sense to have it return a ListenableFuture...

`PersistentBlobProvider#persistToDisk` now returns a `ListenableFuture`, and stream caching is removed. Tests well with a variety of media types and file sizes.  Kept the `openInputStream()` call in an `AsyncTask` due to its blocking > 20ms on larger media files.

Please let me know what you think.  I can rebase to a single commit when ready.

src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
```diff
@@ -60,41 +64,66 @@ private PersistentBlobProvider(Context context) {
   }
 
   public Uri create(@NonNull MasterSecret masterSecret,
```
moxie0: There should be one method that returns ListenableFuture<Uri>

unrulygnu: @moxie0: To clarify, are you meaning that a public method should return the `ListenableFuture` returned by `PersistentBlobProvider#persistToDisk` (which you suggested in your previous comment) back up to the caller in the other class, rather than the caller passing in a listener as an argument?

Or is there something specific about this particular overload?

moxie0: There should be a single public method that returns a future

unrulygnu:
I've done some work on this, but I'd like to wait until #4689 is merged before continuing, for there be conflicts.

unrulygnu:
@moxie0: Updated.  Please let me know what you think.

jkdf2:
I was about to take a look at #5272, but this PR will fix or at least greatly affect that.

unrulygnu:
@jkdf2: I think this PR won't have any effect on that issue.  

The log in that issue indicates that `MediaUtil#getMimeType(Context, Uri)` is called with `null` for the `Uri` parameter, which comes straight from the result `Intent`.  

Signal is expecting a single `Uri` to be returned via `getData()`, for which there's no `null` check at this point.  Note that when a result `Intent` includes multiple `Uri`s, I think `getData()` returns null, with the `Uri`s being available via `getClipData()` instead.

This PR doesn't touch any of that, so if you'd like to submit a new PR on that issue... :+1:

KarolTrzeszczkowski:
Hi! Why isn't this pull request merged yet? I'm not complaining, just asking.

moxie0:
Some proposed changes in #5653

PR #3171, opened 2015-05-15
fix recent emoji pane

mcginty:
1) Make recent list properly update and invalidate.
2) Show most-recently-used first.
3) As a bonus, make drafts and emoji get along.

### DIFF
--- a/src/org/thoughtcrime/securesms/components/emoji/EmojiEditText.java
+++ b/src/org/thoughtcrime/securesms/components/emoji/EmojiEditText.java
@@ -3,36 +3,26 @@ package org.thoughtcrime.securesms.components.emoji;
 import android.content.Context;
 import android.support.v7.widget.AppCompatEditText;
 import android.util.AttributeSet;
-import android.util.Log;
 
-import org.thoughtcrime.securesms.components.emoji.EmojiProvider.InvalidatingPageLoadedListener;
 
 public class EmojiEditText extends AppCompatEditText {
   public EmojiEditText(Context context) {
     super(context);
-    init();
   }
 
   public EmojiEditText(Context context, AttributeSet attrs) {
     super(context, attrs);
-    init();
   }
 
   public EmojiEditText(Context context, AttributeSet attrs,
                        int defStyleAttr)
   {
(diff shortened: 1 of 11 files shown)


src/org/thoughtcrime/securesms/components/emoji/EmojiPageModel.java
```diff
@@ -1,7 +1,12 @@
 package org.thoughtcrime.securesms.components.emoji;
 
-public interface EmojiPageModel {
-  int getIconRes();
-  int[] getCodePoints();
-  void onCodePointSelected(int codePoint);
+public abstract class EmojiPageModel {
+  public abstract int getIconRes();
+  public abstract int[] getCodePoints();
+  public void onCodePointSelected(int codePoint) { }
```
moxie0: does this need to be in the interface?

src/org/thoughtcrime/securesms/components/emoji/RecentEmojiPageModel.java
```diff
@@ -50,10 +51,11 @@ public RecentEmojiPageModel(Context context) {
   }
 
   @Override public int[] getCodePoints() {
```
moxie0: why are these primitive arrays and not lists?

moxie0:
Just looked at EmojiProvider, what is all that executor service invalidate drawable stuff?  I think EmojiProvider should be synchronous, and anything wanting asynchronous loads should be using a loader.

src/org/thoughtcrime/securesms/components/emoji/EmojiEditText.java
```diff
@@ -25,6 +25,13 @@ public EmojiEditText(Context context, AttributeSet attrs,
     init();
   }
 
+  @Override public void setText(CharSequence text, BufferType type) {
```
moxie0: What is the purpose of the insertEmoji method if you're overriding this?

mcginty: getText().replace() should be more efficient, as we don't need to run the regex replacement over the entire text when inserting.

PR #3176, opened 2015-05-15
Add Norwegian Bokmål (nb)

sandakersmann:
Add Norwegian Bokmål (nb) to translations.


mcginty:
Hey @sandakersmann, all our translations are automatically imported from Transifex so no need to manually add the strings.xml file. Thanks for contributing!

mcginty:
cc @moxie0 we might have to add norwegian to the transifex config and our language string arrays though

PR #3183, opened 2015-05-18
Fix image cache bugs.

### LINKED ISSUE #840: Wrong image in chat preview
Hello, 

when I send an image, in the chat there is shown an image I sent before instead of the right one I just have sent. When it is clicked, the right one is shown in full screen.

Even when I delete the conversation and I send a new image, the old wrong picture is shown in chat. (Fullscreen shows the right image). Does Deletion of conversation not delete data of the conversation?

Greets

moxie0:
1) Index SlideDeckListener cache by MMS (id, timestamp) tuple.

2) Index parts by (id, content_id) tuples.

Fixes #840
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/database/MmsDatabase.java
+++ b/src/org/thoughtcrime/securesms/database/MmsDatabase.java
@@ -165,8 +165,8 @@ public class MmsDatabase extends MessagingDatabase {
   };
 
   public static final ExecutorService slideResolver = org.thoughtcrime.securesms.util.Util.newSingleThreadedLifoExecutor();
-  private static final Map<Long, SoftReference<SlideDeck>> slideCache =
-      Collections.synchronizedMap(new LRUCache<Long, SoftReference<SlideDeck>>(20));
+  private static final Map<String, SoftReference<SlideDeck>> slideCache =
+      Collections.synchronizedMap(new LRUCache<String, SoftReference<SlideDeck>>(20));
 
   private final JobManager jobManager;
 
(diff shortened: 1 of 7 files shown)


src/org/thoughtcrime/securesms/database/MmsDatabase.java
```diff
@@ -1128,21 +1129,21 @@ public SlideDeck call() throws Exception {
           SlideDeck    slideDeck    = new SlideDeck(context, masterSecret, body);
 
           if (!body.containsPushInProgress()) {
-            slideCache.put(id, new SoftReference<SlideDeck>(slideDeck));
+            slideCache.put(timestamp + "::" + id, new SoftReference<>(slideDeck));
```
rhodey: how can we cache on `timestamp` (`dateReceived`) for sent messages without wasting a lookup? doesn't this get updated from some unknown condition (sending) to an actual valid timestamp upon delivery? ie: two unique IDs per slide

moxie0: the received timestamp never changes. for outgoing messages, it is always the same as sent.

src/ws/com/google/android/mms/pdu/PduPart.java
```diff
@@ -142,6 +142,7 @@
       */
      public PduPart() {
          mPartHeader = new HashMap<Integer, Object>();
+         setContentId(String.valueOf(System.currentTimeMillis()).getBytes());
```
rhodey: maybe throw some logging statements in to detect this condition and ones like it. I can't think of all the places we might end up with a PduPart that has a fake `contentId` but everytime we do we risk bad utilization of our cache so logging or assertion errors might help us be proactive about that.

moxie0: detect this condition? what condition?

rhodey: maybe I've gotta search through usages more but I recall some places where we'll create PduParts and end up not inserting them into the DB. seems like if we use a PduPart that doesn't come from the database we'll end up with a cache miss because of the way content ID is set, right?

src/org/thoughtcrime/securesms/database/MmsDatabase.java
rhodey: cool, magic :P

src/ws/com/google/android/mms/pdu/PduPart.java
moxie0: if it's not in the database, then there's nothing to load from the database, so there's no cache to miss since it's already in memory?

PR #3184, opened 2015-05-18
Fix for media thumbnails flickering on model updates.

### LINKED ISSUE #1004: Image flickers during send
I just wrote a message to a group and lost my wifi connection. After I hit 'send', an image shown in a message above started flickering rapidly. The flickering only stopped after the message was eventually delivered. Could it be the case that the image is repeatedly redrawn with each attempt to send the message?

### LINKED ISSUE #2663: Attachments are flickering when sending a message
Attachments thumbnails (picture, video and audio) in the ConversationActivity are flickering ~4 times if you send a new textmessage (or another attachment) after #2430 

I guess you are already aware of that, but I thought I open this for completeness...

moxie0:
Only update ImageView contents when they have changed.

Fixes #1004
Fixes #2663

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/components/ThumbnailView.java
+++ b/src/org/thoughtcrime/securesms/components/ThumbnailView.java
@@ -4,6 +4,7 @@ import android.annotation.TargetApi;
 import android.app.Activity;
 import android.content.Context;
 import android.graphics.Bitmap;
+import android.graphics.drawable.Drawable;
 import android.net.Uri;
 import android.os.Build.VERSION;
 import android.os.Build.VERSION_CODES;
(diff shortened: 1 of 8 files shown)


src/org/thoughtcrime/securesms/components/ThumbnailView.java
```diff
@@ -53,31 +58,41 @@ public ThumbnailView(Context context, AttributeSet attrs, int defStyle) {
...
-                               @Nullable MasterSecret masterSecret)
+  public void setImageResource(@Nullable MasterSecret masterSecret,
+                               long id, long timestamp,
+                               @NonNull ListenableFutureTask<SlideDeck> slideDeckFuture)
   {
     if (this.slideDeckFuture != null && this.slideDeckListener != null) {
       this.slideDeckFuture.removeListener(this.slideDeckListener);
     }
 
+    String slideId = id + "::" + timestamp;
```
rhodey: `timestamp` is `MessageRecord.getDateReceived()`, which for a message I send changes from presumably [`-1` or `null` or something] to the actual time of receipt. so, wouldn't this result in the two unique IDs for one slide and thus two non-cache lookups?

rhodey: I guess not two lookups yet, was merging some of this and the image cache fix in my head. however the image drawable would be set to null and then have to be reset once the receive timestamp is updated.

moxie0: the received timestamp is never updated

PR #3186, opened 2015-05-18
Support for direct photo capture from app.

moxie0:
Make the send button toggle to an attach button when the
compose text and attachments are emmpty.

Part of #520

// FREEBIE


rhodey:
the diff looks good to me, I'd rename `cleanup()` to be more descriptive given that we already have a `clear()` but w/e.

switching between the attachment icon and the send button on my N5 causes the layout containing the emoji toggle, edit text, and animating toggle to change in size by a few pixels. also from an empty edit text with the attachment icon present, if I type a letter and then immediately tap the attach icon I can break the toggle state to show a send icon when the edit text is empty.

rhodey:
the change in layout size that I mentioned above also happens in #3198, tagging as possibly related.

McLoo:
shouldn't `<uses-permission android:name="android.permission.CAMERA" />` be added to the manifest.xml?

*really don't know whether I should comment on a pulled PR or the commit - ignoring the question if the comment is helpful anyway:D*

moxie0:
@McLoo As an intermediate step we're using an intent to open the camera app and take the picture for us, so we don't actually need permission to access the camera directly.  @turtlekiosk is working on a PR to do direct camera access for a better experience, and we'll need that permission then.

moxie0:
(and yeah i think commenting on the pr is better than the commit)

Issue #3127, opened 2015-05-08
Crash while trying to display picture message thumbnails


2-4601:
1. Open a conversation that has picture messages
2. If necessary scroll up to find a picture message
3. Crash

https://gist.github.com/23d11c902a98491c22ab
```
E/AndroidRuntime(26145): FATAL EXCEPTION: main
E/AndroidRuntime(26145): Process: org.thoughtcrime.securesms, PID: 26145
E/AndroidRuntime(26145): java.lang.IllegalArgumentException: Unable to find GlideModule implementation
E/AndroidRuntime(26145): 	at com.bumptech.glide.module.ManifestParser.parseModule(Unknown Source)
E/AndroidRuntime(26145): 	at com.bumptech.glide.module.ManifestParser.parse(Unknown Source)
E/AndroidRuntime(26145): 	at com.bumptech.glide.Glide.get(Unknown Source)
E/AndroidRuntime(26145): 	at com.bumptech.glide.RequestManager.<init>(Unknown Source)
E/AndroidRuntime(26145): 	at com.bumptech.glide.RequestManager.<init>(Unknown Source)
E/AndroidRuntime(26145): 	at com.bumptech.glide.manager.RequestManagerRetriever.supportFragmentGet(Unknown Source)
E/AndroidRuntime(26145): 	at com.bumptech.glide.manager.RequestManagerRetriever.get(Unknown Source)
E/AndroidRuntime(26145): 	at com.bumptech.glide.manager.RequestManagerRetriever.get(Unknown Source)
E/AndroidRuntime(26145): 	at com.bumptech.glide.Glide.with(Unknown Source)
E/AndroidRuntime(26145): 	at org.thoughtcrime.securesms.components.ThumbnailView.buildEncryptedPartGlideRequest(Unknown Source)
E/AndroidRuntime(26145): 	at org.thoughtcrime.securesms.components.ThumbnailView.buildThumbnailGlideRequest(Unknown Source)
E/AndroidRuntime(26145): 	at org.thoughtcrime.securesms.components.ThumbnailView.buildGlideRequest(Unknown Source)
E/AndroidRuntime(26145): 	at org.thoughtcrime.securesms.components.ThumbnailView.setImageResource(Unknown Source)
E/AndroidRuntime(26145): 	at org.thoughtcrime.securesms.components.ThumbnailView$SlideDeckListener$1.run(Unknown Source)
[...]

PR #3187, opened 2015-05-18
no more Release.java

rhodey:
1) move PUSH_URL from Release.java into BuildConfig.java
2) avoid redeclaration of proguard files, prevents future #3127
3) move all packagingOptions exclusions into a single closure

the first is necessary for my upcoming espresso test PR, the second and third make our build file more readable which makes espresso test related changes easier to create and understand.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -156,6 +156,8 @@ android {
         exclude 'LICENSE'
         exclude 'NOTICE'
         exclude 'asm-license.txt'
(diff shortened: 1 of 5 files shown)


rhodey:
should these be separate pulls? splitting them into separate commits made sense to me but since we're talking about the build.gradle I figured one pull would be less context switches in between.

PR #3190, opened 2015-05-18
Espresso tests

rhodey:
## espresso prerequisites & cleanup
1) move PUSH_URL from Release.java into BuildConfig.java
2) avoid redeclaration of proguard files, prevents future #3127
3) move all packagingOptions exclusions into a single closure

## espresso tests
1) bumped espresso-core dependency to 2.1
2) removed unecessary testing-support-lib dependency
3) fixed proguard-testing.pro rules
4) use TS staging server for all tests
5) wrote a bunch of espresso tests
6) simplified the androidTest source set definition
7) added rule to optionally include espressoTest/ dir in androidTest source set
8) added gradle.properties and test device numbers to .gitignore

// FREEBIE

we want to keep our espresso tests separate from our regular unit tests, ideally we would do this by defining a new build target called `espresso` or whatever but this doesn't work because although gradle was nice enough to give us `testBuildType` there is only one test source set, `androidTest`, and gradle still considers it to be a magic special case.

so instead we optionally include the `espressoTest/` directory into the `androidTest` source set, defaulting to not included, and overridden by setting the property `espresso` either through a `gradle.properties` file or via command line parameter.

personally I'd like to rename the `androidTest` directory `unitTest` or something more appropriate but kinda getting picky at that point :p

**warning: running the espresso tests will delete all contacts on your test device**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ tests
 lint.xml
 local.properties
 ant.properties
+gradle.properties
 .DS_Store
 build.log
 build-log.xml
@@ -19,3 +20,4 @@ build
 signing.properties
 library/lib/
 library/obj/
+espressoTest/res/values/arrays.xml
--- a/espressoTest/java/org/thoughtcrime/securesms/ConversationListActivityTest.java
+++ b/espressoTest/java/org/thoughtcrime/securesms/ConversationListActivityTest.java
@@ -1,0 +1,279 @@
+/**
+ * Copyright (C) 2015 Open Whisper Systems
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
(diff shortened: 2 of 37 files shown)


rhodey:
I had originally grouped the first 3 commits into their own pull (#3187) but I guess I'm not really sure how to best structure this.

.gitignore
```diff
@@ -19,3 +20,4 @@ build
 signing.properties
 library/lib/
 library/obj/
+espressoTest/res/values/arrays.xml
```
mcginty: what's this used for?

rhodey: that contains all our fake test device numbers + registration codes and is only used for running certain espresso tests. decided not to publish those codes for now even though they only work on staging.

rhodey:
OH! also important to note: **running the espresso tests will delete all contacts on your test device**. contacts are deleted at the start of every test and sometimes added during a test because we have got to test those states!

espressoTest/java/org/thoughtcrime/securesms/ConversationListActivityTest.java
```diff
@@ -0,0 +1,279 @@
...
+    waitOn(ViewLocalIdentityActivity.class);
+  }
+
+  public void testClickSettings() throws Exception {
+    loadAndCheckState();
+    ConversationListActivityActions.clickSettings(getContext());
+    waitOn(ApplicationPreferencesActivity.class);
+  }
+
+  public static Matcher<Object> isThreadFrom(final String contactName) {
```
mcginty: should these be in `ViewMatchers`?

rhodey: if we end up using them anywhere else then yeah but I can't see a use for them outside `ConversationListActivity` related tests

rhodey:
the "espresso prerequisites & cleanup" section of this pull has been moved into #3213 and #3212. I will reopen this once I rebase appropriately.

Issue #2891, opened 2015-04-04
Alignment of the keyboard / Smiley switch for long texts


mad-de:
If you type a long message, the keyboard / Smiley switch gets aligned vertically in the middle, thereby moving upwards the longer your message is.
Whatsapp etc. has a fixed position on the bottom, which makes it easier for unskilled users to notice it's position.
2.10.1
![screenshot_2015-04-04-12-25-18](https://cloud.githubusercontent.com/assets/7095883/6992305/20cda0b6-dac7-11e4-9d68-bfc5024268d6.jpg)

agrajaghh:
I agree, a fixed position at the bottom would look better. Same aplies to the send button...

wp9015362:
Hello,

yes, in Telegram it's also the same as in WhatsApp, the buttons stay fixed at the bottom in both Telegram as well as in WhatsApp.

In the TextSecure screenshot from @mad-de, which is shown above, it simply looks buggy.

So, TextSecure should also have the buttons in a fixed position at the bottom, just like in Telegram and in WhatsApp.

And the font size in the message input box should be the same as the message font size (currently it's not, see: https://github.com/WhisperSystems/TextSecure/issues/2769).

Regards

PR #3191, opened 2015-05-19
fix for issue #2891: Alignment of the keyboard / Smiley switch for long texts

herbig:
FREEBIE

Simple fix for issue #2891 and added ripple effect with the same color for conversation button backgrounds.  Tested on Nexus 6 and Galaxy SII, no visual change to UI (aside from alignment).

Just getting familiar with the codebase, I'll check out something more complex once I'm more acclimated.

### DIFF
--- a/res/layout/conversation_activity.xml
+++ b/res/layout/conversation_activity.xml
@@ -70,8 +70,8 @@
                 <org.thoughtcrime.securesms.components.emoji.EmojiToggle
                         android:id="@+id/emoji_toggle"
                         android:layout_width="wrap_content"
-                        android:layout_height="fill_parent"
-                        android:layout_gravity="center_vertical"
+                        android:layout_height="48dp"
+                        android:layout_gravity="bottom"
                         android:background="@drawable/touch_highlight_background"
                         android:contentDescription="@string/conversation_activity__emoji_toggle_description"
                         android:padding="10dp"/>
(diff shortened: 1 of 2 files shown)


res/layout/conversation_activity.xml
```diff
@@ -70,8 +70,8 @@
                 <org.thoughtcrime.securesms.components.emoji.EmojiToggle
                         android:id="@+id/emoji_toggle"
                         android:layout_width="wrap_content"
-                        android:layout_height="fill_parent"
-                        android:layout_gravity="center_vertical"
+                        android:layout_height="48dp"
```
mcginty: why 48dp instead of wrap_content?

herbig: wrap_content will wrap to the height of the emoji icon / send button assets, making them smaller than they were previously.

The way they were before, with fill_parent, the height constraining element was the height of the EditText, which is 48dp when single line: compose_background is 24dp height + (12dp top + 12dp bottom padding on EditText).

agrajaghh:
I guess the problem with `48dp` is that the heigth of the `EditText` depends on the system fontsize. So for a bigger fontsize the icons will be aligned on the bottom. Should I open another issue for that?

herbig:
@moxie0 I see this was merged with a few edits.  wrap_content doesn't work here, the tap zone of both buttons are now smaller than the height of the row, which looks awkward.

@agrajaghh yeah, for font sizes larger than default, you'd see white space above both buttons when tapping on them

PR #3195, opened 2015-05-19
Fix ever-expanding text box

### LINKED ISSUE #2899: problem with writing long messages
When writing a long message, at some point you become unable to see the conversation and scroll up. It may be my screen that is too small, though it's 4+inches so I guess a lot of people are affected

volcano99:
Fixes #2899. I am not sure why textMultiLine was removed in the first place, but I tested it and works as expected now.

### DIFF
--- a/res/layout/conversation_activity.xml
+++ b/res/layout/conversation_activity.xml
@@ -89,7 +89,7 @@
                         android:paddingRight="0dp"
                         android:paddingLeft="0dp"
                         android:imeOptions="actionSend|flagNoEnterAction|flagNoExtractUi"
-                        android:inputType="textAutoCorrect|textCapSentences"
+                        android:inputType="textAutoCorrect|textCapSentences|textMultiLine"
                         android:maxLength="2000"
                         android:maxLines="4"
                         android:nextFocusForward="@+id/send_button"


rhodey:
looks like it was removed here

rhodey:
looks like it was removed in this commit 2f20bb89f3b64701d047fa3af2aa8b6e0ecfdc34 by @mcginty. did you have any intention behind this @mcginty or just a refactor thing that happened?

https://github.com/volcano99/TextSecure/commit/2f20bb89f3b64701d047fa3af2aa8b6e0ecfdc34#diff-5aaa805a80069b6dba9ba28bf75c3817L91

moxie0:
@volcano99 What exactly is the behavior you're going for?  I tried this, and it doesn't seem to have any effect on my device.

Current master without "enter key sends": The edit text displays multiple lines as expected.
Current master with "enter key sends": The edit text only displays a single line.

Your pr without "enter key sends": same as above.
Your pr with "enter key sends": same as above.

volcano99:
@moxie0 that's weird. On my device, my pr without "enter key sends" displays multiple lines, but not more than 4 (or 5), as expected. This way you can respond to long messages while typing a long message yourself, because you can scroll up and down to see the messages you have received. Currently if i type a long message the text box expands indefinitely, to rhe point it covers the whole screen and i can't see the messages i received
For a better explanation see #2899

PR #3205, opened 2015-05-20
new attach icon

mcginty:
also fixed visual glitch when animating


agrajaghh:
how about replacing it here as well and remove the old one: 
https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/ConversationActivity.java#L537

Thats the only other occurence I found...

Issue #3094, opened 2015-05-01
OOM - Crash while sending picture


ghost:
I was sending a picture when textsecure crashed. The sending also failed by the way.
Log https://gist.github.com/anonymous/74de533cc8d30b66e7d4

moxie0:
This looks like a duplicate of #3092

peterpanondrugs:
still the same ... 
https://gist.githubusercontent.com/anonymous/6723cd79ded93c219999/raw/cat.log

mcginty:
closing since this is old and Glide helped resolve most of these

Issue #3092, opened 2015-04-30
OOM


moxie0:
`````
E/AndroidRuntime( 1086): java.lang.OutOfMemoryError: bitmap size exceeds VM budget
E/AndroidRuntime( 1086): 	at android.graphics.Bitmap.nativeCreate(Native Method)
E/AndroidRuntime( 1086): 	at android.graphics.Bitmap.createBitmap(Bitmap.java:477)
E/AndroidRuntime( 1086): 	at android.graphics.Bitmap.createBitmap(Bitmap.java:444)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.util.BitmapUtil.rotateBitmap(BitmapUtil.java:200)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.util.BitmapUtil.fixOrientation(BitmapUtil.java:191)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.util.BitmapUtil.createScaledBitmap(BitmapUtil.java:97)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.util.BitmapUtil.createScaledBitmap(BitmapUtil.java:86)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.util.BitmapUtil.createScaledBytes(BitmapUtil.java:45)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.mms.MediaConstraints.getResizedMedia(MediaConstraints.java:63)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.jobs.SendJob.getResizedPartData(SendJob.java:93)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.jobs.SendJob.getResolvedPart(SendJob.java:73)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.jobs.SendJob.getResolvedMessage(SendJob.java:51)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.jobs.PushMediaSendJob.deliver(PushMediaSendJob.java:107)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.jobs.PushMediaSendJob.onSend(PushMediaSendJob.java:66)
E/AndroidRuntime( 1086): 	at org.thoughtcrime.securesms.jobs.SendJob.onRun(SendJob.java:39)
[...]

Issue #3084, opened 2015-04-29
OOM: can't update group avatar


ghost:
When i try to update the avatar, after selecting the pic from the gallery, i am displayed the page with the info about the group (members and photo), but the photo is the old one. I can't even crop the picture, it just exits.
https://gist.github.com/d1cca4426eaff1fe536f

rhodey:
there are a bunch of out of memory exceptions in this log, which I expect are the root cause of this bug, but also a few other interesting exceptions too that @moxie0 or @mcginty probably understand better.

first OOM:
```
E/dalvikvm-heap(24127): Out of memory on a 19660816-byte allocation.
I/dalvikvm(24127): "main" prio=5 tid=1 RUNNABLE
I/dalvikvm(24127):   | group="main" sCount=0 dsCount=0 obj=0x40fb7508 self=0x40efd8a8
I/dalvikvm(24127):   | sysTid=24127 nice=0 sched=0/0 cgrp=[no-cpu-subsys] handle=1075855152
I/dalvikvm(24127):   | schedstat=( 773244781598 453929229835 1350017 ) utm=67648 stm=9676 core=1
I/dalvikvm(24127):   at android.graphics.BitmapFactory.nativeDecodeStream(Native Method)
I/dalvikvm(24127):   at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:621)
I/dalvikvm(24127):   at com.soundcloud.android.crop.CropImageActivity.c(CropImageActivity.java:134)
I/dalvikvm(24127):   at com.soundcloud.android.crop.CropImageActivity.onCreate(CropImageActivity.java:79)
I/dalvikvm(24127):   at android.app.Activity.performCreate(Activity.java:5047)
I/dalvikvm(24127):   at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1094)
I/dalvikvm(24127):   at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2056)
I/dalvikvm(24127):   at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2117)
I/dalvikvm(24127):   at android.app.ActivityThread.access$700(ActivityThread.java:134)
I/dalvikvm(24127):   at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1218)
I/dalvikvm(24127):   at android.os.Handler.dispatchMessage(Handler.java:99)
[...]

PR #3214, opened 2015-05-21
Memory leak fix for ContactSelectionListAdapter

herbig:
FREEBIE Relates to #3094 #3092 #3084

Because ContactSelectionListItem is used as a RecipientModifiedListener, it leaks its context (NewConversationActivity/PushContactSelectionActivity), which can lead to OOM crashes.

Here we keep a reference to the inflated views in order to unbind them whenever they become unused.  Because views are recycled while scrolling the adapter, the HashSet stays the same size dependant on how many rows are visible on screen, regardless of how many contacts you've scrolled through.

I used Square's new LeakCanary library to detect the leak and confirm the fix.


moxie0:
why not just remove the listener after the callback?

herbig:
Yeah, that would be much easier.  I assumed the contact could be modified more than once, but is the listener really only for downloading the image?

moxie0:
yeah, once a recipient object is resolved it's done.

herbig:
I see.  So setContactPhotoImage calls setImageDrawable (which could be null at that time) then sets the listener, in case the image is in the process of downloading.  Let me fix the pull request..

herbig:
Ok, that should be better.

mcginty:
This is brand new code so I don't think it's a root cause for #3094 #3092 #3084, but thanks so much and good catch!

herbig:
No problem.  I'd like to do a more thorough run through the app and patch up any leaks I find, if that works for you.  How do you guys normally work, I don't want to step on your toes if you're already looking into things.

mcginty:
We'd love that! Finding bugs is the opposite of stepping on toes. I periodically check for memory leaks if I notice a suspiciously-high memory when testing, but leakcanary looks really promising for more continuous checking (akin to StrictMode).

herbig:
Now that it's not 2am and I'm looking over this again, just that one line won't fix the issue.

Listeners are currently registered on every ContactSelectionListItem, whether they're downloading an image or not.  This means those that already have an image or don't have one at all never call onModified() and thus will never unlisten.

mcginty:
@herbig created a new issue this morning detailing just that :)

herbig:
Nice. My original commit from last night would fix the issue, but it may be overthinking it:

93809c3677746ab671b0a9d28a221545c77c9f05

herbig:
@mcginty Scratch that.  There are a number of leaks associated with RecipientListeners, not just here, and not just within Adapters, so I agree with #3224 in that they should only hold a weak reference.

PR #3222, opened 2015-05-21
migrate to AppCompatDialog

agrajaghh:
This migrates from `MaterialDialog` to `AppCompatDialog` to eventually eliminate this dependency.
The last other occurence of `MaterialDialog` is in `ReceiveKeyDialog`, I'll get red of it in another PR.

I couldn't test the `AlertDialog` in `DeviceProvisioningActivity` and in `PlayServicesProblemFragment`. @rhodey could you help me out there?

PR #3225, opened 2015-05-21
use ConfirmIdentityDialog instead of ReceiveKeyDialog

agrajaghh:
We currently have two different dialogs to confirm a changed Identity:
`ConfirmIdentityDialog` and `ReceiveKeyDialog`. This PR gets rid of the latter. `ConfirmIdentityDialog` seems to handle all message confilcts of a contact, while you have to confirm every single message with `ReceiveKeyDialog`.

This also switches to AppCompatDialog (see #3222)


src/org/thoughtcrime/securesms/ConfirmIdentityDialog.java
```diff
@@ -194,20 +194,17 @@ public void onClick(DialogInterface dialog, int which) {
...
+    private VerifySpan(Context context, IdentityKeyMismatch mismatch) {
+      this.context  = context;
+      this.mismatch = mismatch;
     }
 
     @Override
     public void onClick(View widget) {
       Intent intent = new Intent(context, VerifyIdentityActivity.class);
       intent.putExtra("recipient", mismatch.getRecipientId());
-      intent.putExtra("master_secret", masterSecret);
```
agrajaghh: just to be sure: there aren't any `putExtra("master_secret", masterSecret)` calls after the routing activity change supposed to be in the code if the activity extends `PassphraseRequiredActionBarActivity`, right?

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -369,10 +370,21 @@ private void setContactPhotoForRecipient(final Recipient recipient) {
...
+  private IdentityKeyMismatch getKeyMismatch(final MessageRecord record) {
+    if (record.isIdentityMismatchFailure()) {
+      for (final IdentityKeyMismatch mismatch : record.getIdentityKeyMismatches()) {
+        if (mismatch.getRecipientId() == record.getIndividualRecipient().getRecipientId()) {
+          return mismatch;
+        }
+      }
+    }
+    return null;
+  }
```
agrajaghh: I copied this method from https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/MessageRecipientListItem.java#L142. Where do you think its best to put it? Just making it `public static` there didn't seem like a good idea...

agrajaghh:
This fixes btw a crash which can happen if you confirm messages in the wrong order

src/org/thoughtcrime/securesms/ConfirmIdentityDialog.java
mcginty: that's correct

moxie0:
In 2.17.0, thanks!

PR #3227, opened 2015-05-21
enable LeakCanary in debug builds

mcginty:


herbig:
This is great.  LeakCanary has been pretty amazing in my experience with it thus far.

moxie0:
I'd like to keep the debug and release builds as identical as possible.  Maybe this should be in a different build target?

rhodey:

mcginty:
good idea, will update when i have a free moment

rhodey:
how's about the `debugTest` build target?

PR #3239, opened 2015-05-22
Fix for toggle behavior during rapid clear/fill cycles.

### LINKED ISSUE #3235: Attachment button instead of send arrow
It has happened sometimes (twice, so far) that even though I'm typing a message in the text box I don't get the arrow icon to send -- the attachment icon never goes away. When I click it, the behaviour is that of adding the attachment (so it's not just a problem of displaying the wrong icon, I am actually asked if I want to take a picture and stuff..).
Exiting and re-entering the conversation fixes it. It has happened both when talking to a Signal user and with an SMS one. It seems to be random, I'm trying to figure out how it works but I can't reproduce :/

moxie0:
Fixes #3235
// FREEBIE


mcginty:
lgtm

PR #3244, opened 2015-05-23
Wearable reply

### LINKED ISSUE #1699: Add Android Wear support

chr1shaefn3r:
Added wearable reply action to notifications, so now you can answer an incoming message right from your wrist :)

Tested for the past few days with my own Pebble and I also did check with an actual Android Wear Device from a colleague to test the voice reply.

Fixes #1699 
Follow up to #3010

Thanks to @Karalix for sharing his code in SMSSecure#135, gave me a great headstart.


src/org/thoughtcrime/securesms/notifications/NotificationState.java
```diff
@@ -63,4 +63,28 @@ public PendingIntent getMarkAsReadIntent(Context context, MasterSecret masterSec
...
+    int index          = 0;
+
+    for (long thread : threads) {
+      Log.w("NotificationState", "Added thread: " + thread);
+      threadArray[index++] = thread;
+    }
+
+    Intent intent = new Intent(WearReplyReceiver.REPLY_ACTION);
+    intent.putExtra("thread_ids", threadArray);
+    intent.putExtra("master_secret", masterSecret);
```
moxie0: We can't put this into intents that are leaving the app. We don't know where they're going or where they're stored, but more importantly, if someone locks TS there shouldn't be random copies of this floating around somewhere else.

src/org/thoughtcrime/securesms/notifications/NotificationState.java
```diff
@@ -63,4 +63,28 @@ public PendingIntent getMarkAsReadIntent(Context context, MasterSecret masterSec
 
     return PendingIntent.getBroadcast(context, 0, intent, PendingIntent.FLAG_UPDATE_CURRENT);
   }
+
+  public PendingIntent getReplyIntent(Context context, MasterSecret masterSecret, long recipientId) {
```
moxie0: this all feels a little copy and pastey

src/org/thoughtcrime/securesms/notifications/WearReplyReceiver.java
```diff
@@ -0,0 +1,85 @@
...
+      new AsyncTask<Void, Void, Void>() {
+        @Override
+        protected Void doInBackground(Void... params) {
+          for (long threadId : threadIds) {
+            Log.w(TAG, "Marking as read: " + threadId);
+            DatabaseFactory.getThreadDatabase(context).setRead(threadId);
+          }
+
+          OutgoingTextMessage reply = new OutgoingTextMessage(recipient, responseText.toString());
+          MessageSender.send(context, masterSecret, reply, -1, false);
```
moxie0: you've got the threadid right here...

src/org/thoughtcrime/securesms/notifications/WearReplyReceiver.java
```diff
@@ -0,0 +1,85 @@
...
+
+          MessageNotifier.updateNotification(context, masterSecret);
```
moxie0: What is this for? Looks like you canceled all TS notifications just before the async task.

rhodey:

src/org/thoughtcrime/securesms/notifications/NotificationState.java
chr1shaefn3r: I think I get your point. I'm wondering why the "getMarkAsReadIntent" is allowed to do so?

src/org/thoughtcrime/securesms/notifications/NotificationState.java
chr1shaefn3r: Refactored

src/org/thoughtcrime/securesms/notifications/WearReplyReceiver.java
chr1shaefn3r: Fixed.

src/org/thoughtcrime/securesms/notifications/WearReplyReceiver.java
chr1shaefn3r: I was wondering the same thing, then I looked at https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/notifications/MarkReadReceiver.java and copied it from there.

NickGeek:
Would love to see this added

moxie0:
I sent @chr1shaefn3r a PR with some changes.  Need him to approve them, since I don't have a wear device to test with (donations welcome).

chr1shaefn3r:
I've been running and testing it the whole week, worked flawlessly :+1: 
I merged your commit, rebased everything, should be good to go now.

res/values/strings.xml
```diff
@@ -951,6 +951,14 @@
     <string name="MuteDialog_mute_notifications">Mute notifications</string>
 
 
+    <!-- Wear Reply -->
+    <string name="MessageNotifier_reply">Reply</string>
+    <string-array name="MessageNotifier_wear_replies">
+        <item>Yes</item>
+        <item>No</item>
+        <item>OK</item>
+        <item>Thanks</item>
```
moxie0: What would happen if we put emoji unicode points in here?  Would they render on the watch?  I think it'd be cool to have :+1: and :-1:  instead of "yes and "no".

McLoo: If it worked, :+1: would also cover "ok", right?

NickGeek: In Android Wear the user can already draw emoji with the 5.1 update.

chr1shaefn3r: Android is not happy with putting emoji unicode points here (tested with "\ud83d\udc4d"):
06-22 20:20:13.359  22015-22824/? A/art﹕ art/runtime/check_jni.cc:65] JNI DETECTED ERROR IN APPLICATION: input is not valid Modified UTF-8: illegal start byte 0xf0
06-22 20:20:13.359  22015-22824/? A/art﹕ art/runtime/check_jni.cc:65]     string: '👍'
06-22 20:20:13.359  22015-22824/? A/art﹕ art/runtime/check_jni.cc:65]     in call to NewStringUTF
06-22 20:20:13.359  22015-22824/? A/art﹕ art/runtime/check_jni.cc:65]     from java.lang.String[] android.content.res.AssetManager.getArrayStringResource(int)

As @NickGeek already pointet out, when you are running latest android wear, you can answer with an emoji. Also Pebble has a feature to answer with an emoji, in contrast to answering with a prefilled plaintext.

So all, in all, we should stay with plaintext replies here.

moxie0: I think they'd probably be in the code directly, rather than in the XML here.  Never having seen an Android wear device, I'm not sure I understand the purpose of these strings if you can already quickly respond with arbitrary emoji or other values?

chr1shaefn3r: I didn't find the process of drawing an emoji to be particularly quickly. So I guess most users would opt for those pre-defined text replies if they want to answer really quickly or they go for the voice-recognition (if the environment allows it).

moxie0: Then maybe it'd still be cool if we could replace yes and no with thumbs up and thumbs down? Can you try specifying it in code rather than in xml?

NickGeek: Keep the text. Hangouts has all the text replies with a few common emoji at the bottom. Having emoji replace "Yes", and "No" could be frustrating when texting people with phones that don't support emoji.

Drawing emoji is good as well, IIRC it also has commonly used ones available in the drawing screen.

chr1shaefn3r: Specifiying emojis in code worked, they got rendered correctly on my Pebble Smartwatch.
Like @NickGeek I'm in doubt wether or not it would be the right thing to do.
On my Pebble Smartwatch it makes no sense at all. When I reply to a notification I get confronted with "Emojis" or "Use Templates" and now the templates also have Emojis ...
On Android Wear devices it might make a little more sense, as the option to answer with Emojis is one more swipe away. But I think Google will take great care of their User Experience and when users want to get more possibilities to answer with Emojis Google will add that to Android Wear.

All in all I think we should stay with the text.

moxie0: And none of these devices have a way for system-defined templates?  It seems a little weird that we're the ones choosing the strings a user can respond with.

chr1shaefn3r:
Darn! I found a bug: Answering a text from a group doesn't work properly. I will look into that ASAP.

res/values/strings.xml
chr1shaefn3r: Good point.
With my Pebble Smartwatch the official Pebble-App lets me save five template strings which I will get as an option whenever "Reply" is available. Additionally I see the options the app chooses to offer via "setChoices".
On Android Wear its a little different. There is no configuration-option in the official Android Wear app. But a quick test with Hangouts and WhatsApp-Notifications showed, that the template strings offered to reply are nearly identical. Which makes me assume that Android Wear itself has some pre-configured template strings. (Thanks to my colleague @adracus who actually has a Android Wear Smartwatch and helps with testing)

I think we would be fine with getting rid of the "MessageNotifier_wear_replies" all together.

moxie0:
ok then i'm just waiting to merge this on the group reply issue and whatever you want to do with the template strings.

res/values/strings.xml
NickGeek: I see the same reply template for every app on my Sony Smartwatch 3 (5.1.1).

I wouldn't try to mess with the pre-configured Android Wear settings as they're good.

chr1shaefn3r:
I rebased on master and removed the template strings.
The new NotificationItem.getReplyIntent() is conflicting with NotificationState.getReplyIntent(), the first is pointing to ConversationPopupActivity.class the later to WearReplyReceiver.REPLY_ACTION.
I switched back to NotificationState.getReplyIntent during the rebase, master uses NotificationItem.getReplyIntent (thats why the check is failling, I guess).
@moxie0 Could you share some information about the ConversationPopupActivity, so we can resolve the conflicting methods?

moxie0:
@chr1shaefn3r The quick reply intent in master now is for quick reply on-device. You need a different intent that's just for reply from Wear devices.  Might be easier to leave the PR based off its original branch and let me do the rebase.

What was up with the group reply issue?

chr1shaefn3r:
I did a debugging-session yesterday and failed to pin down the problem.
I know for sure, that with the initial implementation the group reply worked. To me, it looks like the internals of MessageSender.send changed. But q quick check with git blame didn't bring up any recent changes which looked suspicious. I will try to take another look on it tomorrow.

chr1shaefn3r:
I'm puzzled, with the initial implementation (new OutgoingTextMessage and MessageSender.send) it worked just fine (replying to both group and "normal" messages).
Now I can't see any scenario in which MessageSender.send could have ever send a group message (with a OutgoingTextMessage)!?!?
Anyway, I fixed the issue with a similiar implementation as seen in QuickResponseService.

For the sake of overall architecture it might be a good idea to refactor WearReplyReceiver and QuickResponseService (and possible other classes I'm not aware of yet) and centralize the distinction between Group and "normal" message in MessageSender.send.

chr1shaefn3r:
Streamlined the commits, rebased on master and tested reply with TextSecure-, Non-TextSecure aka. SMS- and Group-Messages. All right now :)

moxie0:
@chr1shaefn3r I can't follow what's going on if you modify the history.  I need the unrebased original series of commits, along with the commits for changes you've made since.

chr1shaefn3r:
@moxie0 Sorry for messing with the history, will dig into my dangling-commits and restore the unrebased original commits.

NickGeek:
Awesome! Can't wait for this to get merged!

moxie0:
in 2.22.0, thanks!

Webstas:
I would say no since it only added notification replies and not a native app

bogorad:
What about buzzing when for incoming calls and option to answer?

PR #3262, opened 2015-05-27
make search hint translatable

### LINKED ISSUE #3261: Untranslated string "Search..."
When tapping on the magnifying glass symbol left of the vertical three dots (menu), the original, i. e. untranslated string "Search..." is displayed. Also, there's no entry for that string in the Transifex database.

agrajaghh:
Fixes #3261


moxie0:
in 2.17.0!

PR #3273, opened 2015-05-28
espresso tests & jenkins CI

rhodey:
1) comment out broken PartDatabase tests until fixed
2) move androidTest/ into new test/ directory
3) move PUSH_URL from Release.java into BuildConfig.java
4) use textsecure-service-staging.whispersystems.org for test build
5) move magic SMS mocking code into own util class
6) created a new build flavor for espresso tests
7) create a new source set full of espresso tests
8) updated proguard-testing.pro
9) added test device numbers to .gitignore
10) make running the espresso tests optional via -Pespresso=1 command line arg

I did my best to split this up into small, logical commits. [Jenkins](https://jenkins-ci.org/) + [Spoon](https://github.com/square/spoon) work happily with this setup, using these I've got a build server running espresso tests on all connected android devices for every new push to my branch and publishing test reports to HTML pages.

the espresso build flavor is included conditionally by the project property `espresso` being set to anything. this property can be set from the command line or in a `settings.gradle` file, so by default these new tests will not run on the `connectedCheck` build target.


rhodey:
forgive my build bot for messing with the status of this PR :(

rhodey:
updated so that user now needs to provide `-Pespresso=1` command line arg (or set `espresso=1` in `gradle.properties`) **and** press enter after being warned that our espresso tests will delete all their contacts.

moxie0:
you should have to type something

rhodey:
@moxie0 done. if the user types anything other than `espresso` at the prompt the regular tests will be used. I would prefer a clean exit here instead but all the [recommended ways to stop a build with Gradle](http://stackoverflow.com/questions/10312259/recommended-way-to-stop-a-build-with-gradle) result in 300+ lines of stack traces **and then** a stopped build.

moxie0:
how about `delete all my contacts` instead

src/org/thoughtcrime/securesms/database/PartDatabase.java
```diff
@@ -640,5 +640,15 @@ public String toString() {
...
+
+    @Override
+    public boolean equals(Object object) {
+      if (!(object instanceof PartId)){
+        return false;
+      }
+
+      PartId other = (PartId) object;
+      return rowId == other.rowId && uniqueId == other.uniqueId;
+    }
```
moxie0: please do whatever you need to do without adding this method

rhodey:
@moxie0 done

PR #3283, opened 2015-05-30
ASCII emoji support

### LINKED ISSUE #2943: Where are the old text-only smileys?
Why did you remove the old text based smileys?
I don't like all this emoji crap and preferred the other ones.

mcginty:
Fixes #2943

Simplifies the code a bit since we're dealing with all emoji as Strings now.

* *Why are the emoji back to being stored as Java statics you ask?* Well, Android's resource XML does not support all UTF-16 codepoints so there was no choice if we wanted to represent the emoji as strings.
* I made `EmojiView` since TextView's onDraw is too heavy to work efficiently in older phones. Horrible lag. EmojiView is a nano-ImageView/TextView hybrid that has a very fast draw cycle, making it great for having dozens of them on screen at a time. Seems subjectively faster than even the ImageView we were using before.


moxie0:
in 2.17.0

PR #3292, opened 2015-05-31
fix emoji not being replaced with Editable actions

### LINKED ISSUE #3267: TS emoji and system emoji sizes are different when typing
0. Go to some conversation and tap the input box
1. Type a `BANKNOTE WITH DOLLAR SIGN` (U+1F4B5) with TS emoji drawer
2. Type the same emoji with Lollipop system emoji keyboard
3. Notice that the system dollar is bigger

mcginty dollars

This gets fixed if you the leave conversation (draft gets saved) and then return back to the conversation. Also pasting stuff seems to affect it.
This is only an issue with the input box. Sent and received messages show emojis at one size.

Related: #2937 

TS 2.16.2, Nexus 4, stock Android 5.1.1

mcginty:
also cleaned up the callback stuff to be a bit more object-oriented :).

Fixes #3267


src/org/thoughtcrime/securesms/components/emoji/EmojiEditText.java
```diff
@@ -1,40 +1,48 @@
...
-                                                                               EmojiProvider.EMOJI_SMALL,
-                                                                               callback);
 
-    getText().replace(Math.min(start, end), Math.max(start, end), text);
+    getText().replace(Math.min(start, end), Math.max(start, end), new String(chars));
     setSelection(end + chars.length);
   }
+
+  @Override public void invalidateDrawable(@NonNull Drawable drawable) {
+    if (drawable instanceof EmojiDrawable) invalidate();
```
moxie0: what is this about?

src/org/thoughtcrime/securesms/components/emoji/EmojiTextView.java
```diff
@@ -1,29 +1,37 @@
...
   }
 
   public EmojiTextView(Context context, AttributeSet attrs, int defStyleAttr) {
     super(context, attrs, defStyleAttr);
+    init();
+  }
+
+  public void init() {
+    setFilters(new InputFilter[]{ new EmojiFilter(this) });
   }
```
moxie0: you can use this() in your constructors so you can put your constructor logic into one of them

src/org/thoughtcrime/securesms/components/emoji/EmojiFilter.java
```diff
@@ -0,0 +1,32 @@
...
+    this.view = view;
+  }
+
+  @Override public CharSequence filter(CharSequence source, int start, int end,
+                                       Spanned dest, int dstart, int dend)
+  {
+    char[] v = new char[end - start];
+    TextUtils.getChars(source, start, end, v, 0);
+    Spannable emojified = EmojiProvider.getInstance(view.getContext()).emojify(new String(v), view);
+    if (source instanceof Spanned) {
```
moxie0: what is this about?

src/org/thoughtcrime/securesms/components/emoji/EmojiEditText.java
mcginty: `invalidateDrawable` is called from any drawable related to a view via this method, and we only want to do a full invalidation in the case of an `EmojiDrawable` requesting invalidation. `super.invalidateDrawable` would ignore the EmojiDrawable since it doesn't know it's attached to the view (all Android views implement some form of a `verifyDrawable` method and ignore drawables they don't know), so we need to override it and invalidate ourselves appropriately.

src/org/thoughtcrime/securesms/components/emoji/EmojiFilter.java
mcginty: `source` can be unspannable, but if it does contain spans then it's best to copy the existing spans over so the filter doesn't clear out any other styles. This is done the exact same way in Android's InputFilter (http://grepcode.com/file/repository.grepcode.com/java/ext/com.google.android/android/4.0.1_r1/android/text/InputFilter.java#58).

mcginty:
updated to not use an `init()` method

moxie0:
please rebase

mcginty:
done

PR #3299, opened 2015-06-01
Convert AsyncTask to IntentService

yulin2:
Hi `TextSecure` developers, I'm doing research on Android async programming, particularly on 
AsyncTask and IntentService. AsyncTask can lead to memory leak and losing task result when GUI is recreated during task running (such as orientation change). [This article](http://bon-app-etit.blogspot.com/2013/04/the-dark-side-of-asynctask.html) describes the problems very well. 

We discussed with some Android experts and they agree with this issue, and claim that AsyncTask can be considered only for short tasks (less than 1 second). However, using IntentService (or AsyncTaskLoader) can avoid such problems since their lifecycles are independent from `Activity`/`Fragment`.

In `TextSecure`, the AsyncTasks are declared as non-static inner classes of `Activity`/`Fragment`. So the above issue can occur (especially for relatively long tasks).

I refactored 9 AsyncTasks in `TextSecure` to IntentService (in 8 separate commits), with the help of a refactoring tool we developed. Do you think using IntentService should be better in some of these 9  scenario? Do you want to merge some commits in this pr? Thanks.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -247,6 +247,8 @@
     </service>
 
     <service android:name=".ExportFragment$ExportPlaintextService" android:exported="false"/>
+    <service android:name=".ImportFragment$ImportPlaintextBackupService" android:exported="false"/>
+    <service android:name=".ImportFragment$ImportEncryptedBackupService" android:exported="false"/>
 
 
     <receiver android:name=".gcm.GcmBroadcastReceiver" android:permission="com.google.android.c2dm.permission.SEND" >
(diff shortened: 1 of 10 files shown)


moxie0:
I think we'd probably like to avoid using IntentService.  Most of these cases would probably be better addressed just by making the AsyncTask inner class static, and using an event bus to communicate the result back to the activity.

yulin2:
Thanks for the reply! Are there any specific reasons that you want to avoid using IntentService?

On the other hand, since many AsyncTasks access fields of Activity, I don't think you can make them into static inner classes. Otherwise they cannot be compiled, right?

moxie0:
IntentService is too cumbersome and heavy handed.  If your inner class isn't static (whether an asynctask or a loader), it will retain a reference to the parent activity, which I assume is what you're trying to avoid.  If the task references an instance variable of the enclosing class, that needs to be passed as an argument instead or handled by the callback.

PR #2886, opened 2015-04-03
fix media message forwarding

rhodey:
1) forward media messages using MediaMmsMessageRecord's resolved SlideDeck
2) patch ShareActivity to get type of STREAM_EXTRA from intent when others null
3) patched AudioSlide, VideoSlide, and ImageSlide to work with content:// Uri's from our part database

Fixes #1362
// FREEBIE

PR #2236, opened 2014-12-18
Enable media message forwarding

McLoo:
For images this works without handing the content type through the intents, as it is set to `ContentType.IMAGE_JPEG` hardcoded.
This also unifies AudioSlide's and VideoSlide's `constructPartFromUri` as they were edited anyway.

Fixes #1362 and #1914 

I strongly recommend merging #2234 and #2232 as without them forwarding message is not funny...

PR #3311, opened 2015-06-02
Enable media message forwarding

McLoo:
This is a stew mainly of #2886 with some nice chunks of #2236 

- Tested forwarding to existing as well as new (group-)convos.
- Tested on a 5.0.2 device and KK emulator. Couldn't get a play service enabled GB emu - can't use a new TS installation without play services... :disappointed: 

All media works fine as well as text and text + media.

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/AudioSlide.java
+++ b/src/org/thoughtcrime/securesms/mms/AudioSlide.java
@@ -18,7 +18,6 @@ package org.thoughtcrime.securesms.mms;
 
 import android.content.Context;
 import android.content.res.Resources.Theme;
-import android.database.Cursor;
 import android.net.Uri;
 import android.provider.MediaStore.Audio;
 import android.support.annotation.DrawableRes;
@@ -56,24 +55,15 @@ public class AudioSlide extends Slide {
     return ResUtil.getDrawableRes(theme, R.attr.conversation_icon_attach_audio);
   }
 
-  public static PduPart constructPartFromUri(Context context, Uri uri) throws IOException, MediaTooLargeException {
+  private static PduPart constructPartFromUri(Context context, Uri uri) throws IOException, MediaTooLargeException {
     PduPart part = new PduPart();
 
-    assertMediaSize(context, uri, MmsMediaConstraints.MAX_MESSAGE_SIZE);
-
-    Cursor cursor = null;
-
-    try {
-      cursor = context.getContentResolver().query(uri, new String[]{Audio.Media.MIME_TYPE}, null, null, null);
-
-      if (cursor != null && cursor.moveToFirst())
-        part.setContentType(cursor.getString(0).getBytes());
-      else
-        throw new IOException("Unable to query content type.");
-    } finally {
-      if (cursor != null)
-        cursor.close();
-    } 
+    if (PartAuthority.isPartAuthorityPart(uri)) {
+      part.setContentType(getContentTypeForPartInDb(context, uri));
+    } else {
+      assertMediaSize(context, uri, MmsMediaConstraints.MAX_MESSAGE_SIZE);
+      part.setContentType(getContentTypeForPartOutsideDb(context, uri, Audio.Media.MIME_TYPE));
+    }
 
     part.setDataUri(uri);
     part.setContentId((System.currentTimeMillis()+"").getBytes());
--- a/src/org/thoughtcrime/securesms/mms/PartAuthority.java
+++ b/src/org/thoughtcrime/securesms/mms/PartAuthority.java
@@ -16,8 +16,8 @@ import java.io.InputStream;
 
 public class PartAuthority {
 
-  private static final String PART_URI_STRING   = "content://org.thoughtcrime.securesms/part";
-  private static final String THUMB_URI_STRING  = "content://org.thoughtcrime.securesms/thumb";
+  private static final String PART_URI_STRING   = "content://org.thoughtcrime.provider.securesms/part";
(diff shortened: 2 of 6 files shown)


jeremymasters:
Does this also fix selection of media messages?  In 2.16.3 I can't long-press the image to select it.  I have to long-press in the open space beside it.

2-4601:
@jeremymasters See: #2333, #2488, #2648 (PR).

rhodey:
hey @McLoo, I just checked out this PR, rebased it with master, and ran some tests. every time i try to forward an image message the image preview is the red error icon, tapping send results in no image being sent.

can you rebase this and let me know if you can reproduce? I tested this on my Nexus 5 running stock Android 5.1 Lollipop, rebased into f72cd5b0fffe2465b40213828a9214c8c2517715

McLoo:
@rhodey then you collabs broke it :wink: 

Will check out. The magic is in those lines:
```JAVA
if (isAuthorityPartDb(uri))
  part.setDataUri(PartAuthority.getPublicPartUri(uri));
else
  part.setDataUri(uri);
``` 

Had the triangle before, when not using getPublicPartUri?!

McLoo:
@rhodey tested on CM 12.1
Still works for me. Can you post a debug log? :grin:

rhodey:
hey @McLoo, I just rebased this with 2.18.1 and tested on my stock 2.3.6 Gingerbread device & my stock 4.4.2 device, both work great. my nexus 5 is my daily use phone, I keep my actual install of TextSecure on there, so when I want to test a new build of TS on that device I have to alter the application ID and some other package names in the build.grade and AndroidManifest.xml file-- basically the app that I test on this device is not the exact app that I should be testing.

so, in conclusion it probably doesn't work on my nexus 5 for reasons related to me changing the application ID and package names. interestingly nothing was actually caught in the log file related to the error icon showing up. this PR looks good to me and has tested well, if we could get someone to test this on a stock lollipop device I'd be super confident in merging it.

:+1:

McLoo:
I'm thinking of `assertMediaSize` everytime - not only if not `isAuthorityPartDb`.
There could be changes for media size, after the media went into the DB?!

(`assertMediaSize` should probably move into [sendMediaMessage](https://github.com/WhisperSystems/TextSecure/blob/6caee6b0f076a1dd2e710212435cf950352817db/src/org/thoughtcrime/securesms/ConversationActivity.java#L1108), once bigger files are allowed for push)

McLoo:
Finally I found someone who was willing to try this PR on his daily use phone. :grimacing: 
Titanium Backup works great if you have to juggle with builds while keeping data.

So: on my Sony Z1 compact 5.0.2 (rooted) this also works fine. Couldn't test media playback though cause of #3436

src/org/thoughtcrime/securesms/mms/PartAuthority.java
```diff
@@ -17,7 +17,7 @@
 
   private static final String PART_URI_STRING   = "content://org.thoughtcrime.securesms/part";
   private static final String THUMB_URI_STRING  = "content://org.thoughtcrime.securesms/thumb";
-  private static final Uri    PART_CONTENT_URI  = Uri.parse(PART_URI_STRING);
+  public  static final Uri    PART_CONTENT_URI  = Uri.parse(PART_URI_STRING);
```
moxie0: this is intentionally private

src/org/thoughtcrime/securesms/mms/AudioSlide.java
```diff
@@ -56,24 +55,15 @@ public boolean hasAudio() {
...
-    } finally {
-      if (cursor != null)
-        cursor.close();
-    } 
+    if (isAuthorityPartDb(uri)) {
+      part.setContentType(getContentTypeForPartInDb(context, uri));
+    } else {
+      assertMediaSize(context, uri);
+      part.setContentType(getContentTypeForPartOutsideDb(context, uri, Audio.Media.MIME_TYPE));
+    }
```
moxie0: we shouldn't be distinguishing, a uri is a uri...

McLoo: But PartProvider has 
`CONTENT_URI_STRING = "content://org.thoughtcrime.provider.securesms/part";`
whereas PartAuthority has 
`PART_URI_STRING   = "content://org.thoughtcrime.securesms/part";` 

the latter leads to an error on forwarding
```
org.thoughtcrime.securesms E/ActivityThread﹕ Failed to find provider info for org.thoughtcrime.securesms
```
as there is no such provider in the manifest.

So at some point we have to decide how the uri has to be interpreted. Or am I completely wrong?!
Or does PartAuthority need a change?

McLoo: changing PartAuthority's PART_URI_STRING to "content://org.thoughtcrime.provider.securesms/part";works fine form me

McLoo:
@moxie0 comments addressed

src/org/thoughtcrime/securesms/mms/AudioSlide.java
McLoo: Sorry for spamming. The above comments are concerning Images.

And yet we have to distinguish if we don't want to on fail grabbing parts' contentType

McLoo:
rebased

McLoo:
rebased. reduced refactoring.

McLoo:
:angry: temporarily closing til I'm happy with GIF-detection

getting the MimeType isn't very funny for internal media :/

PR #3320, opened 2015-06-02
fix unit strings

### LINKED ISSUE #3295: units
On the textSecure panel showing the list of threads, each of them has a field showing the time elapsed since the last message. 

typically I read something like "11 mins".

I wonder what the s after min stands for. Unit for minute is min, and units do not take plural form.

Unfortunately I must say that this trend of putting plural mark to units is more and more common in the US, and it appears that it is getting more common in other countries as well. Spreading that nonsense is definitely not good so I suggest that the "plural unit issue" be taken care of as soon as possible.

Check also for the hour, I am pretty sure that soon some "hrs" will show up. "h" would actually be the best, while "hr" being also accepted. 

Thank you

agrajaghh:
Fixes #3295

//FREEBIE

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -158,6 +158,7 @@
 
     <!-- DateUtils -->
     <string name="DateUtils_now">Now</string>
+    <string name="DateUtils_minutes_ago">%d min</string>
 
     <!-- ShareActivity -->
     <string name="ShareActivity_share_with">Share with</string>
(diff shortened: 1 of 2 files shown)


mcginty:
yay net loss! :stuck_out_tongue: code lgtm

res/values/strings.xml
```diff
@@ -694,17 +696,6 @@
...
-    <!-- plurals.xml -->
-    <plurals name="minutes_ago">
-        <item quantity="one">%d min</item>
-        <item quantity="other">%d mins</item>
-    </plurals>
-
-    <plurals name="hours_ago">
-        <item quantity="one">%d hour</item>
-        <item quantity="other">%d hours</item>
-    </plurals>
```
moxie0: I think we should leave hour/hours.  I'm fine with getting rid of mins though.

agrajaghh:
@moxie0 restored the old hour/hours behaviour

PR #3321, opened 2015-06-02
notification sound + vibrate while in call

### LINKED ISSUE #1830: please add option to hear message tone while on a call
hi, please add a option to hear message tone while on a call.  I can not get the vibrate to work and I'm blind so I can't see the notification in the status bar this would be useful.  Your app is extremely accessible to the blind and with Google's screen reader talk back.  I would be willing to donate if you add this option thank you.

rhodey:
play notification sound and vibrate (if enabled) when new message received during a phone call.

Fixes #1830
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -33,7 +33,6 @@ import android.media.MediaPlayer;
 import android.media.Ringtone;
 import android.media.RingtoneManager;
 import android.net.Uri;
-import android.os.Vibrator;
 import android.support.v4.app.NotificationCompat;
 import android.support.v4.app.NotificationCompat.Action;
 import android.support.v4.app.NotificationCompat.BigTextStyle;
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -420,11 +424,28 @@ private static void setNotificationAlarms(Context context,
...
+      AudioManager audioManager = (AudioManager) context.getSystemService(Context.AUDIO_SERVICE);
+      if (audioManager.getRingerMode() == AudioManager.RINGER_MODE_NORMAL && ringtoneUri != null) {
+        Ringtone tone = RingtoneManager.getRingtone(context, Uri.parse(ringtone));
+        tone.setStreamType(AudioManager.STREAM_VOICE_CALL);
+        tone.play();
+      }
+
+      if (signal && vibrate) {
+        Vibrator vibrator = (Vibrator) context.getSystemService(Context.VIBRATOR_SERVICE);
+        vibrator.vibrate(new long[] {0, 250, 250, 250}, -1);
```
rhodey: default found in `NotificationManagerService` source, class is `@hide`n

mcginty: Do we not just want to use the vibrate pattern chosen in settings? Also, is vibrating while a phone is against your face really a feature people want?

wp9015362:
> @mcginty wrote:
> 
> Also, is vibrating while a phone is against your face really a feature people want?

No, that would suck.

It could however be dealt with via an option if it weren't for this restrictive thing:

https://github.com/WhisperSystems/TextSecure/blob/master/contributing.md#development-ideology

...

mcginty:
That "restrictive thing" is there *exactly* so we will never have a face vibrator option.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
rhodey: good point, I'll take out the vibrate. settings only allow led blink pattern ftr.

rhodey: good point, I'll take out the vibrate. settings only allow led blink pattern ftr.

PR #3332, opened 2015-06-03
fix messages with push flag *and* sms flag set...

mcginty:
The insecure message approval logic sets the message as forced SMS but doesn't unset the push flag, so all our view logic was displaying insecure-fallback SMS messages as push without a lock. 

origin: https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/ConversationItem.java#L511


moxie0:
maybe we should unset push at fallback time instead?

PR #3336, opened 2015-06-04
store direct captures internally

mcginty:
Anecdotal evidence suggests that direct capture wasn't working using the external storage directory for some people (would just return to the ConversationActivity with nothing attached).

This seems nicer as it limits the visibility of the captured photo at least a bit better since it's not in the /sdcard/ location.


mcginty:
closing in favor of direct capture pr

PR #3342, opened 2015-06-04
use the Ringtone class instead of interacting with MediaPlayer directly

### LINKED ISSUE #3340: MessageNotifier never calls MediaPlayer.release()
`MessageNotifier` uses `MediaPlayer` for [playing the notification sound for in thread notifications](https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java#L316) but never calls `MediaPlayer.release()` as strongly suggested in the [MediaPlayer documentation](http://developer.android.com/reference/android/media/MediaPlayer.html)

the code should either be updated to call `release()` in a final block or, my preference, stop interacting with the `MediaPlayer` directly and let the `Ringtone` class handle this like so:

```java
Ringtone tone = RingtoneManager.getRingtone(context, uri);
tone.setStreamType(AudioManager.STREAM_NOTIFICATION);
tone.play();
```

`Ringtone` manages a `MediaPlayer` instance internally and takes care of `release()` and all that stuff.

rhodey:
use the Ringtone class instead of interacting with MediaPlayer directly

Fixes #3340
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -29,7 +29,8 @@ import android.graphics.BitmapFactory;
 import android.graphics.Color;
 import android.graphics.drawable.Drawable;
 import android.media.AudioManager;
-import android.media.MediaPlayer;
+import android.media.Ringtone;
+import android.media.RingtoneManager;
 import android.net.Uri;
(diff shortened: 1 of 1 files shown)


rhodey:
this has the added benefit of respecting the volume level set by the user instead of using 25% erretime.

moxie0:
in 3.8.0

PR #3346, opened 2015-06-05
fix checkbox overlap with fastscroll

mcginty:
Styling the fastscroll and changing its size is nontrivial, so I'm going to wait until we convert to a RecyclerView to make it fancier. In the mean time, updated the sticky list headers library to fix a memory leak.


moxie0:
In 2.18.0

PR #3350, opened 2015-06-05
decode bitmaps as RGB565 when low memory

mcginty:
cuts decoding memory in half


agrajaghh:
how about making this conditional on the memclass? On a quick test my photos looked noticeable darker with 16bit.

mcginty:
@agrajaghh great point and thanks for the good sample image. definitely noticeable.

mcginty:
It'd be cool if they supported an RGB_888 when you don't care about alpha.

mcginty:
fixed

moxie0:
in 2.19.0

Issue #2602, opened 2015-03-06
[Feature] Convert known phone numbers in text body into contact names


patcon:
Splitting this out from @brwolfgang's #2582 request for clickable phone numbers.

> It also would be nice if TS would identify the number in text body using my local contacts list and instead of showing the unrecognizable sequence of numbers it showed a (clickable) contact name.
>
> A SMS example: Você recebeu ligações de: 04111123451234 16:59hs
> A processed SMS example: Você recebeu ligações de: : Dad 16:59hs

~~Blocked on #2582.~~

silentmoose:
The function would be simple, simply check the phone number against the stored numbers, however the question is where the best implementation would be. Also the color should be changed on the name to make it known the name is not normal text.

kmindi:
Wait for answers of my just created [question for this on Stack Overflow](https://stackoverflow.com/questions/28960987/how-to-convert-a-phone-number-to-that-contacts-name-in-textview-on-android)

silentmoose:
@kmindi you have to pull it from  a list view. Lot's of tutorials online on how to pull phone numbers from contact list in android. Text secure already has the permission to read them.

KonPap18:
i am interested in fixing this issue with a classmate of mine, for a university project.

moxie0:
I'm down with supporting vcard or something so that contact info can be transmitted and rendered appropriately, but I'm not really down with modifying the content of strings within an otherwise unencoded message.

PR #3352, opened 2015-06-05
contact name is displayed aside known number, in text body(inspired by issue #2602)

KonPap18:
Instead of replacing phone number with the contact name, as requested in issue #2602 we put in aside, inside parenthesis
  Ref #2602
image


agrajaghh:
It seems like you changed a lot of whitespaces unintentionally

moxie0:
Thanks but:

1. There are hella formatting changes unrelated to your change.
2. This does blocking I/O on the UI thread.
3. This is a whole lot of non-model logic in a model.

PR #3353, opened 2015-06-05
add dark theme to emojiDrawer

### LINKED ISSUE #3240: Make the emoji panel back color theme dependent
In other words:
Turn off f the light for the li'l fellas' home, when it's dark outside
 🌞🌚

agrajaghh:
Fixes #3240

screenshot_2015-06-06-01-23-53


2-4601:
Glorious!

src/org/thoughtcrime/securesms/components/emoji/EmojiPageModel.java
```diff
@@ -1,7 +1,8 @@
 package org.thoughtcrime.securesms.components.emoji;
 
 public interface EmojiPageModel {
-  int getIconRes();
+  int getLightIconRes();
```
mcginty: instead of `get<Theme>IconRes()` as independent methods, it might make more sense to do `getIconAttr()` and have the caller resolve it with their context.

agrajaghh: you're right, thats much better! changed it...

res/values/themes.xml
```diff
@@ -203,6 +218,21 @@
         <item name="conversation_attach_photo">@drawable/ic_photo_camera_dark</item>
         <item name="conversation_attach">@drawable/ic_attach_white_24dp</item>
 
+        <item name="emoji_tab_strip_background">@color/gray95</item>
+        <item name="emoji_tab_indicator">#ff595959</item>
```
mcginty: total nitpick since I began the fad of lazily not creating color variables, but since we have so many shades of gray now (well below the amount to make a bad book joke, thankfully) we might label these as @color/grayNN to at least partially standardize these.

agrajaghh:
used `@color/grayNN` now (for the non transparent values at least). For the NN I used (100 - brightness). Is that what you did? Because `<color name="gray5">#ffeeeeee</color>` is (0; 0; 93%) in the HSB model, so it should be `gray7`, no? Or how did you come up with 5?

mcginty:
@agrajaghh I think I just guessed it was 5ish percent. In reality perhaps we should, excluding our brand's `2090EA`-blue, stick with the material guidelines and mark them as "weights" out of 1000? Referring to http://www.google.com/design/spec/style/color.html#color-color-palette, EEEEEE would be `@color/gray200`.

mcginty:
It baffles me but I couldn't find an android color XML version of the material palette, so I made one here if you want to use it: https://gist.github.com/mcginty/95c5ad31a2b2d412ceb2

agrajaghh:
@mcginty I'm a bit confused. How do they get the 200? How can I get the value for e.g. #505050?

jeremymasters:
The only thing I could think of is in the case of #505050 you'd be looking at hex 50 / hex FF which I get .3137 :) which I guess would be 314?

agrajaghh:
@jeremymasters `gray200` is `#eeeeee` so ee=238, and 238/255=0,933. I'm still confused..

jeremymasters:
@agrajaghh wow.  Must be some new math.

PR #3357, opened 2015-06-06
Groupconversation displaying receivers improvement

javacTeam:
This is still in progress and we would like to see what you think about it. We are aware that it may create some bugs so we would love some feedback if posible into what changes to implement.

Thank you in advance,
javacTeam

### DIFF
--- a/res/layout/message_details_activity.xml
+++ b/res/layout/message_details_activity.xml
@@ -11,4 +11,11 @@
               android:layout_height="wrap_content"
               tools:listitem="@layout/message_recipient_list_item" />
 
+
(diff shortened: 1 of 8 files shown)


moxie0:
Please only submit finished PRs here, we can't review works in progress.  A good start would be reading our code style guidelines.

brumsoel:
@javacTeam The guidelines can be found here:
https://github.com/WhisperSystems/TextSecure/blob/master/contributing.md
https://github.com/WhisperSystems/RedPhone/wiki/Code-style-Guidelines

PR #3364, opened 2015-06-07
Couple of cosmetic corrections for strings.xml

RiseT:
I've made a couple of very minor cosmetic corrections (mostly punctuation) of the original strings which I noticed when translating at Transifex.
Each one of them is about consistency and conformity with similar strings.

While I'm pretty sure about the correctness of the changes, I'm an absolute novice to GitHub - sorry in advance if I violate some formal conventions for commiting at GitHub or do something stupid in the process.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -20,8 +20,8 @@
         able to access them.
     </string>
     <string name="ApplicationPreferencesActivity_disable">Disable</string>
-    <string name="ApplicationPreferencesActivity_unregistering">Unregistering...</string>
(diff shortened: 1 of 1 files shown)


mcginty:
Thanks @RiseT, not having ellipses in the title does seem nicer. All these look good to me.

RiseT:
@mcginty I've changed them as pretty much all dialogs of that kind don't have ellipses in the title but they do in the description text. Those are aberrations.

moxie0:
In 2.18.0

Issue #3324, opened 2015-06-03
Display app version in settings dialogue


Norbert80:
The version number is not displayed within the TextSecure app. I suggest to display it at the bottom of the settings dialogue.

Please let me know if I have missed something.

2-4601:
I usually check the version number from _Settings --> Advanced --> Submit Debug Log_

Norbert80:
Thanks. But I still prefer my suggestion. :-)

claenn:
+1

Mihahn:
Duplicate of #1356 #1501 #2053  #2742
But I'd love to see that feature, too!

RiseT:
The fork SMSSecure has already completed this. One possibility is just pulling their code.

PR #3372, opened 2015-06-08
Add footer to preferences with version info

nthomsn:
Ref #3324

Looks like this:
device-2015-06-08-080916


mcginty:
Alternate proposition: we include the version number as a subtitle for "Submit Debug Log" so it's farther away, less intrusive, and seen as a "special" piece of information that's unnecessary for normal use.

nthomsn:
I agree, it fits better there. It's harder to spot the version in the actual debug logs. I'm still orienting myself to the code but it shouldn't be hard to move around.

mcginty:
great, thanks for contributing!

PR #3375, opened 2015-06-09
google photos share URIs, embrace the madness

### LINKED ISSUE #3307: Message disappears on send when sharing photo
I first noticed this after the recent update to the Photos app. When sharing from within that app, a toast notification says "Sharing to TextSecure..." and the message composition process continues as expected. When sent, the message disappears.

This bug does not happen when the sharing process is started from within TextSecure through either the paperclip icon or the "Add attachment" menu option.

Nexus 5, Android 5.1.1 (LMY48B), T-Mobile US

rhodey:
transform google photos nonsense URI scheme to something usable when necessary

Fixes #3307
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ShareActivity.java
+++ b/src/org/thoughtcrime/securesms/ShareActivity.java
@@ -31,6 +31,8 @@ import org.thoughtcrime.securesms.recipients.Recipients;
 import org.thoughtcrime.securesms.util.DynamicLanguage;
 import org.thoughtcrime.securesms.util.DynamicTheme;
 
+import java.net.URLDecoder;
+
 import ws.com.google.android.mms.ContentType;
 
 /**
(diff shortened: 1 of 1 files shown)


moxie0:
i don't understand, why does going through their content provider not work?

rhodey:
@moxie0 it's not clear, but based on stack traces the suspicion is that we need to request some persistable URI permissions in a away similar to the discussion in #3154, however there is currently no documented way to go about this. the only information available on this subject is some very recent stack(exchange/overflow) questions and their conclusion is that you can:

1) make sure to read from the URI only once, and on this single read copy the URI data into a file locally.
2) very recently people noticed that you can parse the decoded URI from the URI and use that.

I'm not certain this is the best way to handle the issue but it definitely is the best way we know of right now and because sharing from google photo's is entirely broken at this time I think it makes sense to use this until more information becomes available on the subject.

PR #1815, opened 2014-08-09
Per-Recipient notification silencing

tinloaf:
This implements a first step towards per-recipient notifications as discussed on the mailing list and wished for in many tickets. So far, I only implemented the possibilities to silence certain people and groups permanently or temporarily, since this seemed to be the minimal consensus achieved on the mailing list.

**WARNING**: This upgrades your database. Although a downgrade should be comparatively easy (delete a table and downgrade the global database version), make sure to have a backup if you test this.

PR #3378, opened 2015-06-09
Support for per-recipient muting, blocking, and ringtones.

### LINKED ISSUE #222: Feature: Blocking
Blocking by user-specified phone number, via something like an email filter rule where the messages from the number are deleted upon receipt and no notification is given.

### LINKED ISSUE #354: Adding per-contact notification preferences
I'm working to add functionality to TextSecure to allow customization of notifications (ringtone, vibrate pattern, led color) on a per-contact basis.  I've got two questions:

1) Is this a feature that you'd be interested in including in mainline TextSecure, and

2) If so, would it be better to let android store the per-contact preferences in the usual preferences location or would it be better to add them to the IdentityDatabase?

Thanks!

### LINKED ISSUE #757: different notification settings for group chats
It would be very nice to be able to change the notification settings for goups independently from other (personal) chats, as group chats might have a lower priority for the user.

moxie0:
Fixes #757
Fixes #354
Fixes #222
Closes #1815

// FREEBIE


agrajaghh:
There is an issue on Lollipop (tested on a Nexus4 with 5.1.1, looks good on 4.4 and 2.3):
screenshot_2015-06-09-17-28-30

I noticed also a minor alignement issue (on every android version):
screenshot_2015-06-09-17-28-35

I guess the mute icon should be vertically aligned with the time or it should be right next to the contact name.

moxie0:
Thanks @agrajaghh , I think that should do it?

agrajaghh:
yep, the black font is gone :blush:  But there is still a minor alignment issue in the conversation list:
screenshot_2015-06-09-22-41-23
The icon is some pixels too high, no? I know nitpick, but it should look good, not like pgp ;)

In the conversation the icon looks a bit weird to me:
screenshot_2015-06-09-22-41-30

how about centering it there as well? (photoshop)
screenshot_2015-06-09-22-41-30 - kopie

I also noticed a bug in the behaviour:
If you have muted Alice and receive a message from Bob (not muted) and now you receive a message from Alice you'll get a notification (incl. sound).

Btw, how is it possible to block contacts? I couldn't figure it out (didn't look at the code yet)...

McLoo:
Started to comment on code lines. Too much for tonight...

In short: 
- Group titles don't update anymore on devices with this feature -  incoming and outgoing updates.
- last received time stamp in conversation list is dark gray rather than fancy blue for unread threads
- Also the group avatar in the conversation list isn't circle cropped, but square. It's circled in cropped in group details though. Maybe of help: I set the avatar from the non PR3378 device. So it's from an incoming update

all tested only with dark theme

agrajaghh:
I found the blocking feature, afaik its too hidden behind the contact/group name. What about an additional entry in the menu (Notification settings, or something like that)? I won't be the only one missing it :wink: I guess it would be enough to show this and get rid of "Mute notification".

The default group avatar is broken there too:
screenshot_2015-06-09-23-10-17

And the alignment of the group name in the conversation is wrong:
screenshot_2015-06-09-23-10-29

5boro:
I noticed it off-centers the group name, as if it was an individual conversation:
Before
ts1
After
ts2
The icon is also off-center:
ts3
But the conversation list looks good
ts4

McLoo:
Good find @agrajaghh really good hidden! 
And didn't you fix slider switches once? 
CM 12.1
screenshot_2015-06-09-23-27-25

agrajaghh:
@McLoo its looking totally different for me:
screenshot_2015-06-09-23-31-47

Why do you have different background colors? could be related...

McLoo:
@agrajaghh oh hell, seems unreverted layout changes got into the test build. Or is CM...

moxie0:
Thanks @agrajaghh those alignment issues should be fixed. @McLoo I can't reproduce the things you posted, did you say you have some local changes?

McLoo:
@moxie0 need to check for local changes. Not within the next 20 hours. Sorry (for spamming)

agrajaghh:
@moxie0 major alignment are fixed for me :blush: 

But I still think the icon is a bit too high:
screenshot_2015-06-10-08-49-50
It should be like that (photoshop):
screenshot_2015-06-10-08-49-50_2
***
Also the red font is hard to read on my phone:
screenshot_2015-06-10-08-45-07

Maybe use the icon as well plus "BLOCKED" in white, like that (bad photoshop):
screenshot_2015-06-10-08-45-07 - kopie
or this:
screenshot_2015-06-10-08-45-07 - kopie - kopie
***
Also I'm experiencing the square cropped avatar like @McLoo. Its just happening when some else is changing the avatar, not when I do it:
screenshot_2015-06-10-09-06-07

oaim:
Maybe the "BLOCKED" font could be black? I don't think it should be white because it's not highlighted enough, but yeah red is a bit hard to see. Maybe black would do the job fine?

McLoo:
@agrajaghh changing the group name works for you?

agrajaghh:
@McLoo yep, tested on 5.1.1 and 2.3, is it not working for you?
gif

2-4601:
> Also the red font is hard to read

this^
Red font on a blue background causes the chromostereoptic effect.

I've seen the red font also in other warnings. For example when one of your contact changes keys.

One text color that works both on black and blue backgrounds is yellow (besides the white that's already used).

5boro:
@agrajaghh Changing the group title works, but not the picture

McLoo:
@moxie0 
update:
- changing the group name works now (mea culpa: local changes :blush: )
- the issues with the slider and the overall-darkness were caused by a dark CM theme/design (Dark Material)
- received avatars are still uncropped (until you reinstall TS, then the are circled)
- @5boro outgoing avatar update works for me

new:
Carl blocks a group "Eee"
- Dana sends a group update for "Eee" --> gets processed at Carls' device: **good imho**
- Carl also gets notified for that: **not good**
- Carl also receives all of Dana's group message in "Eee" and gets notified: **bad**
- Carl blocks Dana in her singel thread directly (single thread). That also blocks Danas' messages in group "Eee": **intended?**

Log after receiving on blocked group: ``Carl = +****15, Dana= +****33``

all on CM 12.1 nightly

moxie0:
Thanks @McLoo. 

1) Square avatar icons.  Fixed -- turns out this is actually an existing bug.  The update won't get displayed until the activity redraws, but that's a whole other problem we have with contact updates in general.

2) Blocking groups. Good catch, we actually can't really support this.  I've disabled the ability to block groups, but left it so that if you block someone *in* a group you don't see messages from them for that group.

oaim:
If you block a contact, will they be able to add you to groups? I'm thinking about a scenario where you keep getting added to a group you don't want to be part of.

moxie0:
@oaim No, you will no longer see group updates from them.  Which is actually also kinda weird, just because if you're in a group with people that *haven't* blocked that contact, you'll potentially develop different views of the group (membership, title, avatar, etc).

oaim:
What matters is that I'm able to leave the group without being added again by that contact, so that's good :)
As for the weirdness that comes from having blocked someone, I think that users expect such a behavior. Or at least I did

moxie0:
In 2.18.0

McLoo:
@moxie0 was it intended to have 'Mute **conversation**' in the pref screen and 'Mute **notifications**' as menu entry? 
Maybe just name it 'Mute'?

PR #3383, opened 2015-06-09
Fix resend button behaviour for failed group messages

### LINKED ISSUE #2639: Incorrect behavior of message resend dialog
Scenario: I am member of a group and a message did not get through (e.g., because I have a bad network connection). I then am presented with a message stating that I should tap for further details.
After tapping I see a message detail dialog together with a list of all group members and separate "send again" buttons for every member.

Problem: Whenever I tap on one of those buttons the message is not only resent to this specific member but to the whole group. To get rid of all buttons I would need to tap them all, thereby causing the message to be sent multiple times to every person.
I see that there might be cases in which some members have got the message while others have not. In that case it would be good if the button would only cause a resend to this specific account.
Often, it will be the case that none of the members got the message because the network issue is on the sender side. In that case, a "send again to whole group" button would be preferable.

haffenloher:
_Edit:_ This PR now replaces the multiple individual resend buttons with a single one (see [screenshot](https://github.com/WhisperSystems/Signal-Android/pull/3383#issuecomment-121348671)) and fixes #2639.

---

I found that [MessageRecipientListItem.java#L185](https://github.com/WhisperSystems/TextSecure/blob/f1e8357131a91c9000bebcd0b8858a2be7c4adb5/src/org/thoughtcrime/securesms/MessageRecipientListItem.java#L185) never gets executed because for group messages, `record.recipients` only contains a recipient with `name` and `number` being "Unknown".

With the proposed change (basically the same as [SmsDatabase.java#L434](https://github.com/WhisperSystems/TextSecure/blob/f1e8357131a91c9000bebcd0b8858a2be7c4adb5/src/org/thoughtcrime/securesms/database/SmsDatabase.java#L434), just without the call to PhoneNumberUtils.formatNumber()), `record.recipients` contains a recipient with `name = "Some Group Name"` and `number = "__textsecure_group__!some_group_id"`.
The resend button now only resends to the selected recipient.

I'm not sure if there's some existing code relying on the current behaviour. At least I found no problems when testing sending, re-sending and receiving one-to-one and group messages with and without media.


brumsoel:
I just found [this older comment](https://github.com/WhisperSystems/TextSecure/issues/2874#issuecomment-94598217) by @rhodey about the same `MmsDatabase` problem with sent messages. When he tried to resolve it - apparently by adding all group members as recipients to the MMS Message record? - "all sorts of things started going wrong"...

rhodey:
I ran this PR through all of our automation tests and it performed fine there, none of the weirdness I ran into before as mentioned in the comment of mine you linked too. then I tested this manually with lots of single recipient and group messages, media included, everything seemed ok. then to test the message resend I setup the following scenario:

1) 3 phones (A, B, C) all added to one group
2) send some messages in the group for sanity check
3) uninstall and reinstall TS on phone A
4) send message from phone B
5) phone C received message, phone A does not because ID key conflict
6) verify ID key conflict from phone B
7) message is resent, only phone A receives the resend as intended

so that went well :D but then I tested the following scenario:

1) 3 phones (A, B, C) all added to one group
2) send some messages in the group for sanity check
3) build new apk with an invalid TS server hostname
4) install apk on phone A
5) send message from phone A
6) error, message not delivered to phone B or C
7) build new apk with valid TS server hostname
8) install apk on phone A
9) open message details, resend to phone B
10) phone B receives message
11) message now has two check marks although phone C never received the message, not possible to resend to phone C because message is no longer in error state

so this comes closer, but in my experience this PR handles message resend for one failed send but is unable to handle the case where a message needs to be resent to more than one recipient.

brumsoel:
You're right, the error state gets removed after the first successful resend to phone B (that's a problem, I'll try and take a look at it), but resending to phone C should still be possible from the message details.

brumsoel:
Now I re-add the error state every time until there are no network failures left :D
Works well in my tests, but there may be a more elegant way.

jeremymasters:
This might be too extensive for now, but a future solution might include tapping the message and seeing who it succeeded to and failed to in the message details and be able to re-send to all failed users, if there are any.

rhodey:
related #3582

brumsoel:
@rhodey Just pushed a fix for that!

I also added a possibility to resend all failed messages at once, as suggested by @jeremymasters:
textsecure_pr3383_resend_all
Maybe there's a better / shorter wording for that button...

Please tell me if you think I should split some of the commits into separate PRs!

agrajaghh:
Maybe just "Resend all"?

gladiac:
I would say "Resend failed" I think it is clear that it is about messages in this case ...

brumsoel:
textsecure_pr3383_resend_all_2

Yep, definitely better. Don't know why I decided against the shorter "Resend all" yesterday

mcginty:
Is there actually a situation where you wouldn't want to send to all recipients that failed to receive?

agrajaghh:
I thought about the same... If we just have one resend button, it should be clear to the user that the others don't receive the message twice.

brumsoel:
Actually, there's none that I can think of. The only thing that has to be solved on a per-recipient-basis are key mismatches.

brumsoel:
There you are:

textsecure_pr3383_resend_all_3

jeremymasters:
Do we need to make sure the resend button can't be hammered repeatedly like a chimp on crack?

brumsoel:
OK, I didn't test this properly, it doesn't work correctly for single recipients yet
@jeremymasters The first thing the onClick callback does is disabling the button, so I guess that should be no problem?

jeremymasters:
@brumsoel That seems like it should do the trick for starters.  Only other question I'd have is, when is it re-enabled?  Forgive me if this is a question that is silly.

brumsoel:
@jeremymasters No, not at all! This is my first try at Android development, so I'll need all the help I can get :)
Aaand you were perfectly right, I forgot to re-enable the button. Fixed that (and the issue with one-to-one conversations, too). Thanks!

brumsoel:
Found and fixed a small bug: when viewing an image you sent to a group through "All images", the group's name instead of "You" was shown as the sender

brumsoel:
Oops, fixed another bug: Both the resend and the verify button were being shown for an identity mismatch in a one-to-one conversation.

gladiac:
Oh, good catch! Patch looks fine for me.

brumsoel:
Rebased, fixed some more problems with wrong image senders being shown in `MediaPreviewActivity` & corrected alignment in `message_details_footer.xml`

brumsoel:
Rebased on top of 3.0.3. I tried to include the newer changes, but Signal crashed when I used MmsDatabase.Reader.getNext():
```
10-23 22:39:06.182: E/AndroidRuntime(8647): java.lang.IllegalArgumentException: column 'attachment_id' does not exist
10-23 22:39:06.182: E/AndroidRuntime(8647): 	at android.database.AbstractCursor.getColumnIndexOrThrow(AbstractCursor.java:303)
10-23 22:39:06.182: E/AndroidRuntime(8647): 	at org.thoughtcrime.securesms.database.AttachmentDatabase.getAttachment(AttachmentDatabase.java:429)
10-23 22:39:06.182: E/AndroidRuntime(8647): 	at org.thoughtcrime.securesms.database.MmsDatabase$Reader.getSlideDeck(MmsDatabase.java:1085)
10-23 22:39:06.182: E/AndroidRuntime(8647): 	at org.thoughtcrime.securesms.database.MmsDatabase$Reader.getMediaMmsMessageRecord(MmsDatabase.java:1019)
10-23 22:39:06.182: E/AndroidRuntime(8647): 	at org.thoughtcrime.securesms.database.MmsDatabase$Reader.getCurrent(MmsDatabase.java:965)
10-23 22:39:06.182: E/AndroidRuntime(8647): 	at org.thoughtcrime.securesms.database.MmsDatabase$Reader.getNext(MmsDatabase.java:956)
10-23 22:39:06.182: E/AndroidRuntime(8647): 	at org.thoughtcrime.securesms.jobs.PushGroupSendJob.onSend(PushGroupSendJob.java:84)
...
```

brumsoel:
Rebased again, with today's commits (c258b4c7?) the crash is gone :)

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -68,6 +68,7 @@
   private MasterSecret masterSecret;
   private Recipients   recipients;
   private long         threadId;
+  private boolean      groupThread;
```
moxie0: is there a reason for this to be an instance variable?

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -100,8 +100,9 @@ private void initializeActionBar() {
     } else {
       relativeTimeSpan = null;
     }
-    getSupportActionBar().setTitle(recipient == null ? getString(R.string.MediaPreviewActivity_you)
-                                                     : recipient.toShortString());
+    getSupportActionBar().setTitle(recipient == null || recipient.isGroupRecipient()
+                                   ? getString(R.string.MediaPreviewActivity_you)
+                                   : recipient.toShortString());
```
moxie0: What's this for?

src/org/thoughtcrime/securesms/jobs/PushGroupSendJob.java
```diff
@@ -77,14 +78,21 @@ public void onSend(MasterSecret masterSecret)
...
       database.markAsPush(messageId);
       database.markAsSecure(messageId);
-      database.markAsSent(messageId);
       markAttachmentsUploaded(messageId, message.getAttachments());
+
+      if (record.hasNetworkFailures() || record.isIdentityMismatchFailure()) {
+        database.markAsSentFailed(messageId);
+      } else {
+        database.markAsSent(messageId);
+      }
```
moxie0: what's this for?

src/org/thoughtcrime/securesms/ConversationFragment.java
brumsoel: I think I reused the value somewhere at some point.. but you're right, it's not needed anymore.

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
brumsoel: After [this change](https://github.com/brumsoel/TextSecure/blob/93841b84ad4b4a0684474ec8e4742d269766530c/src/org/thoughtcrime/securesms/database/MmsDatabase.java#L743), `recipient` isn't always `null` for outgoing messages anymore. Without the additional `isGroupRecipient()` check, opening "All images" and selecting an image I sent to a group results in the group name being shown instead of "You" in the title bar.

src/org/thoughtcrime/securesms/jobs/PushGroupSendJob.java
brumsoel: it adresses the problem rhodey described [in his comment](https://github.com/WhisperSystems/Signal-Android/pull/3383#issuecomment-119750254).
Sending a group message is a single SendJob, but *re*sending is a separate job for every recipient. If resending fails for recipient A and succeeds for B afterwards, the message should not be marked as successfully sent but keep its red exclamation mark and prompt the user to resend again.

moxie0: hmm but does that means there's a race there? if the job is per-recipient, they can execute simultaneously right?

haffenloher: damn, you're totally right.
Should be better now, c7af1115 turns resending into a single job for all network failures. Identity mismatches have to be resolved seperately though, so I think we still need the above if..else statement :/

scienmind:
@haffenloher regarding your last mock-up:
If you go for the single button option, I'd suggest renaming the button to "Resend failed".

Just "Resend" will confuse people - whether you are resending the message to the whole group or just to the failed contacts (as @gladiac already mentioned).

kwantenmechaniker:
@haffenloher could your commits also (partially, at least) be used for addressing #4550 ? it seems somehow related to the issues you fixed.

haffenloher:
@kwantenmechaniker No, I didn't address this - should probably go into its own PR. I'll maybe take a look when I have the time.

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
moxie0: Sorry if I'm out of it, but is that related to this change?  Let's keep it separate if not.

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -157,8 +157,8 @@ public void reloadList() {
   }
 
   private void initializeResources() {
-    this.recipients = RecipientFactory.getRecipientsForIds(getActivity(), getActivity().getIntent().getLongArrayExtra("recipients"), true);
-    this.threadId   = this.getActivity().getIntent().getLongExtra("thread_id", -1);
+    this.recipients  = RecipientFactory.getRecipientsForIds(getActivity(), getActivity().getIntent().getLongArrayExtra("recipients"), true);
+    this.threadId    = this.getActivity().getIntent().getLongExtra("thread_id", -1);
```
moxie0: ?

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
haffenloher: Yep, it fixes a bug I introduced with this PR.

2-4601:
I tested this for one week with Huawei U8800, Android 4.0.4.
It was mostly normal use but also tested identity key changes and groups in poor network conditions.
Did not encounter any issues.

peterpanondrugs:
Still occurs the problem. it happens quite often and its super annoying.
ios log https://gist.github.com/6d881bed7d6fb54faecbdef7c1a913fb 
android log https://gist.github.com/ea564ca69aa6e504036ffdc405a8c538

2-4601:
More testing data:
I've been running these changes on my daily device (Nexus 4, 6.0.1) for over a month now. Everything works great, no issues with resending.

Trolldemorted:
do you require any further code reviews, more testing data, or anything?

large groups are borderline unusable, i am seeing duplicate messages almost every day and sadly contacts of mine have stopped using signal because of this bug.

jeremymasters:
Seems like it would cause a certain amount of unwanted traffic on the servers too if a bunch of duplicate messages get sent.

2-4601:
Not sure if it was this PR or something else but I managed to get a message in a state where it was not delivered yet there was no reason stated and the resend button was missing.

error

Unfortunately don't know the exact steps but here's what approximately happened.
0. I was in a bad coverage area (train)
1. I took a picture
2. Sent it in a three person group (myself included)
3. Message fails to deliver to all recipients
4. I click resend on both recipients.
5. Message gets sent to one recipient but not the other one
6. I click resend for the second contact
7. This failure and tapping the resend button might have happened once or twice more
8. Delivery fails once more but now the resend button has just disappeared

I have confirmed that the user for whom the sending failed did not receive the message.

I was running these changes on top of 3.21.3.

This log does not look very useful:
https://gist.github.com/44e1e555e4869ec999511f65796445c6

haffenloher:
@2-4601 yeah the failure isn't in the log, whatever it was. I think it's probably not related to this PR, but a log would still be interesting.

> I click resend on both recipients.

just to clarify (sorry for the dumb question): you clicked the single resend button below the recipient list, right? If you're running this PR, there should be no individual resend buttons

2-4601:
> just to clarify (sorry for the dumb question): you clicked the single resend button below the recipient list, right? If you're running this PR, there should be no individual resend buttons

Yeah, single resend, sorry for being vague. Just confirmed that I am (and was) really running this PR.

When is the resend button created? Would a log help where I open the message details view? Because I still got the message and there are no buttons

haffenloher:
> When is the resend button created? Would a log help where I open the message details view?

That kind of log wouldn't help unfortunately. The button [is shown if](https://github.com/haffenloher/TextSecure/blob/4b88a8a740e54003c57b2b67e59e78e85f9a1ee3/src/org/thoughtcrime/securesms/MessageDetailsActivity.java#L389-L397) the message record read from the database has NetworkFailures attached to it.

haffenloher:
> Not sure if it was this PR or something else but I managed to get a message in a state where it was not delivered yet there was no reason stated and the resend button was missing.

@2-4601 revisiting this - did your group contain an unregistered user?

2-4601:
> @2-4601 revisiting this - did your group contain an unregistered user?

Nope.

haffenloher:
Rebased and dropped the 3 commits dealing with #3582. They're now in #6171

real-or-random:
I had the same issue as @2-4601 recently (no resend button at all) and I run the version from Play Store. So that issue is probably not related to this PR.

Trolldemorted:
I now also have this behaviour w/o this PR. Cannot tell who received the message, but it was definitely not synced.

Trolldemorted:
@2-4601 could it be that the sync message to your slaves failed?

PR #3391, opened 2015-06-11
quick direct capture

mcginty:
This continues off the work of the original author @turtlekiosk, big shout out.

### what
If a camera is available on the hardware, provide a "quick capture" drawer directly within the app. Unlike when using external intents, captured media will never be visible to other apps. Which is pretty cool.

### summary
* Use CWAC-Camera library for cross-device camera compatibility since there are many gotchas in the Android world. Unfortunately the library is a bit dated, so there are a few classes that needed to be copied and modified to fit our needs.
    * `SurfacePreviewStrategy.java` is copied without modification
    * `TexturePreviewStrategy.java` is copied without modification
    * `CameraView.java` was copied and modified to initialize asynchronously. Unnecessary parts were removed.
* The "keyboard-aware" code was moved out of a ViewGroup and into a helper method so it can be used by whatever needs that info. Makes more sense that way.
* Drafts can now be encrypted or unencrypted, they're treated the same. All input is through PartAuthority (maybe I should rename it to ContentAuthority come to think of it). A new `CaptureProvider` deals with the file management.

### known bugs
* ~~on gingerbread the cropping rectangle isn't set properly, creating a strangely cropped result~~
* dragging animation twitches sometimes, unknown cause


AndroidManifest.xml
```diff
@@ -112,7 +113,7 @@
     <activity android:name=".ConversationActivity"
               android:windowSoftInputMode="stateUnchanged"
               android:launchMode="singleTask"
-              android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize" />
+              android:configChanges="touchscreen|keyboard|keyboardHidden|screenLayout|screenSize" />
```
moxie0: This would have implications for every asynctask in conversationactivity.

mcginty: Hmm, yeah. Mental note to move our AsyncTasks to using an event bus instead of relying on the existence of their starting activities, because I bet there are still some race bugs around that. Also looks like putting AsyncTasks in a Fragment that has `setRetainInstance(true)` will work, but I digress.

mcginty:
fixed orientation issues.

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1100,6 +1123,22 @@ private void updateToggleButtonState() {
     }
   }
 
+  @Override
+  public void onCollapsed() {
```
moxie0: I think these interface method names or arguments should be a little more descriptive if you want the activity itself to implement them.

mcginty: legit

mcginty:
made callback names more descriptive

src/org/thoughtcrime/securesms/components/camera/QuickAttachmentDrawer.java
```diff
@@ -0,0 +1,544 @@
...
+  public QuickAttachmentDrawer(Context context, AttributeSet attrs) {
+    this(context, attrs, 0);
+  }
+
+  public QuickAttachmentDrawer(Context context, AttributeSet attrs, int defStyle) {
+    super(context, attrs, defStyle);
+
+    if (!isDeviceSupported(context)) {
+      throw new IllegalStateException("Tried to create drawer on unsupported device.");
+    }
```
moxie0: looks like this happens before the check in conversation activity

src/org/thoughtcrime/securesms/components/camera/CameraView.java
```diff
@@ -0,0 +1,440 @@
...
+      }
+    }.execute();
+    addView(previewStrategy.getWidget());
+  }
+
+  public void onPause() {
+    if (camera == null || previewSize == null) {
+      synchronized (CameraView.this) {
+        while (camera == null || previewSize == null) {
+          Util.wait(CameraView.this, 0);
```
moxie0: looks like there's a failure case on onResume that would result in this sitting here forever? also, kinda weird that we're blocking, maybe set something for that thread to find instead?

src/org/thoughtcrime/securesms/components/camera/CameraView.java
```diff
@@ -0,0 +1,440 @@
...
+      rotation = (info.orientation - orientation + 360) % 360;
+    }
+    else { // back-facing camera
+      rotation = (info.orientation + orientation) % 360;
+    }
+
+    return rotation;
+  }
+
+  Activity getActivity() {
```
moxie0: why package?

src/org/thoughtcrime/securesms/components/camera/CameraView.java
```diff
@@ -0,0 +1,440 @@
...
+
+  void previewDestroyed() {
+    if (camera != null) {
+      previewStopped();
+      camera.release();
+      camera = null;
+    }
+  }
+
+  void previewReset(int width, int height) {
```
moxie0: package?

src/org/thoughtcrime/securesms/components/camera/CameraView.java
mcginty: unmodified from original, will limit scope further

src/org/thoughtcrime/securesms/components/camera/CameraView.java
mcginty: unmodified from original, will limit scope further

mcginty:
fixed IllegalStateException, agree the onPause blocking is suboptimal, just need to think of a good way to make that async without excessively complicating the task queue situation there

moxie0:
@mcginty It's not possible to let the async task perform the pause/stop/whatever operations when it catches up?

mcginty:
closing to base this off of 2.21.0 instead

PR #3399, opened 2015-06-11
add "Notification settings" to menu in ConversationActivity

### LINKED ISSUE #3395: I can mute from the options menu and the recipient preferences screen?
I knew contact blocking and muting was implemented so went to search for it, found the muting real quick in the options menu but surprised to find that block was not right next to it. it took me ~3 minute of searching followed by a text to @moxie0 to figure out where block was, found it by tapping on contact name.

agrajaghh:
This removes `Mute notifications`from the menu and adds `Notification settings` which will open the `RecipientPreferenceActivity`. When a conversation is muted `Unmute` will be still in the menu.

Fixes #3395

 // FREEBIE

### DIFF
--- a/res/menu/conversation.xml
+++ b/res/menu/conversation.xml
@@ -10,4 +10,6 @@
     <item android:title="@string/conversation__menu_delete_thread"
           android:id="@+id/menu_delete_thread" />
 
+    <item android:title="@string/conversation__menu_notification_settings"
+          android:id="@+id/menu_notification_settings" />
 </menu>
--- a/res/menu/conversation_unmuted.xml
+++ b/res/menu/conversation_unmuted.xml
@@ -1,7 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<menu xmlns:android="http://schemas.android.com/apk/res/android">
-
-    <item android:title="@string/conversation_unmuted__mute_notifications"
-          android:id="@+id/menu_mute_notifications" />
-
-</menu>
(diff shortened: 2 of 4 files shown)


rhodey:
+1 :)

I'm not sure "Notification settings" is the correct title, seems ok, but maybe @mcginty has some vocabulary suggestions

agrajaghh:
yeah, wasn't sure about thath, I'll leave that to the native english speakers :wink:

moxie0:
I'd prefer to leave it the way it is.  Blocking is not an action we want to encourage, while muting is something that we want people to avail themselves of more day-to-day.  The entire notification settings page could be more visible, but I think I'd like to find a way to highlight that without another menu item.

agrajaghh:
To help people find the block option it might work to add a little help text to settings -> privacy -> blocked contacts. But then the custom ringtone and vibrate options are still too hidden..

Mihahn:
How about adding an overlay on the first start explaining how to open those settings? Like it's explained [here](http://developer.android.com/design/patterns/help.html) for gestures?

5boro:
Long tap on conversations?

With [this icon](https://www.google.com/design/icons/#ic_notifications_paused) ? ic_notifications_paused_black_48dp

agrajaghh:
@moxie0 what do you think about a general *Contact Info*/*Group Info* Activity which might combine:
* general infos about the contact (other numbers with call button?)
* group infos/settings (if group conversation)
* Notification settings
* All images
* Verify contact

maybe organized inside different tab?

moxie0:
If we put the circle crop avatar into the ConversationActivity actionbar, it might be more clear that it's something you can tap.

@agrajaghh I think we're probably moving in that direction...

PR #3404, opened 2015-06-11
Add version info in preferences/advanced

nthomsn:
Ref #3324

Now looks like this:
device-2015-06-11-144419

"TextSecure 2.17.0" comes from the android manifest, and "TextSecure" uses the localized app name.


moxie0:
in 2.19.0

PR #3411, opened 2015-06-12
Add developer buildType

mcginty:
This is a buildType that specifically strays from the normal build by having more verbose and obnoxious debug output to help smooth out the production build.

* LeakCanary catches and reports activity leaks and their corresponding heap dumps to identify the cause.
* StrictMode warns for all kinds of blocking operations, leaks, and more. See Android docs for details.


moxie0:
In 2.20.0

rhodey:
fyi: once this is merged into master I'll have to reconfigure jenkins to use the new build target "auotomation" vs "debugTest", this means all PRs that haven't been rebased with this change will fail.

PR #3412, opened 2015-06-12
only open media preview when there is something to preview.

### LINKED ISSUE #3389: media preview of attachment being downloaded
it's possible to open the media preview activity when tapping on an attachment that is in the process of downloading (ie you're receiving it and only see the download icon). in this case you just see an empty (all black) media preview and upon tapping the save icon "error saving to storage".

super picky, but should be easy enough to prevent this from happening and I think it should be done.

reproducible on master and f1137927d4a9441f0dced4840d714f1ad95266d9.

rhodey:
only open media preview when there is something to preview.

Fixes #3389
// FREEBIE


mcginty:
lgtm

PR #3415, opened 2015-06-12
fix resource leaks

mcginty:
Ran Facebook's new Infer static analyzer, and it found some legit cases of potential resource leak.

### DIFF
--- a/src/org/thoughtcrime/securesms/crypto/PreKeyUtil.java
+++ b/src/org/thoughtcrime/securesms/crypto/PreKeyUtil.java
@@ -106,24 +106,36 @@ public class PreKeyUtil {
   }
 
   private static void setNextPreKeyId(Context context, int id) {
+    FileOutputStream fout = null;
     try {
-      File             nextFile = new File(getPreKeysDirectory(context), PreKeyIndex.FILE_NAME);
-      FileOutputStream fout     = new FileOutputStream(nextFile);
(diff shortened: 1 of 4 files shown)


McLoo:
@mcginty 
what about a cursor/reader leak?

`GroupDatabase.java` could use an update
```JAVA
  public GroupRecord getGroup(byte[] groupId) {
    Cursor cursor = null;
    Reader reader = null;
    GroupRecord record = null;

    try {
      cursor = databaseHelper.getReadableDatabase().query(TABLE_NAME, null, GROUP_ID + " = ?",
                                                          new String[] {GroupUtil.getEncodedId(groupId)},
                                                          null, null, null);
      reader = new Reader(cursor);
      record = reader.getNext();

    } finally {
      if (reader != null)
        reader.close();

      if (cursor != null)
        cursor.close();
    }

    return record;
  }
```


Also in `DatabaseFactory.java`
[Cursor mmsCursor](https://github.com/WhisperSystems/TextSecure/blob/ae4f6b1dd4af6b8dce07d6dfe4024ed95f738511/src/org/thoughtcrime/securesms/database/DatabaseFactory.java#L338) and [Cursor partCursor](https://github.com/WhisperSystems/TextSecure/blob/ae4f6b1dd4af6b8dce07d6dfe4024ed95f738511/src/org/thoughtcrime/securesms/database/DatabaseFactory.java#L350) got be cleaned up nicely

PR #3424, opened 2015-06-13
Change 'Title' to 'Group name' in group update message

McLoo:
Feels more descriptive to me. 
`Title` could also be something else - at least for a non native speaker...


mcginty:
in 3.5.0

PR #3437, opened 2015-06-16
Revert thumbnail null check for non image media

McLoo:
You might want to implement some check like "is media present", but for now this gets TS back to work!

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -413,7 +413,7 @@ public class ConversationItem extends LinearLayout {
         intent.putExtra(MediaPreviewActivity.DATE_EXTRA, messageRecord.getDateReceived());
 
         context.startActivity(intent);
-      } else if (slide.getThumbnailUri() != null) {
+      } else {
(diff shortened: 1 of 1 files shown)


rhodey:
tested with photos on my Nexus 5, and photos, video, and audio on my 2.3.6 gingerbread device. LGTM, thanks @McLoo

rhodey:
some context:

I need to stop testing things on my Nexus 5. this pr doesn't actually revert all of aa37d6e7f996761f0d1fd20fcb02f06daece90f9 and it is correct that it does not do that. the second null check in my #3412 pr is entirely unnecessary and also the source of this bug.

when I wrote aa37d6e7f996761f0d1fd20fcb02f06daece90f9 I originally did not include that second null check, but then TS crashed in my tests, I noticed it was related to the secure media dialog and so i was like "oh, of course i should have a null check there too!" and then added it.

adding the null check solved my crash but turns out the source of the crash was actually not a null check issue but a permissions issue caused by the way I install two TextSecure apks side by side on my Nexus 5 (change of application ID and package name).

In conclusion:

1. @McLoo's PR here is not a full reversion but is correct
2. I'm not going to do the application id build flavor hack anymore

Issue #3388, opened 2015-06-10
This one weird trick, the devs don't want you to know, crashes recipients' TS


2-4601:
(Couldn't resist with the title :( ·)

There seems to be no size checking on group name.
1. Update a group's name with _A Modest Proposal_ by _Jonathan Swift_ (Try the _Plain Text UTF-8_ option from [Project Gutenberg](https://www.gutenberg.org/ebooks/1080). Direct links don't work.)
2. After tapping the checkmark to broadcast the update **the other group members' TS will crash**

If you strip the Gutenberg metadata (introduction, license etc.) and just paste the actual text it won't crash. So the sweet spot is somewhere between 20-40KB.

In any case it's still a big wall of text that shouldn't be.
![a modest proposal-jonathan swift](https://cloud.githubusercontent.com/assets/174176/8141636/b5a2ce1e-1172-11e5-8f6a-1e2e6ae356c1.png)
- TS 2.17.0
  - Nexus 4, stock Android 5.1.1
  - Huawei U8800, Android 4.0.4

Debug log: https://gist.github.com/152ae1566c6c3d245ab0

```
E/AndroidRuntime( 4560): FATAL EXCEPTION: Thread-530
E/AndroidRuntime( 4560): Process: org.thoughtcrime.securesms, PID: 4560
E/AndroidRuntime( 4560): java.lang.IllegalArgumentException: byteCount < 0: -26331
E/AndroidRuntime( 4560):    at okio.RealBufferedSource.read(RealBufferedSource.java:46)
E/AndroidRuntime( 4560):    at com.squareup.okhttp.internal.ws.WebSocketReader$FramedMessageSource.read(WebSocketReader.java:249)
E/AndroidRuntime( 4560):    at okio.Buffer.writeAll(Buffer.java:705)
E/AndroidRuntime( 4560):    at okio.RealBufferedSource.readByteArray(RealBufferedSource.java:92)
E/AndroidRuntime( 4560):    at org.whispersystems.textsecure.internal.websocket.OkHttpClientWrapper.onMessage(OkHttpClientWrapper.java:102)
E/AndroidRuntime( 4560):    at com.squareup.okhttp.internal.ws.WebSocketReader.readMessage(WebSocketReader.java:110)
E/AndroidRuntime( 4560):    at com.squareup.okhttp.internal.ws.WebSocket$ReaderRunnable.execute(WebSocket.java:355)
[...]

PR #3438, opened 2015-06-17
fix: bug #3388 large group name crashes for all group members

junitas:
This fix sets the max size of a group name to be 30 chars. Previously large (20-40kb) group names could be inserted, which upon update would crash TS for all group members.


2-4601:
30 characters is **way** too short :)

I have currently a 34 character group name that is fully visible on my tiny Nexus 4 display. On larger devices (and tablets someday) even longer names would be visible.

For example If I want to make a reading group for _"Les horribles et épouvantables faits et prouesses du très renommé Pantagruel Roi des Dipsodes, fils du Grand Géant Gargantua"_ I'd need characters worth  a tweet.

jeremymasters:
If it's wider than my display, does it scroll? Maybe what's needed here is a long length like 255 but that if it doesn't fit in the display completely, it would scroll.  That would be pretty rad.

2-4601:
It does not scroll anymore but a few versions ago it did if the conversation was selected.
Maybe 2.17.x or 2.16.x

junitas:
I increased the max size of a group name to 255 chars, which is of course arbitrary. I didn't realize people used such long group names!

Longer group names now marquee in the conversation title view, and in the recipient preference activity there is now vertical scrolling available for longer names.

rhodey:
looking good @jeremymasters, thanks for the contribution :) I think in addition to preventing the TS client from creating a ridiculously large group name we'll also want to not crash when large group names are received. have you tested this PR to check if it handles receiving group updates that set very long group names? ie: we should prevent large names from being set but also not fail when they somehow do get set.

junitas:
@rhodey, the changes I've made only set a limit to the size when setting a group name, not when receiving. I'll look into adding that check on the receiving end and push a commit when I have something working.

rhodey:
@junitas I'm guessing the 2nd half of this issue will actually be a lot more work and not as straightforward, maybe @moxie0 is willing to take that on and merge this. we'll have to take note of this in #3388 and make sure to keep it open until the 2nd half is also complete.

junitas:
I've put a bit more work into the 2nd half of the issue, about as far as I can take it with my knowledge of the TS code. I'm not sure if more work needs done on it to address potential issues I'm not thinking of. 

When an incoming group create or group update contains a title that is larger than 255 chars, it is trimmed to 255 for the recipient of the group update/create. This partially fixes the bug for the recipient. When tested with an incoming ~40kb group title, it is trimmed to 255 chars and all is well. However, if the sender of the large group name then edits the group name and updates again the recipient's TS will crash. 

This was tested with a Nexus 4 (recipient) running my code, and a Samsung Galaxy S4 (sender) running stock TS. This "secondary" crash would be mitigated if both sender and recipient were running the code that contains the fixes I wrote, but I'm not sure how to stop the sender's TS from sending an abnormally long group title if they somehow worked around the 255 limit of the EditText object in the group update/create menu.

jeremymasters:
I think the 255 character (not bytes!) length here and the scrolling would be inextricably linked.  That many characters won't fit visibly in the title bar.  It would need to scroll or something like that.

rhodey:
looks like our automated testing bot might have incorrectly marked this PR as failing, the test it failed was [testSaveTextDraft](https://github.com/WhisperSystems/TextSecure/blob/master/test/androidTestEspresso/java/org/thoughtcrime/securesms/ConversationListActivityTest.java#L211). I'll have to rerun these tests and add a `@FlakyTest` annotation to that test if it passes, sorry for the trouble, will update ASAP.

rhodey:
the espresso test framework broke `@FlakyTest` :( https://code.google.com/p/android-test-kit/issues/detail?id=153, bummer.

rhodey:
please test this, fickle bot

rhodey:
tested on GB, works as described :) also code is looking good minus some picky style things in the if blocks, ex: https://github.com/WhisperSystems/TextSecure/pull/3438/files#diff-3ac080704bcdf2410005da0e1f360b9eR112

junitas:
I just noticed that this issue is still a problem as of 2.21.0. @rhodey , was wondering about the status of this PR and if theres anything else todo with my code to fix this bug?

rhodey:
we're super picky with our [style guidelines](https://github.com/WhisperSystems/RedPhone/wiki/Code-style-Guidelines) :P. there are only a couple things in this PR that don't fit the guide, if you could correct those in an additional commit then that'd address all my concerns. after that this PR should probably be assigned to @moxie0 for final review, then he'd merge.

thanks for supporting the project :)

rhodey:
sorry i didn't notice the diff changed since last checking this out. i think this all looks great and in accordance with our style guidelines except for the last commit 53ccc9f72601f9230db2a559d2ce348ff0729a57

all the logic in that last commit probably belongs in libtextsecure instead as I [hinted to earlier](https://github.com/WhisperSystems/TextSecure/pull/3438#issuecomment-113238663). if you simply revert that last commit I think we could merge this easily

junitas:
done. I missed what you had said about moxie possibly fixing that logic. thanks for getting back to me!

res/layout/group_create_activity.xml
```diff
@@ -36,6 +36,7 @@
                 android:layout_gravity="center_vertical"
                 android:padding="10dp"
                 android:lines="1"
+                android:maxLength="255"
```
moxie0: I think this should probably be the only line in this PR. We'd like to avoid the marquee behavior if possible...

moxie0:
in 3.8.0

PR #3439, opened 2015-06-17
Pretty up the attachment drawer a little.

moxie0:
1) Rounded corners on the image.

2) A circle X on top of the image, rather than a remove button.

3) Quick 200ms fade on clear.


rhodey:
tested on GB 2.3.6, photo, audio, and video all happy.

PR #3450, opened 2015-06-21
remove brackets from recipient numbers as not allowed for dialing

sseide:
this bugfix resolves issue #1516. Square brackets in numbers prevent the SmsManager from sending the message in some phones.

### DIFF
--- a/src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
+++ b/src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
@@ -91,6 +91,12 @@ public class SmsSendJob extends SendJob {
     ArrayList<PendingIntent> deliveredIntents = constructDeliveredIntents(message.getId(), message.getType(), messages);
     String recipient                          = message.getIndividualRecipient().getNumber();
 
+    // remove characters for visible number formatting - they break sending sms on some phones (issue #1516)
+    // stripSeparator() to rigid? what will people try to send here? - see discussion on #3099
+    //recipient = PhoneNumberUtils.stripSeparators(recipient);
+    // minimum required to fix the issue - brackets are no dialable chars...
+    recipient = recipient.replaceAll("[\\(\\)]", "");
+
     // NOTE 11/04/14 -- There's apparently a bug where for some unknown recipients
     // and messages, this will throw an NPE.  We have no idea why, so we're just
     // catching it and marking the message as a failure.  That way at least it doesn't


McLoo:
As 1516 is yours, I guess you successfully tested this?
I'm pretty sure you'll have to remove your comments, you can put the information in here.

src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
```diff
@@ -91,6 +91,12 @@ private void deliver(SmsMessageRecord message)
     ArrayList<PendingIntent> deliveredIntents = constructDeliveredIntents(message.getId(), message.getType(), messages);
     String recipient                          = message.getIndividualRecipient().getNumber();
 
+    // remove characters for visible number formatting - they break sending sms on some phones (issue #1516)
+    // stripSeparator() to rigid? what will people try to send here? - see discussion on #3099
+    //recipient = PhoneNumberUtils.stripSeparators(recipient);
+    // minimum required to fix the issue - brackets are no dialable chars...
+    recipient = recipient.replaceAll("[\\(\\)]", "");
```
moxie0: Problem is that these parens are part of a valid destination address.  By removing them, you could be mangling one valid address into a different valid address.

sseide: Do you have an example of this valid destination addresses? I have no idea what these could be...

moxie0: A shortcode for example.  Or an email address.

sseide: The replacement is directly done before calling the systems SMSManager. Modified is an copy only used to send this message via SMSManager.

For short codes - i do not know any country where round brackets are used in short codes to distinquish between different ones - as far as i know these are shorter numbers (mostly 3-5 digits) containing digits only, following the ITU dial number plans (e.g. E164, E123 - the last one explicitly recommends () for national numbers). I cannot think of any way brackets come into play here to distinguish different numbers... 
 
And emails are not allowed to contain round brackets within themselfs (as of the RFCs 2881 / 5321). The only form remaining as used/allowed in some programs (even the Android 4.2 SMS-Manager is "alias (user@domain.com)". Here you are right - it will brake if brackets are removed.

Should it be save to change it like?

<pre>
    if (!recipient.contains("@")) {
      recipient = recipient.replaceAll("[\\(\\)]", "");
    }
</pre>

moxie0: Shortcodes aren't only numeric.

sseide: To clear things up - what kind of short code are you talking about? Can you please given an working example? You cannot talk about gsm short codes (the one i'am reffering).

GSM short codes (or UMTS/LTE/telephone network in general) as used by DTEs must conform to E.164 numbering plan - therfore only digits (as well as *#) allowed for transmission. We do not talk about network internal short codes for routing etc. as they can't be send from end devices like mobile phones  (at least as long as the mobile network provider kinda cares about its network)...

Vanity numbers are out of scope too, as the SmsManager on Android (at least on different phones i've testet) does not convert them to digits. Therefor no SMS/MMS can be send giving an vanity number with letters instead of digits. The only letters i have found working are email addresses.

moxie0: Do you need to escape parens when they're inside brackets?

Issue #3453, opened 2015-06-22
crash on sharing video files not indexed by media scanner


sseide:
whenever an intent to share video files is send to TextSecure, the app crashes (NullPointerException) sending the message to the recipient.

<pre>
06-21 20:38:33.388    4959-4959/? W/ConversationActivity﹕ isManual Selection: false
06-21 20:38:33.388    4959-4959/? W/ConversationActivity﹕ forceSms: false
06-21 20:38:33.430    4959-5117/? W/dalvikvm﹕ threadid=17: thread exiting with uncaught exception (group=0x420299a8)
06-21 20:38:33.435    4959-5117/? E/AndroidRuntime﹕ FATAL EXCEPTION: AsyncTask #1
    java.lang.RuntimeException: An error occured while executing doInBackground()
            at android.os.AsyncTask$3.done(AsyncTask.java:299)
            at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:352)
            at java.util.concurrent.FutureTask.setException(FutureTask.java:219)
            at java.util.concurrent.FutureTask.run(FutureTask.java:239)
            at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:230)
            at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1080)
            at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:573)
            at java.lang.Thread.run(Thread.java:838)
     Caused by: java.lang.NullPointerException
            at java.lang.String.<init>(String.java:228)
            at org.thoughtcrime.securesms.util.Util.toIsoString(Util.java:124)
            at org.thoughtcrime.securesms.mms.PartParser.getMessageText(PartParser.java:19)
            at org.thoughtcrime.securesms.database.MmsDatabase.insertMediaMessage(MmsDatabase.java:742)
            at org.thoughtcrime.securesms.database.MmsDatabase.insertMessageOutbox(MmsDatabase.java:724)
            at org.thoughtcrime.securesms.sms.MessageSender.send(MessageSender.java:102)
            at org.thoughtcrime.securesms.ConversationActivity$20.doInBackground(ConversationActivity.java:1130)
            at org.thoughtcrime.securesms.ConversationActivity$20.doInBackground(ConversationActivity.java:1127)
            at android.os.AsyncTask$2.call(AsyncTask.java:287)
[...]

PR #3454, opened 2015-06-22
Attach video npe

sseide:
fix for issue ##3453

Sending media files (audio / video) via Intent to Textsecure, it does check for media type in ShareActivity but does not set the media type correctly when creating the Slides from the URI.

TextSecure only checks the content resolver but fails on files not indexed there.
This patch adds fallback checks to get the mime type in this cases too to prevent crashes later on.

### DIFF
--- a/src/org/thoughtcrime/securesms/ShareActivity.java
+++ b/src/org/thoughtcrime/securesms/ShareActivity.java
@@ -24,12 +24,12 @@ import android.support.annotation.NonNull;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
-import android.webkit.MimeTypeMap;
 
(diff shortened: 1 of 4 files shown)


rhodey:
I'll test this today.

McLoo:
issue and solution confirmed on 2.19.0++ (https://github.com/WhisperSystems/TextSecure/commit/8f922637077242ad50a2804069515337c2c3f7ce) CM 12.1

Code should be unified for audio and video :grinning: 
I put this into `getContentTypeForPartOutsideDb` in #3311 for testing (not committing until @sseide got his reward)

McLoo:
@mcginty feels like this can be closed alongside #3453  - at least I wasn'T able to repro on  3.4.2+ (06f2908)

PR #3455, opened 2015-06-22
Support for quick reply from notifications.

### LINKED ISSUE #483: quick compose quick reply
Any plans on implementing quick compose and quick reply?

moxie0:
Fixes #483


McLoo:
this is a nice one :) 
noticed few things on HTC One S with CM 12.1:

- opening an image from popup and returning leads you to convo list. Had expected the right fullscreen convo at least, but rather the popup again as it does for audio/video and forwarding
- expanding to fullscreen feels slow with ~1 second (maybe cause https://github.com/WhisperSystems/TextSecure/commit/9fc02c8ce1a83f055a00d74ff9409dd9261769f5 isn't merged here?)
- ~~opening the popup immediately on income, the audio/video thumbnail is not shown but the download arrow . didn't notice that for images~~ ah, same in fullscreen --> seperate issue
- emoji drawer is a *little* bit to large for the popup, don't see what a tapped until I close the drawer... Uh, where is the keyboard icon?! :) back button works though.
(might also be an issue with the upcoming in app camera)
- if TS is already open, it's a bit hard to distinguish between the popup an TS - in dark theme
maybe add a border line?!


screenshot should cover point 2 and 3
screenshot_2015-06-24-07-53-53

McLoo:
Popup reply should be disabled for alphanumeric sender ID
(But probably sending to them completely, as they fail immediately. Only thing I can think of is am note function for you mTANs)

moxie0:
> opening an image from popup and returning leads you to convo list. Had expected the right fullscreen convo at least, but rather the popup again as it does for audio/video and forwarding
expanding to fullscreen feels slow with ~1 second (maybe cause 9fc02c8 isn't merged here?)

Hmm, can't reproduce.

> emoji drawer is a little bit to large for the popup, don't see what a tapped until I close the drawer... Uh, where is the keyboard icon?! :) back button works though. (might also be an issue with the upcoming in app camera)

Gonna wait for @mcginty to fix this.

> if TS is already open, it's a bit hard to distinguish between the popup an TS - in dark theme maybe add a border line?!

Yeah I add an alpha layer to the background for devices that support it, but that doesn't really help with dark mode.

moxie0:
Merged in 2.20.0

PR #3456, opened 2015-06-22
Support "quick response" when TextSecure is unlocked.

### LINKED ISSUE #299: Implement "Quick Responses"

moxie0:
Fixes #299

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -460,7 +460,9 @@
     <string name="MessageNotifier_media_message">Media message</string>
 
     <!-- QuickResponseService -->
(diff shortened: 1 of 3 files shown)


McLoo:
On CM 12.1, 
if TS is locked I don't get the toast "Quick response unavailable when TextSecure is locked!" though the code gets executed.
Instead system(?) is toasting "Message sent to *Number*"

moxie0:
Thanks @McLoo can you open an issue for that? This is now merged in 2.20.0

PR #3463, opened 2015-06-24
Contact colors

moxie0:
Colorize conversations and allow for user-selectable contact colors.


McLoo:
in `DatabaseFactory.java` we need a 
```
    if (oldVersion < INTRODUCED_RECIPIENT_COLOR_PREF) {
        db.execSQL("ALTER TABLE recipient_preferences ADD COLUMN  color INTEGER DEFAULT -1");
      }
``` 

to prevent these crashes:
```
     Caused by: java.lang.IllegalArgumentException: column 'color' does not exist
            at android.database.AbstractCursor.getColumnIndexOrThrow(AbstractCursor.java:303)
            at org.thoughtcrime.securesms.database.RecipientPreferenceDatabase.getRecipientsPreferences(RecipientPreferenceDatabase.java:92)
            at org.thoughtcrime.securesms.recipients.RecipientProvider.getRecipientsPreferencesSync(RecipientProvider.java:162)
            at org.thoughtcrime.securesms.recipients.RecipientProvider.getRecipients(RecipientProvider.java:90)
```

McLoo:
Another nice one, thanks!

Colors should not be set by default for a contact that has a photo set in contacts. 
It's a bit surprising (for most ppl) if your dad gets a pink convo :grimacing: 

*btw: prepare for per-user-background-image wishes*

agrajaghh:
@moxie0 I like it! :blush: But I think the contrast between the white text and the light colors (yellow and green) isn't enough. This would be hard to read in sunlight I guess.
screenshot_2015-06-24-23-23-41
screenshot_2015-06-24-23-23-52

@McLoo per-user-background is already supported (next to the hidden mute/block/notification settings)

moxie0:
@McLoo Right you are about the DB migration.

> Colors should not be set by default for a contact that has a photo set in contacts. 
> It's a bit surprising (for most ppl) if your dad gets a pink convo :grimacing:

I dunno, I kind of like the idea that each conversation gets a color, whether there's an avatar icon set or not.  I'll see what @mcginty has to say about this whole thing.

> @McLoo per-user-background is already supported (next to the hidden mute/block/notification settings)

@agrajaghh I think he meant "wallpapers."  =)

>  I like it! :blush: But I think the contrast between the white text and the light colors (yellow and green) isn't enough. This would be hard to read in sunlight I guess.

Hmm, maybe we should switch to the 500 colors instead of the 300 colors.

McLoo:
What about coloring the per-rec-pref header?

> @agrajaghh I think he meant "wallpapers." =)

@moxie0 Correct. And I thought he understood and pointed me to another hidden feature :wink:

agrajaghh:
@McLoo actually I didn't understand :laughing: 

quick and dirty photoshop with 500 colors (just changed the actionbar) Looks better imho
dfc06756-1ac8-11e5-8f1a-581f2feb840d
e20a7222-1ac8-11e5-8885-bf593a7efbf6

moxie0:
The status bar is already at 700, so that's probably the right situation. Not sure if we'd leave the generated avatar backgrounds at 300 or bump those to 500 as well. Though the white fonts are already working there with 300 colors.

agrajaghh:
@McLoo 
> What about coloring the per-rec-pref header?

I thought about that too, but then the background whould have the same color as the avatar. Having different shades would be possible, but don't know how that would look...
But I agree, having e.g. a red conversation and then going in the *per-user-settings* having it blue again is weird...

oaim:
>I thought about that too, but then the background whould have the same color as the avatar.
why not put a white circle around it? would it be too complicated/not worth it?

oaim:
question (I haven't built this version of the app myself yet): do the message bubbles change colors as well? Or is it just about the action bar? It would be cool if the bubbles were more colorful too... I know that technically the blue/green distinction is used for push/green, but maybe it's worth dropping.. most people don't even know the difference anyway, and there's the padlock image for that now that SMS are only unencrypted...

PR #3469, opened 2015-06-25
Translate 'Devices' in preferences

McLoo:
...to be ready when it gets activated.
 :snail: Transifex  :turtle: is sometimes a bit behind


moxie0:
I don't think we need to worry about this yet, it's not even clear that will be the string or that this is its final resting location.

oaim:
Just for the record, I think you should mention to the user how to add a new device. Something like "Nothing to see here. Download our Chrome extension to pair new devices!". Just a note for the future

PR #3493, opened 2015-06-29
fix gingerbread press reaction

### LINKED ISSUE #3392: gingerbread conversation list missing reaction to tap
with the ripple change, it looks like holding on a conversation list item doesn't change its color any more. there should be an indicator.

mcginty:
fixes #3392
// FREEBIE


moxie0:
in 2.21.0

PR #3504, opened 2015-06-30
Refactor colorization to support dark theme.

moxie0:


src/org/thoughtcrime/securesms/color/LightBlueMaterialColor.java
```diff
@@ -0,0 +1,32 @@
+package org.thoughtcrime.securesms.color;
+
+import java.util.HashMap;
+
+class LightBlueMaterialColor extends MaterialColor {
```
mcginty: why not make MaterialColor an enum?

src/org/thoughtcrime/securesms/color/MaterialColor.java
```diff
@@ -0,0 +1,73 @@
...
+public abstract class MaterialColor {
+
+  private final Map<String, Integer> colorWeightMap;
+
+  protected MaterialColor(Map<String, Integer> colorWeightMap) {
+    this.colorWeightMap = colorWeightMap;
+  }
+
+  public int toConversationColor(ThemeType themeType) {
+    if (themeType == ThemeType.DARK) return colorWeightMap.get("900");
```
mcginty: seems like we can avoid this whole ThemeType thing all together by having a <attr name="materialWeight" format="string|reference" /> that is "900" or "500" per theme and resolving it just like any other theme attribute.

proguard-spongycastle.pro
```diff
@@ -1,2 +1,19 @@
...
+-keep class org.spongycastle.crypto.paddings.* {*;}
+-keep class org.spongycastle.crypto.params.* {*;}
+-keep class org.spongycastle.crypto.prng.* {*;}
+-keep class org.spongycastle.crypto.signers.* {*;}
+
+-keep class org.spongycastle.jcajce.provider.digest.** {*;}
+-keep class org.spongycastle.jcajce.provider.keystore.** {*;}
+-keep class org.spongycastle.jcajce.provider.symmetric.** {*;}
+-keep class org.spongycastle.jcajce.spec.* {*;}
+-keep class org.spongycastle.jce.** {*;}
```
mcginty: this should just go straight into 2.21.0 now since GB is broken and it's separate from refactoring colorization

src/org/thoughtcrime/securesms/color/LightBlueMaterialColor.java
moxie0: It has methods that I want the subclasses to be able to override. We'd have to build lists of the enum values anyway, so it doesn't buy anything.

src/org/thoughtcrime/securesms/color/MaterialColor.java
moxie0: I could do that.  My thought was that we'd be able to tweak the return values on a per-color basis this way.

moxie0:
updated to use theme-defined weights

PR #3509, opened 2015-07-01
direct capture in-app

mcginty:


McLoo:
Just read roughly through this... you don't check if you're in quick reply or full screen convo yet, do you?

PR #3516, opened 2015-07-02
in-app direct capture

mcginty:


McLoo:
What about moving the camera button above the emoji button as soon as you reach 3 lines of text? Would save some space.
 On my 4.3 inch display the compose field is already quite limited

mcginty:
@McLoo was thinking about just hiding the camera button the same time that we switch to the "send" button instead of attach.

McLoo:
@mcginty so much more reasonable :see_no_evil:

mcginty:
updated, added the toggle animation

res/values/strings.xml
```diff
@@ -512,7 +512,7 @@
     <string name="conversation_title_view__conversation_muted">Conversation muted</string>
 
     <!-- conversation_activity -->
-    <string name="conversation_activity__type_message_push">Send TextSecure message</string>
+    <string name="conversation_activity__type_message_push">Send via TextSecure</string>
```
moxie0: I kinda like the previous wording better

McLoo: Testing this PR. 
On my 4.3" display there is a line break after TextSecure, putting 'message' in the next line when in quick reply. The new phrase is in one line...

mcginty: Yep, I changed it to fit better on small screens. i also liked the previous wording better, it's just long.

2-4601: There will always be some languages that simply don't have a short translation for this.
Can the string be ellipsized so it doesn't break to a new line on small screens?

mcginty: I thought about that, but I was worried about a language that might show the equivalent of "Send via...", which is a security indication loss for us.

moxie0: Let's switch it back to Send TextSecure message and ellipsize it

2-4601: How about dropping "Send" altogether and just use "TextSecure message" and "Unsecured SMS"?

Issue #3518, opened 2015-07-03
Quick response has problems with country code


zipperle:
If the country code begins with + the number is not decoded correctly. It does not decode %2B correctly to +.

rhodey:
@zipperle I'm unable to reproduce this issue, can you please provide exact steps to reproduce expected symptoms? I'm not sure why the phone number is URL encoded or how to get my phone into a state where this would be so. thanks :)

zipperle:
Hi @rhodey  the [QuickResponseService](https://github.com/WhisperSystems/TextSecure/blob/moxie/quick_response/src/org/thoughtcrime/securesms/service/QuickResponseService.java) gets an intent which contains an [URI](http://developer.android.com/reference/android/telephony/TelephonyManager.html#ACTION_RESPOND_VIA_MESSAGE). This URI contains the sms in the form "sms:\<number\>" and it seems that, if the number begins for example with +49... then the plus is not decoded correctly. This happens on a Samsung S4 with the original os. I hope you  have a better understanding of this issue now.

rhodey:
thanks for the context, I keep confusing quick reply with quick response 0.o

I'm still unable to reproduce the problem with a +1 number but +1 is my actual locale so maybe I need a non-default locale to reproduce this. anyways, I agree that the URI is using RFC 5724 format to pass on the number and can believe that some ROMs may decide to encode the `+` while others not.

PR #3519, opened 2015-07-03
Possible buxfix for #3518

zipperle:
Decode also the path.


zipperle:
Changed target branch.

PR #3520, opened 2015-07-03
Possible buxfix for #3518

zipperle:
Decode also the path.


zipperle:
https://github.com/WhisperSystems/TextSecure/pull/3519 wrong target

PR #3522, opened 2015-07-03
Buxfix for #3518

zipperle:
Decode also the parh.

### DIFF
--- a/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
+++ b/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
@@ -47,7 +47,7 @@ public class Rfc5724Uri {
     String[] parts = uri.split("\\?")[0].split(":", 2);
 
     if (parts.length < 2 || parts[1].isEmpty()) throw new URISyntaxException(uri, "invalid path");
-    else                                        return parts[1];
+    else                                        return URLDecoder.decode(parts[1]);
   }
 
   private Map<String, String> parseQueryParams() throws URISyntaxException {
--- a/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
+++ b/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
@@ -47,11 +47,11 @@ public class Rfc5724Uri {
     String[] parts = uri.split("\\?")[0].split(":", 2);
 
     if (parts.length < 2 || parts[1].isEmpty()){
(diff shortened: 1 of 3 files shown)


rhodey:
this looks good to me, see my comments in #3518 for added context. @zipperle before we merge this could you add a test to [Rfc5724UriTest](https://github.com/WhisperSystems/TextSecure/blob/master/test/androidTest/java/org/thoughtcrime/securesms/util/Rfc5724UriTest.java)? just copy the existing [testGetPath()](https://github.com/WhisperSystems/TextSecure/blob/master/test/androidTest/java/org/thoughtcrime/securesms/util/Rfc5724UriTest.java#L65) unit test in there and modify the test vectors to include an encoded `+` instead.

zipperle:
It seems that Samsung ignores the [recommendation](https://tools.ietf.org/html/rfc5724#section-2.3) that all numbers must begin with +. They use the [Percent-Encoding](https://tools.ietf.org/html/rfc3986#section-2.1) for the number instead. 

The URLDecoder.decode method removes the plus. Therefore it is necessary to check, if the path contains a percentage symbol.

rhodey:
test case looks good :) indentation should be 2 spaces per [guidelines](https://github.com/WhisperSystems/RedPhone/wiki/Code-style-Guidelines) but everything else looks good to me. thanks for the contribution!

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -46,8 +46,13 @@ private String parseSchema() throws URISyntaxException {
   private String parsePath() throws URISyntaxException {
     String[] parts = uri.split("\\?")[0].split(":", 2);
 
-    if (parts.length < 2 || parts[1].isEmpty()) throw new URISyntaxException(uri, "invalid path");
-    else                                        return parts[1];
+    if (parts.length < 2 || parts[1].isEmpty()){
+      throw new URISyntaxException(uri, "invalid path");
+    } else if (parts[1].contains("%")){
```
moxie0: Putting this into the URI class seems kinda gross to me.  This class should either decode or not decode, not base decisions around whether there's a "%" character present.  I assume that always decoding won't work, since it would interpret an unencoded "+" as a space.  So this class probably shouldn't decode, and the caller should be responsible.

moxie0:
in 2.23.0

PR #3532, opened 2015-07-06
use apple emoji

### LINKED ISSUE #3445: Alternatives to Google's emojis
Me and my friends all completely despise the Google emojis. I don't really know why, but the fact still remains that we do. With that said, I think a lot of people would be happy to see alternatives. Twitter's emojis are open source and look quite nice. Of course I would prefer Apple's, but those aren't open source at all so that might be a problem. I don't really see how WhatsApp could get away with using them though. With Signal for Iphone and Signal soon coming to Android, I think it's quite important that both people see the same thing. That's the way Telegram does it, and it makes sure both people see the exact same emojis. Some Apple/Google emojis are so different that it's not even funny. Yeah, I know Samsung for example has their own emojis, but that doesn't exactly make things easier.

Is this being worked on? Is it even being considered? I even tried using Emoji Switcher to get Apple's emojis, but it just isn't working.

mcginty:
fixes #3445
// FREEBIE


McLoo:
Seems to be a plus of about 2MB ?! So much detail...

and, can we keep him at least? 
unbenannt 
:D

mad-de:
Sweet merciful Jesus - just why?

mcginty:
if hairy heart had to go, so does that creature

@mad-de we decided emoji are something we want to be stable across platforms, so we're standardizing on iOS emoji

turtlekiosk:
I usually use the standard keyboard emoji pane instead of the emoji drawer, does this replace the emoji I type using the keyboard?

2-4601:
Especially the smiley faces in the iOS set are like the shitty stock drawings from the nineties CorelDraw.

Well, I'll get through this. Just need to start taking pain medication.

mcginty:
@turtlekiosk yep

@2-4601 time to start a support group

jeremymasters:
Well bless my hairy heart.

moxie0:
in 2.22.0

McLoo:
@mcginty the sacrifices for uniformity: Liberté, harihearté, bluetête  :grimacing:

mad-de:
Phew. Any plans on letting the user decide to use the nonstandard (android set) instead?

Mihahn:
So it's legal to use them? Because afaik you weren't sure if it is...

rainerzufall:
sad to hear that you decided to go with the ugly apple emoji's, but good that there will be the same set across platforms...

rbu:
Oh, I love the currently released (old) emoji set [1]. I understand the Apple set is what WhatsApp, Threema and all the others use, but I actually convinced people to use TextSecure purely on the arguments that it has nicer emojis!

Please bring them back or at least use the Emoji One variants.

[1] https://raw.githubusercontent.com/WhisperSystems/TextSecure/94085798621f4036454fef2a4096ad7588a80793/assets/emoji-people.png

Fanch3n:
I don't like the old emojis at all, especially the faces. The symbols are fine, probably better than the Apple ones, but the faces are just terrible. I'm happy they will be replaced.
Emoji One does look okay for me as well, though.

Thunderpig72:
Another vote for Emoji One here. Maybe if Apple flips out, Emoji One could be plan B?

jellium:
Totally missed the action unfortunately. But I realize that Android/Google's Emoji have been burried, therefore for what it's worth, I vote for Emoji One.

StefanBossbaly:
Are you guys willing to support multiple Emojis and have an option in the settings?  Would you accept a pull request that implements that functionality?

mjpromutico:
I would like to see an option to choose the system default Emoji set. Defaulting it to Apple Emojis and allowing the end user to force system default Emojis would allow for a stable vanilla experience for users.

mcginty:
The mailing list is the place for these kinds of discussions.

PR #3535, opened 2015-07-06
Change TS Media message timestamp from received time to sent time

### LINKED ISSUE #3525: TS Media Messages Received Timestamps Report Different Times
Sometimes I receive an MMS (over push) when I am offline; when I finally come back online, I get the MMS message with two different received timestamps. My guess is that one timestamp corresponds to the time sent, and the other time stamp corresponds to the time received.

This seems to be confusing and unnecessary information for the user, who probably only cares about when the message is sent, and doesn't understand why the time stamps are different.

Device info: Sony Xperia Z3 Tablet Compact, stock android 4.4.4, TextSecure 2.21.0? (I compiled from b7492ebdfa580687b44f1aa31410931f0b92dcdd)

App state: Push registered, incoming SMS disabled, outgoing SMS fully disabled

Debug logs: https://gist.github.com/anonymous/98fe3fb59609cd429909

Example here where the last picture received has a timestamp difference of about 4 minutes, because that's how long I was offline. Both of these screenshots were taken within a minute of each other.

device-2015-07-03-144642-clean
device-2015-07-03-144656-clean

purpleplatypus:
Fixes #3525

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -321,10 +321,7 @@ public class ConversationItem extends LinearLayout
     secureImage.setVisibility(messageRecord.isSecure() ? View.VISIBLE : View.GONE);
     bodyText.setCompoundDrawablesWithIntrinsicBounds(0, 0, messageRecord.isKeyExchange() ? R.drawable.ic_menu_login : 0, 0);
 
-    final long timestamp;
-
-    if (messageRecord.isPush()) timestamp = messageRecord.getDateSent();
-    else                        timestamp = messageRecord.getDateReceived();
+    final long timestamp = messageRecord.getTimestamp();
 
     dateText.setText(DateUtils.getExtendedRelativeTimeSpanString(getContext(), locale, timestamp));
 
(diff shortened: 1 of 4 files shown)


moxie0:
This will just invert the problem.  The deal is this: for textsecure messages, we always display "sent" time in received message timestamps.  For SMS/MMS messages, we always display "received" time in received message timestamps.  We don't do "sent" time everywhere, because the sent time is deeply unreliable in the SMS/MMS world.

So we need to do the same thing in the preview that we do in the message bubbles: display sent time for textsecure messages, and received time for sms/mms messages.

purpleplatypus:
I can't test this with MMS at the moment to ensure nothing bad happened, but I think that dfde3d3 will handle the MMS/TS divide.

purpleplatypus:
I tested this feature with MMS to make sure that everything still works, and it behaves as expected. Let me know if you want to see any test cases or something in particular checked. Otherwise, this should be okay to merge whenever it takes your fancy.

moxie0:
i still don't understand this change, the idea is to get the received date into the mediapreviewactivity, correct? isn't that what's already happening?

purpleplatypus:
The idea is to get the *sent* date into the activity when it's available. That's the whole point, mediapreviewactivity uses the received date, while the conversation thread uses the sent date. It's inconsistent and the inconsistency will confuse users because it's really in their face in the UI (see pictures in #3525).


MediaPreviewActivity is not bothering to check what sort of message it was getting the timestamp for, which is inconsistent with the rest of the code. It should only use the received date if the message was MMS/SMS.

haffenloher:
:+1: Could you also take care of the [MediaPreviewActivity started from "All Images"](https://github.com/WhisperSystems/Signal-Android/blob/b0137c08cbd31507a367cab042f071df5630815a/src/org/thoughtcrime/securesms/ImageMediaAdapter.java#L89)? Same problem, it shows the received instead of the sent date for push messages.

moxie0:
in 3.8.0

PR #3543, opened 2015-07-07
Extend DPAD support. Add softkey support for compose.

Clam-:
Made the UI operate a bit more friendly under DPAD usage.
Specifically navigating the conversation list, entering the country code selecter with DPAD_CENTER, and navigating the import/export options.
Also added right softkey shortcut for compose.

I couldn't figure out why the DPAD doesn't work for the menu popup on a 2.3.3 device. (DPAD works on newer Android versions in the popup menu though.)


rhodey:
I've got a motorola droid with DPAD but don't have any modern devices with this feature, will try this out today :)

rhodey:
I've got a motorola without the large DPAD, there are arrow keys on the keyboard and I wonder if these are treated like a DPAD droid

I'll see if this counts as a DPAD and if so look into it today. unfortunately I have no other devices with the DPAD.

rhodey:
I verified that the arrow keys on the droid 3 keyboard are treated as a DPAD by logging all key codes then tested all the UIs affected by this PR. code looks great and UI functions as expected :)

note: I don't have a device with the `KEYCODE_SOFT_RIGHT` feature (or wan't able to find it) so I had to test that change with `adb shell input keyevent 2`

Clam-:
Cool. I initially tested on a junk device using MonkeyRunner, which is pretty neat but I wish I had of known about ```adb shell input```! 
Thanks. Will come in handy for the future.

moxie0:
I'm not into the right compose shortcut.

Clam-:
I had a feeling you might not be.
My logic was that the compose button was on the right side of the screen, so assigning it to the right softkey would seem intuitive. The only other way to access it would be to scroll down to the end of the conversation list, which is a bit of a bummer. Alternatively letting the compose UI element gain focus via something like ```android:nextFocusRight``` from the conversation list might be more appealing.
Using both would be nice, but I understand if you want to ditch the right soft key assignment.

moxie0:
yeah i'd like to ditch it

Clam-:
I removed the compose shortcut and added navigability from the list items to the fab using DPAD right.
What do you think?

Clam-:
By the way, I don't fully understand the contribution system yet, but does that mean even a tiny pull request like this that isn't related to a bug report needs to have "FREEBE" added to a commit message to not take contributor funds?

agrajaghh:
@Clam- Every commit without "FREEBIE" receives Bitcoins. Have a look at https://github.com/WhisperSystems/BitHub

Clam-:
I see, let me redo this pull request then. Thanks for clarifying.

PR #3552, opened 2015-07-07
Use matching status bar color during action mode

veeti:
Screenshot

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -19,8 +19,11 @@ package org.thoughtcrime.securesms;
 import android.app.ProgressDialog;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.content.res.TypedArray;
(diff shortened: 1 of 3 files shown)


moxie0:
@veeti returns!

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -70,6 +72,7 @@
   private FloatingActionButton fab;
   private Locale               locale;
   private String               queryFilter  = "";
+  private int                  statusBarColor;
```
moxie0: I don't love the idea of maintaining this state here. If there's no other way to "remember," we should break the ActionMode.Callback out into an inner class so that we can correctly scope the state.

veeti:
How's this?

moxie0:
Interesting, that works?  The android attribute somehow gets resolved to a tinted version of our primary color?

veeti:
Yes, the tinted color comes from the `colorPrimaryDark` theme attribute which `statusBarColor` refers to by default.

PR #3580, opened 2015-07-09
add custom color to MessageDetailsActivity

### LINKED ISSUE #3560: Colorize Message Details
2.21.0

Dark theme, private conversation:
dirty harry

----------
Light theme, private conversation:
dirty harry

agrajaghh:
Fixes #3560

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -1,3 +1,19 @@
+/**
+ * Copyright (C) 2015 Whisper Systems
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
 package org.thoughtcrime.securesms;
 
 import android.app.Activity;
@@ -283,6 +299,7 @@ public class ConversationFragment extends Fragment
     intent.putExtra(MessageDetailsActivity.MASTER_SECRET_EXTRA, masterSecret);
     intent.putExtra(MessageDetailsActivity.MESSAGE_ID_EXTRA, message.getId());
     intent.putExtra(MessageDetailsActivity.TYPE_EXTRA, message.isMms() ? MmsSmsDatabase.MMS_TRANSPORT : MmsSmsDatabase.SMS_TRANSPORT);
+    intent.putExtra(MessageDetailsActivity.RECIPIENTS_IDS_EXTRA, recipients.getIds());
     startActivity(intent);
   }
 
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -543,6 +543,7 @@ public class ConversationItem extends LinearLayout
         intent.putExtra(MessageDetailsActivity.MESSAGE_ID_EXTRA, messageRecord.getId());
         intent.putExtra(MessageDetailsActivity.TYPE_EXTRA, messageRecord.isMms() ? MmsSmsDatabase.MMS_TRANSPORT : MmsSmsDatabase.SMS_TRANSPORT);
         intent.putExtra(MessageDetailsActivity.IS_PUSH_GROUP_EXTRA, groupThread && messageRecord.isPush());
+        intent.putExtra(MessageDetailsActivity.RECIPIENTS_IDS_EXTRA, messageRecord.getRecipients().getIds());
         context.startActivity(intent);
       } else if (!messageRecord.isOutgoing() && messageRecord.isIdentityMismatchFailure()) {
         handleApproveIdentity();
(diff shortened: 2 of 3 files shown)


src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -466,6 +467,7 @@ public void onClick(View v) {
         intent.putExtra(MessageDetailsActivity.MESSAGE_ID_EXTRA, messageRecord.getId());
         intent.putExtra(MessageDetailsActivity.TYPE_EXTRA, messageRecord.isMms() ? MmsSmsDatabase.MMS_TRANSPORT : MmsSmsDatabase.SMS_TRANSPORT);
         intent.putExtra(MessageDetailsActivity.IS_PUSH_GROUP_EXTRA, groupThread && pushDestination);
+        intent.putExtra(MessageDetailsActivity.COLOR_EXTRA, groupThread ? MaterialColor.GROUP : messageRecord.getRecipients().getColor());
```
moxie0: This check for a group thread shouldn't be necessary?

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -244,6 +244,7 @@ private void handleDisplayDetails(MessageRecord message) {
     intent.putExtra(MessageDetailsActivity.MASTER_SECRET_EXTRA, masterSecret);
     intent.putExtra(MessageDetailsActivity.MESSAGE_ID_EXTRA, message.getId());
     intent.putExtra(MessageDetailsActivity.TYPE_EXTRA, message.isMms() ? MmsSmsDatabase.MMS_TRANSPORT : MmsSmsDatabase.SMS_TRANSPORT);
+    intent.putExtra(MessageDetailsActivity.COLOR_EXTRA, recipients.getColor());
```
moxie0: Better to pass the recipients ids and let the messagedetailsactivity resolve the color.  It's asynchronously populated, so it could still technically be loading at call time.

agrajaghh: I tried setting the color in `onPostExecute()` but that's too late. You'll see blue first, and this flickering isn't nice. Or where exactly did you mean?

https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/MessageDetailsActivity.java#L288

src/org/thoughtcrime/securesms/ConversationItem.java
agrajaghh: I tried it first without, but for a failed group message `ContactColors.UNKNOWN_COLOR`is returned here: https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/recipients/Recipient.java#L104

src/org/thoughtcrime/securesms/ConversationFragment.java
moxie0: right now in onCreate you're taking the color that was passed in.  instead, pass a recipient id to MessageDetailsActivity, which it uses in onCreate to resolve a Recipients object.  You can then take the color out of the Recipients object directly, and set a listener for recipients changes in case the color had not yet resolved.

agrajaghh:
@moxie0 I'm passing the recipient id now and resolve the color in `MessageDetailsActivity`. Are you sure the listener is necessary? I tried putting `RecipientFactory.clearCache()` before `RecipientFactory.getRecipientForId` but it is still working. 

Using the listener might be a problem, since the callback is not running on the main thread, so its not possible to set the ActionBar and NotificationBar color...

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -283,6 +299,7 @@ private void handleDisplayDetails(MessageRecord message) {
     intent.putExtra(MessageDetailsActivity.MASTER_SECRET_EXTRA, masterSecret);
     intent.putExtra(MessageDetailsActivity.MESSAGE_ID_EXTRA, message.getId());
     intent.putExtra(MessageDetailsActivity.TYPE_EXTRA, message.isMms() ? MmsSmsDatabase.MMS_TRANSPORT : MmsSmsDatabase.SMS_TRANSPORT);
+    intent.putExtra(MessageDetailsActivity.RECIPIENT_ID_EXTRA, recipients.getPrimaryRecipient().getRecipientId());
```
moxie0: I think this can just be recipients.getIds()

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -543,6 +543,7 @@ public void onClick(View v) {
         intent.putExtra(MessageDetailsActivity.MESSAGE_ID_EXTRA, messageRecord.getId());
         intent.putExtra(MessageDetailsActivity.TYPE_EXTRA, messageRecord.isMms() ? MmsSmsDatabase.MMS_TRANSPORT : MmsSmsDatabase.SMS_TRANSPORT);
         intent.putExtra(MessageDetailsActivity.IS_PUSH_GROUP_EXTRA, groupThread && messageRecord.isPush());
+        intent.putExtra(MessageDetailsActivity.RECIPIENT_ID_EXTRA, messageRecord.getRecipients().getPrimaryRecipient().getRecipientId());
```
moxie0: If the idea is that we want the MessageDetailsActivity to be colored the same way was the ConversationActivity, then I think this should be conversationRecipients.getIds().

If the idea is that we want the MessageDetailsActivity to be colored the same way as the indiividual bubble color of the message, then this is correct, but the line in ConversationFragment would need to change.

moxie0:
And yeah please set the listener. You can use Util.runOnMain to handle the callback.

agrajaghh:
done! And you're right (ofc :smirk:), `recipients.getIds()` makes more sense!

a bit offtopic: Is it on purpose that the color of the NotificationBar is changing to black when a message gets selected? 

its currently like that:
screenshot_17-50-11 - kopie

maybe we should leave the NotificationBar colored like that?
screenshot_17-50-11

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -543,6 +543,7 @@ public void onClick(View v) {
         intent.putExtra(MessageDetailsActivity.MESSAGE_ID_EXTRA, messageRecord.getId());
         intent.putExtra(MessageDetailsActivity.TYPE_EXTRA, messageRecord.isMms() ? MmsSmsDatabase.MMS_TRANSPORT : MmsSmsDatabase.SMS_TRANSPORT);
         intent.putExtra(MessageDetailsActivity.IS_PUSH_GROUP_EXTRA, groupThread && messageRecord.isPush());
+        intent.putExtra(MessageDetailsActivity.RECIPIENTS_IDS_EXTRA, messageRecord.getRecipients().getIds());
```
moxie0: If we want this to be the conversation color, it needs to be conversationRecipients.getIds().  If we want it to be the message bubble color, we need to change the line in ConversationFragment.

agrajaghh: using `conversationRecipients.getIds()` now, I read your previous message too fast...

PR #3591, opened 2015-07-09
finish MessageRecipientAsyncTask when context is destroyed

agrajaghh:
Did you guys forgot a return there?
`context = null` would be used otherwiese some lines below: https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/MessageDetailsActivity.java#L265


moxie0:
in 2.25.0 thanks!

PR #3615, opened 2015-07-11
Misc. code clean up tasks

ehartsuyker:
- removed unused imports
- removed unnecessary method modifieds
- replaced x.size() == 0 type comparisons with x.isEmpty()
- removed Intellij 'TODO Auto-generated constructor stub' messages
- added better resource clean up in Base64.java

### DIFF
--- a/src/org/thoughtcrime/securesms/groups/GroupMessageProcessor.java
+++ b/src/org/thoughtcrime/securesms/groups/GroupMessageProcessor.java
@@ -113,7 +113,7 @@ public class GroupMessageProcessor {
       builder.clearMembers();
     }
 
-    if (missingMembers.isEmpty()) {
+    if (!missingMembers.isEmpty()) {
       // TODO We should tell added and missing about each-other.
     }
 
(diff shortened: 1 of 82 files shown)


src/org/thoughtcrime/securesms/groups/GroupMessageProcessor.java
```diff
@@ -113,7 +113,7 @@ private static void handleGroupUpdate(Context context,
       builder.clearMembers();
     }
 
-    if (missingMembers.size() > 0) {
+    if (missingMembers.isEmpty()) {
       // TODO We should tell added and missing about each-other.
```
agrajaghh: you forgot the "!", no?

moxie0:
Thanks @ehartsuyker but we're not really into these kind of diffs.  The changes don't really have tremendous value, but can make rebasing legit pending changes much more annoying.  Cleanup during a legit change is totally welcome, but a drive by of removing missing imports can just make things harder.  Thanks again.

Issue #3565, opened 2015-07-08
thread batch selection only displays selection count after count goes above 1


rhodey:
1) have two or more threads
2) open the main conversation list view
3) long tap on any thread
4) notice the batch mode reads "batch selection mode"
5) tap another thread
6) notice the batch mode now also includes a count of the # selected
7) unselect any of the two threads
8) notice the batch mode still includes a count of the # selected

so steps 4 & 8 represent the same state but are displayed in the UI differently, yeah yeah, minor I know but noticed.

PR #3628, opened 2015-07-12
Closes issue #3565

ehartsuyker:
Correctly sets subtitle on conversation batch select mode during first selection

Edit: Since this is a thing.
// FREEBIE


mcginty:
simplified & rebased, in 3.5.0. thanks!

Issue #3627, opened 2015-07-12
can't scroll all the way up in most conversations


oaim:
Both with ts users and sms conversations. Ts 2.22, android 4.1.2

ComeGitSome:
Same problem here, the bar at the top covers the message.

2-4601:
Confirmed on Nexus 4, stock 5.1.1, TS 2.22.0.

2-4601:
There's also curious behavior when you scroll really fast up and when you reach the top it starts to scroll down instead of stopping at the top. Does not happen always. I guess you have to scroll fast enough.
This also occurs in 2.20.0.

turtlekiosk:
Is the padding (with the height of the action bar) to compensate for the window actionbar overlay mode in ConversationActivity missing?

PR #3635, opened 2015-07-12
add actionbar padding to conversation activity

turtlekiosk:
Adding top padding to the main view of ConversationActivity should compensate for the ActionBar overlay mode that ConversationActivity started using because of direct capture.

Re: Issue #3627 
// FREEBIE


mcginty:
looks good, thanks

PR #3656, opened 2015-07-13
"spirit guide" is appropriative and unnecessary

glyph:
For more details on why to avoid this language, see [this post](http://aquariusmoonrising.tumblr.com/post/58763812891/why-you-shouldnt-use-the-term-spirit-animal).  Plus, it doesn't even really make sense here; even in the bastardized pop-cultural sense, "spirit guide" or "spirit animals" would be *helpful* things and PGP is more in [this vein](http://despair.com/products/mistakes).


glyph:
Is it really necessary to sign a CLA to correct a typo? This change is not copyrightable.

PR #3664, opened 2015-07-14
Re-enable and improve direct capture

mcginty:
* Photo capture is really damn fast now that we write the captured data to a file asynchronously.
* Found and fixed a cyclic dependency that stopped the camera from working if you open and close the camera drawer like a [manic squirrel](https://www.youtube.com/watch?v=Nlvp3i3iM7M).
* Keeping external direct capture, so modified the capturing logic to support both cases.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -11,7 +11,7 @@
                 android:label="Access to TextSecure Secrets"
                 android:protectionLevel="signature" />
 
-    <!--<uses-feature android:name="android.hardware.camera" android:required="false" />-->
+    <uses-feature android:name="android.hardware.camera" android:required="false" />
     <uses-permission android:name="org.thoughtcrime.securesms.ACCESS_SECRETS"/>
     <uses-permission android:name="android.permission.READ_PROFILE"/>
     <uses-permission android:name="android.permission.WRITE_PROFILE"/>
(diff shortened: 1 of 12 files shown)


src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -127,17 +136,44 @@ public boolean isAttachmentPresent() {
...
+          setImage(masterSecret, uri, false);
+          captureUri = uri;
+        } catch (IOException | BitmapDecodingException e) {
+          Log.w(TAG, e);
+        } finally {
+          resolvingCapture = false;
+        }
+      }
+    }.execute(bitmap);
   }
```
moxie0: This way overloaded.  This class shouldn't know anything about anything other than Uris.

src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -116,10 +121,14 @@ public void setMedia(final Slide slide) {
...
 
   public void setMedia(final Slide slide, @Nullable MasterSecret masterSecret) {
+    setMedia(slide, masterSecret, true);
+  }
+
+  public void setMedia(final Slide slide, @Nullable MasterSecret masterSecret, boolean updateThumbnail) {
```
moxie0: For the caller to know whether the attachment manager should update its thumbnail or not feels pretty off.

moxie0:
Is there a reason the "capture provider" needs to write this to a file at all?

mcginty:
drafts

moxie0:
Ok right, then it seems like the CaptureProvider should be responsible for handling the memory->disk transition, but it should just be transparent (via a uri interface) to everything else.

mcginty:
CaptureProvider is now in charge of the memory -> disk transition.

The good news is this way actually turns out to be more memory-efficient once I decided to say fuck it and write a method to rotate YUV420sp image data. No more converting to a Bitmap and back before passing it along to the callback listener, goes straight from YUV -> jpg bytes.

src/org/thoughtcrime/securesms/components/camera/CameraView.java
```diff
@@ -453,20 +455,24 @@ public SerializedAsyncTask() {
...
-        @Override public void run() {
-          onPostMain(result);
-        }
-      });
+        runOnMainSync(new Runnable() {
+          @Override public void run() {
+            onPostMain(result);
+          }
+        });
+      } catch (IllegalStateException ise) {
```
moxie0: why are you catching a runtime exception?

src/org/thoughtcrime/securesms/providers/CaptureProvider.java
```diff
@@ -35,28 +47,68 @@ public static CaptureProvider getInstance(Context context) {
...
 
   public InputStream getStream(MasterSecret masterSecret, long id) throws IOException {
-    return new DecryptingPartInputStream(getFile(id), masterSecret);
+    final byte[] cached = cache.get((int)id);
+    return cached != null ? new ByteArrayInputStream(cached)
+                          : new DecryptingPartInputStream(getFile(id), masterSecret);
+  }
+
+  private int generateId(Recipients recipients) {
+    return Arrays.hashCode(recipients.getIds());
```
moxie0: might want to make this Math.abs

src/org/thoughtcrime/securesms/components/camera/CameraView.java
mcginty: It throws in onMain(). I can throw a more explicit non-runtime exception instead.

src/org/thoughtcrime/securesms/providers/CaptureProvider.java
mcginty: good call

src/org/thoughtcrime/securesms/components/camera/CameraView.java
moxie0: You should only throw a runtime exception if you don't intend for it to be caught.

mcginty: fixed

moxie0:
in 2.23.0

AndroidManifest.xml
```diff
@@ -34,47 +34,47 @@
...
+    <!-- For fixing MMS -->
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE"/>
+    <uses-permission android:name="android.permission.CHANGE_WIFI_STATE"/>
+
+    <!-- Set image as wallpaper -->
+    <uses-permission android:name="android.permission.SET_WALLPAPER"/>
+
+    <!-- Permissions from RedPhone -->
+    <uses-permission android:name="android.permission.RECORD_AUDIO" />
+    <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
```
btcdrak: Why is it necessary for a text messaging application to have access to the microphone and recording functions?

agrajaghh: redphone integration is planed...

@btcdrak have a look at #3771 
Please use the mailing list for questions.

PR #3670, opened 2015-07-15
Prevents empty conversations from appearing

ehartsuyker:
Tested with:
Phone: Moto G 2014
OS: CM 12.1-20150714-NIGHTLY-titan
Android: 5.1.1

Fixes 3581

Edit: Since this is a thing.
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -18,6 +18,7 @@ package org.thoughtcrime.securesms;
 
 import android.content.Intent;
 import android.database.ContentObserver;
+import android.database.Cursor;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.provider.ContactsContract;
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/ConversationListActivity.java
```diff
@@ -227,4 +232,28 @@ public void run() {
     getContentResolver().registerContentObserver(ContactsContract.Contacts.CONTENT_URI,
                                                  true, observer);
   }
+
+  public static void notifyOfGroupCreate() { recentGroupCreate = true; }
+
+  private void removeEmptyThreads() {
+    if (recentGroupCreate) {
+      Log.i("CONVO LIST", "trimming threads", new Exception());
```
agrajaghh: use `TAG` instead of `"CONVO LIST"`

moxie0:
thanks but we shouldn't really be going through elaborate means to delete a side effect, we should just fix the side effect from occurring to begin with

Issue #3604, opened 2015-07-10
Emoji aren't rendered in generated avatar


mcginty:
![img_20150710_130357](https://cloud.githubusercontent.com/assets/373823/8628041/42f92a64-2704-11e5-9360-0fac4040109c.JPG)

agrajaghh:
fixed with https://github.com/WhisperSystems/Signal-Android/commit/67415a8dcbeff4ab3b52757400fe09fb015cca5c

PR #3617, opened 2015-07-11
Improve generated contact photos for names beginning with emoji

danbartram:
Improve the handling of contact names within the `GeneratedContactPhoto` class so that names beginning with emojis or non-language symbols (for any language, not just English) display an appropriate character instead of the missing unicode icon.

This should fix issue #3604.

PR #3684, opened 2015-07-17
Improve generated contact photos with emoji names

danbartram:
*This is a re-submission of pull #3617 as this code should be more readable.*

----

### Issue
This commit fixes Issue #3604.

Contact names which begin with emoji characters cause their generated photo to consist of the "question-mark-in-a-box" UTF-8 character.

This commit fixes the issue by using a new function to strip out any non-language (not just English) characters from their contact name, and instead use the first character of the returned string as their contact photo initial.

----

##### `String getCharacter(String name)`

`(String) name` is the name of the contact that requires their photo to be generated.

This function will return the first character of their name that should be suitable for use in the generated contact photo. Any whitespace, emojis, or non-language characters should be removed from their name and leave only characters suitable for display in their contact photo.

### DIFF
--- a/src/org/thoughtcrime/securesms/contacts/avatars/GeneratedContactPhoto.java
+++ b/src/org/thoughtcrime/securesms/contacts/avatars/GeneratedContactPhoto.java
@@ -33,6 +33,16 @@ public class GeneratedContactPhoto implements ContactPhoto {
                        .height(targetSize)
                        .textColor(inverted ? color : Color.WHITE)
                        .endConfig()
-                       .buildRound(String.valueOf(name.charAt(0)), inverted ? Color.WHITE : color);
+                       .buildRound(getCharacter(name), inverted ? Color.WHITE : color);
+  }
+
+  private String getCharacter(String name) {
+    String cleanedName = name.replaceAll("[^\\p{L}\\p{Nd}]+", "");
+
+    if (cleanedName.isEmpty()) {
+      return "#";
+    } else {
+      return String.valueOf(cleanedName.charAt(0));
+    }
   }
 }


mcginty:
Wouldn't this also exclude punctuation and potentially other displayable characters? Also, has this been tested on a unicode-only alphabet like Chinese to make sure the regex works (I don't remember Java's regex unicode quirkiness)?

danbartram:
This pull request works fine with the Chinese characters I tested.

Chinese character contact photo

However, punctuation and some other characters are removed by choice. I assumed the desire would be to force an alphabet letter to be the first character, but I understand that it might be better to instead allow punctuation and symbols such as currencies (`$`, `£`).

This would use the new regex: `[^\p{L}\p{Nd}\p{P}\p{S}]+` and you can test this on [regex101.com](https://regex101.com/r/nW4oH1/1).

Copying from regex101's explanations for the regex:
```
\p{L} matches any kind of letter from any language
\p{Nd} matches a digit zero through nine in any script except ideographic scripts
\p{P} matches any kind of punctuation character
\p{S} matches math symbols, currency signs, dingbats, box-drawing characters, etc
```

I'm not totally convinced that all of the characters in the `\p{S}` collection will be displayable, but it does allow for some definitely displayable characters to be used too (e.g. the currency symbols).

I'm not sure how you feel about having a regex thrown into the codebase either, and I didn't comment it in this PR as it's fairly "simple" and could be looked up easily online. As well as trying to keep the (potentially unnecessary) comments to a minimum to keep the consistent code style.

Would it be better to include a code comment that briefly summarizes what the regex does?

Would it preferable to use the `[^\p{L}\p{Nd}\p{P}\p{S}]+` pattern instead of the current pull's version - and if so, should it be submitted as a new PR or as a new commit in the same branch?

src/org/thoughtcrime/securesms/contacts/avatars/GeneratedContactPhoto.java
```diff
@@ -33,6 +33,16 @@ public Drawable asDrawable(Context context, int color, boolean inverted) {
                        .height(targetSize)
                        .textColor(inverted ? color : Color.WHITE)
                        .endConfig()
-                       .buildRound(String.valueOf(name.charAt(0)), inverted ? Color.WHITE : color);
+                       .buildRound(getCharacter(name), inverted ? Color.WHITE : color);
+  }
+
+  private String getCharacter(String name) {
+    String cleanedName = name.replaceAll("[^\\p{L}\\p{Nd}]+", "");
```
mcginty: wouldn't replaceFirst work?

mcginty:
this kind of filtering sounds reasonable, especially if we use the slightly more liberal `[^\p{L}\p{Nd}\p{P}\p{S}]+` regex.

@moxie0 thoughts?

moxie0:
seems reasonable

mcginty:
rebased, made slight tweaks, in 3.6.0. Thanks @danbartram for the well-formatted PR and regex references!

PR #3688, opened 2015-07-17
fix missing EmojiTextViews

### LINKED ISSUE #3676: Some views are still using previous emoji set
(Since [the last issue](https://github.com/WhisperSystems/TextSecure/issues/3662) was kidnapped I made this.)

Per-recipient preferences are using the old emoji set.
Screenshot from group preferences:
screenshot_2015-07-14-22-48-27

mcginty:
fixes #3676

### DIFF
--- a/res/layout/contact_selection_list_item.xml
+++ b/res/layout/contact_selection_list_item.xml
@@ -48,7 +48,8 @@
 
     </LinearLayout>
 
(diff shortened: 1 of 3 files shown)


McLoo:
As I can't post in #3676:
Group name and edit-group-actionbar you missed - to late recognized I this

PR #3701, opened 2015-07-18
Transition in RecipientActivity and hide keyboard on ContactSelection (#3523, #3704 and #3706)

### LINKED ISSUE #3523: Missing transition from recipient prefs back to convo
Only when using the ~~"hardware"~~ soft key back button the transition is visible.
Most likely because the action bar back uses a different mode (activity is pulled down)?!

### LINKED ISSUE #3704: Disable keyboard popup on opening the new contact view
I think you should disable the keyboard popup on opening the new contact view as it covers many contacts and many people don't like to search in the first place.

### LINKED ISSUE #3706: Keyboard toggle in contacts view does not open keyboard
Button works for opened keyboard 
But if keyboard is closed, you need to open manually. 
The button just switched is icon

mammuth:
##### This Pullrequest contains 3 very small and trivial fixes: ####
- RecipientActivity: Use the transition that happens on onBackPressed() in onOptionsItemSelected()

As a result the transition moving the contact name in the actionbar of the conversation view is applied not only on a backbutton press, but also when pressing the home-arrow in the actionbar.
Should fix #3523

- ContactSelection: Hide the keyboard when user wants to send a new message and the ContactSelection view is opened.

Most people probably don't want to search for their contacts at the first, but immediately click on them in  their contact list, as stated in the corresponding issue.
Fixes #3704

- ContactSelection: Pressing the toggle button now brings up the keyboard.

Furthermore I changed the keyboard as the default for the toggle button instead of the dial pad, because people would like to input real names instead of phone numbers most of the times and since the keyboard ist hidden by default the keyboard should appear first when clicking the toggle button.
Fixes #3706

All commits are FREEBIES.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -202,6 +202,7 @@
 
     <activity android:name=".NewConversationActivity"
               android:theme="@style/TextSecure.LightNoActionBar"
+              android:windowSoftInputMode="stateHidden"
               android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"/>
 
     <activity android:name=".PushContactSelectionActivity"
--- a/res/layout/contact_selection_activity.xml
+++ b/res/layout/contact_selection_activity.xml
@@ -45,16 +45,16 @@
                     android:layout_gravity="center"
                     android:gravity="center">
 
-                <ImageView android:id="@+id/search_dialpad"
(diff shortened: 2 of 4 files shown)


moxie0:
Thanks.  In general we prefer keeping changes to separate functionality in separate PRs, I merged your first commit and left the other two.

PR #3730, opened 2015-07-20
only create DB readers as needed

mcginty:
Another thing I noticed during the method tracing from scrolling in the conversation.

No matter what the item type, we create both an MmsDatabase reader and an SmsDatabase reader when almost always only one is used. Each of those readers create their own instance of MasterCipher, and each MasterCipher calls Cipher.getInstance() in its constructor which is not a particularly fast call.

I didn't want to edit MasterCipher (although @moxie0, might be worth looking into initializing those ciphers statically?), but this should help lower our object creation during scrolling on the UI thread. Lowers jank.


moxie0:
is this related to a 2.23.1 issue?

mcginty:
nope

moxie0:
in 2.25.0

PR #3735, opened 2015-07-21
fix enter send behaviour

### LINKED ISSUE #3725: "Send TextSecure Message" inside enter button too long (Google Keyboard)
With "enter key sends" enabled, the enter key of the google keyboard displays "Send TextSecure message" which looks really weird. I'm not sure if we can fix this without changing the landscape send button..

screenshot_16-04-06

Tested on a N4 with 5.1.1 and TS 2.23.0

mcginty:
fixes #3725

I could not find a way to make the weird-looking text go away on Google Keyboard in landscape mode. When you set the label, there doesn't appear to be a way to have that not hang around. It's either that or we can just have enter not be send in landscape mode. Which I think would also be acceptable.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -245,7 +245,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     initializeSecurity();
     initializeEnabledCheck();
     initializeMmsEnabledCheck();
-    updateIme();
+    composeText.setTransport(sendButton.getSelectedTransport());
 
     titleView.setTitle(recipients);
     setActionBarColor(recipients.getColor());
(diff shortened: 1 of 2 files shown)


moxie0:
Yeah I think no enter to send in landscape mode makes sense

mcginty:
updated

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -758,23 +761,25 @@ protected void onPostExecute(Boolean isMmsEnabled) {
...
-        if (actionId == EditorInfo.IME_ACTION_SEND) {
-          sendMessage();
-          return true;
-        }
-        return false;
-      }
-    });
+
+    composeText.setInputType(inputType);
+    composeText.setImeOptions(imeOptions);
```
moxie0: Can we put the details in ComposeText and leave the logic out here?

mcginty:
oh yeah that is nicer. updated.

PR #3757, opened 2015-07-22
Fix #3739: Back press always goes to conversation list

### LINKED ISSUE #3739: Pressing back in Conversation doesn't return to conversation list, but just back to the previews app/launcher
#### Reproduce:
- Remove all apps from Androids *"Recent Apps"*
- TS message arrives, notification appears
- Open the conversation by clicking on the notification
- read/answer/whatever and return with the back key

#### Result: 
- Instead of returning to the main view of TS, as expected, the conversation activity closes without opening the main activity
- When wanting to switch to TS using *"Recent Apps"* at any later time the conversion gets opened, also with no ability to come back to the main view

#### Expected:
- Always show the conversation list when pressing back in a conversation

Or is this behavior intentional, because *"well...thats how Android works"*?
[It's not, how Android works](https://developer.android.com/training/implementing-navigation/temporal.html#SynthesizeBackStack) ;-)

mammuth:
fixes #3739

Before this change, pressing the back button would not have returned to the conversation list, as expected, when the conversation list hasn't been in the activity stack. 

This happend when the app has not been running in background, but a conversation is opened through an incoming notification.

Now it's the by Google defined behavior for this scenario: https://developer.android.com/training/implementing-navigation/temporal.html#SynthesizeBackStack

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -392,7 +392,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     } else if (quickAttachmentDrawer.isOpen()) {
       quickAttachmentDrawer.close();
     } else {
-      super.onBackPressed();
+      handleReturnToConversationList();
     }
   }
 


mammuth:
Closing since obviously not wanted.

PR #3758, opened 2015-07-22
👀 the 💥😐

### LINKED ISSUE #3753: All emoji message bubbles can display only one line worth of emojis
1. Send or receive a message that has only emojis and is so long that it does not fit into a single message bubble line
2. TS will display only _e_ number of emojis where _e_ is _the number of emojis sent/received_ `mod` _the number of emojis that fit in one message bubble line_

E.g. If I send 14 cat emojis on my Huawei device which can have a maximum of 12 cat emojis per sent message bubble line then TS will only show 2 (_= 14_ `mod` _12_) cat emojis in the sent message bubble:

emojis one line

TS 2.23.1
Huawei U8800, 4.0.4 (pictured on the left)
Nexus 4, stock 5.1.1 (pictured on the right)

Related #3653

mcginty:
turns out they do want you to modify the FontMetricsInt argument they pass in to getSize(). This works and still solves the issue it was originally meant to solve.

tested on GB/LP

fixes #3753

### DIFF
--- a/src/org/thoughtcrime/securesms/components/emoji/EmojiSpan.java
+++ b/src/org/thoughtcrime/securesms/components/emoji/EmojiSpan.java
@@ -9,17 +9,28 @@ import android.widget.TextView;
 import org.thoughtcrime.securesms.R;
 
 public class EmojiSpan extends AnimatingImageSpan {
+  private final int            size;
+  private final FontMetricsInt fm;
+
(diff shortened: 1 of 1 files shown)


2-4601:
Glorious fix!
Works also on ICS now.
ics
Thanks.

Issue #3752, opened 2015-07-22
Image corruption when using direct capture


jtracey:
Whenever attempting to use the direct capture feature, after taking the picture, it shows up in the draft view (and in the sent picture) as noise. Whatever is going on is deterministic, as visually similar scenes create visually similar corrupted pictures, and visually different scenes create visually different images.

This is the draft of a picture of a grey wall:
![garbage draft](https://cloud.githubusercontent.com/assets/2998852/8819195/b042331a-2ffe-11e5-9d11-106eafd9aea7.png)
This is after it's sent to myself (though this happens to any recipient):
![garbage sent](https://cloud.githubusercontent.com/assets/2998852/8819204/ca1e4706-2ffe-11e5-9e2b-b688c6d4158b.png)
The loading-wheel-of-death is also probably a bug, but unrelated, and only happens when sending to myself, so not really a major concern.
The actual picture it sends:
![textsecure-2015-07-21-225416](https://cloud.githubusercontent.com/assets/2998852/8819218/fd496a34-2ffe-11e5-8ccc-37b3268a3f20.jpeg)
which is obviously a correctly formatted jpeg, it's the content that's wrong.

Happens on TextSecure 2.23 and 2.23.1. I'm using a Samsung Galaxy S II running Cyanogenmod 11 (which is Android 4.4). Sending photos via attachment (both selecting a picture and the "Take Photo" feature) works fine.

mcginty:
that looks like we're getting back image data from your camera in a format other than NV21. Or my NV21 rotate method is broken. Both are possibilities.

jtracey:
I just noticed: It only happens on my back-facing camera, my front-facing camera works fine.

mcginty:
currently don't have enough info to put a fix for this in 2.24.0. once #3760 is out we can get a hint about how to move forward with this.

peterpanondrugs:
S3 LTE int. Android 5.1
![tmp_17866-screenshot_2015-07-24-08-50-39 011219525220](https://cloud.githubusercontent.com/assets/8776733/8869322/9a3af7ce-31e2-11e5-90c8-0d5bcbe28211.png)
https://gist.github.com/30cd23ff29ee5aa81155

I just wanna say that superbeam has the same problem. They told me to use the full screen caption to prevent this bug. Gonna search for more informations

Great great app!

mcginty:
Thanks for the log, @peterpanondrugs! You're saying front camera doesn't work for you either?

[...]

PR #3760, opened 2015-07-22
add previewFormat logging to QuickCamera

mcginty:
to help diagnose #3752. Android's documentation says the preview will always come back in NV21 and all cameras support that format, but now I don't trust that.


moxie0:
in 2.24.0

PR #3762, opened 2015-07-22
asyncify reminderView

mcginty:
on my slow 2.3 device, this saves upwards of 1.5s of UI thread execution on every conversation list resume. Noticeably faster.


moxie0:
in 2.24.0

PR #3763, opened 2015-07-22
inflate transport popup opportunistically

mcginty:
another thing I noticed during trace when opening a conversation. ListPopupWindow inflation can take up to ~100ms, so trimming that down until a user actually needs it since most of the time it won't be necessary.


src/org/thoughtcrime/securesms/components/SendButton.java
```diff
@@ -49,8 +48,11 @@ private TransportOptions initializeTransportOptions(boolean media) {
     return transportOptions;
   }
 
-  private TransportOptionsPopup initializeTransportOptionsPopup() {
-    return new TransportOptionsPopup(getContext(), this, this);
+  private TransportOptionsPopup getTransportOptionsPopup() {
+    if (!transportOptionsPopup.isPresent()) {
+      transportOptionsPopup = Optional.of(new TransportOptionsPopup(getContext(), this, this));
+    }
```
moxie0: maybe this should be synchronized

moxie0:
in 2.24.0

PR #3764, opened 2015-07-22
warm database caches asynchronously at app start

mcginty:
Previously, CanonicalAddressDatabase was reading the entire database synchronously on the UI thread, and TextSecureDirectory was doing main thread I/O every time we checked if a number was a TextSecure user. On my slow phone, this was taking multiple seconds.

This change
* Adds a caching layer for TextSecureDirectory
* Fills the TextSecureDirectory and CanoncalAddressDatabase caches asynchronously on app start
* Asynchronously loads the libphonenumber metadata from disk by calling it's `parse()` method on our local phone number (if it exists). Otherwise, this was another ~1000ms operation on the UI thread the first time a conversation was opened.

### DIFF
--- a/src/org/thoughtcrime/securesms/database/CanonicalAddressDatabase.java
+++ b/src/org/thoughtcrime/securesms/database/CanonicalAddressDatabase.java
@@ -64,17 +64,15 @@ public class CanonicalAddressDatabase {
 
   private CanonicalAddressDatabase(Context context) {
     databaseHelper = new DatabaseHelper(context, DATABASE_NAME, null, DATABASE_VERSION);
-    fillCache();
   }
 
   public void reset(Context context) {
     DatabaseHelper old  = this.databaseHelper;
     this.databaseHelper = new DatabaseHelper(context.getApplicationContext(), DATABASE_NAME, null, DATABASE_VERSION);
     old.close();
-    fillCache();
   }
 
-  private void fillCache() {
+  public void fillCache() {
     Cursor cursor = null;
 
     try {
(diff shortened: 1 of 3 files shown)


moxie0:
I don't really want to add a cache for the TS directory, I can just make accesses async instead.

src/org/thoughtcrime/securesms/database/CanonicalAddressDatabase.java
```diff
@@ -64,17 +64,15 @@ public static CanonicalAddressDatabase getInstance(Context context) {
...
 
   public void reset(Context context) {
     DatabaseHelper old  = this.databaseHelper;
     this.databaseHelper = new DatabaseHelper(context.getApplicationContext(), DATABASE_NAME, null, DATABASE_VERSION);
     old.close();
-    fillCache();
   }
 
-  private void fillCache() {
+  public void fillCache() {
```
moxie0: I don't love the idea of this being publicly accessible.

mcginty: If we don't care about caches and want to make CanonicalAddressDatabase calls async too, we don't need this.

mcginty:
so is the verdict to hold off on touching CanonicalAddressDatabase caching until we can make it async?

even if that's the case, this PR still covers warming the libphonenumber metadata caches.

moxie0:
i'd prefer to fix cadb and the directory the right way, and not to do any cache warming unless it's the bottleneck

mcginty:
it just means the first time we canonicalize a number, an IO op will happen. Usually this is when opening the first conversation after the process starts, the lag is noticeable.

PR #3766, opened 2015-07-22
gifs

mcginty:
* Use Glide for rounding and cropping images *and* gifs off the main thread, finally cutting out our use of RoundedImageView. This was traditionally less memory efficient, but Glide's pooling and reuse of bitmaps make that not the case any more. If anything this should make scrolling faster since there's no shader activity in the ImageView's onDraw().
  * Our new RoundedCorners transformation "cheats" transparency with a hint for what the background it's being drawn on is. This way we don't need an alpha channel for our drawables so can continue using RGB_565 to save memory.
* Add support for animated gifs and a new media constraint allowing them to be up to 1MB.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,7 +44,7 @@ dependencies {
     compile 'org.w3c:smil:1.0.0'
     compile 'org.apache.httpcomponents:httpclient-android:4.3.5'
     compile 'com.github.chrisbanes.photoview:library:1.2.3'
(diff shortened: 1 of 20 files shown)


moxie0:
in 2.24.0

McLoo:
@mcginty the one MB limit works fine, except for the case when you switch to MMS and send.
There you'll get a UndeliverableMessageException cause of the dissatisfying size constraints.

*cool, gif finally - so we can use the surfing feature again :surfer:*

PR #3773, opened 2015-07-23
Update language list

McLoo:
Inspired by https://github.com/SMSSecure/SMSSecure/commit/0f77f7b6eba0681b24189d7c1c3597a3417ad11f

You can choose languages with less translation coverage already, so I added all the imported languages missing.

New:
- Belarusian 86%
- Chinese (traditional) 65% (maybe @corbett is around and could check the Chinese characters :grin:  )
- Euskara (Basque) 73%
- Korean ~~97%~~ 100% (thanks @unrulygnu !!!)
- Macedonian 95%
- Tamil 84%
- Persian 83%
- Vietnamese 50%

some reference: http://www.loc.gov/standards/iso639-2/php/English_list.php

### DIFF
--- a/res/values/arrays.xml
+++ b/res/values/arrays.xml
@@ -5,14 +5,17 @@
       <item>@string/preferences__default</item>
       <item>English</item>
       <item>Arabic العربية</item>
+      <item>Belarusian беларускі</item>
       <item>Bulgarian български</item>
       <item>Burmese မြန်မာစာ</item>
       <item>Català</item>
       <item>Čeština</item>
-      <item>Chinese 中国的</item>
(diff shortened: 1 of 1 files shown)


unrulygnu:
Korean 100% (thanks for the push!)

Also, one minor change:

    <item>Korean 한국어</item>

McLoo:
corrected translated language names for
- Korean (thanks @unrulygnu)
- Chinese (simplified)
- Belarusian
- Persian

*conclusion for translations: google over wikipedia*

moxie0:
I don't want to include new languages in the selector unless they're at 90%.  The ones that are in there now were once at 90% and have fallen behind.

PR #3776, opened 2015-07-23
add toggle to show/hide password

carbin:
This allows for the use of long, complex passwords while reducing the risk of devices being defenestrated.


agrajaghh:
didn't test it yet, but you should include mdpi, hdpi and xxhdpi icons as well

res/layout/prompt_passphrase_activity.xml
```diff
@@ -24,6 +24,17 @@
             android:layout_height="wrap_content"
             android:layout_below="@id/watermark">
 
+            <ImageButton android:id="@+id/passphrase_visibility"
```
mcginty: This looks like a job for `org.thoughtcrime.securesms.components.AnimatingToggle`!

carbin:
Updated to use the very handy AnimatingToggle

src/org/thoughtcrime/securesms/PassphrasePromptActivity.java
```diff
@@ -116,11 +122,27 @@ private void handlePassphrase() {
     }
   }
 
+  private void togglePassphraseVisibility() {
```
mcginty: since you have methods that set the icon explicitly rather than toggle, i think it makes sense for this to be a `setPassphraseVisibility(boolean visibility)` which is also explicit

src/org/thoughtcrime/securesms/PassphrasePromptActivity.java
```diff
@@ -53,6 +56,9 @@
   private DynamicLanguage   dynamicLanguage = new DynamicLanguage();
 
   private EditText passphraseText;
+  private ImageButton showButton;
+  private ImageButton hideButton;
+  private AnimatingToggle visToggle;
```
mcginty: align these instance variables (and in Java fashion, abbreviation is discouraged so `visibilityToggle` is preferable)

carbin:
Updated based on @mcginty's feedback

moxie0:
the code looks good but I don't think the UI looks quite right.  the control isn't vertically aligned correctly, should probably be on the right, and should probably be within the edittext itself.  if you're up for taking a crack at that, post a screenshot with your results.  thanks!

McLoo:
Would also be very useful on PassphraseChangeActivity 
Preventing you from inserting wrong passhrases, that lock everything down

(yes it's possible to enter a new passphrase wrong (not what you wanted) 2 times  :disappointed: )

agrajaghh:
maybe even bundle this into a reusable component? not sure how often this is needed/if it makes sense...

McLoo:
some more input:
https://github.com/iTwenty/Android-PaswordEditText

carbin:
@moxie0 something like this?

device-2015-08-02-232351
device-2015-08-02-232251

moxie0:
Yeah that looks better.  Seems like it might need a little bit of padding, and I think the grey might look better than the black resource.

moxie0:
@carbin When you make incremental changes on a PR we prefer it if you push new commits rather than squashing and force pushing, so that we can see what changed in your update.  Also could you post an updated screenshot with your padding (?) changes?

mcginty:
looks good!

fixed up the padding and some whitespace, shrunk new assets, and put into 3.5.0

PR #3786, opened 2015-07-24
smooth transition for custom keyboards and softkey

### LINKED ISSUE #3780: 2.23.2 higher memory usage
https://gist.github.com/545b3e8b12d586446a18
Memory usage seems noticeably higher now. Was @ 111M at one point. Not doing anything too special. Mostly TS contacts.

mcginty:
no more twitching out when hitting the direct capture button when soft keyboard or emoji keyboard is open.

in the process, I also simplified the view hierarchy for the attachment section of ConversationActivity and made KeyboardAwareLinearLayout a bit more efficient by not doing resource reflection every onMeasure.

Made an effort to simplify some of the QuickAttachmentDrawer math a bit too.

strangely, this refactor seems to fix #3780. could reproduce on 2.24.0, cannot on this branch.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -37,7 +37,6 @@ import android.provider.ContactsContract;
 import android.support.annotation.NonNull;
 import android.support.v4.view.WindowCompat;
 import android.text.Editable;
-import android.text.InputType;
 import android.text.TextWatcher;
 import android.util.Log;
 import android.view.KeyEvent;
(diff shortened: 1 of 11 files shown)


McLoo:
> ~~no more~~ less twitching out 

sorry. same steps: enter convo go landscape go portrait. go emoji. go keybord (redraw). go large emoji.
device-2015-07-24-233603


>  this refactor seems to fix #3780. 

Does so for me (no more hangs). Memory is rising on every rotate action though (faster when SwiftKey is opened)
Just in case this helps (and is not normal behavior) the memory history 
unbenannt (the image above is with SwiftKey oepn all the time)
unbenannt2
unbenannt3

mcginty:
@McLoo this doesn't fix the big emoji drawer problem, i'm putting out a separate PR for that

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -886,6 +886,20 @@ protected void hideEmojiDrawer(boolean expectingKeyboard) {
...
+        quickAttachmentDrawer.close();
+      } else {
+        container.postOnKeyboardOpen(new Runnable() {
+          @Override public void run() {
+            quickAttachmentDrawer.closeInstant();
+          }
+        });
+      }
+    }
+  }
```
moxie0: this feels to me like it should be encapsulated somewhere else

moxie0:
Maybe I'm wrong, but my impression is that there's a "thing" that handles input, which is either a soft keyboard, emoji keyboard, or quick attachment capture. It sort of seems like you can encapsulate all the logic pertaining to switching between those states in that thing.

mcginty:
Yeah, I'm planning on refactoring this to have an InputManager that handles transitions between our now-various virtual and OS-level input views. Began work on that but wanted to put this out first.

mcginty:
updated

moxie0:
in 2.24.0

Issue #3658, opened 2015-07-14
Message Length Limits Are Not Enforced


purpleplatypus:
On Android 4.4.4, TextSecure version 2.21.0 (Sony Xperia Tablet Compact Z3) I found that pasting an extremely long draft message (I was inspired by #3388) into the new message box will bring up the max message length indicator without actually enforcing it. Hence we can get negative characters remaining as shown below.

I was still able to send this message without issue to a Signal app; interestingly it causes Signal to crash while trying to load the message.

I'm assuming that the desired behavior is to actually enforce message length restrictions, so I will look into this.

Steps to Reproduce:
1. Write a really long draft message until you get past 2000 characters.
2. Profit.

![device-2015-07-13-220630](https://cloud.githubusercontent.com/assets/3969963/8665697/8572e574-29ac-11e5-920e-c0d37a00a535.gif)

oaim:
Have a look at https://github.com/WhisperSystems/TextSecure/issues/3362

junitas:
I noticed this as well when playing with a fix for #3388. 2000 chars seems to be like the speed of light, you can approach it, and can exceed it, you just can't ever match it.

ehartsuyker:
Hey. I've mostly solved this here (or so I think):
https://github.com/ehartsuyker/TextSecure/commit/2e3907940457591d3312ad60c1344d2b8efd81c2

I have two smaller things I can't figure out and could use some assitance.

First, is there an image that signifies that a message can't be sent? Right now I'm reusing the `ic_menu_remove_holo_light` which isn't a great idea.

Second, I can't seem to figure out where the icon for the standalone paperclip for tha attachment is controlled. I can get sending being disabled and the `X` from `ic_menu_remove_holo_light` to be displayed, but behind (in front of) the `X` the paperclip is still visible.

agrajaghh:
closing in favor of #5146

Issue #3362, opened 2015-06-07
block typing if max length exceeded


ghost:
When typing a really long message, textsecure used to block the typing when I reached the max length limit. It just happened to me that I exceeded that limit (without noticing) and sent a message that was automatically trimmed while being sent. Please notify the user that the message will be trimmed, or prevent the typing in the first place.
Don't know in which version this problem arose

oaim:
I'm working on a fix

McLoo:
This can easily be reproduce with pasting rather than typing. Paste 3000, get -1000 left and free typing, until something happens to min integer being -2^31 😁

![screenshot_2015-06-26-22-43-43](https://cloud.githubusercontent.com/assets/5326916/8387355/24c3102a-1c58-11e5-8f4b-b704d209a544.png)

*why can't I figure out such  things on a slot machine or cash dispenser? *

ehartsuyker:
If anyone's watching this thread, I have a partial fix [here](https://github.com/WhisperSystems/TextSecure/issues/3658) and could use some help from a better Android dev.

agrajaghh:
closing in favor of #5146

PR #3793, opened 2015-07-25
Block sending on oversized messages

ehartsuyker:
Fixes:
#3658
#3362

Tested on:
OS: CM 12.1-20150716-NIGHTLY-titan
Android: 5.1.1
Phone: 2014 Moto G

Edit: Since this is a thing.
// FREEBIE


turtlekiosk:
Don't use holo icons when you can use material icons.

ehartsuyker:
Can you link me to something explaining that, because I have no idea what
that means or why it should be that way.
Am 25.07.2015 23:25 schrieb "Calvin Hu" <notifications@github.com>:

agrajaghh:
the material design `ic_clear` is a bit thicker than the holo icon you used:
https://github.com/google/material-design-icons/blob/master/content/drawable-xxhdpi/ic_clear_black_48dp.png

turtlekiosk:
@ehartsuyker [Material Design](https://www.google.com/design/spec/material-design/introduction.html) was introduced with Android 5.0 Lollipop and replaces the Holo style. You can find a repo with the Material Design stock icons [here](https://github.com/google/material-design-icons) (although @agrajaghh beat me to it).

ehartsuyker:
@turtlekiosk @agrajaghh Thanks for the info. If the maintainers like the rest of the pull I'll swap it out.

moxie0:
is it not possible to replace all this code with one line "android:maxLength" on the compose edittext?

ehartsuyker:
@moxie0 I thought it made more sense to do it this way that way someone pasting a large amount of text wouldn't unknowlingly send a chopped message. Or someone could type over the limit, see they can't send, and remove text from the beginnging of the field. Also, if I understand it correctly, setting maxLength would prevent someone from writin text that is 20 text messages long (20 * 150 > 2000).

moxie0:
This is a lot of code for a relatively minor outcome, I'd like to start with the simplest thing that could possibly work.

ehartsuyker:
@moxie0 Now solved with less code.

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -793,6 +794,7 @@ private void initializeViews() {
     });
 
     composeText.setOnEditorActionListener(sendButtonListener);
+    composeText.setFilters(new InputFilter[]{new InputFilter.LengthFilter(2000)});
```
mcginty: to bring this diff down from two lines to one: think this can be done in XML with the `android:maxLength` attr :P

ehartsuyker: I tried adding it to the ComposeText element in conversation_activity.xml,
but it didn't work.
Am 03.08.2015 23:03 schrieb "Jake McGinty" <notifications@github.com>:

moxie0: 2000 should probably be a constant

ehartsuyker:
@moxie0 Updated.

moxie0:
right but I think we want that constant to be the same constant that's used in other places the limit is enforced.

ehartsuyker:
With that in mind, the `MmsCharacterCalculator` has a max of 5000, so when a user switches to push, if we had a filter that was set when the messages change, it would truncate message down to 2000. This seems like bad behavior for the end user, so should the Mms max length be reduced to 2000?

ehartsuyker:
@moxie0 So on further investigation my assumption above was wrong. If I set the mode to SMS and past > 2000 characters (with a filter that allows > 2000 on the ComposeText), then switch back to push mode and set the filter to 2000, the send button remains active and if a user hits the delete key, the lose everything back to 2000 chars. This seems non-intuitive. I think the original code will be better for end users.

moxie0:
just cut off the message and toast that it was truncated

moxie0:
Closing for lack of activity. Please reopen if you want to do the simple thing with a toast on truncate.

Issue #3788, opened 2015-07-25
Group name capitalization


dcherian:
For some reason, the group name input field doesn't auto-capitalize the first letter. The "Add Recpient" field on the other hand does auto-capitalize.

To reproduce: Menu -> New Group -> Start typing group name

dcherian:
Should be closed since the pull request was rejected.

PR #3804, opened 2015-07-26
Added textCapSentences to group name

ehartsuyker:
Fixes:
#3788

Y'know, if y'all feel inclined to make this design change.

Edit: Since this is a thing.
// FREEBIE

### DIFF
--- a/res/layout/group_create_activity.xml
+++ b/res/layout/group_create_activity.xml
@@ -36,7 +36,7 @@
                 android:layout_gravity="center_vertical"
                 android:padding="10dp"
                 android:lines="1"
-                android:inputType="textAutoCorrect"
+                android:inputType="textAutoCorrect|textCapWords"
                 android:hint="@string/GroupCreateActivity_group_name_hint" />
             </LinearLayout>
 


samlanning:
FREEBIE needs to be in the commit message not the PR =)

res/layout/group_create_activity.xml
```diff
@@ -36,7 +36,7 @@
                 android:layout_gravity="center_vertical"
                 android:padding="10dp"
                 android:lines="1"
-                android:inputType="textAutoCorrect"
+                android:inputType="textAutoCorrect|textCapWords"
```
brumsoel: I think this was meant to be `textCapSentences`, not `textCapWords` ;)

ehartsuyker: But doesn't "Friends From College" seem like the better formatting than "Friends from college" since it is a title?

agrajaghh: I guess this is language dependend. `textCapWords` looks definitely weird in some languages...

ehartsuyker: I guess someone who has localized an app before might want to weigh in on
what's the best practice. Unless you have and you're saying the standard
thing to do is not use `textCapWords`.

2015-07-27 9:08 GMT+02:00 agrajaghh <notifications@github.com>:

moxie0:
thanks but i don't think we're going to do this

PR #3806, opened 2015-07-26
Added white background to indicator for BLOCKED indicator

### LINKED ISSUE #3524: 'BLOCKED' hard to see on some background colors
E.g. dark theme and _B71C1C_:
kevin j anderson sucks

Suggestions:
- change the font color of _BLOCKED_ to white
- create a white rectangle around it as a background (as is done with the initial avatar bubble)
- create a red rectangle around it as a background and change font color to white (as is done with the _VERIFY_ button when user's key has changed)

Also red does not go that well with the different shades of blue and green.

TS 2.21.0

samlanning:
Also modified text colour in light theme to nicer red

fixes #3524


mcginty:
nice that looks good!

PR #3807, opened 2015-07-26
Make long pressing the attach button open the transport popup

samlanning:
It can be quite frustrating (and confusing) that you can only change transport with long pressing after adding an attachment / writing text. This adds a listener to the attach button to allow for long pressing that too.

### DIFF
--- a/src/org/thoughtcrime/securesms/components/SendButton.java
+++ b/src/org/thoughtcrime/securesms/components/SendButton.java
@@ -93,6 +93,13 @@ public class SendButton extends ImageButton
 
   @Override
   public boolean onLongClick(View v) {
+    return showPopupIfNeeded();
+  }
+
+  /**
+   * @return true iff the popup was opened
+   */
+  public boolean showPopupIfNeeded(){
     if (transportOptions.getEnabledTransports().size() > 1) {
       getTransportOptionsPopup().display(transportOptions.getEnabledTransports());
       return true;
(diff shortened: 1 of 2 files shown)


samlanning:
rebased

samlanning:
This has been opened for a while, any thoughts?

samlanning:
(and rebased again)

TheNain38:
Thanks, is there a reason it isn't merged?

src/org/thoughtcrime/securesms/components/SendButton.java
```diff
@@ -93,6 +93,13 @@ public void onChange(TransportOption newTransport) {
 
   @Override
   public boolean onLongClick(View v) {
+    return showPopupIfNeeded();
+  }
+
+  /**
+   * @return true iff the popup was opened
```
TheNain38: Could you correct your typo please? `iff` to `if`

samlanning: `iff` is a common shorthand that means "if and only if". https://en.wikipedia.org/wiki/If_and_only_if

If it is preferable that this is not used then I can remove it.

TheNain38: Oh, I didn't know that, thanks for showing it to me ;)

mcginty:
It's about time we refactored SendButton, because I'm not sure it makes sense for it to be the owner of this transport selection popup any more. Either way, this is a reasonable solution for the time being with a minimal diff. I minimized it even further to just passthrough the long-click of the attachment button to the send button.

in 3.5.0

PR #3813, opened 2015-07-27
Support for static RecipientDetails for certain shortcodes.

moxie0:
// FREEBIE


src/org/thoughtcrime/securesms/recipients/RecipientProvider.java
```diff
@@ -143,7 +151,8 @@ public RecipientDetails call() throws Exception {
         cursor.close();
     }
 
-    return new RecipientDetails(null, number, null, ContactPhotoFactory.getDefaultContactPhoto(null), color);
+    if (STATIC_DETAILS.containsKey(number)) return STATIC_DETAILS.get(number);
```
mcginty: i feel like the user contact db should override whatever we hardcode

moxie0: it does

mcginty: ah i misread, this is just what it fills in before resolving in the contacts db.

yeah this lgtm

moxie0:
in 2.24.0

PR #3814, opened 2015-07-27
Soften a few colors that are a bit too vibrant.

moxie0:
// FREEBIE


src/org/thoughtcrime/securesms/color/MaterialColors.java
```diff
@@ -29,7 +29,8 @@
     // Amber
```
mcginty: amber comment unneeded now

PR #3822, opened 2015-07-27
support variable keyboard heights

mcginty:
It appears that the problem with SwiftKey was that it really likes resizing itself to nearly the entire screen, then nearly none of the screen, then a sane height. Silly KeyboardAwareLinearLayout had no idea what to do. I just simplified things and seems to work better now. cc @McLoo


moxie0:
in 2.24.0

PR #3823, opened 2015-07-28
fix issue with camera controls going missing

### LINKED ISSUE #3775: Missing camera controls after turning screen on again
2.23.2 and 2.24.0 branch

- open direct capture
- go landscape
- turn screen of / wait till screen turns off
- go portrait
- turn screen on
- have camera preview in full screen portrait without controls
- (#3767 tap focusing works, but shows no circle animation)

mcginty:
fixes #3775


McLoo:
EmojiDrawer should also import InputAwareLayout rather than InputManager

moxie0:
in 2.24.20

PR #3826, opened 2015-07-28
enable half preview in GB

mcginty:
GB works fine with half-preview mode, you just can't move a SurfaceView while it's being rendered to. It's easy to work around that, though, and it helps with testing since there's less difference between the view logic this way.

Also decided to animate *up* the drawer a bit more slowly so that the preview begins approx at when the drawer gets to its final position. Looks nicer that way imo.


moxie0:
in 2.24.0

PR #3833, opened 2015-07-28
assert gif size constraints before OOMEs

### LINKED ISSUE #3778: Large GIFs not thumbnailed or viewed
Tried some GIFs on my HTC One S

for that I had no DRAFT thumbnail while the phone lagged (cause of to many open apps??)
http://laps.noaa.gov/albers/slides/ast/eclipse12_large.gif

for that I had no draft thumbnail, no thumbnail and an empty image view
http://www.diplom.org/Online/maps/20th-large.gif
GIF was perfectly received though on an 2.23.2 device

Log right after added the latter as draft: https://gist.github.com/7754133739d24d662ef9

mcginty:
fixes #3778

this is a temporary restriction to stop ludicrously large GIFs from being loaded and causing OOMEs. It looks like the GifDecider Glide uses doesn't distinguish between animated/single-frame GIFs and even single-frame GIFs are loaded inefficiently (no power-of-two downsampling).

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/GifSlide.java
+++ b/src/org/thoughtcrime/securesms/mms/GifSlide.java
@@ -2,6 +2,7 @@ package org.thoughtcrime.securesms.mms;
 
 import android.content.Context;
 import android.net.Uri;
+import android.provider.MediaStore.Audio.Media;
 
 import org.thoughtcrime.securesms.crypto.MasterSecret;
 import org.thoughtcrime.securesms.util.BitmapDecodingException;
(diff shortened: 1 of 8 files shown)


src/org/thoughtcrime/securesms/mms/GifSlide.java
```diff
@@ -16,9 +17,18 @@ public GifSlide(Context context, MasterSecret masterSecret, PduPart part) {
...
 
   public GifSlide(Context context, MasterSecret masterSecret, Uri uri)
-      throws IOException, BitmapDecodingException
+      throws IOException, BitmapDecodingException, MediaTooLargeException
   {
     super(context, masterSecret, uri);
+    assertMediaSize();
+  }
+
+  private void assertMediaSize() throws MediaTooLargeException, IOException {
```
moxie0: why is this a function?

moxie0:
what happens when you receive a ludicrously large gif?

mcginty:
glide will try to decode it and realize that it can't allocate the correct amount of memory, which I think is reasonable. it's either that or we set some arbitrary limit and refuse to display beyond that.

moxie0:
in 2.24.0

PR #3838, opened 2015-07-29
Transport warning message

samlanning:
This pull request displays a message above the compose window if the current selected transport does not match the transport of the last received message.

This is something I have found myself in dire need of since using TextSecure for my main SMS app again, and having a number of contacts who had TextSecure and now don't, or are iPhone users and sometimes contact with SMS and sometimes TextSecure.

This could go a long way for increasing adoption of TextSecure as it reduces some of the frustration when dealing with contacts with varying usage.

I'm not sure if this is the final look we want to go for or not, screenshots:


moxie0:
thanks but i don't think we want to do this.  maybe highlighting something in the message, but not like this.

PR #3839, opened 2015-07-30
correct lollipop bug leaving new conversations unhighlighted

subverted:
pre-lollipop, textsecure highlighted conversations containing unread messages in a slightly lighter color. when the background was switched to a ripple drawable, this behavior disappeared, making it very difficult to tell which conversations had new messages waiting. this fix returns this behavior to the dark themed textsecure. further work will be required to fix the light theme.


mcginty:
hey @subverted, thanks for the PR. Unfortunately this would break light theme and we try to avoid italics in our UI. I just opened another PR that finishes off where this one started and should work on both light and dark theme.

subverted:
oh yeah; it was about as quick and dirty as they come - the absolute
shortest distance between 'this is broken!!!!' and 'this is fixed for me',
because i was about to start slaughtering goats at it. :)

Issue #3597, opened 2015-07-10
"x" to cancel attachment misaligned


agrajaghh:
If you attach a portrait picture, the "x" isn't next to the picture:
![screenshot_09-47-49](https://cloud.githubusercontent.com/assets/5296073/8614318/d9fd515e-26e8-11e5-950d-e3746b8e52da.png)
![screenshot_09-51-07](https://cloud.githubusercontent.com/assets/5296073/8614366/3d8d73fc-26e9-11e5-85d1-54e778b3bc70.png)

brumsoel:
I think this can be closed now that #3849 is merged :)

agrajaghh:
unfortunately its still not 100% fixed for me:

**good:**
![x_working](https://cloud.githubusercontent.com/assets/5296073/9645289/ebf594f4-51ca-11e5-97e0-fcb69b7692e7.png)

**misplaced:**
![x_not_landscape](https://cloud.githubusercontent.com/assets/5296073/9645291/eeabfa58-51ca-11e5-9f51-819cba5d8e07.png)
![x_not_portrait](https://cloud.githubusercontent.com/assets/5296073/9645295/f01e985a-51ca-11e5-8026-b349bf4ebfd0.png)

rmueller83:
Although @agrajaghh mentioned in #3849 that this should be fixed, the misalignment still occurs for me in 2.26.0

mcginty:
should be fixed in 2.27.0 now

PR #3849, opened 2015-07-30
let ThumbnailView handle remove button placement

mcginty:
Let ThumbnailView figure out when and where to put the "remove" button in the case of drafts.

Since we use ThumbnailView in many places, and most of the time neither the remove button or progress indicators will be shown, I switched those to be lazily inflated since neither ImageViews nor ProgressWheel instances are very cheap to inflate when scrolling quickly.

### DIFF
--- a/res/drawable/conversation_attachment_close_circle.xml
+++ b/res/drawable/conversation_attachment_close_circle.xml
@@ -9,7 +9,7 @@
     <item>
         <shape android:shape="oval">
             <size android:height="21dp" android:width="21dp"/>
-            <solid android:color="@color/textsecure_primary_dark"/>
+            <solid android:color="@color/red_600"/>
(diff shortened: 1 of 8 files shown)


agrajaghh:
just tested, this fixes #3597 :blush: 

Is it (easily) possible to place the "x" directly in the corner like that:
corner
your PR, as comparison:

pr

mcginty:
@agrajaghh yeah it's just a bit more work to shrink the ImageView in the case of the button being open. I'm going to implement that though.

moxie0:
shrink the imageview? shouldn't thumbnailview just have a framelayout wrapping it with the correct padding on the close circle?

mcginty:
thereby shrinking the imageview ;)

PR #3855, opened 2015-07-31
Fix broken group member selection

McLoo:
and as the action button was also hard to hit, increased it's tap area too

Edit: uh, what a whitespace diff :anguished: 
https://github.com/WhisperSystems/TextSecure/pull/3855/files?w=1

Edit 2: forgot to mention the reason for this PR: 
the check mark in the toolbar isn't finishing the activity anymore, as it's taps are hijacked by toggle button now and thus toggles along :)

### DIFF
--- a/res/layout/contact_selection_activity.xml
+++ b/res/layout/contact_selection_activity.xml
@@ -26,56 +26,62 @@
                        android:background="@drawable/circle_touch_highlight_background"
                        android:src="@drawable/ic_search_white_24dp" />
 
-            <EditText android:id="@+id/search_view"
-                      android:layout_height="wrap_content"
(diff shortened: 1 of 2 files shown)


moxie0:
oh, woops, i just made the tap target for the toggle button huge since there was nothing else in the actionbar.  forgot that there is in the case of group member selection.  how about just lowering the radius instead of this?

McLoo:
Hm, not really. Even with a padding of 20 the check mark still is surrounded by the toggle tap area.

Putting the toggle inside the additional LinearLayout draws a clear border and also enables the check mark tap area to be extended.

moxie0:
in 2.25.0

PR #3856, opened 2015-07-31
reduce overdraw

mcginty:
before the direct capture stuff, we had overdraw down to a very reasonable level. It had since jumped up to avoid layout issues, but it looks like I could get it working without those issues and reduce our draw time a good amount.

before:
small1

after:
small2


McLoo:
The total blackness of the background in dark theme wasn't intended?!

moxie0:
in 2.25.0

PR #3862, opened 2015-08-01
Fix Japanese language entry

### LINKED ISSUE #3861: Japanese language entry incorrect
In the language selector in Settings, the Japanese entry is incorrect:

**TextSecure > Settings > Appearance > Language**
ja-lang-entry-incorrect

日本人 means Japanese *person*. 

日本語 means Japanese *language*, which is what should be displayed here.

unrulygnu:
Fixes #3861
// FREEBIE

### DIFF
--- a/res/values/arrays.xml
+++ b/res/values/arrays.xml
@@ -19,7 +19,7 @@
       <item>Hrvatski</item>
       <item>Indonesia</item>
       <item>Italiano</item>
-      <item>Japanese 日本人</item>
+      <item>Japanese 日本語</item>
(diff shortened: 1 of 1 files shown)


f41c0r:
@unrulygnu have you seen https://www.transifex.com/moxie/textsecure-official/

moxie0:
in 2.25.0

unrulygnu:
@f41c0r : Yes, I've spent a bit of time there :) However, this is an array value that persists across locales, not a translatable string resource, so it's not on Transifex.

PR #3871, opened 2015-08-02
Fix Incorrect String: Message Notifications Description (Preferences)

RiseT:
The original string is incorrect as the underlying option affects not only the status bar but TextSecure's message notifications as a whole (status bar, message sound, vibration, LED).

Old string: "Display message notifications in status bar"
New string: "Enable message notifications"

// FREEBIE


moxie0:
in 2.25.0

PR #3873, opened 2015-08-03
fix crash when opening link with no browser installed

### LINKED ISSUE #3726: opening hyperlink with no browser installed crashes TS
https://gist.github.com/40e0ca458f712207dee7

If I send or receive an HTTP/HTTPS link and try to open said link on a device that has no web browser installed, TextSecure will crash.

heyimgay:
so this is the beginning of my attempt to fix #3726

comments/suggestions/hints are very welcome as im new to android programming

### DIFF
--- a/src/org/thoughtcrime/securesms/RegistrationActivity.java
+++ b/src/org/thoughtcrime/securesms/RegistrationActivity.java
@@ -100,7 +100,9 @@ public class RegistrationActivity extends BaseActionBarActivity {
         intent.setAction(Intent.ACTION_VIEW);
         intent.addCategory(Intent.CATEGORY_BROWSABLE);
         intent.setData(Uri.parse("https://twilio.com"));
-        startActivity(intent);
+        if (intent.resolveActivity(getPackageManager()) != null) {
+          startActivity(intent);
+        }
       }
     });
   }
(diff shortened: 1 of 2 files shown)


junitas:
Does this fix work for when a link is clicked in conversation? It seems that this only checks the null status of the Intent when registering or rating the app.

turtlekiosk:
From: http://stackoverflow.com/questions/22479211/intent-resolveactivity-null-but-launching-the-intent-throws-an-activitynotfou
"From what i could see, intent.resolveActivity() will return true if there is ANY activity resolving this intent. Even if this activity is not exported (which makes it unusable for all practical purposes in case it's not from your package)."
It'd be better to catch the ActivityNotFoundException

mammuth:
I highly doubt, your changes will fix the issue you reference. 

In the issue it says TS is crashing, **when clicking on a hyperlink in a received/sent message**. You added checks for opening the PlayStore and registering with twilio?

Maybe @yawnbox might test it?

heyimgay:
Indeed I dont think this is ready yet but i wanted feedback early on, Im currently looking around the source to see what file(s) i have to edit to fix the issue in conversations. pointers welcome :)

moxie0:
@heyimgay Thanks but please don't open PRs unless you've tested them extensively and believe they're ready to be merged.

heyimgay:
@moxie0 OK will (not) do, for future reference should i instead leave a comment on the issue saying im going to work on it? didnt want duplication of work happening - sorry for the inconvenience

FeuRenard:
@heyimgay Do you allow me to use this contribution's code for a new PR?

PR #3876, opened 2015-08-04
DB performance optimizations

mcginty:
First round of database-related performance optimizations - it should cut load times for long conversations roughly in half. It's still kind of slow even given that, but more changes on the way to fix that.

* Add multi-column index that allows quick sorting against thread id and received time, which is how we sort our conversations. Without them, SQLite has been having to sort with temporary B-Trees which is not so efficient. Part of this requires using (column AS name) instead of (expr AS name), so I had to migrate the MMS database to use normal timestamps.
* Use the SQLite EXISTS method for checking if a conversation is empty instead of having to do a full sweep of the table.
* Move unnecessary HashMap and array allocations to static variables for the most-used getConversation() method.
* Stop keeping track of message count in ThreadDatabase since we don't use it and it's slow.


mcginty:
closing for now until i implement the "large query" cursor as wel

moxie0:
This is already a sizable diff, I like the idea of keeping those PRs distinct if there's no compelling reason otherwise?

mcginty:
yeah, i'll break these up into logical chunks

PR #3878, opened 2015-08-04
LargeSQLiteCursor

mcginty:
* getConversation() uses a LargeSQLiteCursor instance, which breaks up queries into chunks of a given window size. This is set to 500 right now, so a cursor load only happens every 500 messages and is pretty quick.
* refactor MmsSmsDatabase a bit to stop excessive object creation when calling getConversation()
* I don't know why we were calling setDistinct on our union subqueries since I'm unaware of any situation where dupe rows would be an issue, but it's slower so I set it to false. If that's a problem, can switch back.

### DIFF
--- a/src/org/thoughtcrime/securesms/database/LargeSQLiteCursor.java
+++ b/src/org/thoughtcrime/securesms/database/LargeSQLiteCursor.java
@@ -1,0 +1,212 @@
+package org.thoughtcrime.securesms.database;
+
+import android.annotation.SuppressLint;
+import android.annotation.TargetApi;
+import android.content.ContentResolver;
+import android.database.AbstractCursor;
+import android.database.ContentObserver;
+import android.database.Cursor;
+import android.database.DataSetObserver;
+import android.database.sqlite.SQLiteDatabase;
+import android.net.Uri;
+import android.os.Build.VERSION_CODES;
+import android.support.annotation.NonNull;
+import android.support.v4.util.LruCache;
+import android.util.Log;
+
+public class LargeSQLiteCursor extends AbstractCursor {
+  private static final String TAG = LargeSQLiteCursor.class.getSimpleName();
+
+  private final SQLiteDatabase db;
+  private final String         query;
+  private final int            windowSize;
+
+  private int                       count;
+  private ContentObserver           contentObserver;
+  private ContentResolver           contentResolver;
+  private DataSetObserver           dataSetObserver;
+  private Uri                       notificationUri;
+  private LruCache<Integer, Cursor> loadedCursors;
+  private Cursor                    current;
+
+  public LargeSQLiteCursor(SQLiteDatabase db, String query, int windowSize) {
+    this.db            = db;
+    this.query         = query;
+    this.windowSize    = windowSize;
+    this.count         = getCount(db, query);
+    this.loadedCursors = new CursorLruCache(4);
+    updateCursor(0);
+  }
+
+  private int getCount(SQLiteDatabase db, String query) {
+    Cursor countCursor = null;
+    try {
+      long start = System.currentTimeMillis();
+      countCursor = db.rawQuery("SELECT COUNT(1) FROM (" + query + ");", null);
+      if (!countCursor.moveToFirst()) {
+        throw new IllegalStateException("couldn't get count of results in query");
+      }
+      final int count = countCursor.getInt(0);
+      Log.w(TAG, "got count " + count + " -> " + (System.currentTimeMillis() - start) + "ms");
+      return count;
+    } finally {
+      if (countCursor != null) countCursor.close();
+    }
+  }
+
+  private String buildWindowedQuery(final int limit, final int offset) {
+    return query + " LIMIT " + limit + " OFFSET " + offset;
+  }
+
+  private void updateCursor(int position) {
+    if (position < 0 || (count > 0 && position >= count)) return;
+
+    final int    offset       = position - (position % windowSize);
+    final Cursor loadedCursor = loadedCursors.get(offset);
+    if (loadedCursor != null) {
+      current = loadedCursor;
+      loadedCursors.remove(offset);
+      loadedCursors.put(offset, loadedCursor);
+    } else {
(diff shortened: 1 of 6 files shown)


jeremymasters:
Using a group by is at least faster in MySQL than a distinct. That might be worth a shot if plan B has to be used.

src/org/thoughtcrime/securesms/database/LargeSQLiteCursor.java
```diff
@@ -0,0 +1,208 @@
...
+    this.count         = getCount(db, query);
+    this.loadedCursors = new CursorLruCache(4);
+    updateCursor(0);
+  }
+
+  private int getCount(SQLiteDatabase db, String query) {
+    long start = System.currentTimeMillis();
+    Cursor countCursor = db.rawQuery("SELECT COUNT(1) FROM (" + query + ");", null);
+    if (!countCursor.moveToFirst()) {
+      throw new IllegalStateException("couldn't get count of results in query");
```
moxie0: just leaked a cursor

moxie0:
there should be hella unit tests for largesqlitecursor

src/org/thoughtcrime/securesms/database/LargeSQLiteCursor.java
mcginty: this is pretty much an assertion since it's throwing a RuntimeException, wasn't sure that we needed to worry about resource leak in that case. will close it anyway

src/org/thoughtcrime/securesms/database/LargeSQLiteCursor.java
```diff
@@ -0,0 +1,212 @@
...
+    } finally {
+      if (countCursor != null) countCursor.close();
+    }
+  }
+
+  private String buildWindowedQuery(final int limit, final int offset) {
+    return query + " LIMIT " + limit + " OFFSET " + offset;
+  }
+
+  private void updateCursor(int position) {
```
moxie0: Unless it's called once for initialization, a private method shouldn't be causing side effects of instance variables if possible.

mcginty:
added unit tests and a bit more infrastructure that statically mocks missing things like Log

src/org/thoughtcrime/securesms/database/MmsSmsDatabase.java
```diff
@@ -38,29 +40,90 @@
...
+    add(MmsSmsColumns.READ);
+    add(MmsSmsColumns.THREAD_ID);
+    add(MmsSmsColumns.RECEIPT_COUNT);
+    add(MmsSmsColumns.MISMATCHED_IDENTITIES);
+    add(SmsDatabase.TYPE);
+    add(SmsDatabase.SUBJECT);
+    add(SmsDatabase.DATE_SENT);
+    add(SmsDatabase.DATE_RECEIVED);
+    add(SmsDatabase.STATUS);
+  }};
```
moxie0: why are these class level now? is all of this used in multiple places?

src/org/thoughtcrime/securesms/database/MmsSmsDatabase.java
```diff
@@ -38,29 +40,90 @@
...
-
-    String order           = MmsSmsColumns.NORMALIZED_DATE_RECEIVED + " ASC";
-
-    String selection       = MmsSmsColumns.THREAD_ID + " = " + threadId;
-
-    Cursor cursor = queryTables(projection, selection, selection, order, null, null);
+    final String selection = MmsSmsColumns.THREAD_ID + " = " + threadId;
+    final String query;
+    if (queryCache.containsKey(threadId)) {
+      query = queryCache.get(threadId);
```
moxie0: is this the bottleneck? if not, drop it

src/org/thoughtcrime/securesms/database/MmsSmsDatabase.java
mcginty: We were initializing a new HashSet object every time that method was called previously, it was just wasteful.

src/org/thoughtcrime/securesms/database/MmsSmsDatabase.java
mcginty: It was slow enough where the cache was well worth it when I was checking the traceview. I think it saves ~100ms on some of the slower phones.

src/org/thoughtcrime/securesms/database/LargeSQLiteCursor.java
```diff
@@ -0,0 +1,218 @@
...
+      @SuppressLint("Recycle")
+      Cursor cursor = db.rawQuery(buildWindowedQuery(windowSize, offset), null);
+      if (notificationUri != null) cursor.setNotificationUri(contentResolver, notificationUri);
+      if (dataSetObserver != null) cursor.registerDataSetObserver(dataSetObserver);
+      if (contentObserver != null) cursor.registerContentObserver(contentObserver);
+      loadedCursors.put(offset, cursor);
+      Log.w(TAG, "loaded new cursor for offset " + offset + ", holding " + loadedCursors.size() + " open cursors.");
+      return cursor;
+    }
+  }
```
moxie0: I don't understand, where is the asynchronous loading and prefetching?  If you scroll to the end of the cursor window, doesn't that result in a blocking query on the UI thread when seeking to the next position?

src/org/thoughtcrime/securesms/database/LargeSQLiteCursor.java
```diff
@@ -0,0 +1,218 @@
...
+
+  private boolean setPosition(int position) {
+    current = getCursorFor(position);
+    return current != null && current.moveToPosition(position % windowSize);
+  }
+
+  @Override public int getCount() {
+    return count;
+  }
```
moxie0: It looks like you have a race condition between calculating the count and executing the query.  I'm pretty sure that callers to this method expect the number of rows returned to be exactly equal to what's returned here.  I don't see how you could guarantee that with this logic.

moxie0:
closing in favor of chunked fetching

PR #3880, opened 2015-08-04
add thread/date multi-col SQLite indices

mcginty:
Because MMS is still stupid, PduHeaders will still use seconds instead of millis, but we'll store it in the DB as millis to allow us to use speedy SQLite indices for SELECTs instead of temporary B-Trees.

### DIFF
--- a/src/org/thoughtcrime/securesms/database/MmsDatabase.java
+++ b/src/org/thoughtcrime/securesms/database/MmsDatabase.java
@@ -152,12 +152,13 @@ public class MmsDatabase extends MessagingDatabase {
     "CREATE INDEX IF NOT EXISTS mms_read_index ON " + TABLE_NAME + " (" + READ + ");",
     "CREATE INDEX IF NOT EXISTS mms_read_and_thread_id_index ON " + TABLE_NAME + "(" + READ + "," + THREAD_ID + ");",
     "CREATE INDEX IF NOT EXISTS mms_message_box_index ON " + TABLE_NAME + " (" + MESSAGE_BOX + ");",
-    "CREATE INDEX IF NOT EXISTS mms_date_sent_index ON " + TABLE_NAME + " (" + DATE_SENT + ");"
+    "CREATE INDEX IF NOT EXISTS mms_date_sent_index ON " + TABLE_NAME + " (" + DATE_SENT + ");",
+    "CREATE INDEX IF NOT EXISTS mms_thread_date_index ON " + TABLE_NAME + " (" + THREAD_ID + ", " + DATE_RECEIVED + ");"
   };
 
   private static final String[] MMS_PROJECTION = new String[] {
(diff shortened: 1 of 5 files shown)


src/org/thoughtcrime/securesms/database/MmsDatabase.java
```diff
@@ -1218,6 +1218,11 @@ public void close() {
     }
   }
 
+  private long getTimestamp() {
+    final long time = System.currentTimeMillis();
+    return time - (time % 1000);
+  }
```
moxie0: this doesn't really seem like the right place for this, it implies that the MmsDatabase has a timestamp

src/org/thoughtcrime/securesms/database/MmsDatabase.java
```diff
@@ -152,12 +152,13 @@
     "CREATE INDEX IF NOT EXISTS mms_read_index ON " + TABLE_NAME + " (" + READ + ");",
     "CREATE INDEX IF NOT EXISTS mms_read_and_thread_id_index ON " + TABLE_NAME + "(" + READ + "," + THREAD_ID + ");",
     "CREATE INDEX IF NOT EXISTS mms_message_box_index ON " + TABLE_NAME + " (" + MESSAGE_BOX + ");",
-    "CREATE INDEX IF NOT EXISTS mms_date_sent_index ON " + TABLE_NAME + " (" + DATE_SENT + ");"
+    "CREATE INDEX IF NOT EXISTS mms_date_sent_index ON " + TABLE_NAME + " (" + DATE_SENT + ");",
+    "CREATE INDEX IF NOT EXISTS mms_thread_date_index ON " + TABLE_NAME + " (" + THREAD_ID + ", " + DATE_RECEIVED + ");"
```
moxie0: The order by needs this index or something? Is it still necessary to have both indexes?

mcginty: i don't know why `mms_date_sent_index` existed, you tell me :stuck_out_tongue:

`mms_thread_date_index` is the only index used when querying with getConversations()

src/org/thoughtcrime/securesms/database/MmsDatabase.java
mcginty: i can rename it to `generateCompatibleTimestamp()`

src/org/thoughtcrime/securesms/database/MmsDatabase.java
moxie0: I think it probably exists because the where clause only mentions the thread id and the intention was to create an index for that query. I guess we can leave it, but it might be consuming space unnecessarily.

mcginty:
updated naming, i vote to keep the index for now until we have more time to decide it's totally unnecessary

PR #3883, opened 2015-08-04
stop conversation list from hogging db

mcginty:
is there a reason ConversationActivity was popup themed?

It looks like because the theme makes the activity transparent, onStop is never called so the conversation list is continuing to reload itself even while in the conversation activity. seems to work fine with the default theming.


moxie0:
went in with the quick reply stuff, does everything quick reply still work without it?

mcginty:
seems like it works fine, ConversationPopupActivity still uses the popup theme

moxie0:
in 2.25.0

PR #3885, opened 2015-08-05
stop sync resolution of recipients on ui thread

mcginty:
Reader.getNext() is called by the adapters to get a MessageRecord, so we shouldn't be resolving recipients synchronously there. I didn't see any parts that rely on resolved recipient details when retrieving a MessageRecord.

the slowdown became more prominent now that recipients query the preference DB synchronously as well.


moxie0:
in 2.25.0

PR #3894, opened 2015-08-05
Beautify strings in Import/Export menu

RiseT:
- Remove redundant periods in description strings
- Remove redundant line feeds in XML (confusing to translators at
Transifex)
- Change app's name from 'SMSBackup And Restore' to its official name 'SMS Backup &
Restore'
- Change "Import encrypted backup" into "Restore encrypted backup" as this is the only option that replaces the complete database rather than add to it (which all the "import" options do). "restore" is the wording used in the description string as well.
- Fix string consistency of "Notification Sound" (see below)

### DIFF
--- a/res/layout/import_fragment.xml
+++ b/res/layout/import_fragment.xml
@@ -80,7 +80,7 @@
                 <TextView android:layout_width="match_parent"
                           android:layout_height="wrap_content"
                           style="@style/Registration.Description"
-                          android:text="@string/import_fragment__import_encrypted_backup"/>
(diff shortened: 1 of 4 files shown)


RiseT:
@McLoo Could you do me a favor and check the ampersand in this PR? Thanks in advance.

McLoo:
:+1: all is well

RiseT:
I accidentally added another commit to this PR which I intended to post as a separate PR ("Fix notification sound string consistency").
Apart from the fact that the title of the primary PR doesn't really fit for this latest addition, it is fine. So I hope it doesn't matter.

PR #3906, opened 2015-08-06
Fix ActionBar title and menu entry for group updates

McLoo:
hope we finally get this nasty behavior fixed
Groups seem to be the lessest liked child :grin:


McLoo:
rebased & enhanced (thanks @DrGrill0r for the MMS hints)

mcginty:
old, should not be an issue any more with the recent group activity refactor

PR #3914, opened 2015-08-07
Emoji-Switcher Implementation

KyleGospo:
Implements an emoji-switcher to swap between Apple and Google emoji, with Apple as the default. This can also be easily expanded to include other emoji sets in the future.

Known Issues:
-Google emoji despite using the sprite sizing settings from prior to switching to Apple emoji seem to have some sizing issues that need to be ironed out in the lower parts of the sprite sheet. If anyone more well aquainted with TextSecure's emoji display is willing to take a quick look at this It'd be much appreciated.
-Had one case where the Emoji sheet failed to switch. A quick app/context restart may need to be applied to the preference.

// FREEBIE


oaim:
I think this would only create more confusion. I don't like apple emojis, but the translation between apple and google emojis just isn't good. Some emojis look different and convey different "feelings". I agree apple emojis are ugly and would rather see emoji one implemented, but it's better if every version of the app (ios, android, chrome extension) uses the same set.

KyleGospo:
I think there's other issues here though. As an Android user, I have Google emoji in the notification, I have Google emoji in the Android Wear app, and I have Google emoji in the Google keyboard which I prefer to use to the built in drawer. When replying with emoji from the watch I have to draw the actual Google emoji.

I'd much rather have consistency than have emoji the same across different operating systems. I don't use them as a language, only to convey rough emotion, so I don't see minor differences between phones as a loss.

The inconsistency in my own ecosystem is dreadful though.

mad-de:
Ohhh yrs. Pull it!

moxie0:
Hey @KyleGospo thanks for the PR, but please don't submit PRs that are unfinished, need testing, or have known issues.  Only clean, final, well-tested PRs should be open here.  If you're submitting something partially finished because you want feedback on whether it would get merged in its final form, please bring that up on the mailing list instead.  Thanks!

Issue #3147, opened 2015-05-10
[Feature request] Indicate a new day [ux]


Mihahn:
I'd like to request a new "feature", which would be some kind of indicator when a new day begins (like it's implemented in other messengers like WhatsApp).
I think this would make searching messages that were sent on a certain day easier and it just structures the conversations...

Below you can see how WhatsApp handles it and I think a similar solution would be great!

![screenshot_2015-05-10-12-09-39](https://cloud.githubusercontent.com/assets/3447340/7553885/c742aeec-f70e-11e4-89fc-29a0c7692549.png)

2-4601:
4c815db076bdad365c785857e690982bfdb63aa3

Issue #3102, opened 2015-05-04
Date format does not respect locale


2-4601:
Currently TS seems to display all dates in the format [MMM d, yyyy](https://github.com/WhisperSystems/TextSecure/blob/d8521637bb2a545d02d2b7a32127a4f01b6d0317/src/org/thoughtcrime/securesms/util/DateUtils.java) irrespective of the locale. The dates are translated correctly though.

Here's some information about date formats in various regions:
https://en.wikipedia.org/wiki/Date_format_by_country

TS 2.13.0
This issue does not depend on whether you have changed the system language or TS language. Neither work.

agrajaghh:
If it's the same issue like #2684 it has been fixed in https://github.com/WhisperSystems/TextSecure/commit/d8521637bb2a545d02d2b7a32127a4f01b6d0317

2-4601:
That should be fixed in 2.13.0 beta release, right?

I searched only open issues before creating this so I didn't see #2684. But that issue seems to be about date translations and not about date format? Before 2.13.0 I also experienced that issue but it seems to be fixed now.

This issue is about date formatting which does not respect the locale. This was introduced in 2.13.0. I did not see it before. Possibly because #2684 prevented me from seeing translated dates.

For example the Finnish date format should be `d. MMM yyyy` and not `MMM d, yyyy`.

TextSecure shows it wrong. Android gets the date formatting right.
![svejk](https://cloud.githubusercontent.com/assets/174176/7457751/9f85e804-f297-11e4-9aff-4145e150e340.jpg)

agrajaghh:
oh sry, yes the commit I mentioned is already in 2.13...

rhodey:
ahhh bummer, I thought [this line](https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/util/DateUtils.java#L44) was supposed to take care of that. thanks for the expected vs actual test vectors, I'll look into it.

rhodey:
```java
private static void testLocalizedSimpleDateFormat(String template, Locale locale) {
  Log.d(DateUtils.class.getName(), "LOCALE >> " + locale + ", TEMPLATE >> " + template);
  SimpleDateFormat dateFormat = new SimpleDateFormat(template, locale);
  Log.d(DateUtils.class.getName(), "LOCALE >> " + locale + ", LOCALE TEMPLATE >> " + dateFormat.toLocalizedPattern());
}
```

```
[...]

Issue #3619, opened 2015-07-11
Messages out of order due to clock change


anshin:
I had a TextSecure (not SMS, didn't try SMS) conversation (other participant was using Signal on iOS in case that matters) this morning with my phone's clock (inadvertently) set an hour ahead of the correct time. Messages received were marked as being sent at the correct time (and were not shown using relative time such as "an hour ago") but as being received at my clock's time. When I corrected the time, the received time did not change, and new messages, marked as being received at the correct (and current) time, were placed in the thread before the older messages rather than appended to the thread.

Device info: LG Optimus Exceed 2, stock android 4.4.2, TextSecure 2.21.0

App state: Push registered, set as default SMS app

wp9015362:
Might be related to the following issues:

https://github.com/WhisperSystems/TextSecure/issues/2385
https://github.com/WhisperSystems/TextSecure/issues/3618

?

malenki:
Android 4.4.2
Signal 3.16.0
I experienced (and still do) a similar issue today.
With one person I exchanged messages on:
Day A, 22:00
**Day B, 14:00**
Day B, 20:30
Day B, 20:35
Day B, 20:40
Day B, 20:45

Now there was obviously a glitch in my Telco providers network which caused my to show 01:37am although it was 17:00 (5pm). Thus I set the clock manually.
Although between 14:00 and 20:30 no messages were exchanged, Signal sorts the messages this way: 
Day A, 22:00
Day B, 20:30
Day B, 20:35
Day B, 20:40
Day B, 20:45
**Day B, 14:00**

The contacts list also shows this person as "last chatted with: 6h ago"

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

Issue #3478, opened 2015-06-25
conversation sort order not updated on message send


rhodey:
tested on my GB device with a25bedff2888399bde6b880a32d64a7cd794fe92

1) send message to person A (new conversation)
2) receive message from person B
3) send another message to person A
4) notice that the conversation with person B is still sorted top

rhodey:
I'm actually seeing some strange message sort behavior on 2.19.1 from the play store on my Nexus 5 also.

riyapenn:
Is this still an issue?

Richtiggenial:
Can't reproduce it.
I think it isn't an open issue anymore.

PR #3926, opened 2015-08-11
Add date indicator above messages

Omikron43:
This adds the date between two messages when the first message was sent/received at minimum one day before the second message.

The date is translated with respect to the language set in TS but also suffers from #3102.
The year is only displayed when the message was sent/received in a year smaller than the current year.

Screenshots with light and dark theme:
date-sep-light
date-sep-dark

### DIFF
--- a/res/drawable/conversation_item_date_separator_background.xml
+++ b/res/drawable/conversation_item_date_separator_background.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android">
+    <solid android:color="@color/gray20"/>
+    <corners android:radius="4dp"/>
+</shape>
--- a/res/drawable/conversation_item_date_separator_background_dark.xml
+++ b/res/drawable/conversation_item_date_separator_background_dark.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android">
+    <solid android:color="@color/gray70"/>
+    <corners android:radius="4dp"/>
+</shape>
--- a/res/layout/conversation_item_date_separator.xml
+++ b/res/layout/conversation_item_date_separator.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                xmlns:android="http://schemas.android.com/apk/res/android"
+                xmlns:tools="http://schemas.android.com/tools">
+
+    <TextView android:id="@+id/conversation_item_date_separator"
+              android:layout_width="wrap_content"
+              android:layout_height="wrap_content"
+              android:layout_centerHorizontal="true"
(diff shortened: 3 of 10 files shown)


Mihahn:
Finally! Closes #3147 :smile:

Omikron43:
Ahh searched but didn't find anything. Thanks!

heyimgay:
Would it be possible to change the indicator so it's under the message? I usually have a rough idea of when the message was sent/received and i would like to use the date indicator to help me quickly find the message while scrolling up.

Another ideaI had is making the indicator stay on the screen at the bottom until the next one pushes it out,

Omikron43:
As the messages are listed in chronological order (or rather should be**) from top to bottom for me the separator indicates: now are following all messages that were sent/received on this date.
I also like the floating indicator (like WhatsApp has it). I'm looking into it but I have no idea when I'll find time for it.

** see for example: #3619, #3478, #3312, #2385

heyimgay:
But you always scroll up to find older messages, starting from the most recent message. From bottom to top. Even if you scroll up and go to a different conversation and come back you are returned to the bottom.

mcginty:
Thanks for the PR. I'd need to be convinced the performance impact is minimal before merging this - at first glance of the code it seems like this could cause a decent amount of extra jank.

Omikron43:
I had the same in mind but did not notice more lag than before when quickly scrolling in a large thread.

I also experimented with the [StickyListHeaders](https://github.com/emilsjolander/StickyListHeaders) library that is already used in the contact selection activity. Sadly, the library does not work with ListFragments or RecyclerListeners. Adapting the ConversationFragment (to be a plain Fragment and without the RecyclerListener) did not boost performance.

Omikron43:
Rebased on master and updated for `RecyclerView` using an `ItemDecoration`.
Sadly, I could not get a floating separator to animate properly so I omitted it.

Also I had only the possibility to test on Android 5.1.1 (Nexus 5 and Galaxy S4).

grandchild:
I don't think this is a good idea, at least not in this form. It could be done taking way less space.
I mean you guys usually have a large phone (I do too), but imagine all this on a 4'' display with the keyboard open, there's not much left to read.
Here's what I would like more
textsecure-date-spacer-mockup-slimversion

heyimgay:
maybe make the date only appear while scrolling, and let them 'push' the bottom one off the screen. for this to work we would have to use footers instead of the proposed headers

mcginty:
This would make sense if we removed the date part of the timestamp inside the bubble (similar to how Telegram and WhatsApp do it), and it'd need to be refactored, likely using StickHeaderAdapter to be DRY. @Omikron43 you still around and wanting to experiment with this? If not I'll close for inactivity and we can take another try at it later.

mcginty:
Going to close pending further updates.

Omikron43:
Sorry for the late response, I'm still around and would like to get this (or a similar feature) merged.
The last time I touched the code it was not very performant (with `ItemDecorator`) and I had too less free time to dig further.
If I remember correctly there were some changes to the `ConversationAdapter`. At the weekend I will have another look and see what has to be rebased andif I can improve the performance. Maybe now there are more advanced sticky header techniques for recycler views.

PR #3930, opened 2015-08-12
Disable compose panel and attachments when group was left

### LINKED ISSUE #3334: emoji keyboard and attachment paperclip still enabled after leaving group
1) create a group
2) leave the group
3) notice that you can't click on message composition field but you can still bring up the emoji keyboard, type emoji, and add attachments

the send button is still disabled as expected though.

McLoo:
This just removes the compose panel and disables attachments from the menu, when you left the group. So the conversation goes down to screen bottom, having the last message display "You have left the group. *(and you can't do anything about it :stuck_out_tongue_winking_eye: )*"

- it's still possible to set notification prefs, as you can be readded to the group
- keeping text drafts: text hidden in compose
- keeping media drafts: media thumbnail visible until you delete it with the (X) button
- not waiting or checking for a group re-add to enable composing while the group convo is open 
(usage/effort/load)

fixes #3334

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -61,7 +61,6 @@ import org.thoughtcrime.securesms.TransportOptions.OnTransportChangedListener;
 import org.thoughtcrime.securesms.color.MaterialColor;
 import org.thoughtcrime.securesms.components.AnimatingToggle;
 import org.thoughtcrime.securesms.components.ComposeText;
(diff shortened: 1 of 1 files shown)


McLoo:
screenshot_2015-08-12-18-57-22

mcginty:
It looks like the UI is broken to me when there's no composition bar or anything down there. I think it'd be better to have some kind of view replacing it (just like when you have blocked someone) letting the user know why something's missing.

McLoo:
:/ had that and didn't like it 
I'll give it another shot

mcginty:
closing, inactive

PR #3947, opened 2015-08-15
Treat delivery failures in MessageDetailsActivity as in-thread notifications

haffenloher:
Until now, when message delivery fails and the currently visible activity is the MessageDetailsActivity of the failed message, you get a notification. Tapping on it brings you back to the ConversationActivity, in which you tap on the failed message to get back to the MessageDetailsActivity you just came from.

Instead I propose to treat the notification as an in-thread notification (only play a sound / do nothing, depending on the user setting for in-thread notifications). Doesn't seem like there's a bug report for this, but I think it's more logical this way.


moxie0:
i think it would probably be good enough just to set the visible thread to the message's thread, rather than having a separate concept of which message is visible

brumsoel:
That sounds reasonable on second thought. Done!

moxie0:
in 3.8.0

PR #3964, opened 2015-08-18
Maintenance: Removed unused code & imports in components pkg

samlanning:


moxie0:
@samlanning Thanks but we generally don't merge this kind of thing, and prefer to clean stuff up as part of deliberate changes.

samlanning:
Ah alright ok, does the same go for my previous PRs?

PR #3975, opened 2015-08-22
Allow saving all attachments of a thread

fehnomenal:
Creates a new button in the `MediaOverviewActivity` to save all images of the conversation.
The diff is big because of different strings for single/multiple images to save.


mcginty:
Cool with this change, a nice convenience button. It wouldn't hurt to have a "save all" icon (maybe just two stacked floppies) to make it clear this is a batch action, though.

mcginty:
quick photoshop

ic_save_white_48dp

transparent version: https://cloud.githubusercontent.com/assets/373823/11358584/bb3a7b4c-9228-11e5-84c4-d7e1ba7539aa.png

honestly it kind of makes it look like an old timey large-format film camera

Omikron43:
Rebased on master and used the batch icon. I really don't know how it should be properly resized so I just threw it into /res/drawable/

agrajaghh:
you need to put different sizes in drawable-mdpi, hdpi, xhdpi, xxhdpi and xxxhdpi. Just look at the normal save icon

Omikron43:
OK, done based on [this answer](https://stackoverflow.com/a/21311302/2505347).

agrajaghh:
action bar icons are 24dp not 48dp I think :stuck_out_tongue_winking_eye:

mcginty:
it's all good, i'll take care of the asset sizing

mcginty:
Right now this iterates through a cursor on the main thread from what I can see, which is to be avoided. I think we'll want to use an ProgressDialogAsyncTask to collect the Attachments we'd like to save (we tend to prefer List instead of arrays), then once they're collected open a warning saying how many attachments they're about to save, and if they accept then spawn a SaveAttachmentTask.

That way all I/O happens off the main thread and the user will be more aware of how many files will be created by hitting that button.

Omikron43:
OK, rebased and reworked according to your comments. I pluralized some strings to not have these ugly `multipleAttachments ? ... : ...` statements.
Regarding the I/O on the main thread: we know beforehand how many images will be saved (the item count of the cursor). This number is printed in the dialog but after accepting there is a very short (probably only a few millisecs with 213 images, Nexus 5) flash of the dialog collecting the images. After the flash the normal save dialog pops up.

src/org/thoughtcrime/securesms/MediaOverviewActivity.java
```diff
@@ -137,12 +146,63 @@ public void onModified(Recipient recipient) {
...
+      public void onClick(DialogInterface dialogInterface, int i) {
+        new ProgressDialogAsyncTask<Void, Void, List<SaveAttachmentTask.Attachment>>(c,
+                                                                                     R.string.ConversationFragment_collecting_attahments,
+                                                                                     R.string.please_wait) {
+          @Override
+          protected List<SaveAttachmentTask.Attachment> doInBackground(Void... params) {
+            Cursor cursor                                   = DatabaseFactory.getImageDatabase(c).getImagesForThread(threadId);
+            List<SaveAttachmentTask.Attachment> attachments = new ArrayList<>(cursor.getCount());
+
+            cursor.moveToFirst();
```
moxie0: while (cursor != null && cursor.moveToNext()) {

Omikron43: done

moxie0:
really hard to follow this when you force push a rebased update rather than adding a commit to address comments. any chance you could separate out the commit?

fehnomenal:
@moxie0 did a bit git-fu and I think I found the original commit that you commented on and extracted the looping commit

moxie0:
cool now as far as i can see there's only one resource in xxxhdpi, rather than sized resources for every density

fehnomenal:
@moxie0 i had a bit trouble with finding the right sizes so @mcginty said in https://github.com/WhisperSystems/Signal-Android/pull/3975#issuecomment-159390560 that he would take care of resizing

agrajaghh:
@fehnomenal I added the missing files and opened a PR in your repo: https://github.com/fehnomenal/TextSecure/pull/1

fehnomenal:
nice thx

moxie0:
in 3.12.0

PR #3991, opened 2015-08-28
media download controls

mcginty:
This is the UX work necessary for handling large file attachments in TS. Sending attachments via streams instead of byte arrays to follow, then we should be good to increase attachment size.

The change allows one to choose what types of attachments to auto-download on mobile/Wi-Fi/roaming, with a download button appearing if auto-dl not allowed.


src/org/thoughtcrime/securesms/components/ThumbnailView.java
```diff
@@ -136,28 +142,53 @@ public void setImageResource(@NonNull Slide slide, @Nullable MasterSecret master
...
+    updateControlsState(slide);
+
     this.slide = slide;
-    if (slide.isInProgress() && showProgress) {
+    buildGlideRequest(slide, masterSecret).into(image);
+    setOnClickListener(new ThumbnailClickDispatcher(thumbnailClickListener, slide));
+    downloadButton.setOnClickListener(new ThumbnailClickDispatcher(downloadClickListener, slide));
+  }
+
+  private void updateControlsState(Slide slide) {
```
moxie0: private methods shouldn't have side effects. maybe this needs to be abstracted out.

src/org/thoughtcrime/securesms/jobs/AttachmentDownloadJob.java
```diff
@@ -40,16 +40,20 @@
 
   @Inject transient TextSecureMessageReceiver messageReceiver;
 
-  private final long messageId;
+  private final long   messageId;
+  private final PartId partId;
```
moxie0: non-transient job variables should only be primitive types

src/org/thoughtcrime/securesms/jobs/requirements/MediaNetworkRequirement.java
```diff
@@ -0,0 +1,29 @@
...
+import org.thoughtcrime.securesms.database.PartDatabase.PartId;
+import org.thoughtcrime.securesms.util.ConnectivityUtil;
+import org.whispersystems.jobqueue.dependencies.ContextDependent;
+import org.whispersystems.jobqueue.requirements.Requirement;
+
+public class MediaNetworkRequirement implements Requirement, ContextDependent {
+  private static final String TAG = MediaNetworkRequirement.class.getSimpleName();
+
+  private transient Context context;
+  private           PartId  partId;
```
moxie0: no complex types

src/org/thoughtcrime/securesms/components/ThumbnailView.java
```diff
@@ -136,28 +142,53 @@ public void setImageResource(@NonNull Slide slide, @Nullable MasterSecret master
...
+    if (hideControls) {
+      progress.setVisibility(GONE);
+      downloadButton.setVisibility(GONE);
+      return;
+    }
+
+    final int     transferProgress  = slide.getPart().getTransferProgress();
+    final boolean autoDownloadReady = ConnectivityUtil.isAutoDownloadAllowed(getContext(), slide.getPart().getPartId());
```
moxie0: looks like this hits the db

mcginty:
updated

src/org/thoughtcrime/securesms/components/ThumbnailView.java
```diff
@@ -321,4 +334,38 @@ public boolean onResourceReady(GlideDrawable resource, Object model, Target<Glid
...
+          transferProgress == PartDatabase.TRANSFER_PROGRESS_FAILED)
+      {
+        progress.setVisibility(GONE);
+        downloadButton.setVisibility(VISIBLE);
+      } else {
+        progress.setVisibility(GONE);
+        downloadButton.setVisibility(GONE);
+      }
+    }
+  }
```
moxie0: Why do you need an async task?  Why isn't this information just on the slide?  This is a View, so it shouldn't have any concept of what's "allowed," it should just be visualizing a model.

mcginty:
updated

src/org/thoughtcrime/securesms/jobs/AttachmentDownloadJob.java
```diff
@@ -41,44 +41,43 @@
   @Inject transient TextSecureMessageReceiver messageReceiver;
 
   private final long messageId;
+  private final long partRowId;
+  private final long partUniqueId;
```
moxie0: If you have to do this, you have to include a serialVersionUID. It will break every existing attachmentdownloadjob, so you also have to handle rescheduling the existing ones.

src/org/thoughtcrime/securesms/jobs/requirements/MediaNetworkRequirement.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.Collections;
+import java.util.Set;
+
+import ws.com.google.android.mms.pdu.PduPart;
+
+public class MediaNetworkRequirement implements Requirement, ContextDependent {
+  private static final String TAG = MediaNetworkRequirement.class.getSimpleName();
+
+  private transient Context context;
```
moxie0: serialVersionUID

src/org/thoughtcrime/securesms/database/PartDatabase.java
```diff
@@ -492,13 +502,23 @@ public void markPartUploaded(long messageId, PduPart part) {
...
   }
 
+  public void setTransferState(long messageId, @NonNull PartId partId, int transferState) {
+    final ContentValues  values   = new ContentValues(1);
+    final SQLiteDatabase database = databaseHelper.getWritableDatabase();
+
+    values.put(TRANSFER_STATE, transferState);
+    database.update(TABLE_NAME, values, PART_ID_WHERE, partId.toStrings());
+    notifyConversationListeners(DatabaseFactory.getMmsDatabase(context).getThreadIdForMessage(messageId));
+    EventBus.getDefault().post(new MediaNetworkRequirementProvider.MediaDownloadControlEvent());
```
moxie0: is there a reason to make this a broadcast? there's only one consumer, right?

src/org/thoughtcrime/securesms/jobs/requirements/MediaNetworkRequirement.java
```diff
@@ -0,0 +1,98 @@
...
+    } else {
+      return Collections.emptySet();
+    }
+  }
+
+  @Override
+  public boolean isPresent() {
+    final PartId       partId = new PartId(partRowId, partUniqueId);
+    final PartDatabase db     = DatabaseFactory.getPartDatabase(context);
+    final PduPart      part   = db.getPart(partId);
```
moxie0: It's unusual for a requirement to make a DB query. This call could potentially happen very rapidly.

mcginty: (discussed offline keeping as is)

mcginty:
updated with migration for existing pending part jobs

src/org/thoughtcrime/securesms/database/PartDatabase.java
mcginty: EventBus is less code and works just as well, so I went with that. Is there a reason *to* make it a broadcast?

moxie0: Hah, no I mean that this seems to be a situation where code in one place wants to trigger code in exactly one other place.  That's typically a function call, I'm asking why it's a broadcast.

src/org/thoughtcrime/securesms/jobs/AttachmentDownloadJob.java
```diff
@@ -35,50 +35,49 @@
 import ws.com.google.android.mms.pdu.PduPart;
 
 public class AttachmentDownloadJob extends MasterSecretJob implements InjectableType {
-
-  private static final String TAG = AttachmentDownloadJob.class.getSimpleName();
+  private static final long   serialVersionUID = -1854008261087503079L;
```
moxie0: this should be 1

src/org/thoughtcrime/securesms/jobs/requirements/MediaNetworkRequirement.java
```diff
@@ -0,0 +1,99 @@
...
+import org.whispersystems.jobqueue.dependencies.ContextDependent;
+import org.whispersystems.jobqueue.requirements.Requirement;
+
+import java.util.Collections;
+import java.util.Set;
+
+import ws.com.google.android.mms.pdu.PduPart;
+
+public class MediaNetworkRequirement implements Requirement, ContextDependent {
+  private static final long   serialVersionUID = 8307059890458646160L;
```
moxie0: version 0

mcginty:
changed versions for UID

PR #3997, opened 2015-08-31
Indicate whether last message is yours in conversation overview

BrokenPhysics:
Since pull request #3012 seems to have died, I have recreated a solution for #1363.


mcginty:
I'm guessing this will break with RTL language translations. Maybe @moxie0 has a strong opinion on this since he deals with translations, but you can use a string like `<b>You: </b> %s` and then use `context.getResources().getText()` instead of `getString` to get the properly spanned text. It does, however, make translation of that string easier to mess up.

Also as I said in another PR related to this, I think the full solution to this problem is to show delivery indicators to the left of a summary instead of "You: ", but that's a change that requires a DB migration most likely.

heyimgay:
This can be closed as we have this: https://github.com/WhisperSystems/Signal-Android/pull/4710

PR #4000, opened 2015-08-31
dagger -> dagger2

mcginty:


moxie0:
let's hold on to this until we have a compelling reason to make the change. not worth breaking the serialization on those jobs just for the upgrade.

PR #4002, opened 2015-09-01
Add branded launch screen

pR0Ps:
When the app starts cold there is currently a blank white screen that shows for about 1/2 a second (longer on slower devices) while the app loads.

This introduces a simple branded launch screen [as defined by the Google design docs](https://www.google.com/design/spec/patterns/launch-screens.html#launch-screens-branded-launch) that shows the TextSecure icon on a black background while the app is initializing.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -135,6 +135,7 @@
 
     <activity android:name=".ConversationListActivity"
           android:label="@string/app_name"
+          android:theme="@style/TextSecure.BrandedLaunch"
           android:launchMode="singleTask"
           android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"
           android:exported="true" />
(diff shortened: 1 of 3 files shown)


moxie0:
Thanks @pR0Ps .  Personally I think those loading screens are kind of garish, what are your thoughts @mcginty ?

jeremymasters:
My only real thought of there being a need for this is slower devices that spend a lot of time between startup and the password screen.  Maybe just fade in the password screen; where instead of the password field it says "Loading..".  Then "Loading..." is replaced with password field.

Dyras:
With the dark theme the white "flash" is really irritating, so I would love to have this feature if the start screen is consistent with the used theme.

pR0Ps:
@Dyras : Unfortunately there's no way to match the screen with the current theme (that I could find), but having it flash black instead of white will help the people using it at night who currently get blinded as it opens.

PR #4006, opened 2015-09-01
Update strings

### LINKED ISSUE #4004: Change notifications string "Content hidden" with "New Message!" when neither name and message are displayed
If notifications are enabled && ( display is not set to show Name and message ) then
- Replace "New TextSecure Message" with "TextSecure"
- Replace "Contents hidden" with "New Message!"

oaim:
Fixes #4004
(Sorry I had another PR open, I realized too late that the issue wanted two strings fixed and not just one)


moxie0:
thanks but the string value needs to be reflected in the string name, you can't just change the value

PR #4012, opened 2015-09-03
Fix PartDatabaseTest

pR0Ps:
Broken by the refactoring in 0c9d9e8dcf6bb42f0377d1d75b2f6c5f5716d4ff


mcginty:
the androidTest set is de-facto deprecated and is unlikely to continue being maintained, we should probably just break these out to unit tests.

Thanks for the PR!

PR #3998, opened 2015-08-31
Typo Correction

BrokenPhysics:
Since #3012 seems to have died, I'm proposing this typo correction again.

Hopefully, I've done everything right by correcting it in every language's *strings.xml*.

PR #4013, opened 2015-09-03
Typo Correction (fixed)

BrokenPhysics:
corrected version of #3998:

Since #3012 seems to have died, I'm proposing this typo correction again.

~~Hopefully, I've done everything right by correcting it in every language's strings.xml.~~ Only the main strings.xml is changed.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -399,7 +399,7 @@
 
     <!-- ThreadRecord -->
     <string name="ThreadRecord_left_the_group">Left the group...</string>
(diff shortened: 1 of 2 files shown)


moxie0:
What is the correlation between this an #3012?  If this just changes a typo in a string variable name and doesn't address any actual bugs, it's not worth changing since translators for each language would have to retranslate that string.

BrokenPhysics:
never mind then. I'm closing this pull request.

PR #4017, opened 2015-09-04
fix crash on pending self-sent media

### LINKED ISSUE #4016: Crash on unlocking - all the time
2.26
https://gist.github.com/a45c8bfd62aee5721179

Tested ~~serious~~ self (stupid auto correct)  send mms 
Had phone & TS locked 

Unlocked. Crash.

mcginty:
fixes #4016

this change makes sure parts copied from outbox to inbox are not pending as well since that doesn't make sense.

tested by queuing up pending self-sends as well as pending normal incoming push attachments, both behaved as expected.


mcginty:
@moxie0 updated

PR #4019, opened 2015-09-05
Large attachments

mcginty:
This change includes asynchronously resolving slide attachment objects since we need to calculate the total size of the media as well as alerting the user about size issues *before* the media is attempted to be sent.

It's a net loss diff since there was a good amount of redundant code that was easy to simplify in the process.

Audio and Video sizes are raised to 100MB, gif raised to 5MB.

PR #4022, opened 2015-09-05
Bump the max sizes for message limits delivered by push

### LINKED ISSUE #700: Limited video size?
Seems that sending video files are limited in size... Don't know if it's pixel size or file size?!?

IMHO there should be exist no size limit.

richo:
I picked totally arbitrary numbers that seemed reasonable sounding at
the time.

closes #700

in https://github.com/whispersystems/TextSecure/issues/700#issuecomment-89356438 @mcginty wrote:

> It's just a matter of limiting auto-downloads of large files depending on the network you're on before we increase the size.

However naively, it seems that a malicious implementor of TextSecure's protocol can already bypass this limit, if the only restrictions are client side at present. Thus I propose raising the limit.

I'm totally onboard with picking other numbers, I literally just eyeballed a minute long video.

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/PushMediaConstraints.java
+++ b/src/org/thoughtcrime/securesms/mms/PushMediaConstraints.java
@@ -32,11 +32,11 @@ public class PushMediaConstraints extends MediaConstraints {
 
   @Override
   public int getVideoMaxSize() {
-    return MmsMediaConstraints.MAX_MESSAGE_SIZE;
+    return 512 * MB;
   }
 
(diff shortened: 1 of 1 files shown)


richo:
More specifically, sending gigantor videos from Signal to TextSecure works today, so unless I'm missing something, this guard isn't protecting anything significant.

rmueller83:
I think this is already addressed by #4019

richo:
Oh whoops. I found my way here from #700 which doesn't mention it.

PR #4033, opened 2015-09-07
fix bubble height change glitch

mcginty:
when going from the "..." to check mark, the height of the bubble changed a bit. This stops that.


moxie0:
in 2.28.0

PR #4043, opened 2015-09-07
Option to use system emoji

mcginty:


src/org/thoughtcrime/securesms/components/emoji/EmojiEditText.java
```diff
@@ -24,7 +25,9 @@ public EmojiEditText(Context context, AttributeSet attrs) {
 
   public EmojiEditText(Context context, AttributeSet attrs, int defStyleAttr) {
     super(context, attrs, defStyleAttr);
-    setFilters(new InputFilter[]{ new EmojiFilter(this) });
+    if (!TextSecurePreferences.isSystemEmojiPreferred(getContext())) {
```
moxie0: You're not nervous about i/o implications here?

mcginty: I am, but it should be cached after the first read, meaning only one I/O hit per app run. I kept it simple so didn't make an async task, but still possible if need be.

mcginty:
looks like this is in 2.27.0

turtlekiosk:
omg it's really happening!!!

PR #4044, opened 2015-09-08
TransferControlView

mcginty:
refactor all the transfer control view logic into its own view, as it probably should have been from the beginning. also take advantage of all the animations we use in AnimatingToggle to make things a bit spiffier with less code redundancy.


src/org/thoughtcrime/securesms/components/AnimatingToggle.java
```diff
@@ -18,7 +19,7 @@
 
   private static final int SPEED_MILLIS = 200;
 
-  private View current;
+  protected View current;
```
moxie0: I'd much prefer that you use methods rather than manipulating this state directly

mcginty:
using methods instead of direct state manipulation

src/org/thoughtcrime/securesms/components/AnimatingToggle.java
```diff
@@ -45,12 +42,18 @@ public void addView(@NonNull View child, int index, ViewGroup.LayoutParams param
...
+    if (view != null)    animateIn(view, AnimationUtils.loadAnimation(getContext(), R.anim.animation_toggle_in));
+
+    current = view;
+  }
 
+  protected void setView(@Nullable View view) {
+    if (current != null) current.setVisibility(GONE);
+    if (view    != null) view.setVisibility(VISIBLE);
     current = view;
   }
```
moxie0: maybe this could be display(View, Boolean) and display(view) could call it with true?

mcginty: sure, renamed

moxie0:
ok lgtm, looks like it doesn't rebase into 2.27 cleanly though

mcginty:
rebased and merged into 2.27.0

PR #4057, opened 2015-09-09
re-introduce unit tests

mcginty:
Allow us to run unit tests with mocking independent of Android devices. Much faster and more foolproof.


moxie0:
in 2.27.0

PR #4063, opened 2015-09-10
download control details

### LINKED ISSUE #4027: Missing thumbnail for pending image download
If an **image** download is pending
there is only the download arrow (and maybe a text)
but no indication (thumbnail) of what will be downloaded.

For audio and video there are the specific built in icons/thumbnails. So maybe we should have that for images too.
Alternative would be to have an ultra low resolution image that is sent with the message as thumbnail.

mcginty:
Show the human-readable content type to the user when the download button appears (eventually showing size too assuming it will be added into the protobuf).

Also fun animations.

fixes #4027

transfercontrols


McLoo:
Will there also be a an optical refresh to the ~~ugly~~ old style "DOWNLOAD" button for MMS?
(That will hide, once the MMS expired?)

McLoo:
screenshot_2015-09-11-10-39-34 01

mcginty:
@McLoo yeah, I'd like to merge those but I think that's a separate PR's worth of diff :).

src/org/thoughtcrime/securesms/components/AnimatingToggle.java
```diff
@@ -60,6 +60,10 @@ protected void display(@Nullable View view, boolean animated) {
     current = view;
   }
 
+  protected @Nullable View getCurrent() {
+    return current;
```
moxie0: this shouldn't be exposed

src/org/thoughtcrime/securesms/components/TransferControlView.java
```diff
@@ -51,25 +69,88 @@ public void setSlide(final @NonNull Slide slide) {
...
+      downloadDetails.setText(slide.getContentDescription());
+      animateIn(downloadDetails);
     } else {
+      setVisibility(GONE);
       display(null, false);
     }
   }
 
+  private void animateIn(@NonNull final View view) {
+    showSmooth();
```
moxie0: some serious side effects

src/org/thoughtcrime/securesms/components/TransferControlView.java
```diff
@@ -79,7 +160,10 @@ public void onEventAsync(final PartProgressEvent event) {
       Util.runOnMain(new Runnable() {
         @Override public void run() {
           progressWheel.setInstantProgress(((float)event.progress) / event.total);
-          if (event.progress >= event.total) display(null);
+          if (event.progress >= event.total) {
+            hideSmooth();
```
moxie0: side effect city

src/org/thoughtcrime/securesms/mms/Slide.java
```diff
@@ -63,6 +63,8 @@ public boolean hasAudio() {
     return false;
   }
 
+  public @NonNull String getContentDescription() { return ""; }
```
moxie0: Is there a reason not to make this abstract?

mcginty: TextSlide didn't need to worry about it, but I can make it abstract and add an unused text description too if you'd prefer that.

moxie0: no this is fine if there's one non-implementer and there's nothing catastrophic about the default behavior

mcginty:
updated

src/org/thoughtcrime/securesms/components/TransferControlView.java
```diff
@@ -65,12 +92,67 @@ public void showProgressSpinner() {
...
+      anim.addListener(new AnimatorListener() {
+        @Override public void onAnimationStart(Animator animation) {}
+        @Override public void onAnimationCancel(Animator animation) {}
+        @Override public void onAnimationRepeat(Animator animation) {}
+        @Override public void onAnimationEnd(Animator animation) {
+          ViewUtil.animateIn(view, inAnimation);
+        }
+      });
+      anim.start();
+    }
```
moxie0: I think this method is probably too long, I can't follow what it does

mcginty:
simplified view logic

PR #4068, opened 2015-09-13
Update BUILDING Document

uberlaggydarwin:
Clarify the mailing list note.

This should make more sense.

// FREEBIE


moxie0:
thanks but i actually like the original wording

PR #4069, opened 2015-09-13
filter duplicate numbers

### LINKED ISSUE #4029: contacts with multiple accounts attached apear twice
After enabling "send sms with textsecure" i tried to select a contact to send the sms to.
Here i noticed that many contacts were appearing twice (or more even more often) which was not the case in the "original" contact app. They only appear twice in the "all contacts"-section, textsecure contacts are only once there.

This appears to be related to the count of numbers attached to a contact - having the mobile number + that contact as whatsapp-contact with the same number has the contact appear twice (with the same number) - if that contact is also on facebook (or has 2 numbers) that contact appears 4 times.

phone: oneplus one, stock 5.0.2 android (as delivered by oneplus), official textsecure app from the appstore (reinstalled today) (2.25.3).

xmatthias:
this fixes #4029 

ContactsContract does not allow to select distinct numbers or grouping functions. Therefore, the query is done as it was, and the duplicate-filtering is done afterwards creating a new cursor with only unique numbers.

### DIFF
--- a/src/org/thoughtcrime/securesms/contacts/ContactsDatabase.java
+++ b/src/org/thoughtcrime/securesms/contacts/ContactsDatabase.java
@@ -197,7 +197,29 @@ public class ContactsDatabase {
                                                        new String[] {"__TS"},
                                                        sort);
 
-    return new ProjectionMappingCursor(cursor, projectionMap,
+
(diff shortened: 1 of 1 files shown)


moxie0:
Thanks, but we can't a) iterate through the entire result set before displaying anything or b) load every single contact into memory

PR #4079, opened 2015-09-17
Listen for recipient changes in conversations and group updates.

moxie0:
This breaks out conversation "updates" (currently just group change info) into a separate list item class.  From there we'll be able to add other things (security indicators, call records, etc...).

It also resolves recipient contact info in the group update messages, and should fix the bug where recipient info sometimes doesn't resolve in conversations.


src/org/thoughtcrime/securesms/ConversationAdapter.java
```diff
@@ -85,13 +85,16 @@ public ConversationAdapter(Context context, MasterSecret masterSecret, Locale lo
...
   @Override
   public void bindView(View view, Context context, Cursor cursor) {
-    ConversationItem item       = (ConversationItem)view;
     long id                     = cursor.getLong(cursor.getColumnIndexOrThrow(SmsDatabase.ID));
     String type                 = cursor.getString(cursor.getColumnIndexOrThrow(MmsSmsDatabase.TRANSPORT));
     MessageRecord messageRecord = getMessageRecord(id, cursor, type);
 
-    item.set(masterSecret, messageRecord, locale, batchSelected, selectionClickListener,
-             groupThread, pushDestination);
+    if (view instanceof ConversationItem) {
```
mcginty: why not use `getItemViewType` to determine the view type?

src/org/thoughtcrime/securesms/ConversationAdapter.java
```diff
@@ -181,6 +184,8 @@ public void toggleBatchSelected(MessageRecord messageRecord) {
 
   @Override
   public void onMovedToScrapHeap(View view) {
-    ((ConversationItem)view).unbind();
+    if (view instanceof ConversationItem) {
+      ((ConversationItem) view).unbind();
```
mcginty: could we just have all views implement an Unbindable interface or something?

src/org/thoughtcrime/securesms/ConversationUpdateItem.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Override
+  public void onModified(Recipient recipient) {
+    Util.runOnMain(new Runnable() {
+      @Override
+      public void run() {
+        set(messageRecord);
+      }
+    });
+  }
```
mcginty: dupe code, throw in a private onModified() that's called from both callbacks?

src/org/thoughtcrime/securesms/util/GroupUtil.java
```diff
@@ -31,19 +34,47 @@ public static boolean isEncodedGroup(String groupId) {
     return groupId.startsWith(ENCODED_GROUP_PREFIX);
   }
 
-  public static String getDescription(Context context, String encodedGroup) {
+  public static GroupDescription getDescription(@NonNull Context context, @Nullable String encodedGroup) {
```
mcginty: result is also `@NonNull`

moxie0:
updated

Issue #3895, opened 2015-08-05
twitched direct capture to crash


McLoo:
As usual I had nothing better to do than testing how robust my display glass is right at the emoji and direct capture button, with rappppid tapping

Just did fast changes between camera and emoji and sometimes keyboard, as you can get from to log.

HTC One S on CM 12.1
2.25.0 branch at https://github.com/WhisperSystems/TextSecure/commit/a820500ae010a5b77e6c456640961925d14a7c39

relevant log part:

```
08-05 23:55:20.455  10855-10855/org.thoughtcrime.securesms D/AndroidRuntime? Shutting down VM
08-05 23:55:20.479  10855-10855/org.thoughtcrime.securesms E/AndroidRuntime? FATAL EXCEPTION: main
    Process: org.thoughtcrime.securesms, PID: 10855
    java.lang.RuntimeException: startPreview failed
            at android.hardware.Camera.startPreview(Native Method)
            at org.thoughtcrime.securesms.components.camera.CameraView.startPreview(CameraView.java:313)
            at org.thoughtcrime.securesms.components.camera.CameraView.access$1200(CameraView.java:47)
            at org.thoughtcrime.securesms.components.camera.CameraView$4.onPostMain(CameraView.java:303)
            at org.thoughtcrime.securesms.components.camera.CameraView$4.onPostMain(CameraView.java:291)
            at org.thoughtcrime.securesms.components.camera.CameraView$SerializedAsyncTask$2.run(CameraView.java:465)
            at org.thoughtcrime.securesms.components.camera.CameraView$SerializedAsyncTask$3.run(CameraView.java:484)
            at android.os.Handler.handleCallback(Handler.java:739)
            at android.os.Handler.dispatchMessage(Handler.java:95)
            at android.os.Looper.loop(Looper.java:135)
            at android.app.ActivityThread.main(ActivityThread.java:5289)
            at java.lang.reflect.Method.invoke(Native Method)
            at java.lang.reflect.Method.invoke(Method.java:372)
            at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
            at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699)
08-05 23:55:35.359  10855-12039/org.thoughtcrime.securesms W/WebSocketConnection? Sending keep alive...
``` 

some more context:
```
[...]

PR #4083, opened 2015-09-17
catch Camera RuntimeExceptions

mcginty:
I was never able to reproduce a RuntimeException from the steps in #3895, but from the logs we know where it was happening. This should be a safer way of dealing with the case where a preview couldn't be started.


moxie0:
cameraview is out of control.  in 2.27.0

PR #4091, opened 2015-09-19
Adding Verizon MSISDN header

### LINKED ISSUE #3150: support additional MSISDN headers
apparently there's also a `x-vzw-mdn` for some verizon APNs, but the value is the number *without* the country code.

char32:
This patch adds the verizon MSISDN x-vzw-mdn, which
is the number without leading country code.

fixes  #3150

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/LegacyMmsConnection.java
+++ b/src/org/thoughtcrime/securesms/mms/LegacyMmsConnection.java
@@ -195,6 +195,7 @@ public abstract class LegacyMmsConnection {
 
   protected List<Header> getBaseHeaders() {
     final String number = TelephonyUtil.getManager(context).getLine1Number();
+    final String number_without_cc = number.substring(2, number.length()); // strip of country code
     return new LinkedList<Header>() {{
       add(new BasicHeader("Accept", "*/*, application/vnd.wap.mms-message, application/vnd.wap.sic"));
       add(new BasicHeader("x-wap-profile", "http://www.google.com/oha/rdf/ua-profile-kila.xml"));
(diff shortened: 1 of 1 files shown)


jtracey:
https://github.com/WhisperSystems/RedPhone/wiki/Code-style-Guidelines#comments

char32:
@jtracey Thanks for reviewing! I read through the coding guidelines. What exactly is against the rules with this patch? I'll fix it directly then.

jtracey:
Just to be clear, I'm not affiliated with OWS. But at the very least, your naming conventions and use of comments don't follow the coding guidelines.

src/org/thoughtcrime/securesms/mms/LegacyMmsConnection.java
```diff
@@ -195,6 +195,7 @@ protected CloseableHttpClient constructHttpClient() throws IOException {
 
   protected List<Header> getBaseHeaders() {
     final String number = TelephonyUtil.getManager(context).getLine1Number();
+    final String number_without_cc = number.substring(2, number.length()); // strip of country code
```
McLoo: Works only for one digit country codes. (+1 I guess)
What about the rest of the world?

moxie0: indeed, you're going to want to use PhoneNumberUtil here

src/org/thoughtcrime/securesms/mms/LegacyMmsConnection.java
```diff
@@ -203,6 +204,9 @@ protected CloseableHttpClient constructHttpClient() throws IOException {
       if (!TextUtils.isEmpty(number)) {
         add(new BasicHeader("x-up-calling-line-id", number));
         add(new BasicHeader("X-MDN", number));
+        
+        // adding Verizon header
```
moxie0: no comments please

char32:
Ok, I will remove the comment, Change the variable naming and add Support for 1, 2 and 3 chars of CountryCodes. I think I will have a patch ready by next weekend.

McLoo:
PhoneNumberUtil will do the country code work for you if you feed it correctly

moxie0:
Cool, please reopen when ready!

Issue #2333, opened 2015-01-13
Difficult to select some messages


johanw666:
After I upgraded from 2.3.3 to 2.4.0 I have difficulties selecting some messages. Not all messages - it seems to be just those messages where TextSecure recognises a phonenumber in the message text. When I long-press such a message and release my finger from the screen the selection disappears.

To select such a message I have to long press, don't release my finger, press another message in the thread so I select 2 messages and then short-press the last one so it leaves the selection.

Update: I tested this on 2 devices, one no-name cheap Android 4.2.2 which does not have this problem. A Sony XPeria Z1 Compact with 4.4.4 does have this problem.

McLoo:
Workaround: If you tap next to the bubble, or at least outside the phonenumber (or hyperlink) it works.

johanw666:
Yes, pressing outside the bubble works well. Thanks.Pressing just outside the phonenumber but within the bubble does not work on the Sony phone, that was why I reported it.

JavaJens:
I can confirm the issue for messages with an image only.
Also can confirm the workaround.

McLoo:
I could swear image long press has work with the PR. 
Besides that, I just noticed K9 Mail uses left swipe over an item to select it. Maybe a good alternative... @moxie0?

JavaJens:
If you already have an image select (e.g. by clicking next to the bubble) it works without a flaw to select by tapping the image, its just the first one...
I do like long tap and find it more intuitive than swiping, but maybe thats just more common for me

hebu:
Same issue with hypelink/phonenumber messages on my device. Good to know, that clicking outside the bubble is a workaround.

Device info from the logs:

> Device : samsung GT-I9305 (m3xx)
> Android: JSS15J.I9305XXUEMKC
> OS Host: SEP-108
> App    : TextSecure 2.4.1

jeremymasters:
Is it different in 2.4.2 now?

hebu:
No.

rhodey:
this is still reproducible on TS 2.6.4 and caused by the same behavior as #2488, ie:

1) press down
2) hold until long press event fires
3) move left or right even the slightest
4) release

2-4601:
Not sure if #2648 will fix this but I just found out that if you come fresh (i.e. from somewhere else) to a conversation you cannot outside-select a message with a link unless you first select another message at least once.

Demonstration:
[...]

PR #4102, opened 2015-09-21
move conversation list to RecyclerView

mcginty:
In preparation for only loading segments of a conversation by default. Gives us animations for free, and I also simplified our selection logic and long-click handling. This new version fixes that annoying bug that caused long-clicking a thumbnail to highlight then immediately unhighlight a conversation item.

### DIFF
--- a/src/org/thoughtcrime/securesms/BindableConversationItem.java
+++ b/src/org/thoughtcrime/securesms/BindableConversationItem.java
@@ -1,0 +1,17 @@
+package org.thoughtcrime.securesms;
+
+import android.support.annotation.NonNull;
+
+import org.thoughtcrime.securesms.crypto.MasterSecret;
+import org.thoughtcrime.securesms.database.model.MessageRecord;
+
+import java.util.Locale;
+import java.util.Set;
+
+public interface BindableConversationItem extends Unbindable {
+  void bind(@NonNull MasterSecret masterSecret,
+            @NonNull MessageRecord messageRecord,
+            @NonNull Locale locale,
+            @NonNull Set<MessageRecord> batchSelected,
+            boolean groupThread, boolean pushDestination);
+}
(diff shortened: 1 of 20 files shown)


brumsoel:
If I understood you correctly, this would close #2488 (and possibly #2333)

mcginty:
thanks @brumsoel, updated commit to include those issues.

src/org/thoughtcrime/securesms/BindableConversationItem.java
```diff
@@ -0,0 +1,17 @@
...
+
+import android.support.annotation.NonNull;
+
+import org.thoughtcrime.securesms.crypto.MasterSecret;
+import org.thoughtcrime.securesms.database.model.MessageRecord;
+
+import java.util.Locale;
+import java.util.Set;
+
+public interface BindableConversationItem extends Unbindable {
```
moxie0: Is there something else that's Unbindable that's not this?

mcginty:
yeah I was going to make ConversationListItem unbindable as well (which I just did in this PR now). The other option is to just make BindableConversationItem include unbind() and since ConversationListItem is the only class we inflate for the conversation list it needn't implement any interfaces to know what it can do.

src/org/thoughtcrime/securesms/BindableConversationItem.java
mcginty: yeah I was going to make ConversationListItem unbindable as well (which I just did in this PR now). The other option is to just make BindableConversationItem include unbind() and since ConversationListItem is the only class we inflate for the conversation list it needn't implement any interfaces to know what it can do.

moxie0:
not sure i follow, my question is why there are two interfaces.  is there something that implements Unbindable that doesn't implement BindableConversationItem?

mcginty:
Yeah, ConversationListItem.

moxie0:
ok then lgtm.  have you tested this EXTENSIVELY?

mcginty:
in 2.28.0

PR #4103, opened 2015-09-22
update appcompat dependencies

mcginty:
* Our "betterpickers" library had an ownership chance, so I switched us to their latest version.
* Updated AppCompat to the most recent version available before the current one (which requires a compile target of Marshmallow, which in turn has a huge set of consequences).
* Start adding the new support annotations to our code


moxie0:
is this necessary for one of your milestones?

mcginty:
this is good hygiene.

moxie0:
are the old dependencies tired?

mcginty:
apparently, because some people can no longer build due to missing betterpickers artifacts

PR #4108, opened 2015-09-22
initialize draft after security

mcginty:
I feel like this should have been this way from the beginning anyway, as if the security state of the recipient changes for any reason between a pause and resume, we want to re-evaluate whether we can still send the attachment.


moxie0:
iirc it used to be that way, and we found some problem with it in practice.  typing a message then adding an attachment caused problems iirc.

PR #4111, opened 2015-09-24
fix click interception logic to handle failures

mcginty:
click handling for failures was getting gobbled up by the new adapter, this fixes it.


src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -120,6 +120,10 @@ public ConversationItem(Context context, AttributeSet attrs) {
     this.context = context;
   }
 
+  @Override public void setOnClickListener(OnClickListener l) {
```
moxie0: @Override \n
public void

moxie0:
in 2.28.0

PR #4122, opened 2015-09-25
fix more batch select click edge cases

mcginty:
This seems to solve almost every case I could run into before this (with the exception of long-clicking on a download button not starting a batch selection). Also fixes that crash.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -161,6 +161,7 @@ public class ConversationItem extends LinearLayout
     mediaThumbnail.setThumbnailClickListener(new ThumbnailClickListener());
     mediaThumbnail.setDownloadClickListener(new ThumbnailDownloadClickListener());
     mediaThumbnail.setOnLongClickListener(passthroughClickListener);
+    mediaThumbnail.setOnClickListener(passthroughClickListener);
     bodyText.setOnLongClickListener(passthroughClickListener);
     bodyText.setOnClickListener(passthroughClickListener);
   }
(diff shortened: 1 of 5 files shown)


src/org/thoughtcrime/securesms/components/ThumbnailView.java
```diff
@@ -97,8 +107,10 @@ private ImageView getRemoveButton() {
   }
 
   private TransferControlView getTransferControls() {
-    if (transferControls == null) transferControls = ViewUtil.inflateStub(this, R.id.transfer_controls_stub);
-    return transferControls;
+    if (!transferControls.isPresent()) {
+      transferControls = Optional.fromNullable((TransferControlView)ViewUtil.inflateStub(this, R.id.transfer_controls_stub));
+    }
+    return transferControls.get();
```
moxie0: it can't be fromNullable if the very next thing you do is call get()

src/org/thoughtcrime/securesms/components/TransferControlView.java
```diff
@@ -65,6 +66,16 @@ public TransferControlView(Context context, AttributeSet attrs, int defStyleAttr
     this.expandedWidth   = getResources().getDimensionPixelSize(R.dimen.transfer_controls_expanded_width);
   }
 
+  @Override public void setFocusable(boolean focusable) {
```
moxie0: @Override \n
public void ...

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -228,7 +229,8 @@ private void setBubbleState(MessageRecord messageRecord, Recipient recipient) {
 
   private void setInteractionState(MessageRecord messageRecord) {
     setSelected(batchSelected.contains(messageRecord));
-    mediaThumbnail.setClickable(!shouldInterceptClicks(messageRecord));
+    mediaThumbnail.setFocusable(!shouldInterceptClicks(messageRecord) && batchSelected.isEmpty());
+    mediaThumbnail.setClickable(!shouldInterceptClicks(messageRecord) && batchSelected.isEmpty());
```
moxie0: As far as I can tell, for every condition where you call shouldInterceptClicks, you additionally check the state of batchSelected.  Makes me think that shouldInterceptClicks should be called something else and either encapsulate the desired batchSelected state as an argument, or not reference batchSelected at all.

Issue #4132, opened 2015-09-27
Lock screen with PIN entry not using FLAG_SECURE


yakovsh:
And therefore it is vulnerable to screen capture on Android 5.0 with MediaProjection API. Sample video shot with Telecine is attached.
![video](https://cloud.githubusercontent.com/assets/667009/10120952/9c8d5410-64a3-11e5-8029-dc2e81016d72.gif)

jodson:
I think that commit solve this problem 

https://github.com/jodson/TextSecure/commit/c0c39c05828a3dd18840af8ae345dd505474cd06

yakovsh:
There is also a second related issue involving the virtual keyboard as well:

https://code.google.com/p/android/issues/detail?id=129285

jodson:
Oh, i don't knew about this. Thank you for show this. So, to create a mechanism to solve that second problem, it's just not show the native keyboard. The password to unlock the screen can be a finger swipe.

What do you think?

yakovsh:
That could be one approach. Another approach I have seen that Google Wallet uses, is that they have a custom PIN entry screen that does not use the keyboard, and that PIN entry screen can be hidden with FLAG_SECURE.

jodson:
I thought in that approach. I am waiting my pull request be accepted. If that happen, I will try implement that custom PIN.

yakovsh:
Thanks!

PR #4135, opened 2015-09-27
Solving the issue #4132

jodson:
In the commit "f629744", I just added the flag_secure.
In the commit "380656d", I created an Intent to call the ConversationListActivity, instead of call super.onBackPressed(). I did this because I received some message of a friend, and I open the message through of the notification window and when I touched in back button, the application goes to background. I think is better to back to the ConversationListActivity like others similar applications does.


frankabjr:
The commit id "380656d" fix de issue #3739.

moxie0:
one change per pr please

jodson:
Sorry. I didn't know about it. In the next pull request I will remember.

Issue #3327, opened 2015-06-03
screen security setting does not apply to "submit debug log" UI


rhodey:
1) enable screen security
2) Settings > Advanced > Submit debug log
3) open recent activities view
4) notice that screen security didn't do its thing

PR #4145, opened 2015-09-29
Change the extends of LogSubmit from BaseActionBar to PassphraseRequi…

frankabjr:
Change the extends of LogSubmit from BaseActionBar to PassphraseRequired, in order to require passphrase to access LogSubmit screen when the TS is locked. 

Fix the issue #3327

### DIFF
--- a/src/org/thoughtcrime/securesms/LogSubmitActivity.java
+++ b/src/org/thoughtcrime/securesms/LogSubmitActivity.java
@@ -1,22 +1,23 @@
 package org.thoughtcrime.securesms;
 
 import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.support.v4.app.FragmentTransaction;
 import android.support.v7.app.ActionBarActivity;
 import android.view.MenuItem;
 import android.widget.Toast;
 
+import org.thoughtcrime.securesms.crypto.MasterSecret;
 import org.whispersystems.libpastelog.SubmitLogFragment;
 
 /**
  * Activity for submitting logcat logs to a pastebin service.
  */
-public class LogSubmitActivity extends BaseActionBarActivity implements SubmitLogFragment.OnLogSubmittedListener {
+public class LogSubmitActivity extends PassphraseRequiredActionBarActivity implements SubmitLogFragment.OnLogSubmittedListener {
   private static final String TAG = LogSubmitActivity.class.getSimpleName();
 
   @Override
-  protected void onCreate(Bundle icicle) {
-    super.onCreate(icicle);
+  protected void onCreate(Bundle state, @NonNull MasterSecret masterSecret) {
     setContentView(R.layout.log_submit_activity);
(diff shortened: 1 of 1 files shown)


mcginty:
We want users to be able to submit debug logs before they enter their password, so unfortunately this isn't the solution.

This comment I left on a (now closed) PR applies here too, I think: https://github.com/jodson/TextSecure/commit/f6297447c9db1c8dd388cb0cea106c36486b230a#commitcomment-13482394.

frankabjr:
I think that if debug information will not have sensitive information about the users anytime, we can only use the initializeScreenshotSecurity. What do think?

mcginty:
The only time it will not be protected is when a user explicitly goes to the settings and turns off screen security. It's on by default. All our activities should probably have the ability to be captured if the user really wants it to be (tutorials, debug screenshots, etc).

frankabjr:
Ok. So, if the BaseActionBarActivity already set the FLAG_SECURE, the issue #3327 will be fixed. Right?

mcginty:
yup

jodson:
Done

https://github.com/WhisperSystems/TextSecure/pull/4152

src/org/thoughtcrime/securesms/LogSubmitActivity.java
```diff
@@ -26,9 +27,7 @@ protected void onCreate(Bundle icicle) {
   }
 
   @Override
-  protected void onResume() {
-    super.onResume();
-  }
+  protected void onResume() { super.onResume(); }
```
mcginty: either leave it as it is or just remove this method if all it does is call super.onResume()

src/org/thoughtcrime/securesms/LogSubmitActivity.java
```diff
@@ -1,22 +1,23 @@
...
 import android.widget.Toast;
 
+import org.thoughtcrime.securesms.crypto.MasterSecret;
 import org.whispersystems.libpastelog.SubmitLogFragment;
 
 /**
  * Activity for submitting logcat logs to a pastebin service.
  */
-public class LogSubmitActivity extends BaseActionBarActivity implements SubmitLogFragment.OnLogSubmittedListener {
+public class LogSubmitActivity extends PassphraseRequiredActionBarActivity implements SubmitLogFragment.OnLogSubmittedListener {
```
mcginty: Like I said, we can't have this extend PassphraseRequiredActionBarActivity since we want users to be able to access this before a passphrase is entered.

src/org/thoughtcrime/securesms/LogSubmitActivity.java
frankabjr: I'm waiting for the response regarding PR #4152 's acceptance. In this case, the only thing that I'll do is remove this method.

mcginty: ahhh gotcha! let's close this in favor of #4152 since I think that should cover this issue too

Issue #3536, opened 2015-07-06
new conversation incorrectly displayed as unread


rhodey:
1) start conversation with A
2) send message to A
3) receive message from A
4) back out to the conversation list UI
5) notice that your conversation with A is shown as unread

PR #4146, opened 2015-09-29
Fixed issue #3536.

GilCol:
Updates the visibleThread in MessageNotifier when sending a new message in a new thread.

Update:
It also solves another bug related to #3536. When receiving a message after sending the first one in a new thread, a push notification was created (shouldn't create push since the thread is open).

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1206,6 +1206,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
 
     if (refreshFragment) {
       fragment.reload(recipients, threadId);
+      MessageNotifier.setVisibleThread(threadId);
     }
 
     fragment.scrollToBottom();


moxie0:
this should go in ConversationActivity

GilCol:
Done

mcginty:
this lgtm, assigning to moxie since he wrote the notifier stuff

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1117,6 +1117,7 @@ protected Void doInBackground(Long... params) {
   protected void sendComplete(long threadId) {
     boolean refreshFragment = (threadId != this.threadId);
     this.threadId = threadId;
+    MessageNotifier.setVisibleThread(threadId);
```
moxie0: This happens in a callback, when the thread might no longer be "visible."

GilCol:
Fixed the callback problem.

src/org/thoughtcrime/securesms/ConversationActivity.java
GilCol: What do you think of the updated solution? @moxie0

moxie0:
that looks appropriate

moxie0:
in 3.3.0

PR #4150, opened 2015-09-30
Back button feature

jodson:


moxie0:
thanks but we should probably do something like this instead: https://developer.android.com/training/implementing-navigation/temporal.html

PR #4154, opened 2015-10-01
Fix "MMS download failed" click intercept

### LINKED ISSUE #4153: Clicking "MMS download failed" launches MessageDetailsActivity instead of retrying download
A bug was introduced in 469c329 that intercepts clicks on "MMS download failed" MMS notifications and launches a `MessageDetailsActivity`, rather than attempting to download the MMS message.

**Steps to Reproduce**

1. Turn off cellular data
2. Coerce someone to send you an MMS message
3. Await the "MMS download failed" message
4. Turn on cellular data
5. Tap the "MMS download failed" message
6. Behold the `MessageDetailsActivity`, along with the confusing `RESEND` button

issue-4153-screenshots

Device info: Nexus 5, stock Android 5.1.1 LMY48M, TextSecure 2.28.1 (ea50714)

App state: TextSecure messages registered, SMS enabled, default SMS app

unrulygnu:
Fixes #4153
// FREEBIE

Addressed only the single case in #4153 ; there may be other conditions where the click should not be intercepted.  Please advise if any such cases should be handled, and I'll update the PR.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -329,7 +329,7 @@ public class ConversationItem extends LinearLayout
 
   private boolean shouldInterceptClicks(MessageRecord messageRecord) {
     return batchSelected.isEmpty() &&
-           (messageRecord.isFailed() ||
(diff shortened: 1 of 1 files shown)


mcginty:
yep, that looks legit. tested. thanks for the PR!

PR #4192, opened 2015-10-03
Add GPL Header

bkerensa:


moxie0:
we'll get these as we do normal changes, thanks

PR #4199, opened 2015-10-04
Back button feature

jodson:


jodson:
@moxie0 

You told me that you will use something like this: 
http://developer.android.com/training/implementing-navigation/temporal.html

Is it better?

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -392,8 +392,10 @@ public boolean onOptionsItemSelected(MenuItem item) {
   @Override
   public void onBackPressed() {
     Log.w(TAG, "onBackPressed()");
-    if (container.isInputOpen()) container.hideCurrentInput(composeText);
-    else                         super.onBackPressed();
+    if (container.isInputOpen()) {
+      container.hideCurrentInput(composeText);
+    }
+    super.onBackPressed();
```
moxie0: why is this change lumped in? we want back just to clear the keyboard before jumping back

moxie0:
closing for lack of activity, feel free to reopen when ready

PR #4200, opened 2015-10-04
Make 'reset secure session' consistent

### LINKED ISSUE #4198: Wording of resetting secure sessions is probably inconsistent & definitely confusing
'**Reset secure session**' in the conversation menu leads to a dialog titled '**End secure session confirmation**' with the content '**Are you sure that you want to end this secure session?**'.
The conversation thread then just displays '**Secure session ended.**' Nevertheless, one immediately can send encrypted messages after 'ending' the secure session. It seems like a new secure session has been silently started afterwards, and the user is not informed about this fact.

As far as I understand it, the wording of '**Reset secure session**' is the one which is correct, i. e. the existing session gets closed and a new session with a new session key is created immediately afterwards. Is this correct?

If so, all involved strings except for '**Reset secure session**' are pretty confusing. They look like the secure session gets closed and that's it.

unrulygnu:
Fixes #4198
// FREEBIE

### DIFF
--- a/res/menu/conversation_secure.xml
+++ b/res/menu/conversation_secure.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <menu xmlns:android="http://schemas.android.com/apk/res/android">
       <item android:title="@string/conversation_secure_verified__menu_reset_secure_session"
-            android:id="@+id/menu_abort_session"/>
+            android:id="@+id/menu_reset_secure_session"/>
 </menu>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -90,8 +90,8 @@
     <!-- ConversationActivity -->
     <string name="ConversationActivity_initiate_secure_session_question">Initiate secure session?</string>
     <string name="ConversationActivity_initiate_secure_session_with_s_question">Initiate secure session with %s?</string>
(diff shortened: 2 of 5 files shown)


RiseT:
Thanks for fixing this!

Since this is one of only two strings in Signal, where a confirmation (unnecessarily) has the word 'confirmation' in its title:
Could you remove 'confirmation'?

unrulygnu:
I saw that and considered it, but since it's unrelated, it's probably better as a separate issue, along with Yes/No for the actions vs. Reset/Cancel.

RiseT:
I see.
Btw - do you happen to know the context of the following - somewhat related - strings?
- ConversationActivity_initiate_secure_session_question ('**Initiate secure session?**')
- ConversationActivity_initiate_secure_session_with_s_question ('**Initiate secure session with %s?**')
- conversation_insecure__menu_start_secure_session ('**Start secure session**')
- KeyExchangeInitiator_youve_already_sent_a_session_initiation_request_to_this_recipient_are_you_sure ('**You've already sent a session initiation request to this recipient, are you sure you'd like to send another? This will invalidate the first request.**')

They obviously aren't used here. Are these leftovers?

unrulygnu:
Not sure, maybe leftover, maybe deliberately left in for downstream.

PR #4235, opened 2015-10-10
Fix minor spelling mistake in RedPhone Constant

WhiteHatTux:
If not fixed, this might cause confusion in the future, because this intent might be exposed through Android Manifest not through a constant, but through copying the string.


moxie0:
in 3.1.0

Issue #2791, opened 2015-03-25
Export backups to a sub-folder.


Wicked0ne:
Right now on my phone TextSecure exports the backup to the main folder on my sdcard. It would be nice if it was exported to a sub-folder, something like /sdcard/TextSecure/backup.xml or even have a setting to allow users to select the folder.

The reason behind this request is that most "backup" apps don't allow backing up a single file. Such apps as BtSync or the open source competitor to BtSync called SyncThing. They act kinda as dropbox but you host the servers and folders they backup to. I update my rom somewhat frequently and like to save my text messages.

mcginty:
Yeah, we should probably organize all our sdcard files in a folder.

agrajaghh:
Saving the attachments in a TextSecure folder on the sdcard might be better as well (like Whatsapp, Threema, Telegram)

Wicked0ne:
I'm going to attempt fixing this. It's been a long time since I've written java. Will submit a pull request if I get it working.

TWfromSWD:
Same as https://github.com/WhisperSystems/TextSecure/issues/1901

SergejKern:
Is this still open?

McLoo:
That tells you the green field under the title, saying: "Open"

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #4236, opened 2015-10-11
Fixing the issue #2791 (Export backups to a sub-folder)

frankabjr:
This pr puts the plain text backup under the folder Signal/Backup/. On the import side, verify if old backup file exist (to oldest clients). Also, I've changed the EncryptedBackupExporter, to point to the new backup path. 

I didn't have tested the EncryptedBackupExporter because this is not working.


src/org/thoughtcrime/securesms/database/EncryptedBackupExporter.java
```diff
@@ -26,15 +26,23 @@
...
 
-public class EncryptedBackupExporter {
+public class EncryptedBackupExporter extends AbstractBackup {
+
+  public static void setup() {
+    File backupDirectory = getDirectoryPath();
+    if (!backupDirectory.exists()) {
+      backupDirectory.mkdirs();
+    }
+  }
```
moxie0: classes should not have public methods called setup() that take no arguments and return no value.  that's what a constructor is for, but in this case it seems like all you need to do is call mkdirs() when you are returning the value

frankabjr: I got it. Fixed in the commit 10314d0.

moxie0:
two space indents

frankabjr:
Ok. I'll close this pr and create a new with all changes in only one commit.

Issue #4237, opened 2015-10-11
Last message from blocked contacts should be blocked


Dyras:
So after being bombarded by a total idiot, I decided to block the creep. Said and done, no more messages. However, I still see the last message written by her on the "front page" of Signal.

The purpose of blocking someone is to make sure that you can't see their messages, so why not block the last message as well? I think Telegram does this, and I love it. Removing the chat completely isn't an option for me.

effigies:
Why not just remove the last message manually?

Dyras:
Yes, I asked myself that too while writing it. But I'm a saver. I want to save everything "just in case".

effigies:
So you want the message removed for you, but don't want to remove the message? Or you want the thread pushed down so that it's not among the first things you see?

McLoo:
Ublock 
Draft or send sth
Block 
Done

Dyras:
That's exactly what I want effigies. I want "BLOCKED" to be shown as an actual message under the person's name, thus not showing me the last message.

McLoo Yeah I'm fully aware of that. The reason I'm requesting this isn't because I'm brain dead or lazy, it's just that I think it would be a neat feature.

McLoo:
Ahh now I got it. Sry. 
Totally makes sense to me. Should be tagged EASY.

mcginty:
This might be solved when archiving functionality is added? If we want to specifically change the update to be "Blocked" it'd only really make sense to me if we added a "You blocked Total Idiot" meta-message into the conversation to be consistent.

samlanning:
How about rather than displaying a message like "Blocked" or "You blocked Total Idiot", what if we just display no text when blocked?

Dyras:
@samlanning As long as I know the user is blocked and I don't have to read their last message I really don't mind. But yeah, sounds good!

dmevdy123:
If I block someone on signal will the messages still appear on the other persons app/phone?

Dyras:
@dmevdy123 What messages? Something you've sent? Absolutely, because their device has already received the message.

dmevdy123:
@Dyras thank you

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #4239, opened 2015-10-11
Show "Blocked" instead of most-recent message as thread-heading message for blocked contacts

effigies:
Addresses #4237. Seemed pretty straightforward change (though it still needs a translatable string). Can't quickly set up a test to make sure it works, so this PR is an ad to anybody who wants to do a build to verify that it does what the reporter was looking for.


moxie0:
please do not post untested prs

PR #4246, opened 2015-10-14
fix crash from incorrect send job

mcginty:
A friend found themselves in a bad state by having an old TS, clicking on "resend" that was erroneously available from the bug described in #4153, and finding themselves in a bad state until they deleted their message in airplane mode.


moxie0:
this would need to be rebased onto 3.1.0

mcginty:
ok done

moxie0:
lol didn't even try to compile this huh?

mcginty:
didn't remember OutgoingMediaMessage was touched. fixed.

PR #4248, opened 2015-10-14
Refactor Part/PDU madness

moxie0:
I did the PartDatabase -> AttachmentDatabase class rename in a separate commit, so that the actual diffs would potentially be easier to follow.


mcginty:
Why are Slides and SlideDecks necessary any more?

moxie0:
We're close to being able to get rid of them and just use Attachments eveywhere. Right now they add a little bit of logic beyond what a strict database model knows about.  For instance, when a gif is added to the attachment database, a thumbnail is generated for it just like everything else.  

But if you use the thumbnail, it doesn't animate.  So the gif slide knows to use the full data instead of the generated thumbnail.  If we get rid of database-stored thumbnails, I think we'd be substantially there in terms of eliminating slide and slidedeck.

moxie0:
Or I guess that gif logic could go into ThumbnailView

mcginty:
we can work that in a separate PR since this one is already crazy big

moxie0:
in 3.1.0

PR #4250, opened 2015-10-14
in-conversation invite reminder

mcginty:
I don't like past-@mcginty's design of the reminder view so I reworked it to look a bit more like the material-friendly Snackbar. Too many icons (especially incongruently designed ones) are a bad thing - this way looks cleaner.

Some other small tweaks to reduce code or improve UX:
* moved from using a Future in initializeSecurity to having an onSecurityChanged() callback in the activity
* the build expiration reminder now has an "upgrade" button that launches our play store page
* when appending the invite, move the cursor to the end to replicate how pasting works in android.

the invite text is placeholder ;P


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -806,14 +792,23 @@ protected void onPostExecute(Pair<Boolean, Boolean> result) {
...
-
-        future.set(true);
+        onSecurityUpdated();
       }
     }.execute(recipients);
+  }
 
-    return future;
+  private void onSecurityUpdated() {
+    initializeDraft();
```
moxie0: this would call initializeDraft for every security update broadcast, rather than just once on create

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1459,4 +1455,28 @@ public void onAttachmentChanged() {
     updateToggleButtonState();
   }
 
+  private class ShowInviteReminderTask extends AsyncTask<Recipients, Void, Pair<Recipients,Boolean>> {
+    @Override protected Pair<Recipients, Boolean> doInBackground(Recipients... recipients) {
```
moxie0: newline

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1459,4 +1455,28 @@ public void onAttachmentChanged() {
...
+      if (recipients.length != 1 || recipients[0] == null) throw new AssertionError("task needs exactly one Recipients object");
+
+      Optional<RecipientsPreferences> prefs = DatabaseFactory.getRecipientPreferenceDatabase(ConversationActivity.this)
+                                                             .getRecipientsPreferences(recipients[0].getIds());
+      return new Pair<>(recipients[0], prefs.isPresent() && prefs.get().hasSeenInviteReminder());
+    }
+
+    @Override protected void onPostExecute(Pair<Recipients, Boolean> result) {
```
moxie0: newline

src/org/thoughtcrime/securesms/recipients/Recipients.java
```diff
@@ -321,6 +321,19 @@ void setStale() {
...
+
+    Recipients that = (Recipients)o;
+
+    return recipients.equals(that.recipients);
+  }
+
+  @Override public int hashCode() {
+    return recipients.hashCode();
+  }
+
```
moxie0: what is all this for?

mcginty: when we finish querying the recipient preferences in an AsyncTask we want to be sure that the current conversation recipients are the same as the one this task was intended for.

mcginty:
updated. initializeSecurity() will call onSecurityUpdated() whenever there is a security change (what we'd want for the invitation link if recipient becomes a push user), but gives back a future for initializing the draft only in onCreate() and onNewIntent().

src/org/thoughtcrime/securesms/recipients/Recipients.java
```diff
@@ -321,6 +321,21 @@ void setStale() {
...
+    Recipients that = (Recipients)o;
+
+    return recipients.equals(that.recipients);
+  }
+
+  @Override
+  public int hashCode() {
+    return recipients.hashCode();
+  }
```
moxie0: if the definition of equality here is that the LinkedList is identity equals between both objects, then it seems like you could just check identity equals of the recipient objects themselves rather than overriding this

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1459,4 +1468,31 @@ public void onAttachmentChanged() {
     updateToggleButtonState();
   }
 
+  private class ShowInviteReminderTask extends AsyncTask<Recipients, Void, Pair<Recipients,Boolean>> {
```
moxie0: is making this static out of the question? like by passing in the reminderview to the constructor?

mcginty: Since it's important to check that the current recipients are the same as the one we started with once the task is finished, it seemed like this should be non-static.

moxie0:
in 3.1.0

PR #4251, opened 2015-10-14
stop listening on old recipient changes

mcginty:
It appears we were "leaking" and creating race conditions by listening to multiple recipients at the same time in the case of receiving a new intent on a ConversationActivity instance that isn't finished.

I haven't been able to consistently reproduce the issue with the wrong contact showing up in the title bar, but since everyone I know who's experienced it has opened from a notification, this would make sense.


moxie0:
in 3.1.0

PR #4252, opened 2015-10-15
only load partial conversation by default

### LINKED ISSUE #3911: Sending mssages to this specific contact almost always crashes textsecure
I have a conversation with this friend of mine that recently has been causing a lot of problems. (This happened on 2.24 too, so it shouldn't be 2.25-related)
So basically:
1. I type a message
2. Hit send
3. Wait...
4. Sometimes the message bubble appears, sometimes it crashes before it does
5. I reopen ts, unlock it, go to the conversation and the message bubble is there. The message is also correctly marked as send and delivered

debug log: https://gist.github.com/a3061a1f4b7edfcfdcab

mcginty:
fixes #3911


moxie0:
in 3.1.0

PR #4257, opened 2015-10-15
unjankify incoming media sharing

### LINKED ISSUE #3989: Cannot share photos from Google Photos
I have a Nexus 6 with stock android 5.1.1, the problem is always reproducible. The steps are:

1. Take a photo with the phone camera;
2. Go to the Photos app, open the image and "share it to TextSecure";
3. TextSecure opens. Chose a recipient. The image appears in the conversation;
4. Press "send". The message disappears, nothing is sent.

Here is the debug log:

https://gist.github.com/anonymous/32aae2e2352b96d4bf40

The initial logged crash is probably not related to this, but to another issue that I'm going to report right after this.

mcginty:
There were some incorrect assumptions about how long we're allowed to read from a Uri passed in via a share intent to us. Turns out not long at all - once the activity that receives the intent finishes, the Uri is no longer guaranteed to be readable.

ShareActivity now writes a copy of the media to our own storage while the media remains relevant.

fixes #3989

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
+++ b/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
@@ -111,12 +111,12 @@ public class AttachmentManager {
     new AsyncTask<Void, Void, Slide>() {
       @Override
       protected void onPreExecute() {
-        if (uri != captureUri) cleanup();
-        else                   cleanupCurrentSlide();
-
         thumbnail.clear();
         thumbnail.showProgressSpinner();
         attachmentView.setVisibility(View.VISIBLE);
+        
+        if (!uri.equals(captureUri)) cleanup();
+        else                         cleanupCurrentSlide();
       }
 
       @Override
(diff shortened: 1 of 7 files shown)


src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -84,26 +87,36 @@ public void onAnimationEnd(Animation animation) {
...
+      cleanupCurrentSlide();
+    }
+  }
+
+  private void cleanupCurrentSlide() {
+    if (getSlideUri() != null && PersistentBlobProvider.isAuthority(getSlideUri())) {
+      PersistentBlobProvider.getInstance(context).delete(getSlideUri());
+    }
+    slide = Optional.absent();
   }
```
moxie0: I like the direction you're going here, but I don't understand why we can't simplify it all the way.  Just store the slide, no reference to the captureUri at all?

mcginty: I tried without it, but it still seems necessary to keep track of the capture intent state without erasing the current working slide (if they hit back after trying to capture via intent, we want the previous slide to not go missing).

moxie0: Hmm, I think something has to get consolidated, there are too many different clear/cleanup methods =)

Issue #4074, opened 2015-09-15
contact name doesn't update after adding contact 


heyimgay:
it requires that you manually exit the conversion and reenter

heyimgay:
probably related to #3711 but it's not system, I have that disabled

XertroV:
I can confirm this is occuring on my G4. I have SMS on. It is triggered regularly.

When I receive a message and open via the notification the conversation has a name mismatch. The heading contains the name of the previous recipient, but the user-icon in the chat accompanying each message reflects the current recipient.

heyimgay:
Thought it was fixed but still happened on 3.7.2

agrajaghh:
Its fixed for me when using "Add to contacts" from the menu, but its not fixed if you add the contact via the avatar (only way in group conversations)

before and after adding to contacts:
![dbc77d64-a9ad-11e5-9380-43574d7a208c](https://cloud.githubusercontent.com/assets/5296073/11984164/4ab9d2fc-a9b9-11e5-9b5d-7b19c066c0be.png)

McLoo:
@agrajaghh "(only way in group conversations)" besides the member list? But I guess it behaves equally...

agrajaghh:
@McLoo wow didn't know you can tap the contacts in the memberlist :grin: This "thing" needs a redesign I guess. Maybe I'll try it during the holidays...

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #4259, opened 2015-10-16
Fixed issue #4074

GilCol:
Fixed issue #4074.

reloadList() is used in order to update the color of the message bubble and identifier (only the action bar color and title is updated without this method call).
I discussed this solution with a friend and we couldn't agree whether this is a good approach or not. Do you have any comments?

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -163,6 +163,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   private static final int PICK_CONTACT_INFO = 4;
   private static final int GROUP_EDIT        = 5;
   private static final int TAKE_PHOTO        = 6;
+  private static final int ADD_CONTACT       = 7;
 
   private   MasterSecret          masterSecret;
   protected ComposeText           composeText;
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -337,6 +338,11 @@ public void onActivityResult(int reqCode, int resultCode, Intent data) {
         setMedia(attachmentManager.getCaptureUri(), MediaType.IMAGE, true);
       }
       break;
+    case ADD_CONTACT:
+      recipients = RecipientFactory.getRecipientsForIds(ConversationActivity.this, recipients.getIds(), true);
+      recipients.addListener(this);
```
moxie0: what are these lines for? what it returns should be identity equals to what we already have, no?

GilCol: @moxie0 
No it is not the same. The recipient object is a new object that has more fields populated (e.g. name, contactUri). We need to therefore get the new (updated) recipient and work with it.

I guess you could work with the same recipient object and update the fields instead of creating a new object, but that's not how it is set up in this project.

moxie0:
in 3.8.0

PR #4260, opened 2015-10-16
Fix two message ordering issues.

### LINKED ISSUE #2385: Non-chronological message order
From time to time, I'm receiving messages in non-chronological order. In rare cases, messages are not transmitted at all which is now better to control since we have the reception confirmation. IMO these things have to do with the server implementation but could be solved in the client and protocol by referencing to the last message ID or similar.

Using the latest Android client available on Google Play store.

### LINKED ISSUE #3618: Offline group messages arrive out of order
when I turn my internet on and lots of new messages are coming in, they are not shown in chronological order. 
I am using Android 2.3.7

moxie0:
1) The group ID for jobs that process received messages was
   previously set to the sender's e164.  This guaranteed
   serialization of messages per-recipient, while allowing
   processing of multiple recipients in parallel.  Unfortunately
   in the case of groups, this results in out of order
   conversations, since the "sender" for each message is
   different.  And we can't determine that it was a group
   message until *after* we process it.  So this change just
   puts all message processing from all senders in one big queue.

2) Synchronization messages were always being displayed before
   received messages, due to the "received time" for those
   being set to the time they were sent.

Fixes #3618
Fixes #2385

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/jobs/PushDecryptJob.java
+++ b/src/org/thoughtcrime/securesms/jobs/PushDecryptJob.java
@@ -82,7 +82,7 @@ public class PushDecryptJob extends ContextJob {
   public PushDecryptJob(Context context, long pushMessageId, long smsMessageId, String sender) {
     super(context, JobParameters.newBuilder()
                                 .withPersistence()
-                                .withGroupId(sender)
+                                .withGroupId("__PUSH_DECRYPT_JOB__")
                                 .withWakeLock(true, 5, TimeUnit.SECONDS)
                                 .create());
     this.messageId    = pushMessageId;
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/jobs/PushDecryptJob.java
```diff
@@ -82,7 +82,7 @@ public PushDecryptJob(Context context, long pushMessageId, String sender) {
   public PushDecryptJob(Context context, long pushMessageId, long smsMessageId, String sender) {
     super(context, JobParameters.newBuilder()
                                 .withPersistence()
-                                .withGroupId(sender)
+                                .withGroupId("__PUSH_DECRYPT_JOB__")
```
mcginty: not using `PushDecryptJob.class.getCanonicalName()` because you don't want it to change if the class name changes and there are persistent jobs?

mcginty:
code lgtm, didn't run it though so trusting it's tested :)

src/org/thoughtcrime/securesms/jobs/PushDecryptJob.java
moxie0: yep

PR #4262, opened 2015-10-16
Display thumbnail previews for images in conversation list.

moxie0:
conversation-list-thumbnails

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationListItem.java
@@ -25,12 +25,16 @@ import android.os.Build.VERSION;
 import android.os.Build.VERSION_CODES;
 import android.os.Handler;
 import android.support.annotation.DrawableRes;
+import android.support.annotation.NonNull;
 import android.util.AttributeSet;
+import android.view.View;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
 
 import org.thoughtcrime.securesms.components.AvatarImageView;
 import org.thoughtcrime.securesms.components.FromTextView;
+import org.thoughtcrime.securesms.components.ThumbnailView;
+import org.thoughtcrime.securesms.crypto.MasterSecret;
 import org.thoughtcrime.securesms.database.model.ThreadRecord;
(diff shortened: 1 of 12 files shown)


mcginty:
Since `ThreadDatabase.update()` is called every single time a message is inserted or updated I'm worried about the extra time this will add to our latency in messages appearing in the conversation on slow-I/O devices. Have you checked how much time this adds onto the `insertMessageOutbox` operation for example?

moxie0:
The slidedeck query has always been running as part of the thread update operation, we're just checking the results of it.  It's an indexed query that I assume should be fast, but I just benchmarked how long the blocking operation for retrieving the slide takes in the thread update.  On my Nexus 5 it was between 1-6 milliseconds, which I'm pretty sure is the cost of the synchronized block.  =)

moxie0:
Also, this moves the thread update out of the message insertion commit, so the change should actually improve time between sent and cursor update.

src/org/thoughtcrime/securesms/ConversationListItem.java
```diff
@@ -136,6 +145,23 @@ public int getDistributionType() {
...
+
+      LayoutParams subjectParams = (RelativeLayout.LayoutParams)this.subjectView.getLayoutParams();
+      subjectParams.addRule(RelativeLayout.LEFT_OF, R.id.thumbnail);
+      this.subjectView.setLayoutParams(subjectParams);
+    } else {
+      this.thumbnailView.setVisibility(View.GONE);
+
+      LayoutParams subjectParams = (RelativeLayout.LayoutParams)this.subjectView.getLayoutParams();
+      subjectParams.addRule(RelativeLayout.LEFT_OF, 0);
+      this.subjectView.setLayoutParams(subjectParams);
```
mcginty: seems like we shouldn't be using a RelativeLayout parent if this is necessary to make it work

moxie0: i think relativelayout is appropriate, but you have to specify the leftof constraint dynamically in order to get the text snippet to terminate at the thumbnail without it terminating at the date when the thumbnail is gone.

PR #4268, opened 2015-10-20
Clarify app upgrade notification

patcon:
My brother is using textsecure and came to me saying that the app was saying it needed to "upgrade" or "update" and would be unusable until that had been done. But he had checked all the menus and settings, and couldn't find out how to do this. In the meantime, he was locked out of messaging. I knew that it meant the had to open up the Play Store and upgrade the app, but he wasn't aware. (He'd turned off auto-updates.)

Perhaps it would be sensible to mention that the upgrade happens through the app store, as he didn't understand and was quite frustrated.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -1037,7 +1037,7 @@
 
     <!-- reminder_header -->
     <string name="reminder_header_expired_build">Your build of Signal has expired!</string>
-    <string name="reminder_header_expired_build_details">Messages will no longer send successfully, please update to the most recent version.</string>
+    <string name="reminder_header_expired_build_details">Messages will no longer send successfully, please update via the app store.</string>
     <string name="reminder_header_sms_default_title">Use as default SMS app?</string>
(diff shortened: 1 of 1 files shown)


jeremymasters:
I think once the version is 90 days old it does this.  Maybe have your friend turn auto-update on for this one app.  Signal for android is not in the play store yet unless you are beta I believe.

patcon:
Removed "to most recent version", as that is extraneous information for anyone using an app store, for which it always updates to the most recent version. Also, seemed like "update" is preferred over "upgrade" in most of the strings, so left it as-is.

patcon:
jeremymasters No hey, I can definitely get him to turn that on, but I still feel the copy could be improved a bit for laypersons :)

jeremymasters:
Nice.  I need to learn how to develop android apps so I can contribute code instead of just my comments, lol.

mcginty:
@patcon I tweaked it a bit in master so there's now an "UPGRADE" button that will take you to the store. That should solve the confusion.

patcon:
:+1: you're awesome, jake

PR #4273, opened 2015-10-22
batch invitations

mcginty:
screenie


res/menu/text_secure_normal.xml
```diff
@@ -1,29 +1,25 @@
...
 
     <item android:title="@string/arrays__my_identity_key"
-          android:id="@+id/menu_my_identity"
-          android:icon="@android:drawable/ic_menu_view" />
+          android:id="@+id/menu_my_identity" />
 
     <item android:title="@string/text_secure_normal__menu_settings"
-          android:id="@+id/menu_settings"
-          android:icon="@android:drawable/ic_menu_preferences" />
+          android:id="@+id/menu_settings" />
```
moxie0: why are you getting rid of these GB icons?

res/layout/reminder_header.xml
```diff
@@ -34,29 +36,23 @@
...
+                android:textColor="@color/white"
+                android:text="@string/reminder_header_close_button" />
+
+        <Button android:id="@+id/accept"
+                style="@style/Widget.AppCompat.Button.Borderless"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:textColor="@color/white"
+                tools:text="ENABLE" />
+    </LinearLayout>
```
moxie0: I'd like to go back to turning the entire view into the tap target, with an X in the top-right, FB-messenger style.

moxie0:
How did you solve the issue of only displaying non-ts users for sms invite?

res/menu/text_secure_normal.xml
mcginty: appcompat doesn't use them any more

mcginty:
I didn't.

moxie0:
that's a pretty essential part of this feature

res/layout/reminder_header.xml
mcginty: I'd like to get better graphic assets done before that happens, otherwise it looks pretty boring imo.

moxie0: i prefer what we have now over this, feels like a step backwards

mcginty: subjectively disagree but sure i'll go back to the original one

McLoo:
Line break before the URL ? Looks a bit scatt
ered...

mcginty:
The only way to cleanly do the filtering for TS users is by asynchronously checking whether the number is in the directory or not already and just disabling the checkmarks for those users. I don't know of a way to filter the query directly - the ways I tried didn't seem to have any effect.

@McLoo I like where you're going with this

PR #4285, opened 2015-10-23
batch invitations

mcginty:
screenie


src/org/thoughtcrime/securesms/contacts/ContactsDatabase.java
```diff
@@ -198,13 +206,37 @@ private void removeTextSecureRawContact(List<ContentProviderOperation> operation
...
+    final MatrixCursor matrix = new MatrixCursor(new String[]{ID_COLUMN,
+                                                              NAME_COLUMN,
+                                                              NUMBER_COLUMN,
+                                                              NUMBER_TYPE_COLUMN,
+                                                              LABEL_COLUMN,
+                                                              CONTACT_TYPE_COLUMN});
+    while (cursor.moveToNext()) {
+      final String number = cursor.getString(cursor.getColumnIndexOrThrow(NUMBER_COLUMN));
+      if (DirectoryHelper.getUserCapabilities(context, RecipientFactory.getRecipientsFromString(context, number, true))
+                         .getTextCapability() != Capability.SUPPORTED)
```
moxie0: This shouldn't have a reverse dependency up to DirectoryHelper. I'd prefer that this aggregation happen in the loader rather than down here in the db.

heyimgay:
should be https://sgnl.link, no? on that note what are the thoughts on self hosting something like https://project.polr.me?

mcginty:
rebased and addressed comments

PR #4292, opened 2015-10-24
Open conversation when clicking on a thumbnail

brumsoel:
The new conversation list thumbnails look really nice!
I noticed that when I click on them, nothing happens because the `ThumbnailView`'s `parentClickListener` isn't set. My expectation was that a click would open the conversation, I hope that's how it was meant to be.


moxie0:
Good catch, but the adapter shouldn't know anything about the children of the view, so I moved the propagation code into ConversationListItem.

brumsoel:
makes sense :+1:

PR #4294, opened 2015-10-24
Consolodate Base64 Classes

samlanning:
both `org.thoughtcrime.redphone.util` and `org.thoughtcrime.securesms.util` have a
`Base64` util class, however, other than whitespace changes, and the `securesms`
one having a couple of extra methods, they are the same.

This commit removes `org.thoughtcrime.redphone.util.Base64` and anything
referencing it now uses `org.thoughtcrime.securesms.util.Base64`


moxie0:
in 3.2.0

PR #4295, opened 2015-10-24
Change text of dialog box confirming blocking.

### LINKED ISSUE #4217: Old string for blocking phone numbers conflicts with recent Signal commit
Because of https://github.com/WhisperSystems/TextSecure/commit/e46caa024def3ef52e05ea41c6879f3f46c3b009, the string

RecipientPreferenceActivity_you_will_no_longer_see_messages_from_this_user ('**You will no longer see messages sent from this user.**')

should be changed in order to include messages and calls (e. g. 'communication').

samlanning:
Blocking users now also blocks incoming calls, so the block confirmation message has been updated.

fixes #4217

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -337,7 +337,7 @@
 
     <!-- RecipientPreferencesActivity -->
     <string name="RecipientPreferenceActivity_block_this_contact_question">Block this contact?</string>
-    <string name="RecipientPreferenceActivity_you_will_no_longer_see_messages_from_this_user">You will no longer see messages sent from this user.</string>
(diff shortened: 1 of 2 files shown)


mangei:
If you rename the key/name of a string value, ALL `string.xml` files for all files need to be adjusted. Use the refactor methods e.g. from Android Studio to be easily change such things.

mangei:
@mcginty : Am I right with my assumption or is it just necessary to update the default/english `string.xml` file?

agrajaghh:
@mangei no, just the english strings.xml should be edited, the translations are done via transifex

mangei:
@agrajaghh thanks for the correction!
@samlanning nevermind my comment ;)

PR #4305, opened 2015-10-26
Change string "Load" to "View"

### LINKED ISSUE #4304: "Load full conversation" string is implementation-detail-ish
In ae97495, a limited number of messages are loaded and displayed in a conversation, with an affordance labeled "Load full conversation" (`load_more_header__load_full_conversation`) provided to load and display the rest.

"Load" is a technical, implementation-detail-ish term which the user doesn't need to know. Platform apps use "view" or "show" in cases like this to clearly state to the everyday user the direct benefit of activating the affordance:  

* View full conversation
* Show full conversation
* View all messages
* Show all messages

unrulygnu:
Fixes #4304
// FREEBIE

Also removed the three full stops, in accordance with https://www.google.com/design/spec/style/writing.html :

    Ellipses
    Omit from menu items or buttons that open a dialog or start some other process.

### DIFF
--- a/res/layout/load_more_header.xml
+++ b/res/layout/load_more_header.xml
@@ -5,5 +5,5 @@
         android:layout_gravity="center"
         android:background="?conversation_item_bubble_background"
         android:textColor="?conversation_item_sent_text_primary_color"
-        android:text="@string/load_more_header__load_full_conversation" />
+        android:text="@string/load_more_header__view_full_conversation" />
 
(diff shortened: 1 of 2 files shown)


mcginty:
Great point. Although I think "View," at least to me, implies opening something in a dedicated viewer. I changed it to "See" which is shorter anyway, hopefully that makes sense.

Thanks much for the PR!

unrulygnu:
Awesome. :+1:   While translating, I came across a few other strings for which I'll post issues.  Not trying to split hairs, just seeking to improve the experience (strings are branding), at your discretion of course.

mcginty:
Confusing language is a serious usability bug, keep 'em coming! Thanks again.

PR #4306, opened 2015-10-26
Canonicalize entries in canonical address database.

moxie0:
Make all our queries E164, so eventually everything in there
will be E164.  Stops thrashing between formats.

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/database/CanonicalAddressDatabase.java
+++ b/src/org/thoughtcrime/securesms/database/CanonicalAddressDatabase.java
@@ -28,7 +28,10 @@ import android.text.TextUtils;
 import android.util.Log;
 
 import org.thoughtcrime.securesms.util.GroupUtil;
+import org.thoughtcrime.securesms.util.TextSecurePreferences;
 import org.thoughtcrime.securesms.util.VisibleForTesting;
+import org.whispersystems.textsecure.api.util.InvalidNumberException;
+import org.whispersystems.textsecure.api.util.PhoneNumberFormatter;
 
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/database/CanonicalAddressDatabase.java
```diff
@@ -46,25 +51,25 @@
...
-  private final Map<String, Long> addressCache = new ConcurrentHashMap<String, Long>();
-  private final Map<Long, String> idCache      = new ConcurrentHashMap<Long, String>();
+  private final Map<String, Long> addressCache = new ConcurrentHashMap<>();
+  private final Map<Long, String> idCache      = new ConcurrentHashMap<>();
 
-  public static CanonicalAddressDatabase getInstance(Context context) {
-    synchronized (lock) {
-      if (instance == null)
-        instance = new CanonicalAddressDatabase(context.getApplicationContext());
+  public synchronized static CanonicalAddressDatabase getInstance(Context context) {
```
mcginty: since we access this DB so much is it worth it to use the double-checked lock pattern here?

src/org/thoughtcrime/securesms/database/CanonicalAddressDatabase.java
```diff
@@ -134,20 +139,31 @@ public void close() {
...
-    return canonicalAddressId;
+      return canonicalAddressId;
+    } catch (InvalidNumberException e) {
+      throw new AssertionError(e);
+    }
   }
 
-  public List<Long> getCanonicalAddressIds(List<String> addresses) {
-    List<Long> addressList = new LinkedList<Long>();
+  public List<Long> getCanonicalAddressIds(@NonNull List<String> addresses) {
```
mcginty: returns `@NonNull`

src/org/thoughtcrime/securesms/database/CanonicalAddressDatabase.java
moxie0: it is never worth it, unless we profile and find it to be the primary bottleneck

mcginty:
in 3.2.0

Issue #4213, opened 2015-10-05
Unregistering dialogs refer only to messaging component


2-4601:
The unregister option in advanced settings has a description:

_"Free private **messaging and calling** to Signal users"_

which conflicts with the subsequent dialogs, e.g.:

_"This will disable Signal **messages** by unregistering you from the server. You will need to re-register your phone number to use Signal **messages** again in the future."_

I.e. the former talks about **messages and calls** while the latter talks only about **messages**.

![screenshot_2015-10-05-21-01-30](https://cloud.githubusercontent.com/assets/174176/10289497/3a121126-6ba7-11e5-8541-fcc2f6dcde27.png)

Affected strings:
```
preferences__use_the_data_channel_for_communication_with_other_signal_users
ApplicationPreferencesActivity_this_will_disable_signal_messages
ApplicationPreferencesActivity_disable_signal_messages
ApplicationPreferencesActivity_unregistering_from_signal_messages
```

3.0.1

2-4601:
Fixed in 3.2.0

PR #4311, opened 2015-10-27
"Signalize" some text strings

RiseT:
Adjusts a couple of text strings in order to include calls AND messages.

**Edit:**
If **title strings** like "Enable Signal calls and messages" or "Disable Signal calls and messages?" should prove to be too long, one could go for something like "Enable Signal communication" or "Disable Signal communication?" as alternatives.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -923,7 +923,7 @@
     <string name="preferences__language">Language</string>
     <string name="preferences__language_summary">Language %s</string>
     <string name="preferences__signal_messages_and_calls">Signal messages and calls</string>
-    <string name="preferences__free_secure_messages_and_calls">Free private messages and calls to Signal users</string>
+    <string name="preferences__free_private_messages_and_calls">Free private messages and calls to Signal users</string>
     <string name="preferences__submit_debug_log">Submit debug log</string>
     <string name="preferences__support_wifi_calling">\'WiFi Calling\' compatibility mode</string>
     <string name="preferences__enable_if_your_device_supports_sms_mms_delivery_over_wifi">Enable if your device uses SMS/MMS delivery over WiFi (only enable when \'WiFi Calling\' is enabled on your device)</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -18,10 +18,10 @@
         Disable lock screen for messages?
     </string>
     <string name="ApplicationPreferencesActivity_disable">Disable</string>
-    <string name="ApplicationPreferencesActivity_unregistering_title">Unregistering</string>
-    <string name="ApplicationPreferencesActivity_unregistering_text">Unregistering from Signal calls and messages...</string>
-    <string name="ApplicationPreferencesActivity_disable_signal_title">Disable Signal calls and messages?</string>
-    <string name="ApplicationPreferencesActivity_disable_signal_text">Disable Signal calls and messages by unregistering from the server. You will need to re-register your phone number to use them again in the future.</string>
+    <string name="ApplicationPreferencesActivity_unregistering">Unregistering</string>
+    <string name="ApplicationPreferencesActivity_unregistering_from_signal_calls_and_messages">Unregistering from Signal calls and messages...</string>
+    <string name="ApplicationPreferencesActivity_disable_signal_calls_and_messages">Disable Signal calls and messages?</string>
+    <string name="ApplicationPreferencesActivity_disable_signal_calls_and_messages_by_unregistering">Disable Signal calls and messages by unregistering from the server. You will need to re-register your phone number to use them again in the future.</string>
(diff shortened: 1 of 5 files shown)


unrulygnu:
From the translation perspective, naming the string resources with their context and purpose, rather than their content, helps tremendously.  This way, we have context to easily reference without digging through the code, and if the string's content changes, not changing the name also keeps the editing history and comments on Transifex.  Thanks!

I was going to open an issue on a couple of these strings.  Since you're touching them already, I'll chime in here.  If it's better to open a separate issue, please let me know.

        <string name="reminder_header_push_text">Upgrade your communication experience.</string>

For an infrequent or otherwise Signal-oh-yeah-what-does-it-do-again user, this description doesn't convey what upgrading the communication experience actually means, especially since "Secure" is no longer in the app name.  What are the benefits to the user?

I understand it's a fine line between too much and too little detail, but something like these might be better:

* Enjoy free secure calls and messaging with other Signal users.
* Upgrade your communication experience with free secure calls and messaging with other Signal users.

.

        <string name="reminder_header_invite_text">Take your communication with %1$s to the next level.</string>

For users who don't regularly interact with salespeople, CXO's, investors, and those who use "ask" as a noun (you know who you are! :stuck_out_tongue_winking_eye: ), "take to the next level" is really confusing. 

And like the string above, it doesn't convey to the casual user the benefits of inviting `%1$s` to Signal.  A couple of suggestions:

* Enjoy free secure calls and messaging with %1$s.
* Take your communication with %1$s to the next level with free secure calls and messaging.

RiseT:
@unrulygnu 
**Naming the string resources with their context and purpose, rather than their content**
I totally agree. I've been under the impression that they have to be named on the basis of their content (i. e. it is wanted that way), but I've noticed that e. g. "reminder_header_push_title" is using a different format, which I greatly prefer for the very reasons you've stated above.

**Content improvements**
I also agree with your suggestions regarding improvements of the strings' contents. I didn't change anything of that as my goal has been to merely correct rather than improve those strings.

One minor quibble:
I'd rather use "messages" than "messaging", because...
- it's in line with the wording of the existing string from c342f0d
- "message" is in line with "calls" regarding form
- quite frankly, translating "messaging" to German is a pain in the a.. :smile:

Apart from that, I will change the strings' contents as per your suggestions. Let's see what the boss says...

On a sidenote, I wonder if "UPDATE" rather than "UPGRADE" would be more appropriate for reminder_header_expired_build_button. I would be happy to change this within this PR if everybody agrees.

unrulygnu:
@RiseT: looks great!  Let's hope the boss agrees.

2-4601:
Fixes: #4213

moxie0:
For string format, I'd like to keep string names as content value.

For the content, "take your conversation to the next level" is exactly the tone we want. It's possible that we can improve on it, but that's the voice we want to use.

jeremymasters:
@moxie0 what about tone like "Signal - Because it's _your_ data" or something like that?  Maybe a little less computer-y word than "data".

RiseT:
@moxie0 
**Strings Names**
While I do understand the disadvantages of using the content value as the string name (especially for translators), I don't really understand the advantages. But I guess they obviously have to exist, so I'll change this.

**String Tone**
OK, I'll revert this. Just one question: Is "Take your **conversation** to the next level" appropriate for **calls and messages**, or would "Take your **communication** to the next level" be better?

**Reminder Expired Build**
For ***reminder_header_expired_build_button***, is this really **upgrading** or rather **updating**?

moxie0:
I think "conversation" is pretty good, but I think you're right that "update" is probably more appropriate.

RiseT:
@moxie0 Done.

RiseT:
Will reopen tomorrow. Just want to make my mind up on one tiny final thing.

RiseT:
OK, as far as I see it, this PR is complete now.

res/values/strings.xml
```diff
@@ -925,10 +922,8 @@
     <string name="preferences__default">Default</string>
     <string name="preferences__language">Language</string>
     <string name="preferences__language_summary">Language %s</string>
-    <string name="preferences__signal">Signal</string>
-    <string name="preferences__use_the_data_channel_for_communication_with_other_signal_users">
-        Free private messaging and calling to Signal users
-    </string>
+    <string name="preferences__signal_messages_and_calls">Signal messages and calls</string>
+    <string name="preferences__free_secure_messages_and_calls">Free secure messages and calls to Signal users</string>
```
moxie0: revert to private

RiseT:
@moxie0 Done.

moxie0:
in 3.2.0

RiseT:
@moxie0 Can I delete the branch? It says "Closed with unmerged commits", but to me it looks like it has been merged with master...?

moxie0:
yes

PR #4314, opened 2015-10-27
Notify when contacts join Signal

moxie0:
// FREEBIE


heyimgay:
any chance we can get a screenshot? I was thinking we could add a notification button to the FAB (floating action button) with the new contacts count and maybe include them at the top (not on first run)

res/xml/syncadapter.xml
```diff
@@ -3,6 +3,6 @@
               android:contentAuthority="com.android.contacts"
               android:accountType="org.thoughtcrime.securesms"
               android:userVisible="true"
-              android:supportsUploading="false"
+              android:supportsUploading="true"
```
mcginty: ?

moxie0: the system wasn't calling the sync adapter on a local modification within a reasonable amount of time otherwise

moxie0:
in 3.2.0

PR #4322, opened 2015-10-28
Warn user before build expires

### LINKED ISSUE #4712: Bad UX with expiring Signal versions
Signal can literally expire any second, so that out of a sudden you are not able to send messages anymore (well, of course only if you do not update it for 90 days).
Immediately updating the app right then might not be possible if you require a wifi connection.

I am sure that this is not a problem for any of the people involved here, but for other non tech-savvy users it can be (firsthand experience).

Reasons for apps not updating automatically:
- new permission required (happened before the RedPhone integration)
- not enough free space on the device available
- for some other reason... it is not uncommon for me to see 10 or so pending app updates on other peoples's devices

There should be a time-triggered warning, or some other trigger to inform the user beforehand.

cascheberg:
Give a user some time before ultimately expiring the build, fixes #4712.

// FREEBIE

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -1118,13 +1118,14 @@
     <string name="text_secure_normal__help">Help</string>
 
     <!-- reminder_header -->
-    <string name="reminder_header_outdated_build">Your version of Signal is outdated!</string>
+    <string name="reminder_header_outdated_build">Your version of Signal is outdated</string>
     <plurals name="reminder_header_outdated_build_details">
-        <item quantity="one">Your version of Signal will expire tomorrow. Please update to the most recent version.</item>
-        <item quantity="other">Your version of Signal will expire in %d days. Please update to the most recent version.</item>
+        <item quantity="one">Your version of Signal will expire in 1 day. Tap to update to the most recent version.</item>
+        <item quantity="other">Your version of Signal will expire in %d days. Tap to update to the most recent version.</item>
     </plurals>
+    <string name="reminder_header_outdated_build_details_today">Your version of Signal will expire today. Tap to update to the most recent version.</string>
     <string name="reminder_header_expired_build">Your version of Signal has expired!</string>
-    <string name="reminder_header_expired_build_details">Messages will no longer send successfully, please update to the most recent version.</string>
+    <string name="reminder_header_expired_build_details">Messages will no longer send successfully. Tap to update to the most recent version.</string>
     <string name="reminder_header_expired_build_button">UPDATE</string>
     <string name="reminder_header_sms_default_title">Use as default SMS app</string>
     <string name="reminder_header_sms_default_text">Tap to make Signal your default SMS app.</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -1120,7 +1120,7 @@
     <!-- reminder_header -->
     <string name="reminder_header_outdated_build">Your version of Signal is outdated!</string>
     <plurals name="reminder_header_outdated_build_details">
-        <item quantity="one">Your version of Signal will expire in one day. Please update to the most recent version.</item>
+        <item quantity="one">Your version of Signal will expire tomorrow. Please update to the most recent version.</item>
         <item quantity="other">Your version of Signal will expire in %d days. Please update to the most recent version.</item>
     </plurals>
     <string name="reminder_header_expired_build">Your version of Signal has expired!</string>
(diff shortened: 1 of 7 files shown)


mcginty:
In this case I think "real soon now" is probably too abstract and we should say "Your version of Signal will expire in %1$d days. Please upgrade."

In general suggest we migrate to "version" instead of "build" around this functionality, since I don't know that normies understand "build" is in this context.

cascheberg:
Thanks for your feedback. I thought of that, but I just wanted to improve this a bit and keep the changes simple, but I can work on that.

cascheberg:
@mcginty I updated the pull request according to your suggestion.

RiseT:
@mcginty I've discussed with Moxie elsewhere if it should rather be "update" than "upgrade" (he agreed). Is there some special reason that you're using the term "upgrade" in this case?

cascheberg:
I did not modify that part of the existing expiry strings, so the wording in this PR is still "update".

mcginty:
@RiseT "update" is fine, I was focusing on showing the number of days until Signal would become useless :).

res/values/strings.xml
```diff
@@ -1036,7 +1036,12 @@
     <string name="text_secure_normal__mark_all_as_read">Mark all read</string>
 
     <!-- reminder_header -->
-    <string name="reminder_header_expired_build">Your build of Signal has expired!</string>
+    <string name="reminder_header_outdated_build">Your version of Signal is outdated!</string>
+    <plurals name="reminder_header_outdated_build_details">
+        <item quantity="one">Your version of Signal will expire in one day. Please update to the most recent version.</item>
```
mcginty: "will expire tomorrow"

res/values/strings.xml
```diff
@@ -1036,7 +1036,12 @@
...
+        <item quantity="other">Your version of Signal will expire in %d days. Please update to the most recent version.</item>
+    </plurals>
+    <string name="reminder_header_expired_build">Your version of Signal has expired!</string>
```
mcginty: we should make these less excited sounding

src/org/thoughtcrime/securesms/components/reminder/OutdatedBuildReminder.java
```diff
@@ -0,0 +1,24 @@
+package org.thoughtcrime.securesms.components.reminder;
+
+import android.content.Context;
+import android.content.res.Resources;
+
+import org.thoughtcrime.securesms.R;
+import org.thoughtcrime.securesms.util.Util;
+
+public class OutdatedBuildReminder extends ExpiredBuildReminder {
```
mcginty: I think this should just be its own Reminder (not extending ExpiredBuildReminder) that sets its dismissability independently and calls the normal super(String, String) constructor.

cascheberg: I did this to inherit the play store update button, and I assumed inheriting the dismissability would be fine.
Also, setting up the plural strings in the super() constructor call is quite ugly.

res/values/strings.xml
cascheberg: I am open for suggestions of course

cascheberg:
@mcginty I updated the PR to make `OutdatedBuildReminder` not extend `ExpiredBuildReminder`, as you suggested. Please let me know if that is ok / better.

cascheberg:
@mcginty any new thoughts on this? @moxie0 maybe?

This is my first PR here, so I do not yet quite know how you handle things. But I don't want this to just lead to nothing.

Screenshot:
signal_expired

moxie0:
won't this display "will expire in 1 day" all the way up to the moment of expiration?

cascheberg:
Could you please explain your concern in more detail?
It should start at 10 days and count down until it says that it will expire "tomorrow" (which is not quite exact to be honest). I have conducted a short test of this feature.

moxie0:
I guess that's what I mean, it'll say "tomorrow" all the way until it expires, which could be in the middle of the day, right?

cascheberg:
That is correct. Would "less that one day" be ok? Or "any time now"?

moxie0:
how about "today?"

cascheberg:
I'd say that is really good.
nit-picking: I suggest to then slightly modify it so that it says:
3. will expire in 2 days
2. will expire in 1 day
1. will expire today
Does that make sense?

moxie0:
cool

cascheberg:
PR updated accordingly

moxie0:
in 3.8.0

PR #4324, opened 2015-10-28
Make 'notification sound' strings consistent

RiseT:
Make two inconsistent '**notification sound**' strings consistent with the other corresponding strings.

// FREEBIE

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -674,7 +674,7 @@
     <!-- recipient_preferences -->
     <string name="recipient_preferences__mute_conversation">Mute conversation</string>
     <string name="recipient_preferences__disable_notifications_for_this_conversation">Disable notifications for this conversation</string>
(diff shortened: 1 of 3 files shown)


res/values/strings.xml
```diff
@@ -871,7 +871,7 @@
     <string name="preferences__pref_led_blink_custom_pattern_on_for">On for:</string>
     <string name="preferences__pref_led_blink_custom_pattern_off_for">Off for:</string>
     <string name="preferences__pref_led_blink_custom_pattern_set">Custom LED blink pattern set!</string>
-    <string name="preferences__sound">Sound</string>
+    <string name="preferences__notification_sound">Notification sound</string>
```
moxie0: i'd like this to remain "sound" since it's under the notification section.  we don't preface "led color" with "notification led color" for instance

RiseT:
@moxie0 Done.

moxie0:
in 3.2.0

PR #4326, opened 2015-10-28
Refactor CameraView

mcginty:
* Entirely drop dependency on deprecated CWAC-Camera library
* Merged QuickCamera and CameraView since there wasn't a clean hierarchy of functionality as it was.
* selfie by default

Tested on every device I ran it on (2.3, 6.0) with no loss in functionality. I was careful around the orientation code and left it as is, since that's still relatively magic as far as I'm concerned. I'll take care of that in another PR.

(I left the commit log messy for now in case a bug is found and I need to bisect).


moxie0:
Cool, is this ready to be squashed into 3.2 then? Or multi-commit is fine if you need the history, just rewrite them to be meaningful.

mcginty:
Let me squash into discrete nice commits.

mcginty:
in 3.2.0

PR #4329, opened 2015-10-29
batch invite

mcginty:
(opened against 3.2.0)


src/org/thoughtcrime/securesms/components/reminder/ShareReminder.java
```diff
@@ -0,0 +1,36 @@
...
+    setOkListener(new OnClickListener() {
+      @Override public void onClick(View v) {
+        TextSecurePreferences.setPromptedShare(context, true);
+        context.startActivity(new Intent(context, InviteActivity.class));
+      }
+    });
+  }
+
+  public static boolean isEligible(final @NonNull Context context) {
+    return !TextSecurePreferences.hasPromptedShare(context);
```
moxie0: You don't think this should wait a little bit after an initial install?

mcginty: If I install I want to talk to my friends on Signal

moxie0: Maybe I'm an exception, but I don't think I would ever *mass invite* friends to an app before I'd done *anything* with it yet.  If I tap the new conversation fab, I'd want the ability to select an individual friend to invite to evaluate it with, but I probably wouldn't mass invite until I'd at least sent a message.

moxie0: ...this would also create a scenario where you potentially have three things in a row to immediately get through when you install. But I could be convinced otherwise.

agrajaghh: I agree. I think users get really easy annoyed in the beginning. Pushing them to invite others when they just want to get used to the app might not be a good idea

mcginty:
updated

moxie0:
looks like this needs to be rebased again

mcginty:
in 3.3.0

Issue #4278, opened 2015-10-23
Change String "Broadcast" to "Broadcast MMS" 


AdrianAulbach:
When I recieve a group MMS but use broadcast mode to reply two problems occur:

1. Even though my answer is plaintext only and sent to each recipient individually it still seems to get sent as MMS instead of SMS. This doesn't only increase costs but can also be an issue with some people not being able to read the message.
2. I message is sent to myself as well.

It would also be nice to have an option to reply only to the sender of the original message.

riyapenn:
1. Renaming issue to change string, to make it clear what broadcast does.

2. @AdrianAulbach Add a debug log immediately after this happens to https://github.com/WhisperSystems/TextSecure/issues/3936 if you are willing to test or if you come across this again.

franciscod:
close this?

2-4601:
@franciscod Why close? The commit was not accepted (#4331) or am I missing something obvious?

franciscod:
take a look here https://github.com/davidmascharka/TextSecure/commit/71ec62ea9c7e2c9fb1ab6d253fe4b8a7c908c51b

franciscod:
oh, it's from `master` branch... not this repo, but `davidmascharka`'s. my bad!

ZeekHuge:
`davidmascharka's` commit has been added to this repo, so probably this issue should be closed now ?

nrizzio:
71ec62e is only in the repository because a pull request was opened, but it was never was actually included into master (note the lack of any version numbers or branch names under the commit message). Moxie rejected #4331, so this should be closed.

> thanks but I think it's fine the way it is, we put 'MMS' in the compose textview

PR #4331, opened 2015-10-29
Update strings.xml to fix issue #4278

davidmascharka:
Change "Broadcast" in group messages to "Broadcast MMS" so the user knows that they will send an MMS message. Lets the user know of potential costs.


moxie0:
thanks but I think it's fine the way it is, we put 'MMS' in the compose textview

PR #4334, opened 2015-10-31
fix failed incoming MMS regression

mcginty:
happened when https://github.com/WhisperSystems/TextSecure/commit/db9656c70cbd65ea9cc84728bfbebf68c0678423 copied the indicatorText view from outgoing messages and ConversationItem previously relied on its existence (or lack thereof) to determine whether it should be capable of displaying it.


moxie0:
in 3.2.0

PR #4339, opened 2015-11-02
un-jank Google Photos sharing

mcginty:
(as well as sharing from many other apps, most likely)

We had incorrect assumptions about how temporary URI permissions worked when passed in via share intents.

This change copies over the file in ShareActivity to hold on to until the user dismisses it. Also includes some AttachmentManager refactoring and simplifications.

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
+++ b/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
@@ -38,9 +38,11 @@ import org.thoughtcrime.securesms.components.AudioView;
 import org.thoughtcrime.securesms.components.RemovableMediaView;
 import org.thoughtcrime.securesms.components.ThumbnailView;
 import org.thoughtcrime.securesms.crypto.MasterSecret;
-import org.thoughtcrime.securesms.providers.CaptureProvider;
+import org.thoughtcrime.securesms.providers.PersistentBlobProvider;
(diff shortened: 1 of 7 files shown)


src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -53,18 +55,17 @@
   private final @NonNull RemovableMediaView removableMediaView;
   private final @NonNull ThumbnailView      thumbnail;
   private final @NonNull AudioView          audioView;
-  private final @NonNull SlideDeck          slideDeck;
   private final @NonNull AttachmentListener attachmentListener;
 
-  private Uri captureUri;
+  private @NonNull  Optional<Slide> slide = Optional.absent();
+  private @Nullable Uri             captureUri;
```
moxie0: why do you still need this again?

mcginty: It seemed less of a hack than having AttachmentManager distinguish which slides had URIs that were yet-to-be-valid and we need some way to know that a capture is in progress and which URI is the one that we sent out as an intent to be captured.

I could imagine a few edge cases where the wrong media might be displayed through some fidgetous (should be a word) task switching, and figured this to be safer.

moxie0:
in 3.3.0

Issue #2971, opened 2015-04-13
CrashBrowser: ConversationActivity NPE - onNewIntent before onCreate 0.o


moxie0:
`````
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean org.thoughtcrime.securesms.mms.AttachmentManager.isAttachmentPresent()' on a null object reference
at org.thoughtcrime.securesms.ConversationActivity.onNewIntent(Unknown Source)
at android.app.Instrumentation.callActivityOnNewIntent(Instrumentation.java:1211)
at android.app.Instrumentation.callActivityOnNewIntent(Instrumentation.java:1223)
at android.app.ActivityThread.deliverNewIntents(ActivityThread.java:2456)
at android.app.ActivityThread.performNewIntents(ActivityThread.java:2468)
at android.app.ActivityThread.handleNewIntent(ActivityThread.java:2477)
at android.app.ActivityThread.access$1600(ActivityThread.java:151)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1375)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:135)
at android.app.ActivityThread.main(ActivityThread.java:5254)
at java.lang.reflect.Method.invoke(Native Method)
at java.lang.reflect.Method.invoke(Method.java:372)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
`````

mcginty:
did not know onNewIntent could be called before onCreate.

mcginty:
In fact I still can't say I understand this one. `onNewIntent` should only be called after an Activity has been created, paused, then is relaunched with a new intent. AttachmentManager is set in onCreate and never unset.

mcginty:
If it works like this, then pretty much all the logic in `onNewIntent` is incorrect, but Android's documentation contradicts what's actually happening here.

rhodey:
the only occurrence I'm aware of in which we mess around with Android's lifecycle ideas is [here, in DynamicLanguage](https://github.com/WhisperSystems/TextSecure/blob/master/src/org/thoughtcrime/securesms/util/DynamicLanguage.java#L25)

it's a stretch, but only guess I have at this time.

moxie0:
[...]

PR #4342, opened 2015-11-02
extra logging for activity lifecycle

mcginty:
to help debug #2971


src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
```diff
@@ -38,6 +38,7 @@
 
   @Override
   protected final void onCreate(Bundle savedInstanceState) {
+    Log.w(TAG, "onCreate(" + savedInstanceState + ")");
```
moxie0: the toString of Bundle is going to be safe to print?

moxie0:
in 3.3.0

PR #4367, opened 2015-11-03
Nicer looking attachment type selector

moxie0:


res/layout/attachment_type_selector.xml
```diff
@@ -0,0 +1,171 @@
...
+        </LinearLayout>
+
+        <LinearLayout android:layout_width="match_parent"
+                      android:layout_height="wrap_content"
+                      android:layout_weight="1"
+                      android:gravity="center"
+                      android:orientation="vertical"
+                      android:layout_marginLeft="16dp">
+
+        </LinearLayout>
```
mcginty: useless?

res/layout/attachment_type_selector.xml
```diff
@@ -0,0 +1,171 @@
...
+
+
+        <LinearLayout android:layout_width="match_parent"
+                      android:layout_height="wrap_content"
+                      android:layout_weight="1"
+                      android:gravity="center"
+                      android:orientation="vertical"
+                      android:layout_marginLeft="16dp">
+
+        </LinearLayout>
```
mcginty: useless?

res/layout/attachment_type_selector.xml
```diff
@@ -0,0 +1,171 @@
...
+
+        <LinearLayout android:layout_width="match_parent"
+                      android:layout_height="wrap_content"
+                      android:layout_weight="1"
+                      android:gravity="center"
+                      android:orientation="vertical"
+                      android:layout_marginLeft="16dp">
+
+        </LinearLayout>
```
mcginty: useless?

res/layout/attachment_type_selector.xml
```diff
@@ -0,0 +1,171 @@
...
+                    android:elevation="4dp"
+                    app:circleColor="@color/green_400"/>
+
+            <TextView android:layout_marginTop="10dp"
+                      android:layout_width="wrap_content"
+                      android:layout_height="wrap_content"
+                      style="@style/AttachmentTypeLabel"
+                      android:text="@string/attachment_type_selector__camera"/>
+
+        </LinearLayout>
```
mcginty: this is copy pasted enough where it should probably be its own custom view. It should be possible to have a single TextView with a compound upper drawable instead of an imageview and a textview inside a LinearLayout.

res/layout/attachment_type_selector.xml
```diff
@@ -0,0 +1,171 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
mcginty: since these rows are hardcoded, have you checked how this looks across particularly small and particularly large screens?

res/layout/attachment_type_selector.xml
moxie0: placeholders for the remaining bottom three attachment types. doing it this way makes the gravity for the bottom row align perfectly with the top row

mcginty:
What do you think about putting the popup above the composition text area so it's easier to cancel out of that popup in order to type something?

res/layout/attachment_type_selector.xml
moxie0: I'm fine with it being redundant within a single layout if there's no logic to compartmentalize, and it seems unlikely to be reused elsewhere.  Can't really use a compound drawable instead of an imageview, because it's not possible to adjust the padding between text and image independently

res/layout/attachment_type_selector.xml
moxie0: for the smallest and largest i have

moxie0:
I like the positioning the way it is, feels almost like a keyboard to me right now.

PR #4375, opened 2015-11-04
Update regular expression pattern for sms verification

rafael81:
In South Korea, SMS message include an extra text in front of body message in first line.

The extra text provide information that where message comes from.

So SMS verification cannot be passed in South Korea while registering.

It will cause failure of registration so that they can only choose ARS.

### DIFF
--- a/src/org/thoughtcrime/securesms/service/SmsListener.java
+++ b/src/org/thoughtcrime/securesms/service/SmsListener.java
@@ -38,7 +38,7 @@ public class SmsListener extends BroadcastReceiver {
   private static final String SMS_RECEIVED_ACTION  = Telephony.Sms.Intents.SMS_RECEIVED_ACTION;
   private static final String SMS_DELIVERED_ACTION = Telephony.Sms.Intents.SMS_DELIVER_ACTION;
 
-  private static final Pattern CHALLENGE_PATTERN = Pattern.compile(".*Your TextSecure verification code: ([0-9]{3,4})-([0-9]{3,4}).*");
+  private static final Pattern CHALLENGE_PATTERN = Pattern.compile(".*\\s*Your TextSecure verification code: ([0-9]{3,4})-([0-9]{3,4}).*");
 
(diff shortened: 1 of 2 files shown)


moxie0:
why doesn't the existing regular expression work? should already account for extra text in front of the message

rafael81:
Because it has new line in front of the body message like below.
>[국제발신]
>Your TextSecure verification code: 999-999
So the existing regular expression didn't match new line in prefix of body message.

moxie0:
'.' doesn't match newline? I don't think the regular expression you've included would work, since not every prepended value is guaranteed to end with whitespace.

rafael81:
'.' doesn't match newline. So I added "\s*" to the expression NOT "\s+".

2-4601:
Maybe I'm too late for this party but wouldn't the `DOTALL` flag be more general and more readable?
> In dotall mode, the expression . matches any character, including a line terminator. By default this expression does not match line terminators. [1]

Thus the pattern in this commit:
`private static final Pattern CHALLENGE_PATTERN = Pattern.compile(".*\\s*Your TextSecure verification code: ([0-9]{3,4})-([0-9]{3,4}).*");`
would be expressed as:
`private static final Pattern CHALLENGE_PATTERN = Pattern.compile(".*Your TextSecure verification code: ([0-9]{3,4})-([0-9]{3,4}).*", Pattern.DOTALL);`

This `DOTALL` version will also catch cases such as:
```
"XXX\nXXXYour TextSecure verification code: 1337-1337"
"XXX\nXXX\nYour TextSecure verification code: 1337-1337"
```

[1] https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#DOTALL

moxie0:
good idea

rafael81:
@2-4601 suggested good idea. Can I make the changes ?

2-4601:
@rafael81 already done: 8d4f3803dfbc660da0a822cd587e0e9b05480f78

rafael81:
Good to know :-)

PR #4381, opened 2015-11-04
Update the README's Travis build status image with the new name

aprescott:
TextSecure -> Signal. The current one shows a broken image. This one shows an "Unknown" for the build status, which is better, although Travis doesn't actually seem to know about Signal-Android yet.


aprescott:
This seems like something that doesn't need a CLA signature. Is one required to merge this?

moxie0:
afaik we've never used travis, so i'd be inclined just to delete it

aprescott:
:+1:

PR #4369, opened 2015-11-04
Replace 'my' with 'your'

RiseT:
// FREEBIE

**Description:**
Bringing strings in line with Google's Design Guidelines by replacing 'my' with 'your'.

**Quote:**
*Speak directly to users as “you,” not “me.” Don’t put words in the users’ mouths, with phrases that use “I” or “my.”*

**See**
https://www.google.com/design/spec/style/writing.html#writing-language

PR #4395, opened 2015-11-05
Replace 'my' with 'your' (rebased 3.3.0)

RiseT:
#4369 rebased against 3.3.0

// FREEBIE

**Description:**
Bringing strings in line with Google's Design Guidelines by replacing
'my' with 'your'.

**Quote:**
*Speak directly to users as “you,” not “me.” Don’t put words in the
users’ mouths, with phrases that use “I” or “my.”*

**See**
https://www.google.com/design/spec/style/writing.html#writing-language

### DIFF
--- a/res/menu/text_secure_normal.xml
+++ b/res/menu/text_secure_normal.xml
@@ -16,7 +16,7 @@
     <item android:title="@string/arrays__import_export"
           android:id="@+id/menu_import_export" />
 
-    <item android:title="@string/arrays__my_identity_key"
+    <item android:title="@string/arrays__your_identity_key"
           android:id="@+id/menu_my_identity" />
 
     <item android:title="@string/text_secure_normal__menu_settings"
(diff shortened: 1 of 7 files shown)


mcginty:
This looks good to me. Set for 3.4.0 so translators have time.

RiseT:
@mcginty As far as I know, the strings get pulled from Transifex (are pushed to Transifex...? Whatever) only after the PR has been committed to master. While I could be wrong, that's how it looked like to me in the past.

mcginty:
@RiseT that's right, but 3.3.0 has been moved to master and we don't want to add more strings in since its release to beta channel is imminent.

RiseT:
@mcginty Ah - ok. Thanks. I don't care if it is in 3.3.0 or 3.4.0, just wanted to mention it, as the time for translation has occassionally pretty short in the past (e. g. 1 hour).

moxie0:
in 3.4.0

PR #4317, opened 2015-10-27
Change 'SD card' to 'storage' & cosmetic corrections

RiseT:
Fixes #3920 

Change **'SD card'** in strings to **'storage'** for clarity & consistency.

Furthermore, I've done a couple of minor cosmetic string corrections.
I've decided to include these here rather than in a separate PR as there is some overlap with the 'SD card' stuff, and I'd like to save the translators from translating these strings twice.

- Change app's name from 'SMSBackup And Restore' to its official name **'SMS Backup & Restore'**
- Change 'Import encrypted backup' into '**Restore** encrypted backup' as this is the only option that replaces the complete database rather than add to it (which all the 'import' options do). 'restore' is the wording used in the description string as well.
- Remove **redundant periods** at the end of single sentences of some related description strings (in line with Google's design guidelines and the rest of Signal's description strings)
- Remove some **redundant line feeds** in XML (confusing to translators at Transifex)

PR #4404, opened 2015-11-05
Change 'SD card' to 'storage' & cosmetic corrections (rebased 3.3.0)

### LINKED ISSUE #3920: Wording: 'SD Card' vs. 'Storage'
The wording of several strings is inconsistent regarding the use of 'SD Card' and 'Storage', although all of these strings refer to the very same thing (as seen in the source code).

I'd like to do a small PR correcting that, but want to gather information first what term is prefered.

**Some things to consider:**
- As far as I know, TextSecure is not able to access a 'real' SD card when using KitKat or Lollipop.
- The term 'SD card' could be confusing to users with no SD card slot on their phone. The concept of a 'emulated SD card' could be even more confusing to those users ('There are no power users').

**Possible terms:**
- SD card
- storage
- device storage
- external storage (thx, @2-4601)
- anything else?

I'd appreciate your input!

RiseT:
Rebased against master (3.3.0)
Replaces #4317
Fixes #3920

Change **'SD card'** in strings to **'storage'** for clarity & consistency.

Furthermore, I've done a couple of minor cosmetic string corrections.
I've decided to include these here rather than in a separate PR as there is some overlap with the 'SD card' stuff, and I'd like to save the translators from translating these strings twice.

- Change app's name from 'SMSBackup And Restore' to its official name **'SMS Backup & Restore'**
- Change 'Import encrypted backup' into '**Restore** encrypted backup' as this is the only option that replaces the complete database rather than add to it (which all the 'import' options do). 'restore' is the wording used in the description string as well.
- Remove **redundant periods** at the end of single sentences of some related description strings (in line with Google's design guidelines and the rest of Signal's description strings)
- Remove some **redundant line feeds** in XML (confusing to translators at
Transifex)

### DIFF
--- a/res/layout/import_fragment.xml
+++ b/res/layout/import_fragment.xml
@@ -80,7 +80,7 @@
                 <TextView android:layout_width="match_parent"
                           android:layout_height="wrap_content"
                           style="@style/Registration.Description"
-                          android:text="@string/import_fragment__import_encrypted_backup"/>
+                          android:text="@string/import_fragment__restore_encrypted_backup"/>
 
                 <TextView android:layout_width="match_parent"
                           android:layout_height="wrap_content"
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -140,7 +140,7 @@
         <item quantity="other">This will permanently delete all %1$d selected messages.</item>
     </plurals>
     <string name="ConversationFragment_save_to_sd_card">Save to storage?</string>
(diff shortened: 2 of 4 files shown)


RiseT:
@mcginty You've added #4317 to the 3.4.0 milestone. I've closed that one - please add this instead. Rebased against 3.3.0/latest master per @moxie0.

mcginty:
done :)

moxie0:
in 3.4.0

PR #4407, opened 2015-11-06
Refactor group logic

mcginty:
This is just moving around the "business" logic, separate from UI refresh.

* `GroupManager` class (sibling of `MessageSender`) handles all network- and DB-related group management work and is nicely annotated this time to be more confidently NPE-safe.
* All the AsyncTasks are now static and have their variables explicitly passed into their constructor
* Fixed the cases of changing instance variables off the main thread since that was bound to lead to danger.
* The Adapter now keeps track of the members list, none of that weird "syncing" between the list in the activity and the adapter and the ArrayAdapter underneath that. Just a single BaseAdapter that has a list that manages it so the user sees what will be used in the logic.

### DIFF
--- a/src/org/thoughtcrime/securesms/GroupCreateActivity.java
+++ b/src/org/thoughtcrime/securesms/GroupCreateActivity.java
@@ -21,16 +21,12 @@ import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.graphics.Bitmap;
-import android.graphics.BitmapFactory;
 import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
-import android.text.Editable;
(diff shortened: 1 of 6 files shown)


src/org/thoughtcrime/securesms/GroupCreateActivity.java
```diff
@@ -547,145 +378,165 @@ protected void onProgressUpdate(Void... values) {
...
+                           Set<Recipient>      members)
+    {
+      this.activity     = activity;
+      this.masterSecret = masterSecret;
+      this.avatar       = avatar;
+      this.name         = name;
+      this.members      = members;
+    }
+
+    @Override protected void onPreExecute() {
```
moxie0: @ Override\n, always

mcginty: slipped through the cracks. my IntelliJ was still re-formatting them to be on the same line...

Issue #1911, opened 2014-09-13
Import/export of public keys with sharing interface


effigies:
This is similar to but not quite the same as #802, which seems to be aimed at importing/exporting key files. And it may be violating rule 3, so apologies in advance if this is deemed too power-usery.

Importing/exporting public keys via the sharing interface would be useful for allowing people to easily construct alternative out-of-band verifications.

For example, for people who have verified PGP keys in the past, but are not now close enough to verify TextSecure keys:

1) Share public key -> Sign/Encrypt with APG -> Share via Email/whatever
2) Receive -> Verify/decrypt with APG -> Share to TextSecure, importing the key

Right now, you can export your fingerprint QR code, sign, send and have them verify and scan it, but it's not quite smooth, and the QR seems like a complicating step.

Another possibility is making a URI protocol that TextSecure is assigned to, like `tsfpr:<PHONENUMBER>:<FINGERPRINT>`. This would resolve the import problem altogether, and could resolve something like #1016 in passing.

FeuRenard:
I assume that the equivalent of 2014's "public keys" are "safety numbers" (2016). The sharing interface is now implemented in several ways

- https://github.com/WhisperSystems/Signal-Android/commit/b55a7ff5c05adc0b2e5c5c0b2b54f116f964cec3: Add copy and compare safety numbers from clipboard actions
- https://github.com/WhisperSystems/Signal-Android/commit/92aeafc26172fbbce036cdf99b82f8313acb204b: Add share button for safety numbers

So if I understand this issue correctly, it will be resolved.

Issue #826, opened 2014-02-27
Fingeprint Exchange


IlonkaPingel:
Currently you can only confirm a fingerprint, if you read a QR-Code from anothers phone. This requires both chatters to be in the same place. This is at least the safest way, but there are other ways to do this nearly as secure as this. You could e.g. talk with someone on the phone, who dictates his fingerprint, so that you can confirm it by simply checking, if the code he's dictating matches the one, your phone is showing. So there should be an option, to confirm another ones fingerprint by simply telling the app "Yes, I've checked/confirmed the fingerprint of my opposite." This would be similar to the way PGP fingerprint exchanges can work.

Aeronaut:
Simply checking if the fingerprints are the same isn't enough! The way how it's implemented yet is practically unusable. If the fingerprint is altered by a man in the middle you would never now if you don't check frequently....

So doing it like in OTR: Saving each fingerprint and display the trusted state in Chat...

2-4601:
This seems to be about two different issues: 1) sharing the fingerprint and 2) saving the verification status
1) is a duplicate of #817
2) is the same as #910

PR #4426, opened 2015-11-06
Tap the local key to Share it

chadmiller:
This touches on #1911 and #826 .  When you're staring at your own
fingerprint, what should tapping on it do?

Nothing? That's dumb. In Android, We expect to be able to copy data
between apps, and a fingerprint is exactly the kind of thing a user
would want to share with someone.

So, tapping your local fingerprint now prompts you to send your
fingerprint to any app you have that accepts plain/text. If you already
have a communication channel you trust, this simplifies your universe,
and this makes it dead easy to publish it on Twitter et c.


moxie0:
thanks for thinking about this but we're going to revamp this whole flow soon

PR #4451, opened 2015-11-08
https invite links

heyimgay:
//FREEBIE


brumsoel:
Connecting to [sgnl.link via HTTPS](https://sgnl.link) is not possible.

heyimgay:
Indeed and as far as I can find it does not seem to be possible. one solution is to self-host a [polr](https://polr.me) instance, for instance.

moxie0:
thanks for your interest, but please don't submit prs that aren't functional

Issue #3354, opened 2015-06-06
swipe to delete messages


fazlearefin:
Need a feature to delete messages by swiping

agrajaghh:
I like the way gmail is doing it, without asking for confirmation but offering a "undo" action instead:
![undo](https://cloud.githubusercontent.com/assets/5296073/8019261/a73c56c4-0c45-11e5-88b9-3828fefd47a6.gif)

oaim:
When you say "messages", do you mean "conversations"?
I would rather want the possibility to archive conversations by swiping them.

fazlearefin:
By message I mean the full thread/conversation with a particular contact. Yes, archiving is also an additional possibility. Swiping from left to right (mimicing a cross out a work when writing on a piece of paper) would delete and swiping from right to left would archive a conversation.

aibaraiduas:
Swiping is a bit problematic because it makes it easier to delete conversations by accident than the current way of holding down the conversation then hitting the trash can icon.

At the very least, the user should always have to *actively confirm* that yes they do want to delete the conversation after swiping—not simply have an undo button that they might miss if they swiped by mistake when putting the phone into their pocket or something. The undo option in gmail and other email clients works fine only because there is a trash folder so the user can undelete a message if it was mistakenly deleted. With TextSecure there is no such thing.

ldematte:
I would like to give it a go. 
I would propose to implement it like gmail: let the user choose if swiping does nothing, archive, or delete. I would add a "undo" option, instead of active confirmation, as it seems more natural (and it follows Material design guidelines).

ChaoticMind:
This library might be useful https://github.com/hudomju/android-swipe-to-dismiss-undo

ldematte:
Thank you for the suggestion; I looked at the library, but I preferred to code support directly inside ConversationList*, and do not take an external dependency. Also, it works nicely with touch, long press and scroll (not an easy combination!)
I have a feature branch here: https://github.com/ldematte/TextSecure/tree/swipe_to_delete

So far, it works, but I have some questions:
- Archive, delete, or let the user choose? (Through preferences) (right now both swipe-to-left and swipe-to-right delete)
- Undo: right now it is just a delay for delete: an item is scheduled for delete in 5 seconds, if the user do press "Undo" deletion is cancelled. Alternative: alter conversation/thread tables to mark an item as deleted (like a "delete time"), periodically sweep tables.
[...]

Issue #1141, opened 2014-03-13
Archive


andypust:
I see the option of deleting messages and limiting conversation limits in the settings, so I don't know if the practice of regularly deleting messages is a practice in security. But I like to keep an archive of my messages, separate from my "current" messages. I think this is a pretty common feature in most messaging platforms. 

![archive](https://f.cloud.github.com/assets/6936150/2405528/f7775dba-aa52-11e3-9d05-21b7848dfd60.png)

manhole11:
Any update on this?

andy11:
Sorry for dropping out on this. 

I think I can re-implement these changes on the latest branch this weekend.

Two things that are still needed are:
- An image to use as a background image when there is nothing in your inbox
- Icons to use for "Archive" and "Move to inbox"

Can anyone help out with these?

agrajaghh:
For the background image: How about just a gray text saying something like:
> You don't have any messages yet.
> To write a message, click on the "+" in the bottom right corner.

These are the archive and inbox icons from https://github.com/google/material-design-icons/ :

ic_archive_black_48dp.png
![ic_archive_black_48dp](https://cloud.githubusercontent.com/assets/5296073/6091777/97f7c0ea-aebc-11e4-9316-2190296a91f7.png)

ic_inbox_black_48dp
![ic_inbox_black_48dp](https://cloud.githubusercontent.com/assets/5296073/6091778/9a0373e8-aebc-11e4-9e21-016275b04ede.png)

The archive icon looks good to me, but I'm not convinced by the inbox icon. What do you think?

andy11:
In my opinion, I think the "move to inbox" icon would be better with the arrow flipped (up arrow to me implies you're raising the priority which matches what I would typically want when I move something from the archive to the inbox). Flipping the arrow would also make it look more different than the "archive" button (initially I thought both those pictures were for the "archive" action). 

I like the idea of having background text instead of an image when the inbox is empty, so I'll go with that.

Doing merge now, hopefully will have it working again later today.

agrajaghh:
Or what about using the first icon (ic_archive_black_48dp) for both: arrow point down for "putting in archive" and arrow up for "getting out of archive"?

![97f7c0ea-aebc-11e4-9316-2190296a91f7](https://cloud.githubusercontent.com/assets/5296073/6093062/8f2b3392-aef9-11e4-8b0a-53774af0650d.png)

[...]

PR #4463, opened 2015-11-09
Swipe to delete

ldematte:
I took a shot at issue #3354 (swipe to delete).

For now, it implements the swiping gestures and events (inside ConversationListItem, following the same pattern used for click and long-press). It replaces the "old" click and long-press listeners, as swiping has complex interactions with touch events handling (and handling all of them in one place make it more consistent and easier). Gestures are modelled after the GestureDetector class; specific cases like multi-finger touches are handled too.

It also plays well with the parent RecycleView scrolling; I tried to make it work nicely and avoid to subclass it and it seems to have worked.
It reproduces as closely as possible gmail behaviour.

What is missing, and IMO separate, broader issues, is what to do after swiping: 
- delete, "archive", or let the user choose (with a "do nothiing" option too)? (Through preferences, maybe?) (right now both swipe-to-left and swipe-to-right do a "delete")
- Undo: right now it is just a delay for delete: an item is scheduled for delete in 5 seconds, if users press "Undo", deletion is cancelled. An interesting alternative is to do not actually delete the message(s): alter MessagingDatabase tables to mark an item as deleted (add something like a "delete time"); then, periodically sweep tables.
- This second option may be the starting point for an "archive" or recycle bin functionality (which I believe would be an entirely new (interesting) feature/issue).

I tried to stay as close to your coding style as possible.
Let me know what you think!


build.gradle
```diff
@@ -32,6 +32,9 @@ repositories {
     maven { // cwac-camera
         url 'https://repo.commonsware.com.s3.amazonaws.com'
     }
+    maven { //material-dialogs
```
agrajaghh: this shouldn't be in this PR I guess

build.gradle
```diff
@@ -209,6 +211,7 @@ android {
                           'proguard-shortcutbadger.pro',
                           'proguard-retrofit.pro',
                           'proguard.cfg'
+            applicationIdSuffix ".debug"
```
agrajaghh: same here

build.gradle
ldematte: I only modified the build.gradle to make it compile with com.afollestad:material-dialogs (there was a dep problem, already pointed out in a separate issue).
I probably should exclude the whole file from the PR, thanks

TheNain38:
Would there be a new setting to disable that behavior and get the old one ?

agrajaghh:
> the answer is never more options

and what old behaviour? Nothing happens when you swipe...

TheNain38:
Yes, but if you mistakely swipe when you did not meant to, then it could be annoying

ldematte:
(I take gmail as an example here as a very used application with swiping features, but the behaviour I have observed is quite consistent).

There is an undo option, like suggested in the issue and like in gmail. So if you did it by mistake, and you realize it right away, you can undo it.
I see three options here:
1. configuration option: choose action for swipe left/right (nothing, delete, archive) (gmail does that)
Note that the third option requires extensive work to Signal: there is no archive atm.
2. swiping (deletion, in general!) will move messages to a recycle bin, instead of directly deleting them (also like in gmail). Very easy.. if you add a recycle bin to Signal!
3. Add a confirmation message.

Personally, I like 1. and 2.
I think they are both separate issues, but very complementary to this PR, and will be great to have them. 
Personally, I think that this PR plus 1. (with safe defaults: right to archive, left to none) and 2. will make a great UX.
If the WhisperSystem dev team likes the current PR, I would love to open issues for 1. and 2. and add them to the app. If they reject it... what's the purpose? (well, n.2 has a great value on its own, TBH)
I hate n.3, but this is OSS, so if someone else wants to add it, free to work on it ;)

PS: I understand "the answer is never more options" , but my feeling is that this advice is more related to the encryption/security aspects of Signal. But if you really hate options, than a recycle bin/archive is the answer.

TheNain38:
For example: gmail is annoying me with this because, sometimes I meant to go up or down and instead it was deleting the email -_-

agrajaghh:
for reference: in #1141 there was already some work done on archive..

ldematte:
@TheNain38 gmail has options to make a swipe go to archive  (default) or disable it completely. That's what n.1 was about.

@agrajaghh: I think #1141 plus this would be great!

TheNain38:
@ldematte Ok, thanks

ldematte:
@TheNain38 glad to help!

moxie0:
Thanks we have swipe to archive in now, and are pausing to evaluate whether we'd like to include swipe to delete.  If we do, we'll probably use the recyclerview stuff rather than the gesture stuff.  Thanks!

Stathread:
Please re-open this. I am about to un-install Signal for this reason alone. I don't want to archive my messages, I want to delete them. I am a user of Signal and like the Google swipe to delete.

Obviously you wouldn't make this default. You put a setting in to allow people to switch from "Swipe to Archive" to "Swipe to Delete" that way everyone complaining about "accidentally" deleting messages can stop complaining.

dcherian:
Swipe left to delete. Swipe right to archive?

PR #4473, opened 2015-11-10
Always play "sonar" on call initialization

brumsoel:
If "sonar" was the last sound played on the last call, it wasn't being played on the next calling attempt due to [this line in OutgoingRinger](https://github.com/WhisperSystems/Signal-Android/blob/ca5da2cf2ad038c9ee7a505e5b4ef0e323159a4b/src/org/thoughtcrime/redphone/audio/OutgoingRinger.java#L82).

Steps to reproduce:
- B has no internet connection
- A tries to call B, hears "sonar" ringer
- A hangs up, tries to call B again
- no "sonar" ringer :(


moxie0:
good one, in 3.4.0

PR #4488, opened 2015-11-11
Android Wear notifications

makubi:
Successor of https://github.com/WhisperSystems/Signal-Android/pull/4472

One stack group element per contact is displayed.
Messages appear in the order they were received (latest message bottom).

Notifications are ordered by the time of latest message arrival.
Contacts with the newest message appear on top.

#### Bug on Android 5.0 and 5.1

If an alarm (ringtone, vibration) is set on the summary notification,
every single thread message appears on the handheld.
If an alarm is set on the single thread notifications, the more
notifications arrive, the more alarms are kicked off.

To work around the issue, the summary notification's priority is set to
'default' instead of 'high', which suppresses the heads-up notification.
This notification is shown in the Android notification bar nevertheless.

Users enabling this workaround will often read notifications on the
wearable and therefore not miss the heads-up notification.

On older / newer devices, this setting has no effect, the summary
notification is displayed as usual.

See https://code.google.com/p/android/issues/detail?id=130689

Developed with Sebastian Chlan <sebastian.chlan@gmail.com> - @dr03lf

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1211,8 +1211,9 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     new AsyncTask<Long, Void, Void>() {
       @Override
       protected Void doInBackground(Long... params) {
-        DatabaseFactory.getThreadDatabase(ConversationActivity.this).setRead(params[0]);
-        MessageNotifier.updateNotification(ConversationActivity.this, masterSecret);
+        Long threadId = params[0];
+        DatabaseFactory.getThreadDatabase(ConversationActivity.this).setRead(threadId);
+        MessageNotifier.updateNotificationCancelRead(ConversationActivity.this, masterSecret, threadId);
         return null;
       }
     }.execute(threadId);
--- a/src/org/thoughtcrime/securesms/jobs/TrimThreadJob.java
+++ b/src/org/thoughtcrime/securesms/jobs/TrimThreadJob.java
@@ -19,7 +19,10 @@ package org.thoughtcrime.securesms.jobs;
 import android.content.Context;
 import android.util.Log;
 
+import org.thoughtcrime.securesms.crypto.MasterSecret;
 import org.thoughtcrime.securesms.database.DatabaseFactory;
+import org.thoughtcrime.securesms.notifications.MessageNotifier;
+import org.thoughtcrime.securesms.service.KeyCachingService;
 import org.thoughtcrime.securesms.util.TextSecurePreferences;
 import org.whispersystems.jobqueue.Job;
 import org.whispersystems.jobqueue.JobParameters;
@@ -51,6 +54,9 @@ public class TrimThreadJob extends Job {
       return;
 
     DatabaseFactory.getThreadDatabase(context).trimThread(threadId, threadLengthLimit);
+
+    final MasterSecret masterSecret = KeyCachingService.getMasterSecret(context);
+    MessageNotifier.updateNotificationCancelRead(context, masterSecret, threadId);
   }
 
   @Override
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -17,6 +17,7 @@
 package org.thoughtcrime.securesms.notifications;
 
 import android.app.AlarmManager;
+import android.app.Notification;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.content.BroadcastReceiver;
@@ -32,6 +33,8 @@ import android.os.AsyncTask;
 import android.os.Build;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.support.v4.app.NotificationCompat;
+import android.support.v4.app.NotificationManagerCompat;
 import android.text.Spannable;
 import android.text.SpannableString;
 import android.text.TextUtils;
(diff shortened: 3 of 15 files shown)


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -655,6 +654,7 @@ private void handleDeleteThread() {
       public void onClick(DialogInterface dialog, int which) {
         if (threadId > 0) {
           DatabaseFactory.getThreadDatabase(ConversationActivity.this).deleteConversation(threadId);
+          MessageNotifier.updateNotificationCancelRead(ConversationActivity.this, masterSecret, threadId); // TODO necessary?
```
agrajaghh: TODO? :stuck_out_tongue_winking_eye:

src/org/thoughtcrime/securesms/jobs/TrimThreadJob.java
```diff
@@ -51,6 +52,7 @@ public void onRun() {
       return;
 
     DatabaseFactory.getThreadDatabase(context).trimThread(threadId, threadLengthLimit);
+    MessageNotifier.updateNotificationCancelRead(context, null, threadId); // TODO update notifications here?
```
agrajaghh: same

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -218,22 +275,93 @@ private static void sendSingleThreadNotification(@NonNull  Context context,
...
+   * <br><br>
+   * This bug was reported at https://code.google.com/p/android/issues/detail?id=159947
+   * @return <b>true</b> if the device is affected, <b>false</b> otherwise
+   */
+  private static boolean doesNotDisplayNonSummaryGroupNotifications() {
+    return Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP;
+  }
+
+  /*
+    TODO check in master: On 4.4, the text without sender is shown
```
agrajaghh: same

agrajaghh:
afaik, there shouldn't be any comments in code, see https://github.com/WhisperSystems/Signal-Android/wiki/Code-Style-Guidelines#comments

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
makubi: Uhm, sorry, forgot to remove that one.

src/org/thoughtcrime/securesms/ConversationActivity.java
makubi: Yep, notifications were not yet updated on that action but I think they have to. Can you confirm this?

makubi:
In genereal, yes, but in some situations they make sense. The comments are just short ones or do you mean the JavaDoc?

agrajaghh:
I think @moxie0 doesn't like both, but maybe wait for his answer...

moxie0:
Quick high-level feedback: please no comments, and no additional preferences.

Also keep in mind that if your *first* PR to the project is a 600-line complex change, it is very unlikely that it will be merged. I'd recommend starting with smaller easy changes to get a feeling for the organization and style of the code base, building up some shared understanding, and then trying to take on larger things after a few months of regular small contributions.

makubi:
Hi @moxie0,

I can remove the comments, but still think they make sense in some place.

I'd rather prefer not adding any new preferences, but this means setting the priority of the summary notification to default which prevents heads-up notifications (on the given devices).
We were unsure which way to go for the workaround.

I understand your note about starting with smaller merge requests, but in this case I think the changes are appropriate (some code was just factored to new methods).
In my opinion, the code needs some refactoring, but I tried not to change too much on the current code base.

We chose this feature because we did not see somebody working on issue https://github.com/WhisperSystems/Signal-Android/issues/2525 and thought it would be a good way to start.

#### tl:dr
* Shall we remove the setting and set the priority to default for the summary notification (for the affected devices)?
* I would really like to see wear support added to Signal.

moxie0:
> I can remove the comments, but still think they make sense in some place.

We disagree.

> I'd rather prefer not adding any new preferences, but this means setting the priority of the summary notification to default which prevents heads-up notifications (on the given devices). We were unsure which way to go for the workaround.

I know nothing about this, but if I receive a PR from someone I've never interacted with before, it means that I can't fully trust that there's no way to do this without an additional preference.  Which means that I have to go and look into all of the gotchas and details myself, at which point it would probably be faster just to make the change myself.  If you start with some smaller changes, I would have more confidence.

> In my opinion, the code needs some refactoring, but I tried not to change too much on the current code base.

The notification code is probably the worst in the entire repository, which is all the more reason why making a 600 line change to it as your first contribution probably isn't going to work out.

makubi:
Well, I understand that this MR is harder to review than others, but the benefit is (independent of the outcome of this MR) that you don't have to go through all the problems of the Android notification system as we pointed some out. :)

I would be happy to refactor the code, but as you stated, this would introduce much more changes.

The current problem is that you are unable to read messages on wearables if more than one contact sends an message, so everybody using e.g. a watch has to open the app on the handheld.


Of course I can shorten the MR by removing some code (Classes I introduced), but I think this does not increase the readability.
What about the JavaDoc?

How should we continue? Are you going to decline the MR either way?

mariussoutier:
+1

SuperTux88:
This fixes also notifications on the pebble smartwatch :+1:

makubi:
Hi,

I updated the PR according to your comments. Can you please have a look at it again?

vinilox:
This is very important for android wear users. Without this Signal is unusable as soon as you get 2+ messages.

fbe:
"please no comments,"
image

moxie0:
This change is very difficult to follow and doesn't conform to the coding guidelines for this project, I'm not able to merge this.  It would probably be better if you'd document exactly what change you'd like to see in the original issue (no more summary notifications? something else?) than trying to submit a PR for this without having done something smaller first.

Issue #4160, opened 2015-10-01
Allow call handling from notifiaction


McLoo:
Right now you get the full screen incoming call activity.

If you then (accidentally) switch the app or go to home screen, it's still ringing and you have to tap the notification "Signal call in progress", that brings you back to the call activity. There should be the buttons "Dismiss" and "Answer" on the notification.

jeremymasters:
It might also be good to have a "Dismiss/Reject with message" option here if it could fit into the notification.

PR #4490, opened 2015-11-11
add action buttons to the call notification

agrajaghh:
while its ringing:
screenshot_15-44-13

during a call:
screenshot_15-44-24


McLoo:
fixes #4160
sweet!

agrajaghh:
I was looking for that issue but couldn't find it :grin:

moxie0:
Looking good. Personally I like it when we preserve the original material icon file names, that way it's clear that we don't end up with duplicate resources.

agrajaghh:
thats what I did :stuck_out_tongue: I just removed the suffix (since its not correct):
`ic_phone_48px` -> `ic_phone`
Or would you like to keep it?

moxie0:
The diff I see has added rather than moved files? My assumption was that you copied something like ic_phone_gray600_48dp.png over from material icons.

agrajaghh:
I'm using this [script](https://github.com/sebkur/android-res-utils) to generate the png's from the material icon svg since there is no 32dp version in the icon repo

btw, I used black with 54% opacity. I couldn't find offical info about the color of action icons. is this correct? I just found [this](https://www.google.com/design/spec/style/icons.html#icons-system-icons) (at the bottom)

dcherian:
How about changing the string to "Incoming Signal call" rather than "Signal call in progress" for the first screenshot. Seems weird to say "in progress" when I haven't picked up.

agrajaghh:
@dcherian sounds good, I changed it...

I think it would be even better if the contact name (or number, if no contact) would be displayed. But I'll leave that to another PR..

agrajaghh:
Damm, just realized I forgot outgoing calls. Will fix it soon...

jeremymasters:
@agrajaghh What about a button for denying with text response?  Thought I would ask since Christmas is approaching...;)

agrajaghh:
outgoing calls should be fixed now.

@jeremymasters hm, I think its best to do this seperately, maybe open an issue for that?

outgoing while ringing:
out
I'm not happy with "Signal call in progress". What would be better?

incoming:
inc

established:
established

mawed:
Shouldn't there be a little lock close to the phone to show the encrypted call as it is shown in the conversation wit another Signal user?

dcherian:
Some options for outgoing calls:
 - "Outgoing Signal call"
 - "Establishing Signal call"
 - "Signal: calling NAME_OF_CONTACT"

agrajaghh:
Using "Establishing Signal call" now, but maybe too technical :smirk:
Also using the lock phone icon from the action bar for the notification.

outgoing while ringing:
out

incoming:
inc

established:
established

dcherian:
Does the text string need to be repeated? Could be replaced with contact name in the next PR I guess

agrajaghh:
I thought it's complicated to include the contact names, but it was pretty easy (if I did it correctly :smirk:)

outgoing while ringing:
out

incoming:
inc

established:
established

dcherian:
Looks sweet!

mcginty:
@agrajaghh this looks great! Definitely better with contact names prioritized.

(also for future-times, I found this Android Studio plugin for material asset generation: https://github.com/konifar/android-material-design-icon-generator-plugin)

src/org/thoughtcrime/redphone/RedPhone.java
```diff
@@ -63,6 +63,10 @@
   private static final int STANDARD_DELAY_FINISH    = 1000;
   public  static final int BUSY_SIGNAL_DELAY_FINISH = 5500;
 
+  public static final String ANSWER_ACTION   = RedPhone.class.getName() + ".ANSWER_ACTION";
+  public static final String DENY_ACTION     = RedPhone.class.getName() + ".DENY_ACTION";
+  public static final String END_CALL_ACTION = RedPhone.class.getName() + ".END_CALL_ACTION";
```
mcginty: small nitpick: getCanonicalName() is more consistent with how other action strings are constructed in other parts of the app (even though getName() would return the same string in this case)

src/org/thoughtcrime/redphone/RedPhone.java
```diff
@@ -89,6 +93,16 @@ public void onResume() {
     registerBluetoothReceiver();
   }
 
+  @Override
+  public void onNewIntent(Intent intent){
+    if (ANSWER_ACTION.equals(intent.getAction())) {
```
mcginty: Java 7 to the rescue! `switch(intent.getAction())`

src/org/thoughtcrime/redphone/RedPhone.java
agrajaghh: done

src/org/thoughtcrime/redphone/RedPhone.java
agrajaghh: problem is that `ANSWER_ACTION` etc. are not constant expressions. So either I leave this like it is, or I change the actions to actual strings. What do you prefer?

agrajaghh:
the material asset plugin looks good for the future :blush: But unfortuantely it doesn't support 32dp as well :grin:

agrajaghh:
@mcginty should I include xxxhdpi icons as well? just saw them in #4553...

mcginty:
@agrajaghh yep, we're in the world where xxxhdpi is a thing on new devices now

agrajaghh:
okay, included them now

PR #4492, opened 2015-11-11
Update contact DB on incoming messages

### LINKED ISSUE #3949: textsecure contacts do not auto update on receiving a message from a newly registered TS who is in your contact list
it requires a manual refresh of the contact list (with the not so obvious pull to refresh mechanism)

moxie0:
When we receive a Signal message from a previously unregistered
user, reflect that in the contact DB.

Fixes #3949
// FREEBIE


moxie0:
in 3.4.0

PR #4494, opened 2015-11-11
change cursor color to text color

heyimgay:
//FREEBIE


mcginty:
At least on android 4.2, this change makes the cursor so thin (looks like 1px wide) it's awkwardly difficult to see. I agree the colors aren't wonderful, but I want to make sure the blinking cursor doesn't decrease in visibility. We'll probably need a different solution, unfortunately, as simple a solution as this would have been.

Thanks!

PR #4495, opened 2015-11-11
Colorize "add attachment"-button

### LINKED ISSUE #3521: Colorize "add attachment"-button
I think you should colorize the "add attachment" button, too, to have a consistent UI... :smiley:

heyimgay:
Fixes #3521

//FREEBIE


moxie0:
thanks but I don't think we're going to do this

Issue #3053, opened 2015-04-24
Sending a deleted image gets stuck in pending state


2-4601:
This came up in #3046 briefly.

Sometimes when you delete a picture it will leave around ghost thumbnails that are visible in the system gallery and also in TS attachment gallery. This can lead to the following issue if you try to send the deleted image:
1. Delete a picture that's on your device with shell or a file manager (or create a ghost thumbnail in some other way)
2. Attach that picture to a TS message using the attachment gallery's _Recent_ tab (the gallery does not know that the picture was deleted)
3. You will get a missing thumbnail icon
4. Tap send
5. TS will create an "empty" message that gets stuck in a pending state for eternity

If you try to do the same in Google Hangouts you will get a 'Can't find picture' notification and you cannot send the image.
TextSecure should also check if the file exists and not trust the gallery viewer.

.gif demonstration:
![output](https://cloud.githubusercontent.com/assets/174176/7310772/6e7ec4a2-ea3a-11e4-9159-6965298c8ab9.gif)

TS 2.12.0, stock Android 5.1 and Android 4.0.4

mcginty:
Nice find

https://github.com/WhisperSystems/TextSecure/pull/3059 will at least make the conversation and media previews show an error icon instead of just nothing. Will keep this issue open though to also add some checks for drafts if the file is missing and fail earlier.

moxie0:
in 3.4.0

PR #4499, opened 2015-11-12
fail faster in FileNotFoundException case

mcginty:
I actually couldn't reproduce #3053 in 3.3.1, but either way it does make more sense to fail immediately without retry if we receive a FileNotFoundException, so adding this in for extra sanity.


moxie0:
in 3.4.0

PR #4501, opened 2015-11-12
Clarify instructions in BUILDING.md

offby1:
I'm pretty much a noob when it comes to Java and Android; the things I've fixed here either prevented me from building, or led me to waste effort.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -19,7 +19,7 @@ The following steps should help you (re)build Signal from the command line.
 2. Make sure you have the [Android SDK](https://developer.android.com/sdk/index.html) installed.
 3. Ensure that the following packages are installed from the Android SDK manager:
     * Android SDK Build Tools
-    * SDK Platform
+    * SDK Platform (API level 22)
     * Android Support Repository
     * Google Repository
 4. Create a local.properties file at the root of your source checkout and add an sdk.dir entry to it.  For example:
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -5,7 +5,8 @@ Basics
 ------
 
 Signal uses [Gradle](http://gradle.org) to build the project and to maintain
-dependencies.
+dependencies.  However, you needn't install it yourself; the
+"gradle wrapper" `gradlew`, mentioned below, will do that for you.
 
 Building Signal
 ---------------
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -19,12 +20,12 @@ The following steps should help you (re)build Signal from the command line.
 2. Make sure you have the [Android SDK](https://developer.android.com/sdk/index.html) installed.
 3. Ensure that the following packages are installed from the Android SDK manager:
     * Android SDK Build Tools
-    * SDK Platform
+    * SDK Platform (API level 22)
```
odeke-em: This might be misleading as there are multiple API levels that can be used. Is "level 22" the minimum?

BUILDING.md
```diff
@@ -19,12 +20,12 @@ The following steps should help you (re)build Signal from the command line.
...
     * Android Support Repository
     * Google Repository
-4. Create a local.properties file at the root of your source checkout and add an sdk.dir entry to it.
+4. Create a local.properties file at the root of your source checkout and add an sdk.dir entry to it.  For example:
 
-        sdk.dir=\<path to your sdk installation\>
+        sdk.dir=/Application/android-sdk-macosx
```
odeke-em: Maybe too single platform centric to Mac OSX?

BUILDING.md
```diff
@@ -5,7 +5,8 @@ Basics
 ------
 
 Signal uses [Gradle](http://gradle.org) to build the project and to maintain
-dependencies.
+dependencies.  However, you needn't install it yourself; the
```
odeke-em: Would you mind taking a look at the Wiki at https://github.com/WhisperSystems/Signal-Android/wiki/How-to-build-Signal-from-the-sources?
If needed maybe make an amend here?

BUILDING.md
offby1: I think it's both the minimum and the maximum -- I no longer remember for sure, but I think there's something in the source that requires that exact version.

BUILDING.md
offby1: Sure, it's specific to OS X, but it's an example; any example will be specific to some OS.  The point is that I didn't realize that I wasn't meant to type the `\<` and `\>` verbatim; I thought an example would be clearer.

PR #4503, opened 2015-11-12
adhere to android plural standards

mcginty:
using "ONE" changes the button sizes too much, and no need to accentuate the fact that someone's only inviting one friend ;)


moxie0:
in 3.4.0

PR #4507, opened 2015-11-12
Don't show images without data under "all images"

haffenloher:
This affects images
- that have yet to be downloaded (using media download controls)
- that are currently being downloaded
- that failed to download

Before:
screenshot from 2015-11-11 13-46-30

After:
screenshot from 2015-11-12 23-57-39


mcginty:
glorious!

thanks so much, great change

PR #4513, opened 2015-11-13
Improve Delete Thread, Reset Secure Session confirmation dialogs

### LINKED ISSUE #4512: "Delete thread" and "Reset secure session" dialog strings need updating
@RiseT [pointed out](https://github.com/WhisperSystems/TextSecure/pull/4200#issuecomment-145327577) that two dialogs have the word, "confirmation," unnecessarily cluttering their titles:

dialogs-side-by-each

The text and button strings could also be made more consistent with other Signal dialogs.  For example:

fixed-dialogs-side-by-each-2

Device info: Nexus 5, stock Android 6.0 MRA58K, TextSecure 3.3.2 (0c75e65)

App state: Signal registered, SMS enabled, default SMS app

unrulygnu:
Fixes #4512
// FREEBIE

Before:
dialogs-side-by-each

After:
fixed-dialogs-side-by-each-2

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -89,10 +89,11 @@
     <string name="ConversationItem_unable_to_open_media">Can\'t find an app able to open this media.</string>
 
     <!-- ConversationActivity -->
-    <string name="ConversationActivity_reset_secure_session_confirmation">Reset secure session confirmation</string>
-    <string name="ConversationActivity_are_you_sure_that_you_want_to_reset_this_secure_session_question">Are you sure that you want to reset this secure session?</string>
-    <string name="ConversationActivity_delete_thread_confirmation">Delete thread confirmation</string>
-    <string name="ConversationActivity_are_you_sure_that_you_want_to_permanently_delete_this_conversation_question">Are you sure that you want to permanently delete this conversation?</string>
+    <string name="ConversationActivity_reset_secure_session_question">Reset secure session?</string>
+    <string name="ConversationActivity_this_may_help_if_youre_having_encryption_problems">This may help if you\'re having encryption problems in this conversation. Your messages will be kept.</string>
+    <string name="ConversationActivity_reset">Reset</string>
+    <string name="ConversationActivity_delete_thread_question">Delete conversation?</string>
+    <string name="ConversationActivity_this_will_permanently_delete_all_messages_in_this_conversation">This will permanently delete all messages in this conversation.</string>
     <string name="ConversationActivity_add_attachment">Add attachment</string>
     <string name="ConversationActivity_select_contact_info">Select contact info</string>
     <string name="ConversationActivity_compose_message">Compose message</string>
@@ -1039,7 +1040,7 @@
     <string name="conversation__menu_add_attachment">Add attachment</string>
     <string name="conversation__menu_update_group">Update group</string>
     <string name="conversation__menu_leave_group">Leave group</string>
-    <string name="conversation__menu_delete_thread">Delete thread</string>
+    <string name="conversation__menu_delete_thread">Delete conversation</string>
     <string name="conversation__menu_view_media">All images</string>
     <string name="conversation__menu_conversation_settings">Conversation settings</string>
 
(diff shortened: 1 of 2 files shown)


res/values/strings.xml
```diff
@@ -89,10 +89,11 @@
     <string name="ConversationItem_unable_to_open_media">Can\'t find an app able to open this media.</string>
 
     <!-- ConversationActivity -->
-    <string name="ConversationActivity_reset_secure_session_confirmation">Reset secure session confirmation</string>
-    <string name="ConversationActivity_are_you_sure_that_you_want_to_reset_this_secure_session_question">Are you sure that you want to reset this secure session?</string>
-    <string name="ConversationActivity_delete_thread_confirmation">Delete thread confirmation</string>
-    <string name="ConversationActivity_are_you_sure_that_you_want_to_permanently_delete_this_conversation_question">Are you sure that you want to permanently delete this conversation?</string>
+    <string name="ConversationActivity_reset_secure_session_question">Reset secure session?</string>
+    <string name="ConversationActivity_this_will_start_a_new_secure_session">This will start a new secure session for this conversation. Your messages will be kept.</string>
```
RiseT: I've been targetting this string as well, as there seems to exist a lot of confusion about the purpose of its underlying function.
Thus, in the German translation I've translated this string in a way that clearly explains what the function actually does.

Translated back to English, this would come down to something like this (as I've done some prior research, I strongly believe the technical details to be 100% correct):

**"End current secure session and establish a new one on occassion of the next message. Helpful for solving encryption problems."**

While the wording could possibly improved upon, I believe something like this is really necessary in order to finally put all confusion and misunderstandings about the purpose of this function to rest.

This is also in line with Google's Design Guidelines, as the purpose of an alert's text is to explain its title in in a more detailed & precise way.

unrulygnu:
Yeah, I also dug in and annoyed the hell out of a friend with "Secure session reset" messages while I played in the debugger trying to understand what was going on.  :smile: 

After asking several friends about their thoughts about resetting the secure session, the main concern was "Is this going to delete my messages?"  So I went with that as the main explanatory point of the dialog text, keeping the rest as behind-the-curtain implementation details.

I can add more text about being helpful for solving encryption problems, but even libaxolotl's description confidently states the following:

> There is no need to ever tear down a session once one has been established.

So is this feature here only for dev purposes?  Or maybe to clear up possible issues during protocol upgrades?  Maybe the boss can shed more light and advise on whether more detail is needed in the dialog text.

RiseT:
@unrulygnu I think I I can roughly describe the most frequent scenario where this feature is helpful.

User A establishes a secure session with user B (i. e. they chat...). By doing this, session key 'Y' is established. Now User A does a Nandroid backup of his phone (backing up the current session key with it). Afterwards, user A and user B chat for a couple more minutes. After some time, the secure session establishes a new session key 'Z' automatically. No problem so far, as user A and user B use the same session key. Now user A flashes his phone with the new SuperGeek custom mod, of which he heard good things. But he doesn't like it, so he restores his prior Nandroid backup. By doing this, user A restores the session key 'Y' as well. But his friend, user B, is using session key 'Z'. This leads to a badly encrypted message when they try to exchange the next message.

unrulygnu:
@RiseT : huge thanks for the review and thorough explanation.  I have one question:  do you know what messages/strings are shown to the user in such a scenario?  I'd like to update the dialog text with something relevant to what the user sees and avoiding words like "encryption" ([no. 2](https://github.com/WhisperSystems/Signal-Android/blob/master/contributing.md#user-content-development-ideology)), if possible.

RiseT:
@unrulygnu I think this would be **MessageDisplayHelper_bad_encrypted_message** ("Bad encrypted message...")

unrulygnu:
@RiseT , thanks again for the good info.  What do you think of this for the text? "This may help if you're having encryption problems in this conversation. Your messages will be kept."

fixed-dialog-reset-2-sm

RiseT:
@unrulygnu Moxie is taking no prisoners today :smile:

PR #4516, opened 2015-11-13
add recipients.removeListener in onDestroy()

agrajaghh:
I added the listener in #3580 without adding `removeListener()`, but I guess this is necessary?


src/org/thoughtcrime/securesms/MessageDetailsActivity.java
```diff
@@ -138,6 +139,11 @@ public void run() {
     });
   }
 
+  @Override protected void onDestroy() {
```
moxie0: @ Override \n

moxie0:
but yeah this shouldn't be necessary, since it's stored as a weakreference

agrajaghh:
fixed the newline in case you want to merge it anyway

moxie0:
thanks i'm going to let the weak reference handle things in this case

PR #4522, opened 2015-11-13
Already enable proximity sensor while outgoing call is ringing

### LINKED ISSUE #4519: Proximity sensor not enabled while dialing
The screen does not turn black while an outgoing call is still dialing. It will turn black as soon as the call is answered. Tested on two different phones running Signal 3.3.2.

Steps to reproduce:
1. make an outgoing call, called person should not answer the call
2. put your phone to your ear to hear the dialing sound

Expected:
- screen should be locked to avoid accidentally pressing anything

Result:
- screen is not locked

I think this is a problem in the code, see [here](https://github.com/WhisperSystems/Signal-Android/blob/HEAD/src/org/thoughtcrime/redphone/RedPhoneService.java#L210) and [here](https://github.com/WhisperSystems/Signal-Android/blob/HEAD/src/org/thoughtcrime/redphone/call/LockManager.java#L89).
I can still provide a debug log on request.

haffenloher:
Fixes #4519

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/redphone/RedPhoneService.java
+++ b/src/org/thoughtcrime/redphone/RedPhoneService.java
@@ -207,7 +207,7 @@ public class RedPhoneService extends Service implements CallStateListener, CallS
     sendMessage(Type.OUTGOING_CALL, getRecipient(), null);
 
     state = STATE_DIALING;
-    lockManager.updatePhoneState(LockManager.PhoneState.INTERACTIVE);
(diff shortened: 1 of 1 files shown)


generalmanager:
Probably fixes https://github.com/WhisperSystems/Signal-Android/issues/4434 too, of which #4519 seems to be a duplicate.

moxie0:
in 3.8.0

PR #4528, opened 2015-11-14
Remove funky halloween easter egg

### LINKED ISSUE #4420: Halloween easter egg: Muting notifications in group chat messes up avatars & names

haffenloher:
Fixes #4420

// FREEBIE

Not entirely sure if this is a good way to fix this - feels a bit more like a workaround - but I think there's no context in which updating the conversation bubbles & avatars with group recipient info makes sense, so at least this shouldn't break anything.


moxie0:
hmm yeah thanks but I think I'd like to try to figure out the underlying cause instead

PR #4530, opened 2015-11-14
Make sure the compose text field always has focus

### LINKED ISSUE #4527: text box focus after contact selection
This is a terrible title but here are the steps:

1. Click FAB to get contact selection screen.
2. Use keyboard to type letters and narrow down selection.
3. Choose a contact
4. Conversation window opens with keyboard open.
5. Text box doesn't have focus. So, you can type on the keyboard without the letters actually going anywhere.

Not sure if this is Signal's fault, or just Android.
Signal 3.4.0, Moto G. Android 4.4.4

EDIT: *last* word typed appears after tapping on the "text input" box. This is after following above steps.

haffenloher:
Fixes #4527

// FREEBIE

Side note: This adds a blinking cursor in the text field, even when entering the conversation the normal way. I hope this is acceptable.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -925,6 +925,8 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     composeText.setOnClickListener(composeKeyPressedListener);
     composeText.setOnFocusChangeListener(composeKeyPressedListener);
 
+    composeText.requestFocus();
+
     if (QuickAttachmentDrawer.isDeviceSupported(this)) {
       quickAttachmentDrawer.setListener(this);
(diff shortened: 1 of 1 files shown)


mammuth:
So you're trying to fix a issue regarding bad UX by introducing new UX problems? :wink: 

Back to serious: You could try to wrap your requestFocus() into *if(container.isInputOpen())*. 

I didn't test it, but I think this should probably fix your problem.

haffenloher:
I personally don't think it's a problem at all, but I wanted to mention it in case someone else thought it was.

> You could try to wrap your requestFocus() into if(container.isInputOpen()).

That condition seems to always be false unfortunately. With this, the text field no longer gets focus.

haffenloher:
Closing in favor of #4555

PR #4533, opened 2015-11-14
Add gitter.im link to readme

samlanning:
As we have a gitter.im room set up and added to the README for [Signal-iOS](https://github.com/WhisperSystems/Signal-iOS), and we have an unused but activated room for this repository, thought it would be a nice idea to add this link to direct people over to this room for development related chat? (anything not mailinglist worthy).


moxie0:
I actually think this might serve better as a channel that isn't extensively advertised

samlanning:
That's a fair comment. I have seen a number of people hop on `Signal-iOS` on gitter to ask support-related questions etc...

How about putting it in `contributing.md`, and doing the same in `Signal-iOS`, moving it from README to CONTRIBUTING?

moxie0:
done

PR #4540, opened 2015-11-15
Removed Crypto.mk~

makubi:
Added *~ to .gitignore

Thx @SuperTux88

// FREEBIE


moxie0:
thanks but this is all from upstream

makubi:
Alright, did you apply the patch to upstream yet? What is **upstream** here?

PR #4553, opened 2015-11-16
Support for in-app device provisioning

moxie0:
// FREEBIE


src/org/thoughtcrime/securesms/DeviceAddFragment.java
```diff
@@ -0,0 +1,245 @@
...
+                                                                       0, 0, width, height,
+                                                                       false);
+
+        BinaryBitmap bitmap = new BinaryBitmap(new HybridBinarizer(source));
+
+        Result result = reader.decode(bitmap);
+
+        if (result != null) return result.getText();
+
+      } catch (NullPointerException | ChecksumException | FormatException e) {
```
mcginty: when does NullPointerException happen?

moxie0: There are reports that sometimes zxing's reader will throw it.  I never saw it myself, but figured I'd include it.

mcginty:
needs to be rebased

PR #4566, opened 2015-11-17
Fix notification strings (#4004)

char32:
This patch fixes the notification strings in the
case that neither name nor message should be shown.

This changes the string names and values for all
language and in the respective source file.

Additionally the german translation is adjusted too.

Fixes issue #4004

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/SingleRecipientNotificationBuilder.java
+++ b/src/org/thoughtcrime/securesms/notifications/SingleRecipientNotificationBuilder.java
@@ -131,9 +131,9 @@ public class SingleRecipientNotificationBuilder extends AbstractNotificationBuil
     if (privacy.isDisplayMessage()) {
       setTicker(getStyledMessage(recipient, message));
     } else if (privacy.isDisplayContact()) {
-      setTicker(getStyledMessage(recipient, context.getString(R.string.SingleRecipientNotificationBuilder_signal)));
+      setTicker(getStyledMessage(recipient, context.getString(R.string.SingleRecipientNotificationBuilder_new_message)));
     } else {
-      setTicker(context.getString(R.string.SingleRecipientNotificationBuilder_signal));
+      setTicker(context.getString(R.string.SingleRecipientNotificationBuilder_new_message));
     }
   }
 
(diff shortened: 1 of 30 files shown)


haffenloher:
Please don't change the translations, `res/values/strings.xml` is enough :-)
The other XML files are generated through [Transifex](https://www.transifex.com/moxie/textsecure-official/).

char32:
@haffenloher Thanks for the Review! :-)

Ok I will remove the Translation today in the eveing. But the Change of the string_name itself in all language files (to SingleRecipientNotificationBuilder_new_message) should remain, right? So it is enough to remove the Change of the german Translation value, right?

ChaoticMind:
@char32 no, the localized files are generated whole when exporting the translations from Transifex, including the string names. You only need to change `res/values/strings.xml`.

char32:
@ChaoticMind Thanks! That makes it a lot easier. Didn't know about TransiFlex before. But yeah I am still learning how everything in TextSecure works together :-)
I will Change it later today then and update this  PR.

char32:
Ok, the proposed changes are done :-)

src/org/thoughtcrime/securesms/notifications/SingleRecipientNotificationBuilder.java
```diff
@@ -131,9 +131,9 @@ public void setTicker(@NonNull Recipient recipient, @Nullable CharSequence messa
     if (privacy.isDisplayMessage()) {
       setTicker(getStyledMessage(recipient, message));
     } else if (privacy.isDisplayContact()) {
-      setTicker(getStyledMessage(recipient, context.getString(R.string.SingleRecipientNotificationBuilder_new_signal_message)));
+      setTicker(getStyledMessage(recipient, context.getString(R.string.SingleRecipientNotificationBuilder_signal)));
     } else {
-      setTicker(context.getString(R.string.SingleRecipientNotificationBuilder_new_signal_message));
+      setTicker(context.getString(R.string.SingleRecipientNotificationBuilder_signal));
```
moxie0: shouldn't these be "new message"?

char32: @moxie0 
I am not sure. According to #4004 every occourance of
"New TextSecure Message" should be replaced with "TextSecure". Ok now with "Signal".

Should I change it to SingleRecipientNotificationBuilder_new_message ? From the context I would would preffer SingleRecipientNotificationBuilder_new_message, too.

char32:
Ok, I looked over it again and changed it accordingly. The two occourances should be new_message , like you stated, for sure! :-) Changed that accidentally in the first & second patchset....

Issue #4548, opened 2015-11-16
After a call the audio of the recipient's phone is stuck to earpiece speaker


2-4601:
Steps to reproduce:
* Jerry calls George with Signal
* George answers
* They end the call
* Now the audio of George's phone is stuck to the handset earpiece speaker
 * everything (music, videos, notifications) play through the earpiece
 * every time George tries to adjust the volume of his phone he sees a bluetooth volume control overlay instead of the normal volume control overlay

This does not seem to happen to Jerry's phone (i.e. the initiator of the call).

To fix this George can call someone with Signal or system dialer and end the call immediately. Now audio will play through the loud speaker again and normal volume controls become available.

This has been an issue for me since Signal 3.3.x (currently using 3.4.1)

Affects both of my devices:
 * Nexus 4, stock 5.1.1
 * Huawei U8800, 4.0.4

2-4601:
@agrajaghh: Could you help me to test this by trying to reproduce this with your Nexus 4?

Initial state of Nexus 4: stock 5.1.1, Signal 3.4.1, All sounds mode (i.e. the normal state :bell:) and all volumes are non-zero.

Test:
* use another device to make a Signal call to your Nexus 4
* answer the call with your Nexus 4
* end the call from your another device

After the call has ended:
* try to adjust the volume of your Nexus 4 using the physical volume slider at the side of the phone
 * do you see a bluetooth volume control or the regular Lollipop volume controls?
* try to play some audio (e.g. YouTube video)
 * does the sound come from the speaker behind the phone or from the earpiece speaker at the top front?

agrajaghh:
Sure! I'm running master with latest commit: https://github.com/WhisperSystems/Signal-Android/commit/69f5aacd2514ba3ea90d1b3cfba3908e229bc40b Is this 3.4.1?

>try to adjust the volume of your Nexus 4 by using the physical volume slider at the side of the phone> 
> * do you see a bluetooth volume control or the regular Lollipop volume controls?

bluethooth volume control

> try to play some audio (e.g. YouTube video)
> 
> * does the sound come from the speaker behind the phone or from the earpiece speaker at the top front?

earpiece speaker :flushed: 

> To fix this George can call someone with Signal or system dialer and end the call immediately. Now audio will play through the loud speaker again and normal volume controls become available.

confirmed

wow thats fucked up.... :laughing:

haffenloher:
[...]

PR #4567, opened 2015-11-17
Reset audio mode when stopping incoming ringer

haffenloher:
I admit this is mainly the result of trial and error and I do not fully understand why it works (and other things didn't), but it seems to fix both symptoms from #4548 (bluetooth symbol when adjusting volume after the call; sound playing through the earpiece speaker).

### DIFF
--- a/src/org/thoughtcrime/redphone/audio/IncomingRinger.java
+++ b/src/org/thoughtcrime/redphone/audio/IncomingRinger.java
@@ -29,6 +29,8 @@ import android.os.Vibrator;
 import android.provider.Settings;
 import android.util.Log;
 
+import org.thoughtcrime.securesms.util.ServiceUtil;
+
 import java.io.IOException;
 
 /**
@@ -74,7 +76,7 @@ public class IncomingRinger {
   }
 
   public void start() {
-    AudioManager audioManager = (AudioManager)context.getSystemService(Context.AUDIO_SERVICE);
+    AudioManager audioManager = ServiceUtil.getAudioManager(context);
     audioManager.requestAudioFocus(new AudioManager.OnAudioFocusChangeListener() {
                                      @Override
                                      public void onAudioFocusChange(int focusChange) {}
@@ -120,6 +122,10 @@ public class IncomingRinger {
     }
     Log.d(TAG, "Cancelling vibrator");
     vibrator.cancel();
+
+    AudioManager audioManager = ServiceUtil.getAudioManager(context);
+    audioManager.setMode(AudioManager.MODE_NORMAL);
+    audioManager.abandonAudioFocus(null);
   }
 
   private boolean shouldVibrate() {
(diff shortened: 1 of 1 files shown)


2-4601:
Tested on Huawei U8800, Android 4.0.4. Fixes both symptoms.

moxie0:
I think this is a good hint, but I want to make sure that all the audio focus stuff is requested in the correct order.  Right now we're requesting focus twice, once for the in-call conversation and once for the ringer.  It looks like they overlap, when they probably shouldn't.

haffenloher:
Yeah, they do overlap:

A calls B.
B's phone on incoming call:
1) `RedPhoneService.initializeAudio()` requests temporary/transient focus for `STREAM_VOICE_CALL`
2) `IncomingRinger.start()` requests focus for `STREAM_MUSIC`

On accepting call:
3) `IncomingRinger.stop()` <-- without the change, `IncomingRinger` doesn't abandon focus here

When the call ends:
4) `IncomingRinger.stop()`, again
5) `RedPhoneService.shutdownAudio()` abandons focus

motatuc:
I have to add that the problem still exists for me with cm12.1 from 20.11.2015 and signal from today. I still have the bluetooth symbol after placing a call with signal. I then cannot adjust volume anymore.

haffenloher:
Some more info: When testing without this PR, I noticed that for incoming calls, changing the audio mode via `setMode(AudioManager.MODE_NORMAL)` in `RedPhoneService.shutdownAudio()` fails (!??); `getMode()` still returns `MODE_IN_COMMUNICATION`. This exception can be found in the system logs:

```
/system_process E/JavaBinder: *** Uncaught remote exception!  (Exceptions are not yet supported across processes.)
                java.util.NoSuchElementException: Death link does not exist
                    at android.os.BinderProxy.unlinkToDeath(Native Method)
                    at android.media.AudioService.setModeInt(AudioService.java:2133)
                    at android.media.AudioService.setMode(AudioService.java:2105)
                    at android.media.IAudioService$Stub.onTransact(IAudioService.java:348)
                    at android.os.Binder.execTransact(Binder.java:446)
```

I found only [one other mention](https://stackoverflow.com/questions/28148538/audiomanager-and-nosuchelement-exception) of this problem, with no solution :(

Maybe it would be possible to stop using `setMode()` altogether. Apparently, [it's discouraged for VoIP applications](https://code.google.com/p/sipdroid/issues/detail?id=270).

haffenloher:
I've now removed the extra focus request for the ringer, so no more overlapping. The RedPhoneService already requests focus before the ringer starts.

Calling `audioManager.setMode(AudioManager.MODE_NORMAL)` in `IncomingRinger.stop()` is still necessary to prevent the strange [NoSuchElementException](https://github.com/WhisperSystems/Signal-Android/pull/4567#issuecomment-159114706).
Apparently, Android does not like the transition
MODE_NORMAL > MODE_RINGTONE > MODE_IN_COMMUNICATION > MODE_NORMAL, but
MODE_NORMAL > MODE_RINGTONE > MODE_NORMAL > MODE_IN_COMMUNICATION > MODE_NORMAL works fine :confused:

PR #4575, opened 2015-11-17
Update donation image in readme

### LINKED ISSUE #4542: Donate button image not loading
In README.md appears donate button is broken.

mammuth:
Since the button-generator by coinbase seems to fail from time to time (like at the moment, where the donation image results in a 404) a better solution is to have a normal image as donation button.
Furthermore this one is more eye-catching I guess.

The image is a free promotional graphic and is taken from https://en.bitcoin.it/wiki/Promotional_graphics

fixes #4542
//FREEBIE

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ whispersystems@lists.riseup.net
 https://lists.riseup.net/www/info/whispersystems
 
 ## Contributing Funds
-[![Donate](https://www.coinbase.com/assets/buttons/donation_large-5e1b50d6490970e32b80023f3070b1d77afc621b9e64ac996596a67a4671967b.png)](https://www.coinbase.com/checkouts/51dac699e660a4d773216b5ad94d6a0b)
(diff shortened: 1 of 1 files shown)


odeke-em:
If you don't mind, here is a "Before" and "After" to provide visual context of the PR

### Before
screen shot 2015-11-18 at 8 31 02 am


### After
screen shot 2015-11-18 at 8 30 24 am

mammuth:
Cool, thank you @odeke-em

moxie0:
i think this might be a little too eye catching

agrajaghh:
maybe just half size?
d2652e24-8dce-11e5-8cbf-7e0ec48d491e

mammuth:
Alright, I made the image smaller so the width **matches the width of the BitHub image** some lines above and rebased it into one commit.

moxie0:
in master

PR #4580, opened 2015-11-18
Use separate menu title strings for insecure/secure calls

### LINKED ISSUE #4579: Unsecure/secure call menu title uses same string, confusing
Unsecure (non-Signal users) and secure (Signal users) call menu icons are different, but the title strings are the same:

before-sidebyeach

This can be confusing, especially in accessibility conditions. Distinctly labeling a Signal call would be preferable:

after-sidebyeach

unrulygnu:
Fixes #4579
// FREEBIE

**Before**
before-sidebyeach

**After**
after-sidebyeach


unrulygnu:
Also, if you'd like me to add `unsecure` to the `menu_call_icon` attr and `ic_call_white_24dp` drawable names to keep things consistent under the hood, let me know.

PR #4591, opened 2015-11-19
Add location sharing to the attachment menu

Fuzion24:
This should close https://github.com/WhisperSystems/Signal-Android/issues/638.

This pull request still needs a few things:
 -  [ ] A white icon for GPS 
 -  [ ] Color choice for the background
 -  [ ] Production API key setup for Google's Location API

Some other points of discussion:
Since this is effectively leaking GPS coords to Google, we may want to add a warning stating so.
Maybe there can be an additional setting where no Google maps/etc are relied on and it just pulls the GPS coords directly from the location provider.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,8 @@ repositories {
 dependencies {
     compile 'me.leolin:ShortcutBadger:1.1.0-WS1'
     compile 'se.emilsjolander:stickylistheaders:2.7.0'
-    compile 'com.google.android.gms:play-services-base:6.5.87'
+    compile 'com.google.android.gms:play-services-gcm:8.1.0' // https://developers.google.com/android/guides/setup
+    compile 'com.google.android.gms:play-services-location:8.1.0'
     compile 'com.jpardogo.materialtabstrip:library:1.0.9'
     compile 'org.w3c:smil:1.0.0'
     compile 'org.apache.httpcomponents:httpclient-android:4.3.5'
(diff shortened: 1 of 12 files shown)


Fuzion24:
When someone reviews this, I can rebase on current master.

Omikron43:
Just by reading the source it seems like you set the message text to a google maps link containing the coordinates and optionally (if found) a description of the address. 
I think it would be nice to have a new message type or whatever where a snippet of the map centering the shared location is send/displayed maybe even with additional custom text.

I don't know what would be the best way to implement such a new message type. Some time ago I thought of just sending a [geo uri](https://en.wikipedia.org/wiki/Geo_URI_scheme) together with whatever the user has typed into the message field. Once the message is sent/received it is parsed for a geo/gmaps/osm/bing/whatever url so that we know a location was shared. Based on the url a map preview is generated and displayed like the current image thumbnails but on a click the default map application would be launched.
I found some obstacles that I was too lazy to research:
- what to do if the message was a SMS and no data connection is available? Maybe just display the url and try to obtain the preview at a later time when data is available again (hopefully)
- how to add the preview into the database?

If there would be a new message type however the preview could be directly generated by he sender and will get sent bundled with the message.


I read something about the `PlacePicker` from Google and found this ([source](https://developers.google.com/places/android-api/placepicker#attributions)):
> When your app displays information obtained via the place picker, the app must also display attributions. See the documentation on [attributions](https://developers.google.com/places/android-api/attributions).

I don't see where you put any attributions and I don't think that people using Signal will be happy with using Google's APIs (see your linked issue).
There is a [project to display OpenStreetMaps](https://github.com/osmdroid/osmdroid) on android which I think would be the best alternative to google together with [this location provider](https://github.com/mapzen/LOST/).

build.gradle
```diff
@@ -173,6 +175,8 @@ android {
         buildConfigField "String", "REDPHONE_RELAY_HOST", "\"relay.whispersystems.org\""
         buildConfigField "String", "REDPHONE_PREFIX_NAME", "\".whispersystems.org\""
         buildConfigField "boolean", "DEV_BUILD", "false"
+
+        multiDexEnabled true
```
moxie0: multi dex is a no-go

moxie0:
I think we want to send a "tile" which is just an image with the link under it, for compatibility with MMS

heyimgay:
We could populate the draft with both the geo uri and a link using a third party service like osm along with the tile image.

As for the icon see [here](https://github.com/google/material-design-icons/blob/master/device/drawable-xxxhdpi/ic_gps_fixed_white_48dp.png)

Fuzion24:
@moxie0, but to be clear, you're ok with using Google maps & Placepicker?

moxie0:
yep

heyimgay:
Note that we need to use the geo: uri to support the map app that is installed. Looking into this I found https://github.com/guardianproject/LocationPrivacy which is capable of converting between many links (including gmaps ofc) and geo: too.

moxie0:
I pushed a moxie/location branch which uses the placepicker to generate a image thumbnail and link

moxie0:
Ok I merged that location picker branch into 3.9.0, let me know what you think.  Thanks for discovering the place picker and putting this together!

PR #4597, opened 2015-11-19
Improve date localization for API >= 18 (4.3, JELLY_BEAN_MR2)

### LINKED ISSUE #3102: Date format does not respect locale

unrulygnu:
Fixes #3102 

**Before**
before

**After** (only Android 4.3 and above)
after

See https://github.com/WhisperSystems/Signal-Android/issues/3102#issuecomment-157965514 for more details.

### DIFF
--- a/res/layout/message_details_header.xml
+++ b/res/layout/message_details_header.xml
@@ -34,7 +34,8 @@
 
         <TableLayout android:id="@+id/metadata_container"
                      android:layout_width="match_parent"
-                     android:layout_height="match_parent">
+                     android:layout_height="match_parent"
+                     android:shrinkColumns="1">
 
             <TableRow android:id="@+id/sent_container"
                       android:layout_width="match_parent"
@@ -51,7 +52,7 @@
                           android:layout_width="wrap_content"
                           android:layout_height="wrap_content"
                           android:layout_marginLeft="@dimen/message_details_table_row_pad"
-                          tools:text="Jan 18, 9:29AM" />
+                          tools:text="Jan 18, 2015, 12:29:37 AM GMT-08:00" />
 
             </TableRow>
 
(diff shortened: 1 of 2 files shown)


mcginty:
With how long these date strings can get, I'm more and more inclined to just have times in the message bubble and a sticky date separator header.

2-4601:
Yeah, they can get pretty long (especially when some languages refuse to comply with the MMM string format).
Separator headers in the thread would be more clean and remove a lot of repetitive clutter from the bubbles.

But this does fix a long standing annoyance with the dates. Up till now many languages have been erroneously stuck in the `month, day` format. This releases them to their true form `day, month`.
Plus other small fixes (like wrong punctuation, separators, invalid grammar) that were just wrong.

unrulygnu:
@mcginty: this PR accommodates the current design, as well as moving dates to a separator header in the future: running custom date patterns through `getLocalizedPattern` fixes the issues @2-4601 noted in his comment for API >=18.

mcginty:
Yep, didn't intend to say this change was unnecessary (it totally is), it just made me think about ways to avoid particularly long date strings from expanding the message bubble size drastically.

Let's merge this into 2.6.0 now that that branch is live and I'll be working on adding some sticky date headers into this release as well to keep message bubble sizes safe from the infamous "death by finnish timestamp length".

mcginty:
f1869e436fd28c3ed9dabc2d39ff3c541b686c4e

PR #4616, opened 2015-11-20
update camera preview strategy

mcginty:
@McLoo @2-4601 would you mind trying this branch out to see if it alleviates some of the problems you saw?

### DIFF
--- a/src/org/thoughtcrime/securesms/components/camera/CameraUtils.java
+++ b/src/org/thoughtcrime/securesms/components/camera/CameraUtils.java
@@ -3,6 +3,7 @@ package org.thoughtcrime.securesms.components.camera;
 import android.app.Activity;
 import android.hardware.Camera;
 import android.hardware.Camera.CameraInfo;
+import android.hardware.Camera.Parameters;
 import android.hardware.Camera.Size;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


2-4601:
Sure.
* Quickcapture still sometimes hangs (on a still shot) at start but then proceeds if I tap the screen or make the mode full screen.
* Device provisioning camera works now! (But it also has that hanging issue sometimes at start)

mcginty:
@2-4601 thanks so much for the live action testing

I just updated this branch with an attempt to stop this hanging you're describing.

2-4601:
Ok, will test in a bit. Just re-tested with 3.4.2 to make sure it has worked before. The answer is yes. There's some quick stuttering at the beginning even there but it never gets stuck on a still frame.

mcginty:
Thanksthanks

Yeah, it will be near-impossible to remove all stuttering because some of it seems to happen internally by Android when spinning up the Camera on some devices (regardless of if we call the initializers off the main thread).

2-4601:
I see you have done some more while I was testing but results based on https://github.com/mcginty/TextSecure/commit/036e89f922cca7fe48389421525a09bef9737291
* I think it got stuck but maybe 1/50 times. I tested so many times that it's also possible that my brain hanged instead of the quick camera. At the end I was not able to make it hang at all.
* The very first time I tried the QR camera it definitely hanged on a still pic. But after that it never got stuck.

2-4601:
Ok, with the latest commit the camera _"has become unstuck in time."_ I.e. I can't get it to hang anymore.

moxie0:
in master

PR #4619, opened 2015-11-20
make RecyclerView unique IDs per row id/recipients combination

mcginty:


moxie0:
adding a new cache makes me nervous this late in the game

mcginty:
removed caching

PR #4620, opened 2015-11-20
run Camera.startPreview() off main thread

mcginty:
Tested on GB and a few phones, works as expected and removes 500ms of main thread work in the worst case I tested (Galaxy S3 and Galaxy Nexus).

It also happens to simplify the number of times we need catchall exception handling.


2-4601:
Tested this with Huawei U8800. Seems to work as good (or better) as current master at 4e8e8978f44d94beb94c92fe74027c3365a0bd2f

PR #4622, opened 2015-11-20
Java 8 deprecates _ as identifier, Java 9 disallows it.

kyanha:
see https://bugs.openjdk.java.net/browse/JDK-8061549

// FREEBIE


moxie0:
Thanks but this isn't our code, and we're a long ways away from Java 9 (or Java 8, for that matter).

PR #4651, opened 2015-11-21
Change "Forget from memory" to simply "Forget"

amire80:
Forgetting is usually from memory, so this makes
the message less redundant and easier to understand and to translate.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -916,7 +916,7 @@
     <string name="preferences__timeout_passphrase">Timeout passphrase</string>
     <string name="preferences__pref_timeout_interval_dialogtitle">Select passphrase timeout</string>
     <string name="preferences__pref_timeout_interval_title">Timeout interval</string>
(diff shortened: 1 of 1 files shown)


mcginty:
Even the word "forget," while technically accurate, sounds kind of confusing. Really the user wants Signal to re-lock itself after some amount of time, and they likely don't care what Signal "knows" in memory.

What about "Auto-lock"?

amire80:
Sounds OK. "Forget from memory" sounded really weird to me when I was translating. But if these strings are to be completely rewritten, it will be better if somebody who is more experienced than I am with using Signal does, because for now I am still more experienced with translating Signal than with using it.

RiseT:
@amire80 That's simple.

**"Auto-lock Signal after some timeout interval"**
**"Amount of time to wait before auto-locking Signal"** (ditched the imho redundant "The")

By the way - you have to adjust the string names as well (i. e. that's what Moxie wants). In strings.xml as well as in all code segments that are referencing those strings.

moxie0:
closing for lack of activity, feel free to reopen when ready

RiseT:
@moxie0 I could do a PR with this if desired.

PR #4662, opened 2015-11-22
updated dependency

cameronaaron:
Updated Dependancies and Gradle version *Currently this is not mergeable because the checksums don’t match with the ones manually provided and I can’t really figure out how to get the md5 of the updated dependencies i would be thankful if someone would lend me a hand*

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.2.3'
+        classpath 'com.android.tools.build:gradle:1.5.0'
         classpath files('libs/gradle-witness.jar')
     }
 }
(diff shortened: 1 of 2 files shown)


f41c0r:
you need to take the sha256 of the corresponding file for each dependency that gets downloaded. You can check in the gradle cache for dependencies you downloaded, so for example, for org.whispersystems.curve25519-java, you would find the corresponding dependency at 
`~/.gradle/caches/modules-2/files-2.1/org.whispersystems/curve25519-java/0.2.2/2dd03424fc32f806917ac01a98da82f03385db98/curve25519-java-0.2.2.jar' `

Or a dependency could be in  ~/.m2/ if you're looking for something in mavenLocal (e.g. a dependency you compiled and installed by hand in your local maven repo).

 Or you skip all that, delete all the verifications and just do:
`gradle -q calculateChecksums`

which is a gradle task that prints all the checksums of each dependency.

cameronaaron:
Got It thank you that makes sense! I appreciate it

moxie0:
Thanks but we don't update dependencies just to update them.  If they're stable, we use them until there's something we want from a new version, at which point we test the changes rigorously.

PR #4673, opened 2015-11-22
Remove unnecessary linebreaks from res/values/strings.xml

### LINKED ISSUE #4672: A lot of translatable messages have linebreaks that seem unnecessary
A lot of translatable messages have linebreaks that seem unnecessary. As just a few examples:
* registration_progress_activity__your_device_needs_network_connectivity
* registration_progress_activity__no_network_connectivity
* registration_progress_activity__waiting_for_sms_verification
* registration_activity__your_country_code_and_phone_number

There are dozens of others.

Are these linebreaks actually useful for how the app runs, or are they there just to keep the line lengths in res/values/strings.xml under a certain limit? If it's the latter, then maybe they should be removed, because when translating in Transifex, they appear in a rather weird way, and the Transifex interface sometimes complains that they are not translated correctly, even when it appears completely artificial.

amire80:
fixes #4672

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -14,9 +14,7 @@
     <string name="ApplicationPreferencesActivity_are_you_sure_you_would_like_to_immediately_trim_all_conversation_threads_to_the_s_most_recent_messages">Are you sure you would like to immediately trim all conversation threads to the %s most recent messages?</string>
     <string name="ApplicationPreferencesActivity_delete">Delete</string>
     <string name="ApplicationPreferencesActivity_disable_passphrase">Disable passphrase?</string>
-    <string name="ApplicationPreferencesActivity_disable_lock_screen">
-        Disable lock screen for messages?
-    </string>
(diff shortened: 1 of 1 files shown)


moxie0:
My hesitation is that there's a good chance this change would require translators to re-translate all of these strings in transifex?

amire80:
Yes, but see my comment in #4672 - it should take less than ten minutes, and if they don't do it, the current strings will just continue to work. (Unless Transifex is very silly and it deletes outdated strings. I'm not sure.)

moxie0:
I think transifex will delete outdated strings, which is probably the correct thing for it to do.  Since a lot of the translation teams aren't as on the ball as you are, I'm going to defer on making this change.  Let's just do it incrementally as the strings themselves change.  Thanks!

RiseT:
@moxie0 Yes, Transifex deletes outdated strings.

amire80:
That's silly :/

Oh well.

RiseT:
Well - to be more precise, it doesn't delete the string but the (old) translation of the string.

PR #4683, opened 2015-11-22
Hide devices preference pane. 

uberlaggydarwin:
Signal-Browser is not ready yet (uses the staging server) so to avoid user confusion remove the device provisioning menu.

// FREEBIE


bartz1:
It's always been there (but hidden) so if the developers are displaying it now it probably means that the browser extension is about to be released.

moxie0:
yep

PR #4689, opened 2015-11-23
Add MIME type to PersistentBlobProvider

### LINKED ISSUE #4536: Wrong file suffix on saving videos
I just received a video within a group chat in Signal 3.4.0. When I save that video, it is saved to sdcard\Movies\textsecure-2015-11-14-131819.attach

To make it play in other apps I have to open a file manager and rename the video to textsecure-2015-11-14-131819.mp4

I would expect Signal to directly save the file as textsecure-2015-11-14-131819.mp4

unrulygnu:
Fixes #4536

See https://github.com/WhisperSystems/Signal-Android/issues/4536#issuecomment-158854716 for long-winded details.

### DIFF
--- a/src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
+++ b/src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
@@ -1,6 +1,7 @@
 package org.thoughtcrime.securesms.providers;
 
 import android.annotation.SuppressLint;
+import android.content.ContentUris;
 import android.content.Context;
 import android.content.UriMatcher;
 import android.net.Uri;
@@ -32,7 +33,8 @@ public class PersistentBlobProvider {
   private static final String     URI_STRING            = "content://org.thoughtcrime.securesms/capture";
   public  static final Uri        CONTENT_URI           = Uri.parse(URI_STRING);
   public  static final String     AUTHORITY             = "org.thoughtcrime.securesms";
-  public  static final String     EXPECTED_PATH         = "capture/*/*";
+  public  static final String     EXPECTED_PATH         = "capture/*/*/#";
+  private static final int        MIMETYPE_PATH_SEGMENT = 1;
   private static final String     BLOB_EXTENSION        = "blob";
   private static final int        MATCH                 = 1;
   private static final UriMatcher MATCHER               = new UriMatcher(UriMatcher.NO_MATCH) {{
@@ -79,10 +81,11 @@ public class PersistentBlobProvider {
 
   private Uri create(MasterSecret masterSecret, InputStream input, long id, String mimeType) {
     persistToDisk(masterSecret, id, input);
-    return CONTENT_URI.buildUpon()
-                      .appendEncodedPath(String.valueOf(System.currentTimeMillis()))
-                      .appendEncodedPath(id + "." + getExtensionFromMimeType(mimeType))
-                      .build();
+    final Uri uniqueUri = CONTENT_URI.buildUpon()
+                                     .appendPath(mimeType)
+                                     .appendEncodedPath(String.valueOf(System.currentTimeMillis()))
+                                     .build();
+    return ContentUris.withAppendedId(uniqueUri, id);
   }
 
   private void persistToDisk(final MasterSecret masterSecret, final long id, final InputStream input) {
@@ -111,30 +114,18 @@ public class PersistentBlobProvider {
   public boolean delete(@NonNull Uri uri) {
     switch (MATCHER.match(uri)) {
     case MATCH:
-      long id = getId(uri);
+      long id = ContentUris.parseId(uri);
       cache.remove(id);
-      return getFile(id).delete();
+      return getFile(ContentUris.parseId(uri)).delete();
     default:
       return new File(uri.getPath()).delete();
     }
   }
 
-  public @Nullable InputStream getStream(MasterSecret masterSecret, Uri persistentBlobUri) throws IOException {
-    return MATCHER.match(persistentBlobUri) == MATCH ? getStream(masterSecret, getId(persistentBlobUri))
-                                                     : null;
-  }
-
-  private @NonNull InputStream getStream(MasterSecret masterSecret, long id) throws IOException {
+  public @NonNull InputStream getStream(MasterSecret masterSecret, long id) throws IOException {
     final byte[] cached = cache.get(id);
     return cached != null ? new ByteArrayInputStream(cached)
                           : new DecryptingPartInputStream(getFile(id), masterSecret);
-  }
-
-  private static long getId(Uri persistentBlobUri) {
-    final String filename = persistentBlobUri.getLastPathSegment();
-    int index = filename.lastIndexOf(".");
-    return index == -1 ? Long.parseLong(filename)
-                       : Long.parseLong(filename.substring(0, index));
   }
 
   private File getFile(long id) {
@@ -143,8 +134,7 @@ public class PersistentBlobProvider {
 
   public static @Nullable String getMimeType(@NonNull Context context, @NonNull Uri persistentBlobUri) {
     if (!isAuthority(context, persistentBlobUri)) return null;
-    return MimeTypeMap.getSingleton()
-                      .getMimeTypeFromExtension(MimeTypeMap.getFileExtensionFromUrl(persistentBlobUri.toString()));
+    return persistentBlobUri.getPathSegments().get(MIMETYPE_PATH_SEGMENT);
   }
 
   private static @NonNull String getExtensionFromMimeType(String mimeType) {
--- a/src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
+++ b/src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
@@ -33,14 +33,12 @@ public class PersistentBlobProvider {
   private static final String     URI_STRING            = "content://org.thoughtcrime.securesms/capture";
   public  static final Uri        CONTENT_URI           = Uri.parse(URI_STRING);
   public  static final String     AUTHORITY             = "org.thoughtcrime.securesms";
-  public  static final String     LEGACY_EXPECTED_PATH  = "capture/#/#";
   public  static final String     EXPECTED_PATH         = "capture/*/*/#";
   private static final int        MIMETYPE_PATH_SEGMENT = 1;
   private static final String     BLOB_DIRECTORY        = "captures";
   private static final String     BLOB_EXTENSION        = "blob";
   private static final int        MATCH                 = 1;
   private static final UriMatcher MATCHER               = new UriMatcher(UriMatcher.NO_MATCH) {{
-    addURI(AUTHORITY, LEGACY_EXPECTED_PATH, MATCH);
     addURI(AUTHORITY, EXPECTED_PATH, MATCH);
   }};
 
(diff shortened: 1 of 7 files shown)


unrulygnu:
Rebased on 3.5.1 (6a18824)

moxie0:
I don't 100% understand the rationale behind this change yet, but at first look we don't use shared preferences for things like that.  Shared prefs should be for actual preferences, or for managing state that there is exactly one of (registered or not, registered phone number, etc...) We don't want to use them to emulate a database.

unrulygnu:
@moxie0 , given the name, "Shared Preferences," I would agree, so I was surprised by the following from [Saving Key-Value Sets | Android Developers](https://developer.android.com/training/basics/data-storage/shared-preferences.html):

> If you have a relatively small collection of key-values that you'd like to save, you should use the SharedPreferences APIs.

That's exactly what is needed here, so I went with it given that it's a recommendation from the platform team.  I totally understand if you still don't want to use it for this purpose.  Since submitting this PR, I've also seen that attachment content type is stored encrypted in the database, so that's another concern that needs to be addressed.

The rationale is this:  without MIME type in PersistentBlobProvider, it gets completely lost when a persisted draft media file becomes an official attachment.  For audio and video, this means that they don't get the right extension on save, among potential other issues.

Since `PersistentBlobProvider`, at this point, keeps only files on storage (and an in-memory cache while writing), there is no existing data structure to which to append the MIME type.  Hence the quickie.

I can add a new database table for `PersistentBlobProvider` and update the PR if you'd like.  Let me know which direction you'd like to see this go.

moxie0:
@unrulygnu In this case I think it's a little too dynamic for shared prefs.  If I had exactly 10 of something, with 10 consistent key names, I'd consider using shared prefs for that.

If what we're mostly concerned about is the file extension, why don't we encode the mime type into the file extension of the file on disk that PersistentBlobProvider manages?

unrulygnu:
@moxie0 
> why don't we encode the mime type into the file extension of the file on disk that PersistentBlobProvider manages

Thanks for the feedback, especially today, you're awesome! I will look into this.  One last question: does the attachment's content type being encrypted in the database suggest that this data (and its relationship with the URI) should be secured?  Or are we ok tagging the encrypted blob with its actual content type in plaintext?

Updating the PR now with a rebase.  Will update within the next day or two with the requested solution.  

Thanks again!

moxie0:
In the AttachmentDatabase? I think only the content is encrypted there, which is mostly our policy with local storage.  I have no problem with tagging the encrypted blob with its content type.

unrulygnu:
> In the AttachmentDatabase?

`DraftDatabase.java`

    values.put(DRAFT_TYPE, masterCipher.encryptBody(draft.getType()));

> I have no problem with tagging the encrypted blob with its content type.

Thanks again for the clarification. I'm on it.

unrulygnu:
@moxie0 :  updated the PR, removing the use of SharedPreferences, and instead using the file extension of the blob file to reference the blob's type.  

Note that the use of `FilenameFilter` in `PersistentBlobProvider#getFile` also accommodates the transition which this PR introduces:  previously all blobs were appended with a `.jpg` extension, after merge the extensions will be variable. Since the `FilenameFilter` matches the ID portion of the filename (ignoring the extension), any existing blobs from before this PR was merged will be returned normally.

Problem types in issue [#4536](https://github.com/WhisperSystems/Signal-Android/issues/4536#issuecomment-158854716) were audio and video.  Tested draft saves, sending, saving attachments, and forwarding with several subtypes of each, as well as images.  All worked well on a Nexus 5.

There's an unrelated regression I discovered in testing which I believe was introduced by bcd0895, where blob files are not being deleted when attachment send is complete.  I will document this in a separate issue.

Please let me know what you think.

src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
```diff
@@ -29,12 +33,14 @@
...
-  private static final int        MATCH         = 1;
-  private static final UriMatcher MATCHER       = new UriMatcher(UriMatcher.NO_MATCH) {{
+  private static final String     URI_STRING             = "content://org.thoughtcrime.securesms/capture";
+  public  static final Uri        CONTENT_URI            = Uri.parse(URI_STRING);
+  public  static final String     AUTHORITY              = "org.thoughtcrime.securesms";
+  public  static final String     EXPECTED_PATH          = "capture/*/#";
+  private static final String     BLOB_DIRECTORY         = "captures";
+  private static final String     DEFAULT_EXTENSION      = "blob";
+  private static final int        MATCH                  = 1;
+  private static final UriMatcher MATCHER                = new UriMatcher(UriMatcher.NO_MATCH) {{
```
moxie0: alignment?

src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
```diff
@@ -121,12 +135,51 @@ public boolean delete(@NonNull Uri uri) {
...
+  }
+
+  private @Nullable File getFile(final long id) {
+    final File dir = context.getDir(BLOB_DIRECTORY, Context.MODE_PRIVATE);
+    final File[] files = dir.listFiles(new FilenameFilter() {
+      @Override
+      public boolean accept(File dir, String filename) {
+        return filename.startsWith(Long.toString(id));
+      }
+    });
```
moxie0: Iterating through the list of files doesn't feel great. We should probably encode the type/extension into the URI itself.

src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
unrulygnu: Fixed and pushed.  Sorry for the obvious mistake.

unrulygnu:
> Iterating through the list of files doesn't feel great. We should probably encode the type/extension into the URI itself.

I will get started on this tonight.

unrulygnu:
@moxie0: Sorry for the delay.  Persistent blob files of all types now save with just a generic `.blob` extension to avoid iterating through them.  Full, specific MIME types are now stored/queried in persistent blob URI's.

Drafts with persistent blobs created before this change will use the previous URI format, as well as the previously hardcoded `.jpg` extension for the blob files.  This has been accommodated in its own commit for specific review if needed.

Tested with video and audio shared from third party apps via `ACTION_SEND`, quick camera, and drafts created both before and after the commits in this PR.

Let me know what you think, and whether you need any further edits or squashing.

moxie0:
I'd like to encode the mime type into the file extension, not the path. I also don't want us to have some legacy URI support around forever, so let's keep the path the same and just leave those be.

unrulygnu:
@moxie0: I know your hands are probably full, and my recent intermittence is probably making things harder. Here's a summary of the commit/review history:

>>>>>>(unrulygnu) See https://github.com/WhisperSystems/Signal-Android/issues/4536#issuecomment-158854716 for long-winded details. (excerpt: I'm submitting a PR with an implementation using SharedPreferences)

>>>>>(moxie0) In this case I think it's a little too dynamic for shared prefs. \<snip\> why don't we encode the mime type into the file extension of the file on disk that PersistentBlobProvider manages?

>>>>(unrulygnu) updated the PR, removing the use of SharedPreferences, and instead using the file extension of the blob file to reference the blob's type. 

>>>(moxie0) Iterating through the list of files doesn't feel great. We should probably encode the type/extension into the URI itself.

>>(unrulygnu) Full, specific MIME types are now stored/queried in persistent blob URI's.

>(moxie0) I'd like to encode the mime type into the file extension, not the path.

Then would you like me to revert the recent commits and go back to the previous solution, using the blobs' file extensions, despite your reservations about iterating through the file list?  Note that the number of persistent blob files should be very few at any given time since they exist only while the media message is in draft state.

Or are you suggesting that we append a file extension to the numeric ID in the Content URI, breaking the [Android Content URI convention](https://developer.android.com/reference/android/content/ContentUris.html)?  In my opinion, this feels less great than iterating through the persistent blob files.

While I was trying to keep the impact of this PR to a minimum -- especially considering the simplicity of `PersistentBlobProvider` in its current state -- going to the database is another option.

Having spent a lot of time in `PersistentBlobProvider` lately, I feel that storing the MIME type in the URI path feels the most elegant.  The type stays with the URI in a clean one-to-one relationship, requiring no potentially janky querying/iterating elsewhere.  And the code handling the legacy URI format (which is important because it prevents user data loss on update) can be easily reverted after ample time is allowed for user migration.

Please let me know how you'd like to proceed.

moxie0:
Yes, I'd like the extension to be in the URI. I don't want to have a legacy URI format.

unrulygnu:
@moxie0: Removed the MIME Type from the URI path, appending relevant file extension to the ID (last path segment) instead.

Tested with media shared from third party apps, as well as quick camera.  Attachments save with the correct extension after sending.

What do you think?

unrulygnu:
Just rebased on current master (72064d8), no changes required.

src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
```diff
@@ -29,12 +29,14 @@
...
-  public  static final Uri        CONTENT_URI   = Uri.parse(URI_STRING);
-  public  static final String     AUTHORITY     = "org.thoughtcrime.securesms";
-  public  static final String     EXPECTED_PATH = "capture/*/#";
-  private static final int        MATCH         = 1;
-  private static final UriMatcher MATCHER       = new UriMatcher(UriMatcher.NO_MATCH) {{
+  private static final String     URI_STRING            = "content://org.thoughtcrime.securesms/capture";
+  public  static final Uri        CONTENT_URI           = Uri.parse(URI_STRING);
+  public  static final String     AUTHORITY             = "org.thoughtcrime.securesms";
+  public  static final String     EXPECTED_PATH         = "capture/*/*";
+  private static final String     BLOB_DIRECTORY        = "captures";
```
moxie0: no need for a constant if it's only used in one place, and is designed not to be accessible from anywhere else

moxie0:
Hmm, breaking `parseId` doesn't seem great.  Right now you have the file extension in the URI but not in the file.  What if you did the opposite, and put the file extension in the file but not the URI?

unrulygnu:
@moxie0:
> What if you did the opposite, and put the file extension in the file but not the URI?

We did this:

https://github.com/WhisperSystems/Signal-Android/pull/4689#issuecomment-159952025
> If what we're mostly concerned about is the file extension, why don't we encode the mime type into the file extension of the file on disk that PersistentBlobProvider manages?

https://github.com/WhisperSystems/Signal-Android/pull/4689#issuecomment-160397498
> updated the PR, removing the use of SharedPreferences, and instead using the file extension of the blob file to reference the blob's type. 

https://github.com/WhisperSystems/Signal-Android/pull/4689#discussion-diff-46447586
> Iterating through the list of files doesn't feel great. We should probably encode the type/extension into the URI itself.

src/org/thoughtcrime/securesms/providers/PersistentBlobProvider.java
unrulygnu: Reverted to string literal in 7e4eb1d.

moxie0:
lol, ok sorry for the back and forth. let's try it your way then

unrulygnu:
Updated, tested.  Note again that the issue this PR fixes concerns video/audio shared from third party apps to Signal via `Intent.ACTION_SEND`.  Attaching from within Signal is unaffected.

Also note that per your previous directive, existing drafts with persistent blobs will be nuked when users update Signal with this PR applied, due to the change in persistent blob URI:

https://github.com/WhisperSystems/Signal-Android/pull/4689#issuecomment-167116574
>  I don't want to have a legacy URI format.

Please let me know if you want me to squash the commits or need anything else.  Thanks!

moxie0:
in 3.13.0

PR #4692, opened 2015-11-23
Improve GroupCreateActivity

AsamK:
This PR fixes one sever (vanishing avatar when updating group) and two minor issues in the group create activty.

See commit messages for more info.

### DIFF
--- a/res/layout/group_create_activity.xml
+++ b/res/layout/group_create_activity.xml
@@ -29,7 +29,8 @@
                 android:src="@drawable/ic_group_photo"
                 android:contentDescription="@string/GroupCreateActivity_avatar_content_description" />
 
-            <EditText android:id="@+id/group_name"
+            <org.thoughtcrime.securesms.components.emoji.EmojiEditText
+                android:id="@+id/group_name"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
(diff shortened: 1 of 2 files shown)


res/layout/group_create_activity.xml
```diff
@@ -29,7 +29,8 @@
                 android:src="@drawable/ic_group_photo"
                 android:contentDescription="@string/GroupCreateActivity_avatar_content_description" />
 
-            <EditText android:id="@+id/group_name"
+            <org.thoughtcrime.securesms.components.ComposeText
```
mcginty: why ComposeText for this and not EmojiTextView?

AsamK: Because if i'm not mistaken, EmojiTextView can't be edited. And this is the field where the user inputs the group name.

mcginty: Sorry, I mean EmojiEditText

AsamK: Ah, ComposeText was the first one I saw, I'll change it to EmojiEditText.

AsamK:
Changed the last commit to use EmojiEditText instead of ComposeText

mcginty:
made some small tweaks, in 3.6.0 now. Thanks much!

PR #4694, opened 2015-11-23
select *all* threads

### LINKED ISSUE #4565: Mark all conversations - only two are marked then
When I long-press the "mark" icon, "mark all" is shown, but when I press "mark all", only two of my many conversations are marked.

This happens on all my devices running Signal 3.3.3 and I regard this a regression.

mcginty:
fixes #4565


moxie0:
needs rebase

mcginty:
in 3.5.1

PR #4702, opened 2015-11-24
fix "partial scroll" when sending in certain cases

### LINKED ISSUE #2856: Conversation doesn't always scroll to bottom on sending
- when the last message was a (**edit: n outgoing**)  media message 
- and new message ist text only 
- and keyboard is open when sending 

Second text message scrolls down correctly.

This drove me nuts for quite a while  😲
Feature  introduced in #1216

mcginty:
fixes #2856

do not ask me why a method called "isRunning" takes a callback that tells you when animations are completed because i don't have a good answer for you. it is one of the many great mysteries of the Android API naming convention choices.

also sped up animations because it looks more *lightning fast* this way.


moxie0:
in 3.6.0

PR #4707, opened 2015-11-24
Add Hebrew to language list

McLoo:
Now that Hebrew is 100% translated and pulled from transifex, let's enabled it in the language list.
Shalom.

### DIFF
--- a/res/values/arrays.xml
+++ b/res/values/arrays.xml
@@ -13,6 +13,7 @@
       <item>Español</item>
       <item>Français</item>
       <item>Greek ελληνικά</item>
+      <item>Hebrew עברית</item>
       <item>Hrvatski</item>
       <item>Indonesia</item>
       <item>Italiano</item>
@@ -49,6 +50,7 @@
       <item>es</item>
       <item>fr</item>
       <item>el</item>
+      <item>he</item>
       <item>hr</item>
       <item>in</item>
       <item>it</item>


res/values/arrays.xml
```diff
@@ -49,6 +50,7 @@
       <item>es</item>
       <item>fr</item>
       <item>el</item>
+      <item>he</item>
```
moxie0: think it's iw?

agrajaghh: http://stackoverflow.com/a/8470980/3755928
Android :imp:

McLoo:
@moxie0 that would explain the lang_map [he:iw](https://github.com/WhisperSystems/Signal-Android/blob/master/.tx/config#L3)

According to that stackoverflow discussion, we should go the HE way. If it would be possible to pull hebrew twice, we were on the 100% side if the language preference is set to default. 
(Maybe by defining a second lang_map entry *he:he* alongside the existing *he:iw*)

@amire80 would you mind sharing your expertise on that?

amire80:
Thanks for asking :)

First of all, after the latest upgrade from the Play store, Hebrew strings appear, so whatever you already did does work.

I know a few things about localization expert, but I'm not an Android development expert at all.

I do know that "iw" is an old and deprecated ISO code. "he" is now used almost everywhere, but "iw" somehow made it to the Java libraries that Android uses. Android apps is probably the last area of modern software where it still pops up.

Most of my work is around MediaWiki localization. When we export the translations to the Android app's source repository, it is saved as "iw". These are the files for the Wikipedia Android app: http://git.wikimedia.org/tree/apps%2Fandroid%2Fwikipedia.git/master/app%2Fsrc%2Fmain%2Fres . You can see that it uses values-iw, like Signal does now.

If you can change it to "he" without breaking anything else, it would be great, but don't waste too much time on that. Issue #4714, for example, is much more important for all RTL languages - Hebrew, Arabic, Persian, Urdu, and others.

(And, as I already wrote in #4714, I want to thank you again for making the translation process so smooth.)

moxie0:
@amire80 is there a yiddish translation forthcoming?

amire80:
Not that I know - I don't know Yiddish (unfortunately).

But Arabic, Persian, Urdu and Pashto should be very important (all of them have more speakers than Hebrew).

moxie0:
@amire80 ah, I thought I saw you request yiddish on transifex, got excited

McLoo:
@amire80 thanks for elaborating!

McLoo:
In case no one recognized:
Android (5.1.1) is happy with HE and IW - doubled checked that. 
Odd. Sanely handled though.

PR #4710, opened 2015-11-24
Inbox delivery status icons

### LINKED ISSUE #1363: Indicating in the conservation overview whether the last message is yours or from your conversation partner

### LINKED ISSUE #4049: delivery indicator in conversation list item
the conversation summary should show the most recent message's delivery status if it's outgoing

haffenloher:
This is based on @agrajaghh's nice [mockup](https://github.com/WhisperSystems/Signal-Android/pull/3012#issuecomment-94916759).

:warning: **This upgrades your database!** :warning:
Don't test this on your main device, you won't be able to roll back easily!

Screenshot:

signal inbox status icons

Fixes #1363
Fixes #4049


mcginty:
YESSSSSSSSSSS fantastic!

Personally I think the indicator would look better to the left of the snippet (where "You: " would have been).

src/org/thoughtcrime/securesms/database/ThreadDatabase.java
```diff
@@ -124,14 +128,17 @@ private long createThreadForRecipients(String recipients, int recipientCount, in
...
+                            long date, int deliveryStatus, int receiptCount, long type)
   {
-    ContentValues contentValues = new ContentValues(4);
+    ContentValues contentValues = new ContentValues(7);
     contentValues.put(DATE, date - date % 1000);
     contentValues.put(MESSAGE_COUNT, count);
     contentValues.put(SNIPPET, body);
     contentValues.put(SNIPPET_URI, attachment == null ? null : attachment.toString());
     contentValues.put(SNIPPET_TYPE, type);
+    contentValues.put(DELIVERY_STATUS, deliveryStatus);
```
mcginty: Changing the final static ints in DisplayRecord in this current setup would require a DB migration, which is an awkward dependency.

See `SmsMessageRecord.getGenericDeliveryStatus()` as a similar example - the constants stored in the database should be managed by the database and then translated appropriately afterwards for the display model.

src/org/thoughtcrime/securesms/database/model/MessageRecord.java
```diff
@@ -65,12 +58,11 @@
                 List<IdentityKeyMismatch> mismatches,
                 List<NetworkFailure> networkFailures)
   {
-    super(context, body, recipients, dateSent, dateReceived, threadId, type);
+    super(context, body, recipients, dateSent, dateReceived, threadId, deliveryStatus, receiptCount,
+            type);
```
mcginty: alignment

src/org/thoughtcrime/securesms/util/StatusManager.java
```diff
@@ -0,0 +1,76 @@
+package org.thoughtcrime.securesms.util;
+
+import android.view.View;
+
+public class StatusManager {
```
mcginty: You're totally right to make a custom view for the indicators, and now I think the logic for showing a specific indicator belongs in that view: `setPending()`, `setDelivered()`, `setSent()`, `hide()`, etc.

RiseT:
Excellent!
Any chance of putting the lock symbol next to the checkmarks if the last message is encrypted or is this off-topic regarding this PR?

agrajaghh:
I'm not sure the lock symbol is really important in the conversation list... I think thats too much information

haffenloher:
Thanks for all your comments :)
I'd also prefer not to include the lock symbol. Whether a conversation is encrypted or not is more of a long-term property, it usually doesn't change with every message.

New screenshot with indicators on the left:

inbox2

:neutral_face: 
I think I'm slightly in favor of having them on the right. It would also allow to simply use the same padding for the icons everywhere (both in the inbox & conversation views).

agrajaghh:
total nitpick: the red warning icon seem a bit too high. They don't look centered like the checkmarks :grin:

RiseT:
Well - while I agree that the encryption status *usually* doesn't change with every message or during a conversation, it certainly *can do* so. So I would see it as a per message property.

But nevermind. That's just what I *personally* would like to see, and I'm certainly no expert on user interfaces :smile: Still a great PR.

src/org/thoughtcrime/securesms/database/ThreadDatabase.java
haffenloher: Ah, I already wondered what that method was for, thanks for explaining!
Changed it, but now I have to translate the constants twice: MessageRecord -> ThreadDB -> ThreadRecord (or is there a better way?)

mcginty:
maybe I'm just a weirdo for liking it on the left still. I've seen it both ways in messengers (WhatsApp is left, Telegram and FB Messenger are right). If we do go with right (which is fine with me if I'm in the minority :) it probably makes sense to also have a "You: " prepended to outgoing message snippets as well, which there's a separate PR out for.

haffenloher:
I don't know, let's wait for some additional opinions :)

Adressed the rest of the comments, new screenshot for @agrajaghh:

inbox3

2-4601:
left... right... let the chicken decide

mawed:
I would prefer the right side.


Does pics and other media geht also signs?

agrajaghh:
hm, I think I prefer the right side as well...

warning icon looks better now :blush: 

how about adding a preference for the side? :smiling_imp: muhahahaha

RiseT:
While I intuitively would have put it on the right side, the left side doesn't look bad either. So no opinion, I guess... Another question is if additional symbols in the future are likely (e. g. an optional read status). That might change the picture (left vs. right side).

cascheberg:
Nice!
In the conversations the text is aligned left, and date and status is in the bottom right corner, so I find it much more consistent to have both on the right side in the conversation list as well.
And for me it's easier to distinguish the right-sided icons from the text at a glance, especially in the given short "Off Line - Are you there?" example.

RiseT:
*In the conversations the text is aligned left, and date and status is in the bottom right corner, so I find it much more consistent to have both on the right side in the conversation list as well.*

Consistency is certainly a valid argument that makes a lot of sense.

ChaoticMind:
What happens if the last message is a "special message", like a call? (e.g. x called you)

RiseT:
*What happens if the last message is a "special message", like a call? (e.g. x called you)*

I guess no symbol at all. One could use specific symbols for each of these events, though, like a phone symbol for calls.

haffenloher:
Fixed checkmark padding & chicken spelling
And added an outgoing call for @ChaoticMind:

inbox4

src/org/thoughtcrime/securesms/ConversationListItem.java
```diff
@@ -89,9 +105,26 @@ public ConversationListItem(Context context, AttributeSet attrs) {
...
+
+    this.statusManager = new StatusManager(deliveryStatusIndicator, failedIndicator,
+                                           pendingApprovalIndicator);
+
+    sentIndicator     .setPadding(0, sentIndicator.getPaddingTop(),
+                                  Math.round(sentIndicator.getPaddingLeft() * 0.7f),
+                                  sentIndicator.getPaddingBottom());
+    deliveredIndicator.setPadding(0, deliveredIndicator.getPaddingTop(),
+                                  Math.round(deliveredIndicator.getPaddingLeft() * 1.8f),
+                                  deliveredIndicator.getPaddingBottom());
```
moxie0: This view shouldn't have any visibility "inside" DeliveryStatusView, which should be a completely opaque thing that manages its sub-views.

haffenloher: yeah, I agree, I have to find a better solution for this. It was just a temporary change along with moving the icons from right to left, which makes different paddings necessary

src/org/thoughtcrime/securesms/util/StatusManager.java
```diff
@@ -0,0 +1,55 @@
+package org.thoughtcrime.securesms.util;
+
+import android.view.View;
+
+import org.thoughtcrime.securesms.components.DeliveryStatusView;
+
+public class StatusManager {
```
moxie0: I don't think this class should exist any longer

src/org/thoughtcrime/securesms/database/ThreadDatabase.java
moxie0: We should probably just stop translating back and forth and have the model store the raw SMS delivery status value, rather than a "simplified" version.  The only thing that actually accesses the delivery status other than the model itself is the exporter (which erroneously exports the simplified type), so I think it might be simpler just to have the model do the calculations rather than translating back and forth.

moxie0:
I think we should try alerts on the left and pending, sent, delivered on the right

mawed:
Coming back to my question: Does the picture also gets some delivery status icon? (Or is it an incoming message on the screenshot, so without any icon?)

Second: Does incoming messages get some icons?

haffenloher:
I think I like it:
inbox6

@mawed: I included an outgoing media message in the screenshot. No icons for incoming messages so far, my idea was that this would be enough to distinguish them from outgoing ones.

Will take a look at storing the raw SMS delivery status in the database tomorrow

mcginty:
Looking good!

That thin exclamation mark in the red warning circle is adorably tiny but probably hard to read on lower-density devices :).

McLoo:
We can haz a dark theme shoz?

The warning/failed indicator (exclamation mark)/i should be the same as in-convo and in notifications I think. (And IMO they all should be round and red)

cascheberg:
Is there any particular reason why the alerts should be on the left side?
From my intuition there should be distinct columns, to tell apart text from status at a glance: left is always text, rightmost is always status. Even if there is no status icon, maybe that rightmost space should not be occupied by text.

(After all someone might just send confusing emojis in a message: :x: :heavy_exclamation_mark: :no_entry_sign:)

Bastelbursche:
I think delivery status should stay on the left together with the message because it's meaning is explicitly linked to the specific message which is shown.

Warning circle could stay on the right because it's an special and important notice to the user, so its place should be special, too... and because of confusing emojis.

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -351,9 +342,10 @@ private void setFailedStatusIcons() {
   }
 
   private void setFallbackStatusIcons() {
-    statusManager.displayPendingApproval();
-    indicatorText.setVisibility(View.VISIBLE);
-    indicatorText.setText(R.string.ConversationItem_click_to_approve_unencrypted);
+    alertView              .setPendingApproval();
+    deliveryStatusIndicator.setNone();
+    indicatorText          .setVisibility(View.VISIBLE);
+    indicatorText          .setText(R.string.ConversationItem_click_to_approve_unencrypted);
```
moxie0: we're not this crazy with the alignment (aligning method names), just equals and definitions

smichel17:
I'm with moxie on this one. I think check marks and pending need to go on the right for visual consistency with inside a thread.

Error notifications could go on ether side.

src/org/thoughtcrime/securesms/ConversationListItem.java
```diff
@@ -74,6 +79,7 @@
   private final @DrawableRes int unreadBackround;
 
   private final Handler handler = new Handler();
+  private final Context context;
```
moxie0: what's this for? getContext() is always available to your view if you need it

src/org/thoughtcrime/securesms/components/DeliveryStatusView.java
```diff
@@ -0,0 +1,98 @@
...
+      inflater.inflate(R.layout.conversation_item_pending, pendingIndicatorStub, true);
+      TextView pendingIndicator = (TextView) findViewById(R.id.pending_indicator);
+      pendingIndicator.setTextColor(iconColor);
+    }
+
+    if (variableWidth) {
+      ViewGroup.LayoutParams params = pendingIndicatorStub.getLayoutParams();
+      params.width = ViewGroup.LayoutParams.WRAP_CONTENT;
+      pendingIndicatorStub.setLayoutParams(params);
+    }
```
moxie0: what is this for?

moxie0:
Getting close, go less crazy on the alignment, ditch the simplified "status" values/translation and you'll be substantially there.

jocelynthode:
Having everything aligned on the right is my personal preference.

haffenloher:
Updated - thanks for the feedback!

> That thin exclamation mark in the red warning circle is adorably tiny but probably hard to read on lower-density devices :).

Yep, I made a new icon (using ic_error_white_18dp), should be better now.

> The warning/failed indicator (exclamation mark)/i should be the same as in-convo and in notifications I think. (And IMO they all should be round and red)

Yeah, I already switched to the round icons in my last screenshot ;) The notifications actually use this instead:
ic_action_warning_red
The gray alert is for an insecure SMS fallback: A doesn't know B just unregistered and tries to send a secure message to B - will probably never happen for most people.

inbox7
inbox7dark

psyfx:
Love it! I would vote for everything on the right. If a message fails to send it is an important information for the user. Therefore error messages should ihmo aligned to right, since a failed message can be mistaken for an incoming message or an emoji, if the user checks his messages and does not look on the right *and* left side. If everything is right aligned, a quick check on the right corner tells a user the status of the message and is consistent with the conversation view.

Dyras:
I'm also voting for everything being on the right side.

However, more importantly I feel that the (!) icon on the dark theme isn't visible enough.

p4nci:
New on haffenloher's based mockup:

signal-convo-haffenloher-based-mockup-2015 11 27

McLoo:
To add more noise: The left/right combination looks great. All on the left is too distracting.
If you care for that status, move your eyes right. But you'd probably always care about errors...

Does RTL need to be handled here (swap left/right)?

scienmind:
I'd suggest adding an icon marking incoming messages as such.
This way it is consistent and straightforward - every conversation will have a status.

Regarding placement:
Reserving a screen area to hold all "message status info" info will be more clear and consistent:
One area with its boundaries holds message content preview, and another holds the status.

Something like this:
image

This design may be carried consistently to fix #4744 as well, placing "Archived" as a status there.

This also will improve RTL/LTR language handling:
RTL languages' text is right aligned (even with English UI, which is OK), and if we'll have a dedicated area (lets say, left column) just for the "status infos", the whole Conversations List will look much more streamlined and organized, even when mixed languages used (some messages LTR, some in RTL). 


(At first, split positioning sounded good, but it its less clear and can be confused with emojis as already mentioned, and multilingual texting is jumping right-left enough already...) 

(@p4nci 's suggestion is not a bad option either, though it looks a bit too info-dense on one side)

haffenloher:
Rebased.

We won't be able to reach a consensus on icon placement, that's for sure :grin: I think pretty much all of the proposed options are acceptable, it's mostly a matter of taste and little subtleties.

@p4nci's mockup is interesting, but I don't really see the benefit over other designs that would justify wasting so much space on the right.

Same for the emoji/icon confusion concerns. In my opinion this is a theoretical thing that won't be a real problem for people. The icons only appear for outgoing messages, so stuff you wrote yourself; they look different enough from any emoji; and even in the (unlikely) case that someone can trick you into believing an emoji in his/her message is actually a status icon, what would be the worst possible outcome? You tap on the inbox entry, read the message and realize your mistake. It's completely harmless.

@scienmind: I find it hard to understand the RTL problems without a screenshot (but I think they'll probably belong to the scope of #4714, e.g. replacing `toLeftOf` with `toStartOf`, `alignParentLeft` with `alignParentStart` etc. to simply flip the inbox layout.)

> more importantly I feel that the (i) icon on the dark theme isn't visible enough.

Hmm, maybe on bad screens? It's the standard icon for "pending insecure SMS fallback approval", the same as the in-convo one. Changing it would probably belong into another PR.

scienmind:
@haffenloher : The multilingual stuff I'm talking about, is not related to the #4714 (which is concerned with the UI translation and layout). I'm talking about using "normal" English interface, just texting messages in RTL languages occasionally.

Here is how it works now:
image

And as mentioned - no bugs implied, it's pretty much the expected behavior.
I just think that reserving a screen section for dates and status infos will also make multi-lingual messaging look more consistent and nicer:
image

smichel17:
Upon further thought, I'm with @scienmind 

Consistent UI is good. Moreover, when there is a media preview, the area marked as 'status' is already wasted; In that case, having an icon to the left would make the text area even smaller.

p4nci:
I'm for consistent UI to. To clarify my mockup:
all fields (except optional "Media preview") have well-defined structure + more space for different date formats (short/long).

signal-convo-2015 11 29

p4nci:
sorry, my bad: "Media preview" should be in the message block.

signal-convo-2015 11 30

heyimgay:
The media preview should be larger for symmetry reasons. And we can use [these](https://www.google.com/design/icons/#ic_call_made) icons for calls.

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -336,14 +321,21 @@ private void setStatusIcons(MessageRecord messageRecord) {
 
     if      (messageRecord.isFailed())                     setFailedStatusIcons();
     else if (messageRecord.isPendingInsecureSmsFallback()) setFallbackStatusIcons();
-    else if (messageRecord.isPending())                    statusManager.displayPending();
-    else if (messageRecord.isDelivered())                  statusManager.displayDelivered();
-    else                                                   statusManager.displaySent();
+    else {
```
moxie0: purely stylistic, but if this one has braces the ones above it need them too. alternately inlined else conditions with setOutgoingStatusIcons, setPendingStatusIcons, etc... would work

res/layout/conversation_list_item_view.xml
```diff
@@ -111,6 +112,13 @@
                   android:textSize="12sp"
                 />
 
+        <org.thoughtcrime.securesms.components.DeliveryStatusView
+                  android:id="@+id/delivery_status"
+                  android:layout_width="wrap_content"
+                  android:layout_height="wrap_content"
+                  android:layout_below="@+id/date"
+                  android:layout_alignParentRight="true"
+                  app:iconColor="?attr/conversation_list_item_subject_color"/>
```
moxie0: Shouldn't this be to the left of archived, rather than on the other side of it?

src/org/thoughtcrime/securesms/components/AlertView.java
```diff
@@ -0,0 +1,69 @@
...
+    failedIndicator   = (ImageView) findViewById(R.id.sms_failed_indicator);
+
+    if (attrs != null) {
+      TypedArray typedArray = getContext().getTheme().obtainStyledAttributes(attrs, R.styleable.AlertView, 0, 0);
+      boolean useSmallIcon = typedArray.getBoolean(R.styleable.AlertView_useSmallIcon, false);
+      typedArray.recycle();
+
+      if (useSmallIcon) {
+        failedIndicator.setImageDrawable(ContextCompat.getDrawable(getContext(), R.drawable.ic_error_red_18dp));
+      }
```
moxie0: What's this about?  This feels a little specific for a generic view to me, but I don't fully get the rationale.

res/layout/conversation_list_item_view.xml
haffenloher: Currently, archived + delivery_status never get shown at the same time, so I simply used alignParentRight for both.

Using `toLeftOf="@id/archived"` on delivery_status unfortunately leads to this:
screenshot from 2015-12-01 22-45-15

src/org/thoughtcrime/securesms/components/AlertView.java
haffenloher: [mcginty noted](https://github.com/WhisperSystems/Signal-Android/pull/4710#issuecomment-159751967) that the original in-convo error icon looks bad when scaled down for use in the conversation list, so I added a separate, smaller error icon with a larger exclamation mark. `useSmallIcon` switches it on or off.

p4nci:
How about a (split) lines (see also advanced settings) between items/contacts, regardless of the discussion left/right, only right, only left.

signal-convo-2015 12 01

res/layout/conversation_list_item_view.xml
moxie0: Dang, yeah this layout is getting... dynamic. So what's the rationale for not showing delivery status in the archived view?

haffenloher: Yay, just learned about `android:layout_alignWithParentIfMissing`:
> If set to true, the parent will be used as the anchor when the anchor cannot be be found for layout_toLeftOf, layout_toRightOf, etc.

it's perfect for the leftOf problem :)

> what's the rationale for not showing delivery status in the archived view?

There is none, I guess I was just being lazy.
Here are two different placements, it gets a little cluttered when there's a thumbnail:
archived1
archived2

cascheberg:
How about moving the "Archived" indicator up to the left or right side of the username?

haffenloher:
@cascheberg's 3rd possibility
archived3

I'm too stupid to put "Archived" to the right of the name atm :(

2-4601:
Isn't the _Archived_ box a bit redundant anyway? I mean all the conversations in the **Conversations archive** are in fact archived. Therefore the text box doesn't bring any new information.
Well, I guess a user might forget that they are in the archive and the box could be seen as an additional reminder... but yeah, would help also this pr if the box was dropped completely.

p4nci:
I agree with 2-4601: additional "Archived" i's a bit too much.
Since the whole conversation is archived, how about icons like this (or similar):
signal-archived

moxie0:
@haffenloher  let's go with https://cloud.githubusercontent.com/assets/264472/11517587/c5f96026-988a-11e5-9696-861e451f0c27.png

haffenloher:
:+1: done

moxie0:
in 3.7.0

PR #4721, opened 2015-11-26
Make strings more consistent, "3G" -> "mobile data"

### LINKED ISSUE #4720: Strings inconsistent, "3G" vs "mobile data"
3634ba0 (2013-07-08) included a string inconsistency, "3G" vs "mobile data":

        <string name="registration_progress_activity__your_device_needs_network_connectivity">Your
            device needs network connectivity in order to use this TextSecure feature. Check to ensure
            that it is connected to 3G or Wifi.
        </string>
        
        <string name="registration_progress_activity__if_you_are_connected_via_wifi_its_possible_that_there_is_a_firewall">
            If you are connected via wifi, it\'s possible that there is a firewall blocking access to
            the TextSecure server. Try another network or mobile data.
        </string>

This is the only instance of "3G" in the string resources at this time (e4e8511). Since "mobile data" covers all cellular data connections, it would be better if we roll with that. A later commit, aaf3d6a (2015-09-04), also uses "mobile data":

        <string name="preferences_chats__when_using_mobile_data">When using mobile data</string>

unrulygnu:
Fixes #4720
// FREEBIE

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -797,7 +797,7 @@
     </string>
     <string name="registration_progress_activity__your_device_needs_network_connectivity">Your
         device needs network connectivity in order to use this Signal feature. Check to ensure
-        that it is connected to 3G or Wi-Fi.
+        that it is connected to mobile data or Wi-Fi.
     </string>
     <string name="registration_progress_activity__restrictive_firewall">Restrictive firewall.
     </string>


RiseT:
@unrulygnu Could you please remove line feeds when you come across them?

See e. g. https://github.com/WhisperSystems/Signal-Android/pull/4673

unrulygnu:
:+1:

PR #4728, opened 2015-11-26
Added german translation strings (Play store description). Used 'Text…

vanitasvitae:
…Secure' as appname, like the english default strings did


vanitasvitae:
Oh just saw that german playstore is already translated.

PR #4741, opened 2015-11-28
Clarify how to link devices.

### LINKED ISSUE #4649: Tell the user how to actually link devices
Assume that:
1. the user has no idea there's a Signal browser extension
2. the user opens the "device" section
What would they think? I think that instead of displaying "No devices linked.." you should display something like "[...] you can download our Chrome extension to add new devices! <link>". Though I get it, right now there's no official link to get the chrome extension - but I assume there will be a way soon, or you wouldn't have set this menu entry as visible? Anyway, consider this as a note for the future.

uberlaggydarwin:
Fixes #4649

The Chrome extension is not up yet so please do not merge yet. I"m not sure of the best way to link users towards a Chrome extension (they don't want to open the page on their Android smartphone).

// FREEBIE

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -672,9 +672,8 @@
     <string name="device_link_fragment__link_device">Link device</string>
 
     <!-- device_list_fragment -->
-    <string name="device_list_fragment__no_devices_linked">No devices linked...</string>
+    <string name="device_list_fragment__no_devices_linked">No devices linked...\n \nYou can download our Chrome extension to add a device!</string>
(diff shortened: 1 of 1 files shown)


uberlaggydarwin:
The Chrome extension is not up yet so please do not merge yet. I"m not sure of the best way to link users towards a Chrome extension (they don't want to open the page on their Android smartphone).

Possibly useful: https://developer.chrome.com/webstore/branding

moxie0:
Thanks, if it's important to include this string, I think we'd want a separate textview rather than using newlines in this case.

PR #4746, opened 2015-11-28
Increase swipe-to-archive snackbar duration

### LINKED ISSUE #4745: Swipe-to-archive snackbar disappears too fast, inconsistent
When swiping a conversation to archive, the snackbar's duration is set to peekaboo (`LENGTH_SHORT`):

            new SnackbarAsyncTask<Long>(getView(),
                                        getString(R.string.ConversationListFragment_archived_conversation),
                                        getString(R.string.ConversationListFragment_undo),
                                        getResources().getColor(R.color.amber_500),
                                        Snackbar.LENGTH_SHORT, false)

When archiving conversations via the menu, however, the snackbar's duration is set to `LENGTH_LONG`, which gives the user more time to read, comprehend, and take action on the snackbar:

        new SnackbarAsyncTask<Void>(getView(), snackBarTitle,
                                    getString(R.string.ConversationListFragment_undo),
                                    getResources().getColor(R.color.amber_500),
                                    Snackbar.LENGTH_LONG, true)
        {

It would be best if they both were set to `LENGTH_LONG`.

unrulygnu:
Fixes #4745
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -440,7 +440,7 @@ public class ConversationListFragment extends Fragment
                                     getString(R.string.ConversationListFragment_moved_conversation_to_inbox),
                                     getString(R.string.ConversationListFragment_undo),
                                     getResources().getColor(R.color.amber_500),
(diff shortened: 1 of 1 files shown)


moxie0:
in 3.7.0

Issue #3609, opened 2015-07-11
"Send TextSecure message" should be renamed to something like "Type TextSecure message" and should have a lighter gray


wp9015362:
Hello,

when the message input box is empty, then there is a text in there in gray which says "Send TextSecure message".

Wouldn't it make more sense to rename this to "Type TextSecure message"?

For comparison:

In WhatsApp for Android it says "Type a message". And in Telegram for Android it simply says "Message".

Also:

In both WhatsApp for Android as well as in Telegram for Android, this text has a lighter gray color compared to TextSecure.

In TextSecure the text almost is as dark as the text that is actually typed into the message input box.

It would be nice if you could give the text a lighter gray as well to differentiate it more from the text that is actually being typed into the message input box.

And, as already mentioned in https://github.com/WhisperSystems/TextSecure/issues/2769, it would also be nice if you could adjust it's font size to match that of the messages in the bubbles.

Regards

2-4601:
More comparisons:

**kik:**
_Type a new message_

**Viber:**
_Type to compose_

**Facebook Messenger:**
_Write a new message_

**Snapchat:**
_Send a chat_

**Stock 4.0 SMS app**
_Type message_

**Google Messenger SMS app:**
_Send message_

**Hangouts:**
_Send Hangouts message_

**Signal iOS** ![1f610](https://cloud.githubusercontent.com/assets/174176/8633602/ea35538c-27d8-11e5-9ba7-3ce3f3ddffe7.png)
_New message_

McLoo:
And in landscape mode the button text then would be 'type TextSecure message' 😀

wp9015362:
> @McLoo wrote:
> 
> And in landscape mode the button text then would be 'type TextSecure message' :grinning:

What do you mean? What does it have to do with landscape mode?

McLoo:
![screenshot_2015-07-11-14-43-29](https://cloud.githubusercontent.com/assets/5326916/8634348/b325722e-27f5-11e5-95ba-e73b44c92a40.png)

wp9015362:
@McLoo :

You seem to be using an ancient version of TextSecure or something seems to be wrong with your TextSecure installation.

TextSecure 2.21.0 is using material design and does not have a button which says "Send TextSecure message".

TextSecure 2.21.0 has a round button with blue color and with a logo inside it, not with a text.

What version of TextSecure are you using?

dkasak:
[...]

PR #4755, opened 2015-11-28
Wording: send -> type (conversation's input text field)

RiseT:
// FREEBIE

- Replace '*send*' with '*type*' in the conversation's input text field

**Reason:** The current wording doesn't make much sense to me, as tapping on the text field containing one of these strings leads to *typing* rather than *sending* the message.

**Alternative:** An alternative to '*type*' would be '*write*'. If that is prefered, I'll be happy to change it accordingly.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -626,9 +626,9 @@
     <string name="conversation_title_view__conversation_muted">Conversation muted</string>
 
     <!-- conversation_activity -->
-    <string name="conversation_activity__type_message_push">Send Signal message</string>
-    <string name="conversation_activity__type_message_sms_insecure">Send unsecured SMS</string>
-    <string name="conversation_activity__type_message_mms_insecure">Send unsecured MMS</string>
+    <string name="conversation_activity__type_signal_message">Type Signal message</string>
+    <string name="conversation_activity__type_unsecured_sms">Type unsecured SMS</string>
+    <string name="conversation_activity__type_unsecured_mms">Type unsecured MMS</string>
     <string name="conversation_activity__send">Send</string>
     <string name="conversation_activity__remove">Remove</string>
     <string name="conversation_activity__window_description">Conversation with %1$s</string>
--- a/src/org/thoughtcrime/securesms/TransportOptions.java
+++ b/src/org/thoughtcrime/securesms/TransportOptions.java
@@ -92,20 +92,20 @@ public class TransportOptions {
       results.add(new TransportOption(Type.SMS, R.drawable.ic_send_sms_white_24dp,
                                       context.getResources().getColor(R.color.grey_600),
                                       context.getString(R.string.ConversationActivity_transport_insecure_mms),
-                                      context.getString(R.string.conversation_activity__type_message_mms_insecure),
+                                      context.getString(R.string.conversation_activity__type_unsecured_mms),
                                       new MmsCharacterCalculator()));
     } else {
       results.add(new TransportOption(Type.SMS, R.drawable.ic_send_sms_white_24dp,
                                       context.getResources().getColor(R.color.grey_600),
                                       context.getString(R.string.ConversationActivity_transport_insecure_sms),
-                                      context.getString(R.string.conversation_activity__type_message_sms_insecure),
+                                      context.getString(R.string.conversation_activity__type_unsecured_sms),
                                       new SmsCharacterCalculator()));
     }
 
     results.add(new TransportOption(Type.TEXTSECURE, R.drawable.ic_send_push_white_24dp,
                                     context.getResources().getColor(R.color.textsecure_primary),
                                     context.getString(R.string.ConversationActivity_transport_signal),
-                                    context.getString(R.string.conversation_activity__type_message_push),
+                                    context.getString(R.string.conversation_activity__type_signal_message),
                                     new PushCharacterCalculator()));
 
     return results;


2-4601:
AFAIK the reason for using the word _'send'_ is that the string turns to the name of the send button in landscape. See #3609

RiseT:
@2-4601 Mh. I see what you mean. This has a pretty workaround-ish look to me. I don't really like the general look in landscape to be honest.

In landscape mode, the string is displayed twice (text field & button), which looks sort of stupid (sorry...). Furthermore, the dark theme isn't respected (as I've written elsewhere).

I guess - when it's done the way it currently is - there's no way to use separate strings for the text field and for the button in landscape mode, right?

RiseT:
Thanks for the info @2-4601 , I close the PR and will open an issue instead soon.

RiseT:
@unrulygnu Do you think you can do a PR for this, i. e. separating the text field string from the button string? Doesn't sound too difficult but unfortunately I have zero Java knowledge.

unrulygnu:
@RiseT: 

In landscape, it looks like we are relying completely on the IME (keyboard) to generate the input UI, and all we can do is feed it text to apply to specific views.

> there's no way to use separate strings for the text field and for the button in landscape mode, right?

No, you can use separate strings. See `src/org/thoughtcrime/securesms/components/ComposeText.java`:

        if (isLandscape()) setImeActionLabel(transport.getComposeHint(), EditorInfo.IME_ACTION_SEND);
        // ...
        setHint(transport.getComposeHint());

There is the send button label, and the hint (which disappears once your start typing).  

> I suggest using the existing 3 strings for the text (input) field, and for the button a new string with just 'Send'. 

If we change the button label to "Send", then there is nothing indicating to the user whether this is a Signal or SMS message once text has been entered by the user, since the compose hint disappears.  Also consider the lack of information when returning to a saved draft in landscape mode.

So your bonus nice-to-have of using some other visual indicator of the transport, in my opinion, would be more of a must-have necessity if we took the transport name out of the button.  In order to do this, however, we would have to build and maintain the landscape data entry UI, to which there was reluctance in other comments I've seen.

One doable way to give some screen real estate back to the message input would be to put some newlines in the button string, while leaving the compose hint as-is:

landscape-compose-text-newlines-signal-message

There are two options to implement this:

1. Separate string resource: would require some plumbing, and the translators would have to be informed that the newlines are required after each word for this particular string. Worst case scenario is some locales have a big button like they do now, if the translators omit the newlines.

2. Programmatically replace spaces with newlines in the button: no translator overhead, but may cause unforeseen problems in some locales.

What do you think?  Too ugly?  Awesome?  Worth opening an issue to get the boss's feedback?

RiseT:
@unrulygnu Looking much better than the current state, that's for sure.

What kind of newlines do you mean? The Transifex (red arrows)/XML file ones or "\n"? Signal already uses "\n" in some strings, so that shouldn't be a problem for translators.

RiseT:
But the strings in/for portrait mode stay as they are, i. e. without newlines, right?

unrulygnu:
> What kind of newlines do you mean? 

`"\n"`

In this case, I used `"Send\nSignal\nmessage"`.  If you've cloned the repo and are able to build, you can test out various strings, make screenshots, etc.  Just refer to the file and lines I referenced in my comment above, replacing `transport.getComposeHint()` with whatever strings you want.

> But the strings in/for portrait mode stay as they are

The conditional on `isLandscape()` in the referenced line above assigns the landscape send button text, and nothing else.  Note that there's no conditional for setting the hint, though, so the way the code is currently structured, the compose hint will use the same string in both portrait and landscape.

RiseT:
So if I understand this correctly,

- "**Type** Signal message" (or the 2 other forms of that string) could be used for portrait, and
- "**Send**\nSignal\nmessage" could be used for landscape,

correct?

unrulygnu:
Sure.  Or you can use "**Type** Signal/SMS/MMS Message" for the compose hint for both portrait and landscape, and change only the landscape send button text to "**Send**\nSignal/SMS/MMS\nMessage".  

Both options would require some work to implement, since the code isn't set up to just drop the strings in to do like this now, but it shouldn't be a huge deal.  Just up to boss approval.

I've gotta be up in 4 hours, so I'm gonna :sleeping:  I'll get back with you tomorrow on any other items.

RiseT:
*so the way the code is currently structured, the compose hint will use the same string in both portrait and landscape.*

That's fine, I guess, no? What I was trying to say above (and totally f#k it up): What would be of importance, though, is that the compose hint uses different strings from the button, i. e. "Type [...]" for the compose hint strings and "Send [...]" for the button text strings.

RiseT:
Totally agree with your latest post. Perfect solution (that is, without coloring that button blue)!

Sleep well.

RiseT:
Ok - let's sum this up.

**Compose Hint Strings (portrait & landscape):**
- Type Signal message
- Type unsecured SMS
- Type unsecured MMS

**Button Text Strings (landscape):**
- Send\nSignal\nmessage
- Send\nunsecured\nSMS
- Send\nunsecured\nMMS

RiseT:
Btw - thanks very much for your nice explanation of the corresponding code. Just had the time to look at that in some detail.

So *setImeActionLabel()* is the method used for the button text and *setHint()* the one for the compose hint, correct?

unrulygnu:
Sorry for the delay.  Do you want to post a new issue on your strings suggestion for feedback?  I can probably put a PR together in a couple of days.

> So *setImeActionLabel()* is the method used for setting the button text and *setHint()* the one for the compose hint, correct?

Correct.

> PS: What does IME stand for?

[Input Method Editor](https://en.wikipedia.org/wiki/Input_method).

RiseT:
@unrulygnu Thanks & done.

RiseT:
@unrulygnu Another, unrelated question. There is a pretty wild mixture of blue & turquois regarding the button text color in Signal's alerts. One of many examples would be. Settings >> Appearance >> Theme uses a blue, Settings >> Advanced >> Signal messages and calls uses turquois (I prefer blue btw). Is this a technical limitation or an oversight?

unrulygnu:
@RiseT :  Cool! I'll hack on it tomorrow.

> Settings >> Appearance >> Theme uses a blue,

CANCEL button render color: 1c7ac5

> Settings >> Advanced >> Signal messages and calls uses turquois

Switch render color: 1b78c1
CANCEL / OK buttons render color: 80cbc4

> Is this a technical limitation or an oversight?

Without digging in, I can't think of a technical reason. It looks like a theme/style issue.  Good catch!

RiseT:
@unrulygnu Thanks for the info.
While we're at it - is there a technical reason that dialogs like Settings >> Notifications >> Sound still have a white background when using the dark theme? Are these some standard dialogs of Android that cannot be modified easily (like the blue button in landscape mode)?

heyimgay:
Its a RingtonePreference element which you can't theme. https://stackoverflow.com/questions/16589467/ringtonepreference-theme#16702655

RiseT:
@heyimgay I understand. Thanks for the info.

RiseT:
@unrulygnu Is standardizing the rendering colors mentioned above to $1b78c1 (or $1c7ac5...? Don't know) on your to-do list, i. e. are you planning doing this? :grinning:

unrulygnu:
@RiseT: Not yet.  Were you planning to submit an issue on it? That could allow others to pick it up if I'm not quick enough. :sunglasses:

RiseT:
@unrulygnu Ok, cool - I'll submit an issue. No hurry.

PR #4770, opened 2015-12-01
Return to archive on back button press in archived convo

haffenloher:
When the action bar's back/up/home button is pressed in an archived conversation, return to the archive instead of the inbox (to match the behavior of the phone's own back button).

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -163,6 +163,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
 
   public static final String RECIPIENTS_EXTRA        = "recipients";
   public static final String THREAD_ID_EXTRA         = "thread_id";
+  public static final String IS_ARCHIVED_EXTRA       = "is_archived";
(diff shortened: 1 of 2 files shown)


2-4601:
What should the _back_ behaviour be in this case:
1. Open an archived conversation
2. Send a new message in it (Note: This will move it to inbox)
3. Press back

Where should it go? Inbox? Archived conversations?

heyimgay:
Inbox because we assume the reason the user went into the archived view was to get to that specific conversation.

haffenloher:
I expect *back* to bring me back to the previous view (the archive), even after sending a message.
On returning to the archive view, the un-archived convo isn't gone immediately (which could be confusing), but there's a nice animation of the conversation item disappearing from the list.

heyimgay:
But sending a message changes the location of the conversation while you are in it. So when you get out of the view you are where the conversation is.

2-4601:
My main worry is that the user does not know that sending a message moves the conversation back to the inbox. But if there is an animation it will solve this as it gives **direct** feedback to the user what happened to the conversation.

heyimgay:
There already is an animation on entry of the conversations view. What is missing is to scroll back to the top.

PR #4779, opened 2015-12-02
Refactorings

plafer:


moxie0:
I'll take the SmsTransportDetails commit if you'd like to submit a PR with just that change, but I'm not interested in the others.  In general we prefer one logical change per PR. Thanks!

Issue #4752, opened 2015-11-28
Dates have wrong locale in 'Manage linked devices'


2-4601:
3.6.0
### System language: Korean, Signal language: Default
![screenshot_2015-11-28-17-21-20](https://cloud.githubusercontent.com/assets/174176/11453026/c0f20e22-9603-11e5-8f29-79923478a03a.png)

--------------------
### System language: English (US), Signal language: Korean
![screenshot_2015-11-28-17-22-35](https://cloud.githubusercontent.com/assets/174176/11453027/c55f47c2-9603-11e5-809a-1667bcf504ab.png)
(oh yeah, the title^ has some localization issues too... it's translated but does not change immediately after Signal language change)

PR #4780, opened 2015-12-02
Extract "Today" in device list item to string resource

unrulygnu:
Related to #4752
// FREEBIE

Fully fixing #4752 requires integrating dynamic language.


moxie0:
in 3.7.0

PR #4790, opened 2015-12-02
Linkify geo URIs

### LINKED ISSUE #4206: Linkify geo-schemed uris
I'm using Signal 3.0.1 beta on Android on a Nexus 5. Debug log (though I doubt you'll want it) is https://gist.github.com/526f0bbf1de80801e008

It would be great if geo: uris in messages that I receive turned into hyperlinks that, when clicked, gave the choice of opening the geo: link in whatever applications the user has installed that deal with them.

geo: uris are a way of encoding location information in a universal way, instead of doing so in a mapping-application specific way. See:
http://geouri.org/
https://en.wikipedia.org/wiki/Geo_URI_scheme
https://developer.android.com/guide/components/intents-common.html#Maps

The geo intent appears to be supported on Android (if I click on a geo link in Chrome, it asks me which of my various mapping applications I would like to open it with), but geo: uris within messages in Signal do not turn into hyperlinks and cannot be clicked to load up the location on a map. I commonly see this with OSMAnd links, which generate a geo: uri by default when one uses it to send a location. I believe that GeoSMS also sends geo: links to other people via SMS.

To a degree this is a privacy issue. A key reason I use OSMAnd is because it does offline mapping and geo: uris allow you to send and resolve locations without sending them to a remote webserver (like Google Maps). 

I'm not an Android developer, but it looks to me like the 7-odd lines of example code in the developer.android.com link above should go a long way to making these geo: uris work.

rmueller83:
Fixes #4206

The geo URI RegEx is from https://codecov.io/github/zxing/zxing/core/src/main/java/com/google/zxing/client/result/GeoResultParser.java

The idea of LinkifyHelper is borrowed from ChatSecure and could be extended to (their) other URI types: https://github.com/guardianproject/ChatSecureAndroid/blob/master/src/info/guardianproject/util/LinkifyHelper.java

Example geo URIs for testing can be found here: https://en.wikipedia.org/wiki/Geo_URI_scheme

The next step would be Signal to generate geo URIs using the location API.

### DIFF
--- a/res/layout/conversation_item_received.xml
+++ b/res/layout/conversation_item_received.xml
@@ -75,7 +75,6 @@
                     android:textColor="?conversation_item_received_text_primary_color"
                     android:textColorLink="?conversation_item_received_text_primary_color"
                     android:textSize="@dimen/conversation_item_body_text_size"
-                    android:autoLink="all"
                     android:linksClickable="true" />
 
             <LinearLayout android:id="@+id/mms_download_controls"
--- a/res/layout/conversation_item_sent.xml
+++ b/res/layout/conversation_item_sent.xml
@@ -77,7 +77,6 @@
 
             <org.thoughtcrime.securesms.components.emoji.EmojiTextView
                     android:id="@+id/conversation_item_body"
-                    android:autoLink="all"
                     android:layout_width="wrap_content"
                     android:layout_height="wrap_content"
                     android:paddingLeft="4dp"
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -28,7 +28,6 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v7.app.AlertDialog;
 import android.text.TextUtils;
-import android.text.util.Linkify;
 import android.util.AttributeSet;
 import android.util.Log;
 import android.view.LayoutInflater;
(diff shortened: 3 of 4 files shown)


moxie0:
Can you clarify why we want support for geo uris?

dcherian:
It is a standard way of sharing location information. The OS should let you open a mapping application if you tap on the link.

https://en.wikipedia.org/wiki/Geo_URI_scheme

rmueller83:
Exactly. This is related to #4591 where @moxie0 wrote "I think we want to send a "tile" which is just an image with the link under it". The geo URI would be the link then. Clicking on it opens an external application like Google Maps which for example allows the user to navigate to the received location.

heyimgay:
Relevant: https://github.com/guardianproject/LocationPrivacy

moxie0:
I think we'll just encode the location as a google maps uri rather than messing with self-linkifying to get geo uris.  Thanks for looking into this!

heyimgay:
@moxie0 this feature might still be desired when sharing the location from another app through signal

snaggen:
Geo uri is a generic solution that might be used to invoke all kinds of map
apps, where Google Maps is one of the apps that support it. Encoding it as
a Google Maps uri feels like a much more specific solution. In practice it
might not be much of a difference, since I guess most map apps register to
handle Google Maps links... but geo uri feels like the more standardized
way to go.

/Snaggen

tis 22 dec. 2015 14:26 heyimgay <notifications@github.com> skrev:

HolgerJeromin:
> In practice it
might not be much of a difference, since I guess most map apps register to
handle Google Maps links... 

Osmand does not generates google maps links. "only" geo: and an own url.

PR #4802, opened 2015-12-04
Wording ambiguity: Archived conversation -> Conversation archived

RiseT:
// FREEBIE

**Issue:**
There are 3 occurrences of 'Archived conversation[s]' in Signal - with 2 different meanings.

- *'Archived conversations'* means *'Conversations have been archived'* (i. e. archiving completed)
- *'Archived conversation'* means *'Conversation has been archived'* (i. e. archiving completed)

But:
- *'Archived conversations (%d)'* means 'Conversations ***which*** *have been archived'*

I've seen incorrect translations at Transifex because of this ambiguity.

**Solution:**
- Change the first two strings into *'Conversations archived'* and *'Conversation archived'* for differentiation.

I would think this is less ambiguous, but I'm not a native speaker. At the very least, it uses 2 different wordings for the 2 different meanings. So it's your judgment if this makes sense.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -180,7 +180,7 @@
     </activity-alias>
 
     <activity android:name=".ConversationListArchiveActivity"
-              android:label="@string/AndroidManifest_conversations_archive"
+              android:label="@string/AndroidManifest_archived_conversations"
               android:launchMode="singleTask"
               android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"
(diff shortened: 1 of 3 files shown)


2-4601:
I think it would also be good for consistency if the title of the archive (_'Conversations archive'_) was named the same as the _'Archived conversations (%d)'_ at the bottom of inbox, i.e. the title should be _'Archived conversations'_ (without the number).

RiseT:
@2-4601 Thanks & done.

moxie0:
in 3.8.0

PR #4809, opened 2015-12-04
Add inbox delivery status migration

### LINKED ISSUE #4799: Conversation list's delivery status icon shows only one checkmark until conversation opened
Seems to affect only conversations with the last message sent before updating to 3.7.0.  

After updating to 3.7.0, sending a message then quickly tapping Back before the conversation's delivery status icon is updated works as expected:  the conversation list's delivery status icon is updated in realtime.

**Steps to reproduce**

1. Launch Signal / navigate to the conversation list
2. Find pre-3.7.0 update Signal conversation with only one checkmark
3. Tap conversation
4. If the last message indeed has only one checkmark, repeat steps 1 through 3
5. Observe two checkmarks
6. Tap back to return to the conversation list
7. Behold the updated delivery status icon with two checkmarks

Device info: Nexus 5, stock Android 6.0 MRA58K, Signal 3.7.0 (b99c4e3)

App state: Signal registered, SMS enabled, default SMS app

haffenloher:
Fixes #4799

// FREEBIE

~~I used the version code of the upcoming version (176) to make sure the migration also runs for those with 3.7.0 installed.~~

Tested on an emulator and on my phone, works as expected.

### DIFF
--- a/src/org/thoughtcrime/securesms/database/DatabaseFactory.java
+++ b/src/org/thoughtcrime/securesms/database/DatabaseFactory.java
@@ -31,7 +31,6 @@ import org.thoughtcrime.securesms.crypto.DecryptingPartInputStream;
 import org.thoughtcrime.securesms.crypto.MasterCipher;
 import org.thoughtcrime.securesms.crypto.MasterSecret;
 import org.thoughtcrime.securesms.crypto.MasterSecretUtil;
-import org.thoughtcrime.securesms.database.model.ThreadRecord;
 import org.thoughtcrime.securesms.notifications.MessageNotifier;
 import org.thoughtcrime.securesms.util.Base64;
 import org.thoughtcrime.securesms.util.Util;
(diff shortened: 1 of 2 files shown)


moxie0:
Oh we probably don't want this to be an activity-level migration, those are generally only for migrations that require a mastersecret.  Can't we do this one in sql?

haffenloher:
no problem, I'll change it

haffenloher:
done, sorry for the delay. Please tell me if you prefer the migration to be added to the existing `INTRODUCED_CONVERSATION_LIST_STATUS_VERSION` instead of incrementing the db version a second time.

src/org/thoughtcrime/securesms/database/DatabaseFactory.java
```diff
@@ -790,6 +791,28 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
...
+        Cursor threadCursor = db.query("thread", new String[] {"_id"}, null, null, null, null, null);
+
+        while (threadCursor != null && threadCursor.moveToNext()) {
+          long threadId = threadCursor.getLong(threadCursor.getColumnIndexOrThrow("_id"));
+
+          Cursor cursor = db.rawQuery("SELECT DISTINCT date AS date_received, status, " +
+                                      "delivery_receipt_count FROM sms WHERE (thread_id = ?1) " +
+                                      "UNION ALL SELECT DISTINCT date_received, -1 AS status, " +
+                                      "delivery_receipt_count FROM mms WHERE (thread_id = ?1) " +
+                                      "ORDER BY date_received DESC LIMIT 1", new String[]{threadId + ""});
```
moxie0: Is there a reason to have the date in the projection?

moxie0:
@haffenloher Other than my date question, this looks good. Separate version is the way to go.

src/org/thoughtcrime/securesms/database/DatabaseFactory.java
haffenloher: There doesn't seem to be a way of sorting by date (sms) / date_received (mms) without adding them to the result set under the same name.

moxie0: ah, of course

PR #4815, opened 2015-12-05
Fix plurals

### LINKED ISSUE #4811: "the %s most recent messages" needs plural support
Hi,

The string ApplicationPreferencesActivity_this_will_immediately_trim_all_conversations_to_the_s_most_recent_messages says "This will immediately trim all conversations to the %s most recent messages."

It needs proper plural support. For example, in Russian "messages" is translated differently for "23 messages" and "36 messages".

cascheberg:
Fixes #4811 and other occurences

// FREEBIE

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -9,9 +9,15 @@
     <!-- ApplicationPreferencesActivity -->
     <string name="ApplicationPreferencesActivity_currently_s">Currently: %s</string>
     <string name="ApplicationPreferenceActivity_you_havent_set_a_passphrase_yet">You haven\'t set a passphrase yet!</string>
-    <string name="ApplicationPreferencesActivity_messages_per_conversation">%s messages per conversation</string>
+    <plurals name="ApplicationPreferencesActivity_messages_per_conversation">
+        <item quantity="one">1 message per conversation</item>
+        <item quantity="other">%d messages per conversation</item>
+    </plurals>
     <string name="ApplicationPreferencesActivity_delete_all_old_messages_now">Delete all old messages now?</string>
-    <string name="ApplicationPreferencesActivity_this_will_immediately_trim_all_conversations_to_the_s_most_recent_messages">This will immediately trim all conversations to the %s most recent messages.</string>
+    <plurals name="ApplicationPreferencesActivity_this_will_immediately_trim_all_conversations_to_the_d_most_recent_messages">
+        <item quantity="one">This will immediately trim all conversations to the most recent message.</item>
+        <item quantity="other">This will immediately trim all conversations to the %d most recent messages.</item>
+    </plurals>
     <string name="ApplicationPreferencesActivity_delete">Delete</string>
     <string name="ApplicationPreferencesActivity_disable_passphrase">Disable passphrase?</string>
     <string name="ApplicationPreferencesActivity_disable_lock_screen">
@@ -34,7 +40,10 @@
     <string name="ApplicationPreferencesActivity_sms_mms_summary">SMS %1$s, MMS %2$s</string>
 
     <!-- AppProtectionPreferenceFragment -->
-    <string name="AppProtectionPreferenceFragment_minutes">%d minutes</string>
+    <plurals name="AppProtectionPreferenceFragment_minutes">
+        <item quantity="one">1 minute</item>
+        <item quantity="other">%d minutes</item>
+    </plurals>
 
     <!-- DraftDatabase -->
     <string name="DraftDatabase_Draft_image_snippet">(image)</string>
@@ -109,8 +118,10 @@
     <string name="ConversationActivity_group_members">Group members</string>
     <string name="ConversationActivity_group_conversation">Group conversation</string>
     <string name="ConversationActivity_unnamed_group">Unnamed group</string>
-    <string name="ConversationActivity_d_recipients_in_group">%d members</string>
-    <string name="ConversationActivity_d_recipients_in_group_singular">1 member</string>
+    <plurals name="ConversationActivity_d_recipients_in_group">
+        <item quantity="one">1 member</item>
+        <item quantity="other">%d members</item>
+    </plurals>
     <string name="ConversationActivity_saved_draft">Saved draft</string>
     <string name="ConversationActivity_invalid_recipient">Invalid recipient!</string>
     <string name="ConversationActivity_calls_not_supported">Calls not supported</string>
--- a/src/org/thoughtcrime/securesms/ConversationTitleView.java
+++ b/src/org/thoughtcrime/securesms/ConversationTitleView.java
@@ -82,8 +82,7 @@ public class ConversationTitleView extends LinearLayout {
     int size = recipients.getRecipientsList().size();
 
     title.setText(getContext().getString(R.string.ConversationActivity_group_conversation));
-    subtitle.setText(getContext().getResources().getQuantityString(
-            R.plurals.ConversationActivity_d_recipients_in_group, size, size));
+    subtitle.setText(getContext().getResources().getQuantityString(R.plurals.ConversationActivity_d_recipients_in_group, size, size));
     subtitle.setVisibility(View.VISIBLE);
   }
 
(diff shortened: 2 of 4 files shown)


res/values/strings.xml
```diff
@@ -193,7 +204,10 @@
 
     <!-- DateUtils -->
     <string name="DateUtils_now">Now</string>
-    <string name="DateUtils_minutes_ago">%d min</string>
+    <plurals name="DateUtils_minutes_ago">
+        <item quantity="one">1 min</item>
+        <item quantity="other">%d min</item>
```
RiseT: There is no plural form for this.
See https://github.com/WhisperSystems/Signal-Android/issues/3295

cascheberg: That issue is specifically about the use of units which is ok.
But I think / it seems that there are some languages that do not treat this as a unit, but as a short or written-out form, depending on whatever is more common there. For example Polish seems to do that, and maybe non-latin languages? And it is done with "hours" instead of "h".
I see this as an option that translators can make use of. And I did not use "mins", so that should not spread.

RiseT: Ok, been wondering about non-latin languages as well, although I have been under the impression that these scientifc units are international standards.

haffenloher: If we don't know if there really is a problem, wouldn't it be better to delay changing that string until there's a translator actually complaining about it?

cascheberg: Well, nobody has complained about the other occurrences yet, also they are problematic even in English. Maybe someone should ask a polish translator, would be interesting to know ;)

RiseT:
The sort of "danger" (might be a tad strong) I'm seeing here is that if you ask translators (at Transifex) for a plural form that actually doesn't exist, chances are that they will come up with one nevertheless. Look at the issue I've refered to in my initial post. Back then, it has been "mins" in the English string, although a plural form for the unit "min" doesn't exist. Many people aren't aware of this (myself included back then).

res/values/strings.xml
RiseT: The sort of "danger" (might be a tad strong) I'm seeing here is that if you ask translators (at Transifex) for a plural form that actually doesn't exist, chances are that they will come up with one nevertheless. Look at the issue I've refered to in my initial post. Back then, it has been "mins" in the English string, although a plural form for the unit "min" doesn't exist. Many people aren't aware of this (myself included back then).

moxie0: @cascheberg I'd like you to make the most conservative change possible to solve the outstanding issue.

cascheberg:
@moxie0 Updated the PR to not include the "%d min" change that was discussed here.

src/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
```diff
@@ -76,7 +76,8 @@ private void initializePlatformSpecificOptions() {
   private void initializeTimeoutSummary() {
     int timeoutMinutes = TextSecurePreferences.getPassphraseTimeoutInterval(getActivity());
     this.findPreference(TextSecurePreferences.PASSPHRASE_TIMEOUT_INTERVAL_PREF)
-        .setSummary(getString(R.string.AppProtectionPreferenceFragment_minutes, timeoutMinutes));
+        .setSummary(getResources().getQuantityString(
+                R.plurals.AppProtectionPreferenceFragment_minutes, timeoutMinutes, timeoutMinutes));
```
moxie0: one line

src/org/thoughtcrime/securesms/ConversationTitleView.java
```diff
@@ -82,8 +82,8 @@ private void setRecipientsTitle(Recipients recipients) {
     int size = recipients.getRecipientsList().size();
 
     title.setText(getContext().getString(R.string.ConversationActivity_group_conversation));
-    subtitle.setText((size == 1) ? getContext().getString(R.string.ConversationActivity_d_recipients_in_group_singular) :
-                         String.format(getContext().getString(R.string.ConversationActivity_d_recipients_in_group), size));
+    subtitle.setText(getContext().getResources().getQuantityString(
+            R.plurals.ConversationActivity_d_recipients_in_group, size, size));
```
moxie0: one line

src/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
```diff
@@ -77,8 +77,8 @@ public boolean onPreferenceClick(Preference preference) {
       final int threadLengthLimit = TextSecurePreferences.getThreadTrimLength(getActivity());
       AlertDialog.Builder builder = new AlertDialog.Builder(getActivity());
       builder.setTitle(R.string.ApplicationPreferencesActivity_delete_all_old_messages_now);
-      builder.setMessage(getString(R.string.ApplicationPreferencesActivity_this_will_immediately_trim_all_conversations_to_the_s_most_recent_messages,
-                                   threadLengthLimit));
+      builder.setMessage(getResources().getQuantityString(R.plurals.ApplicationPreferencesActivity_this_will_immediately_trim_all_conversations_to_the_d_most_recent_messages,
+              threadLengthLimit, threadLengthLimit));
```
moxie0: align second line

src/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
```diff
@@ -107,7 +107,7 @@ public void onClick(DialogInterface dialog, int which) {
 
     public TrimLengthValidationListener() {
       EditTextPreference preference = (EditTextPreference)findPreference(TextSecurePreferences.THREAD_TRIM_LENGTH);
-      preference.setSummary(getString(R.string.ApplicationPreferencesActivity_messages_per_conversation, preference.getText()));
+      onPreferenceChange(preference, preference.getText());
```
moxie0: is this necessary for the minimal possible change?

src/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
```diff
@@ -116,18 +116,20 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
...
       }
 
-      if (Integer.parseInt((String)newValue) < 1) {
+      if (value < 1) {
         return false;
       }
 
-      preference.setSummary(getString(R.string.ApplicationPreferencesActivity_messages_per_conversation, newValue));
+      preference.setSummary(getResources().getQuantityString(
+              R.plurals.ApplicationPreferencesActivity_messages_per_conversation, value, value));
```
moxie0: one line

src/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
RiseT: Only if someone ever sets this to 1 -- which is pretty unlikely, as it doesn't make too much sense. Wondering about that one as well.

cascheberg: The `getQuantityString` for plurals requires an actual integer, which `preference.getText()` is not. So it is simple to re-use all the conversion and type checking stuff in `onPreferenceChange`.
It is of course also possible to just cast the String to Int right there, depending on what is preferred as a minimal change.

cascheberg: @RiseT well yeah, that is a different argument then, which I cannot judge.

moxie0:
in 3.8.0 thanks

src/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
cascheberg: @RiseT Update: actually, that is not even true, because some languages differentiate more than just one and other.

RiseT: You're right - didn't think about that.

RiseT:
@cascheberg Could you take a look at
- ConversationListFragment_conversations_archived
- ConversationListFragment_conversation_archived
- ConversationListFragment_moved_conversation_to_inbox
- ConversationListFragment_moved_conversations_to_inbox

This doesn't seem to work in 3.8.0, i. e. when archiving 1 conversation, ConversationListFragment_conversations_archived is displayed.

I'm aware you've had nothing to do with the creation of these strings, but you've done that singular/plural stuff before.

2-4601:
I think 3.8.0 has an issue when using the batch selection.
* long press a conversation (1 selected)
* archive it from the top bar icon
* "Conversations archived" string shows up

But if you archive with a swipe the singular string is correctly shown.

PR #4846, opened 2015-12-09
Fix wrong Chinese language entry

### LINKED ISSUE #4845: Wrong Chinese language entry
The current entry for Chinese when selecting languages in Settings is "中国的" which means "China's" (possessive).

A better entry would be "Chinese (Simplified) 中文 (简体)", which does mean what it says (a peek at Transifex shows that the Chinese (Traditional) 中文 (繁體) translation is coming soon).

darrenfoong:
Fixes #4845

// FREEBIE


moxie0:
in 3.8.0

PR #4861, opened 2015-12-11
Change separator-stripping behaviour for SMS recipients to fix #4833

### LINKED ISSUE #4833: Can’t send SMS to phone numbers with brackets after upgrade to 3.3+
Messages (SMS, not MMS, not data) appear to be sent successfully in the UI but are never received by the contact after upgrading to any version later than 3.1.1. Messages from other contacts are still received, only sending is broken. Manually downgrading to 3.1.1 apk restores SMS sending functionality. Other SMS apps continue to work. Latest 3.6.1 release from Play Store is still broken. Disabling Signal messages in Advanced prefs does not fix the problem. Uninstalling and reinstalling does not fix the problem.

Device info: Samsung Galaxy S5 Dev Ed, stock Android 4.4.4, Signal 3.6.1
App state: Push registered, SMS Enabled, SMS deliver disabled, Wi-Fi Calling compatibility disabled (phone is not Wi-Fi Calling enabled), MMS Default settings, Media auto-download disabled

Log, starting Signal & sending SMS: https://gist.github.com/csnover/dd39624d381160536798

This may be related to the shortcode sending problem reported around the same time but I am *not* sending to shortcode contacts, these are full 10-digit (US) numbers.

csnover:
This uses the standard Android API for stripping separators to avoid omitting any potentially valid characters. It also moves recipient validation down after the mangling, so at least if it ruins something, users hopefully will not have their messages simply being sent into the abyss.

Refs #1516, #3099, #4833. Follow-on to discussion at https://github.com/WhisperSystems/Signal-Android/commit/bde838d35a59d7e3791ecf6fc7fe5b00ccd7b2f1.

### DIFF
--- a/src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
+++ b/src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
@@ -20,7 +20,6 @@ import org.thoughtcrime.securesms.jobs.requirements.ServiceRequirement;
 import org.thoughtcrime.securesms.notifications.MessageNotifier;
 import org.thoughtcrime.securesms.recipients.Recipients;
 import org.thoughtcrime.securesms.service.SmsDeliveryListener;
-import org.thoughtcrime.securesms.transport.InsecureFallbackApprovalException;
 import org.thoughtcrime.securesms.transport.UndeliverableMessageException;
 import org.thoughtcrime.securesms.util.NumberUtil;
 import org.thoughtcrime.securesms.util.TextSecurePreferences;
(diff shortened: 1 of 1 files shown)


csnover:
I did end up adding convertKeypadLettersToDigits since it appeared that it is what would normally end up happening anyway. However, given that I don’t know the intimate details of every thing that might be accepted by a network, if this PR is accepted as-is it would probably be a good idea to put something specific in the release notes like “Behaviour when sending SMS to non-numeric recipients has changed, so please let us know if you run into any problems!”

moxie0:
looks good, you just gotta rebase it

csnover:
I have no idea how that was so out of date. It’s rebased against master now.

PR #3443, opened 2015-06-18
Enable preview for image drafts

McLoo:
This is pretty cool in three ways:
1. you can (p)review and save your TS-photographs (*Edit:in full resolution*)
2. you can (p)review (and save) your gallery picked images
3. McLoo is happy having this feature now

Could also be useful if somewhere in the future we have in app audio or video recording...

PR #4863, opened 2015-12-11
Enable preview for image drafts

McLoo:
as seen in #3443
(Only collabs can reopen if they had closed a PR/ an issue. Also, that closed PR does not seem to update - so much git left to learn...)

Had to use `setOnClickListener()` because a ThumbnailClickListener [only get's dispatched if](https://github.com/WhisperSystems/Signal-Android/blob/946c43940b1e5ddb06db4bf7e86e09284cea95ff/src/org/thoughtcrime/securesms/components/ThumbnailView.java#L184)
`slide.getTransferState()          == AttachmentDatabase.TRANSFER_PROGRESS_DONE`, which won't ever happen for a draft. It that remains in state `TRANSFER_PROGRESS_STARTED`.

Quite happy with this solution :smiley:

### DIFF
--- a/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
+++ b/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
@@ -138,7 +138,7 @@ public class MediaPreviewActivity extends PassphraseRequiredActionBarActivity im
 
     mediaUri     = getIntent().getData();
     mediaType    = getIntent().getType();
-    date         = getIntent().getLongExtra(DATE_EXTRA, -1);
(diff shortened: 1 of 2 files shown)


McLoo:
Removed the put of ´putExtra(MediaPreviewActivity.DATE_EXTRA´ and instead set the default date in MediaPreviewActivity.java. As there is no reason i can think of why it should by -1 instead.

That way
- there is no confusing "0 minutes ago" in action bar subtitle
- we still have the correct timestamp used for the filename
- we can save from MediaPreview without crashing even if - for whatever reason - we don't get a proper timestamp (crashing is what it does when no date is set, due to `
throw new AssertionError("uri, content type, and date must all be specified");`)

PR #4872, opened 2015-12-13
Fix crash when adding a contact from a conversation

jarteaga:
(First contribution, so let's start small.)

This line from #4259 seems to have introduced a NPE crash when adding a contact from a conversation. A null Bundle ends up being passed to this function:
```Java
  @Override
  public Loader<Cursor> onCreateLoader(int id, Bundle args) {
    return new ConversationLoader(getActivity(), threadId, args.getLong("limit", PARTIAL_CONVERSATION_LIMIT));
  }
```

Fix tested on a Nexus 7 (2012) running stock Android 5.1.1, on top of e79ee7803fea85ec999e0f6dc803fc1227096cd0.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -153,7 +153,7 @@ public class ConversationFragment extends Fragment
   }
 
   public void reloadList() {
-    getLoaderManager().restartLoader(0, null, this);
+    getLoaderManager().restartLoader(0, Bundle.EMPTY, this);
   }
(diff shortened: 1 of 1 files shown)


cascheberg:
I can confirm that the original code crashes, and that this fixes it.
@moxie0 please consider this for a 3.8.1 release.

(CC @GilCol)

GilCol:
I can also confirm it crashes now...
When I did the changes, I copied that line from onNewIntent() in ConversationFragment (and it worked fine). But I can see that someone fixed that line as well.

PR #4877, opened 2015-12-13
Support for group nicknames

### LINKED ISSUE #4852: Feature Request: Nicknames in groups
When joining a new group it's sometimes hard to find out which people belong to which phone number. Asking everyone in the group who they are is tedious.
It would be nice to be able to set a nickname in the app, that's sent in an extra field in each group message. The receiving clients could show this nickname if the number has no associated contact. Setting a nickname would of course be optional and the nickname is only sent in group messages.

Other messengers support this as well and it's very useful to find out who the person is that just sent a message in a group.

AsamK:
Caution: This PR updates the database!

This PR adds support for nicknames in groups. The user can set a custom nickname in the chat preferences. This nickname will be included in all group messages. The recipients will show the nickname if they don't have the sender in their contact list.
Nicknames from other people are stored in the RecipientPreferenceDatabase.

Example:
group_nickname

Needs WhisperSystems/libtextsecure-java#19

Fixes #4852

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -1019,6 +1019,9 @@
     <string name="preferences__enable_if_your_device_supports_sms_mms_delivery_over_wifi">Enable if your device uses SMS/MMS delivery over WiFi (only enable when \'WiFi Calling\' is enabled on your device)</string>
     <string name="preferences_app_protection__blocked_contacts">Blocked contacts</string>
     <string name="preferences_notifications__display_in_notifications">Display in notifications</string>
+    <string name="preferences_chats__nickname">Nickname</string>
+    <string name="preferences_chats__set_nickname">Set a nickname to be shown to other group members</string>
+    <string name="preferences_chats__current_nickname">Your nickname in groups: %s</string>
     <string name="preferences_chats__when_using_mobile_data">When using mobile data</string>
     <string name="preferences_chats__when_using_wifi">When using Wi-Fi</string>
     <string name="preferences_chats__when_roaming">When roaming</string>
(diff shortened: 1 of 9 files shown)


moxie0:
Thanks but I don't think we want to do this in this way.  We'll probably create profiles that we register with the server rather than bundling the nickname with every message.  That way nicknames are visible in the group membership list as well as individual conversations.

AsamK:
@moxie0 I agree this is not an optimal solution. However I'm not in favour of having profiles on the server. That would mean that anyone who has my phone number can query the server for my nickname. With my approach only people who are in the same group as me can see my nickname.

With my approach also, the nickname is visible in the membership list and individual conversations after receiving the first message from this person. I also regard not having the nickname before receiving a message, a minor issue, because I usually care about knowing who wrote something and not who is idling in a group.

If the increased data usage by bundling the nickname is a big issue, this could be optimized. e.g by sending the nickname only once a day or only after the group has been updated.

moxie0:
I understand your concerns, but that's the way we're going to do it.

strugee:
@moxie0 something to consider here - it might make sense to limit this per-group. I'm in a Facebook Messenger chat right now where everyone has ridiculous nicknames, all of which are inside jokes. E.g. someone's named "Bullet Train Conductor," which is a joke that only makes sense in that particular group chat. I wouldn't want that contact showing up that way in my conversations list as "Bullet Train Conductor" - I'd want their real name.

Or maybe we should implement both? I.e. have server profiles be "names" and also have per-group "nicknames"? I'm happy to file a new issue on the latter if that's something you'd support.

AbigailBuccaneer:
> We'll probably create profiles that we register with the server rather than bundling the nickname with every message. That way nicknames are visible in the group membership list as well as individual conversations.

Was this ever followed through with? Signal is still unusable for very large groups because of a lack of names. It'd be good to know if this is still planned.

jeanlucthumm:
Any updates on this?

PR #4879, opened 2015-12-14
Prevent auto-backup on Marshmallow by targeting an older version

jan-kiszka:
https://developer.android.com/training/backup/autosyncapi.html states
that only apps targeting API level 23 or higher will participate in the
auto-backup. By targeting level 22 for now, we can re-enable ADB-based
backup/restore without risking to leak data to the backup cloud.

For the long run, a proper encrypted backup/restore remains the better
solution. Once that is available, we can consider disabling backups
again.

Signed-off-by: Jan Kiszka <jan.kiszka@web.de>

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -6,6 +6,7 @@
       android:versionName="3.7.2">
 
     <uses-sdk tools:overrideLibrary="com.amulyakhare.textdrawable,com.astuetz.pagerslidingtabstrip,pl.tajchert.waitingdots,com.h6ah4i.android.multiselectlistpreferencecompat"/>
+    <uses-sdk andoid:targetSdkVersion="22" />
 
     <permission android:name="org.thoughtcrime.securesms.ACCESS_SECRETS"
                 android:label="Access to TextSecure Secrets"
@@ -92,8 +93,6 @@
     <application android:name=".ApplicationContext"
                  android:icon="@drawable/icon"
                  android:label="@string/app_name"
-                 tools:replace="android:allowBackup"
-                 android:allowBackup="false"
                  android:theme="@style/TextSecure.LightTheme">
 
     <meta-data android:name="com.google.android.gms.version"


jan-kiszka:
Even if this patch is likely too trivial to argue over: grant of copyright license is a clear red line. You are a commercial entity, not the FSF or such. Consider using http://developercertificate.org instead.

heyimgay:
The CLA is only if you want your changes accepted into the main branch - you are free to fork away and do whatever you please.

And open whisper systems is not charging anyone. Yes, they are a legally defined entity but that is the reality of a capitalist society.

jan-kiszka:
@heyimgay No need to explain the obvious. When contributing, I clearly prefer upstream over forking for good reasons. I'm running various OSS projects for fun and for profit (for a large company). Moreover, I'm dealing with CLAs regularly, specifically at work, with their motivations, side-effects and limitations. I'm fine with formally confirming the OSS license of my contribution, and I'm fine with discussing reasonable relicensing requests as needed. But I'm not signing blank checks (nor does my employer when they face such a request).

AndroidManifest.xml
```diff
@@ -6,6 +6,7 @@
       android:versionName="3.7.2">
 
     <uses-sdk tools:overrideLibrary="com.amulyakhare.textdrawable,com.astuetz.pagerslidingtabstrip,pl.tajchert.waitingdots,com.h6ah4i.android.multiselectlistpreferencecompat"/>
+    <uses-sdk andoid:targetSdkVersion="22" />
```
agrajaghh: this is not enough?
https://github.com/WhisperSystems/Signal-Android/blob/master/build.gradle#L158

2-4601:
@jan-kiszka Have you tested this on an Marshmallow device? The Android documentation you referenced in the first post is not very clear.

jan-kiszka:
Not yet, I need to update my SDK first. Rebuilding Signal is scheduled for spare time next week.

jtracey:
@jan-kiszka One reason OWS makes you sign a CLA is so they can integrate any and all of the codebase into other projects as needed. For example, if they didn't have a CLA, doing the axolotl integration with Whatsapp described here:
https://whispersystems.org/blog/whatsapp/
would be significantly more difficult, as all non-CLA commits would have to be removed before integration. The code is GPL'd here, so you can use it as you see fit, but if you don't want OWS to be able to use that code, then that's a very strong disincentive for them to integrate it, since at that point it would be necessary to keep two projects going where there could be one.

AndroidManifest.xml
jan-kiszka: Good point - if that already declares the app to target version 22, reverting 3bec127020 should be enough. Sorry, I'm not very deep into such Android details anymore.

jan-kiszka:
@jtracey I may have overread the point, but the CLA doesn't seem to explain the reason for the requested copyright assignment. If it is integration into projects that are GPL-incompatible, they can consider choosing a different license (LGPL, BSD etc.) for the to-be-shared core, along with providing the reasoning. That happens, that can be a valid move and is not an issue then. But leaving it open makes one only wonder about the intentions and plans behind the assignment.

jtracey:
@jan-kiszka Again, you haven't given any reason why they wouldn't ask for a CLA. I'm not with OWS, so I can't speak for them, but at least to me it clearly makes more sense to use a copyleft license and then integrate the code into other projects when it is worthwhile, rather than let whoever come by and make non-free forks of the projects. In any case, this is probably not very well suited to a PR discussion and if you have further questions, they should probably be posted to the mailing list. Point being, your code will not be merged if you do not sign the CLA.

jan-kiszka:
If I contribute to project with a clear license, I can foresee what can happen with my code /wrt usage in this or other OSS projects, in potential commercializations. If I sign a copyright assignment like this CLA requests, I can't. I'm not against commercialization per se, but I'm against hidden or at least unclear agendas.

Yes, I'm also not a fan of CLAs because most of them needlessly complicate the contribution process (proven here). Many successful OSS projects without such a process demonstrate that such thing is not inherently needed.

And the reason that this discussion ended up here is the unfortunate fact that the need for the CLA was only raised by a CI check, not by  explanation in CONTRIBUTING.md e.g.

moxie0:
lol omg wtf is wrong with people, you're trying to get us to change our license policy for an incorrect one-line change to a config file that you didn't test.  Please, if our CLA is preventing you from submitting PRs, then I consider that a success.

Issue #4648, opened 2015-11-21
Try again in device management doesn't try again


bartz1:
1. phone is not connected to the internet (no wifi no data)
2. open device management
3. error displays "no network", which is correct. there's also a button "try again"
4. turn on data
5. press "try again" a thousand times, but it doesn't do anything - the error popup is still there

can't send a debug log but it's pretty consistent so you should be able to reproduce

PR #4882, opened 2015-12-14
Device management - try again fix

kaladivo:
Fix for #4648


src/org/thoughtcrime/securesms/DeviceListFragment.java
```diff
@@ -142,6 +142,7 @@ private void handleLoaderFailed() {
                               new DialogInterface.OnClickListener() {
       @Override
       public void onClick(DialogInterface dialog, int which) {
+        getLoaderManager().getLoader(0).forceLoad();
         getLoaderManager().initLoader(0, null, DeviceListFragment.this);
```
moxie0: is this line still necessary?

kaladivo: It is, because forceLoad() method only loads data set. It does't call LoaderCallbacks.

PR #4884, opened 2015-12-14
Device management AlertDialog set not cancelable

kaladivo:
When the user enters to the device management in the settings and the network connection fails the AlertDialog with the message and option to try again is displayed to him. Currently user can cancel the dialog in which case he will see only the blank screen. That might be quite confusing to him.

I have fixed it by adding a cancel button, OnCancelListener and set the dialog to not be cancelable by clicking outside of it.

### DIFF
--- a/src/org/thoughtcrime/securesms/DeviceListFragment.java
+++ b/src/org/thoughtcrime/securesms/DeviceListFragment.java
@@ -145,7 +145,25 @@ public class DeviceListFragment extends ListFragment
         getLoaderManager().initLoader(0, null, DeviceListFragment.this);
       }
     });
-    builder.show();
+
+    builder.setNegativeButton(android.R.string.cancel,
+            new DialogInterface.OnClickListener() {
+      @Override
+      public void onClick(DialogInterface dialog, int which) {
+        DeviceListFragment.this.getActivity().onBackPressed();
+      }
+    });
+    builder.setOnCancelListener(new DialogInterface.OnCancelListener() {
+      @Override
+      public void onCancel(DialogInterface dialog) {
+        DeviceListFragment.this.getActivity().onBackPressed();
+      }
+    });
+
+    AlertDialog dialog = builder.create();
+    // This method is not available in the Builder class.
+    dialog.setCanceledOnTouchOutside(false);
+    dialog.show();
   }
 
   private void handleDisconnectDevice(final long deviceId) {
--- a/src/org/thoughtcrime/securesms/DeviceListFragment.java
+++ b/src/org/thoughtcrime/securesms/DeviceListFragment.java
@@ -159,9 +159,7 @@ public class DeviceListFragment extends ListFragment
       }
     });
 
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/DeviceListFragment.java
```diff
@@ -145,7 +145,25 @@ public void onClick(DialogInterface dialog, int which) {
         getLoaderManager().initLoader(0, null, DeviceListFragment.this);
       }
     });
-    builder.show();
+
+    builder.setNegativeButton(android.R.string.cancel,
+            new DialogInterface.OnClickListener() {
```
moxie0: one line

src/org/thoughtcrime/securesms/DeviceListFragment.java
```diff
@@ -145,7 +145,25 @@ public void onClick(DialogInterface dialog, int which) {
...
+    });
+    builder.setOnCancelListener(new DialogInterface.OnCancelListener() {
+      @Override
+      public void onCancel(DialogInterface dialog) {
+        DeviceListFragment.this.getActivity().onBackPressed();
+      }
+    });
+
+    AlertDialog dialog = builder.create();
+    // This method is not available in the Builder class.
```
moxie0: no comments please

src/org/thoughtcrime/securesms/DeviceListFragment.java
```diff
@@ -145,7 +145,25 @@ public void onClick(DialogInterface dialog, int which) {
...
+    dialog.setCanceledOnTouchOutside(false);
```
moxie0: this should remain true

src/org/thoughtcrime/securesms/DeviceListFragment.java
```diff
@@ -145,7 +145,23 @@ public void onClick(DialogInterface dialog, int which) {
...
+    });
+    builder.setOnCancelListener(new DialogInterface.OnCancelListener() {
+      @Override
+      public void onCancel(DialogInterface dialog) {
+        DeviceListFragment.this.getActivity().onBackPressed();
+      }
+    });
+
+    AlertDialog dialog = builder.create();
+    dialog.setCanceledOnTouchOutside(true);
```
moxie0: why are you calling this?

kaladivo: To prevent user from stucking on a blank screen. When the user selects cancel, he sould be directed backwards instead of leaving him at a blank screen. I described the issue in the pull request description.

moxie0: The method setCanceledOnTouchOutside does all that? In the last diff you explicitly set it to false, now you're explicitly setting it to true. Why do you have to explicitly set it?

kaladivo:
You're right. My bad. Sorry.

moxie0:
8c36e206815d2799feaa1ec8233047737bb50f7e

Issue #4733, opened 2015-11-27
Low sound in call ( Signal 3.5.2+)


jocelynthode:
Hey,

I still have the problem of the closed issue #4244. The workaround of enabling/disabling speakers still does the trick.

Device info: Samsung I9300 , CM Lolipop 5.1.1, Signal 3.5.2

Debug log : https://gist.github.com/anonymous/b09ff201c845cddd5da1

In the log, I phoned a friend, confirmed that I could barely hear him, then did the workaround. Then I confirmed I could hear him well and stopped the call.

mcginty:
tentatively fixed by https://github.com/WhisperSystems/Signal-Android/pull/4737

2-4601:
Do you still experience this in 3.6.0?

peterpanondrugs:
i do. https://gist.github.com/7baf7dd095bb9076650a
Same is for just the "ringing"
If it starts ringing (#1) then use the speaker on -> off trick, afterwards ringing volume (#2) is louder then #1.

jocelynthode:
I can't tell you yet if the problem is fixed in 3.6.0 as this version isn't yet available for me on the play store.

AlfonsoMuskedunder:
@2-4601 I do, too.
In case a log provides useful information: https://gist.github.com/616e3dab2632d1b5c8da
(I just made a call where this happened - including the workaround - on a Moto G, Android 5.1, Signal 3.6.0; other end on a Samsung SIII Mini, Android 4.1.2, Signal 3.6.0.)

7adietri:
Also encountering this on a Fairphone 1 (Android 4.2.2, Signal 3.6.1). The speaker workaround fixes this for me too.

jocelynthode:
Problem still present for me on Signal 3.6.1

@moxie0  any idea why ?

2-4601:
Everyone experiencing this please test pr #4885.

jocelynthode:
@2-4601 Sorry I don't have an environment to test with a dev build.

2-4601:
0caad7f348ae51fb4fec36ebc934065302c7303e should fix this

cascheberg:
Please check if the issue is fixed in 3.9.0 if you can. It is available on the Google Play Store as a beta release.

AlfonsoMuskedunder:
[...]

PR #4885, opened 2015-12-14
Fix call volume

### LINKED ISSUE #4188: Speaker setting switches off
If I make a Signal call and set the device on speaker immediately after I press call, the speaker setting switches off as soon as the call starts ringing. When I set it back on speaker, the speaker setting switches off again as soon as the other side picks up. When I set it on speaker again it keeps on during the entire call.

cascheberg:
I am also affected by the low sound volume in calls, see #4733.
In my case that is because of the `setBluetoothScoOn(true)` for all calls. It will switch to bluetooth (although not connected) and thus show the bluetooth icon for the current volume setting.
The workaround that was mentioned, which is enabling and disabling the speaker, works because it will call `setBluetoothScoOn(false)` in `enableDefaultRouting`.
Furthermore `stopBluetoothSco()` was never called.
~~I also removed the `initializeAudio()`, because the ringers will now request the audio focus themselves.
And for every audio mode setting I added a focus request, and vice versa.~~
Also fixes #4188.

### DIFF
--- a/src/org/thoughtcrime/redphone/RedPhoneService.java
+++ b/src/org/thoughtcrime/redphone/RedPhoneService.java
@@ -293,12 +293,14 @@ public class RedPhoneService extends Service implements CallStateListener, CallS
     AudioManager audioManager = ServiceUtil.getAudioManager(this);
     AudioUtils.resetConfiguration(this);
 
+    Log.d(TAG, "request STREAM_VOICE_CALL audio focus");
     audioManager.requestAudioFocus(null, AudioManager.STREAM_VOICE_CALL,
                                    AudioManager.AUDIOFOCUS_GAIN_TRANSIENT);
   }
 
   private void shutdownAudio() {
-    AudioManager am = (AudioManager) getSystemService(AUDIO_SERVICE);
+    Log.d(TAG, "reset audio mode and abandon focus");
+    AudioManager am = ServiceUtil.getAudioManager(this);
     am.setMode(AudioManager.MODE_NORMAL);
     am.abandonAudioFocus(null);
   }
--- a/src/org/thoughtcrime/redphone/RedPhoneService.java
+++ b/src/org/thoughtcrime/redphone/RedPhoneService.java
@@ -294,15 +294,16 @@ public class RedPhoneService extends Service implements CallStateListener, CallS
     AudioUtils.resetConfiguration(this);
 
     Log.d(TAG, "request STREAM_VOICE_CALL audio focus");
-    audioManager.requestAudioFocus(null, AudioManager.STREAM_VOICE_CALL,
-                                   AudioManager.AUDIOFOCUS_GAIN_TRANSIENT);
+    audioManager.requestAudioFocus(null, AudioManager.STREAM_VOICE_CALL, AudioManager.AUDIOFOCUS_GAIN);
   }
 
   private void shutdownAudio() {
     Log.d(TAG, "reset audio mode and abandon focus");
+    AudioUtils.resetConfiguration(this);
     AudioManager am = ServiceUtil.getAudioManager(this);
     am.setMode(AudioManager.MODE_NORMAL);
     am.abandonAudioFocus(null);
+    am.stopBluetoothSco();
   }
 
   public int getState() {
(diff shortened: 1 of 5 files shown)


2-4601:
This seems to be a very sensitive area and feels like just thinking about those lines will cause some weird bugs.

However I tested this with my Huawei U8800 device (4.0.4) and nothing horrible happened. I'm glad to see the bluetooth volume control going away.
One small drawback was that if the volume level was <50% I could hear practically nothing. But at >50% levels it was loud enough. Without this commit I can hear something even with the lowest non-zero volume level. 

(FTR, I never had the problems described in #4733.)

moxie0:
Hey @cascheberg thanks for looking into this, but I'd like this PR to be much smaller if possible.  

If this is a device-specific bug caused by setBluetoothScoOn, then ideally this PR would be limited to that single line.  However, doesn't changing that mean Signal calls won't automatically connect to people's bluetooth headsets?

panjamo:
On my device, OnePlus, I have the same Issue. Pressing volume buttons while doing a phone call with signal, the blue tooth volume panel is shown (blue tooth is deactivated!), and volume dosen't change. This behaviour persits until reboot even for normal phone calls.

cascheberg:
@moxie0 I updated the PR to only add some logging and not automatically connect to bluetooth (two separate commits).
I think `isBluetoothScoAvailableOffCall()` is checking a device capability, but not a connection status.
I do not really know yet how all the bluetooth stuff works, but I'd say you need to at least wait for a `ACTION_SCO_AUDIO_STATE_UPDATED ` intent (961b9b882f1a36b2be38d310e6f7faec2133a7c6), or maybe even check all bonded bluetooth devices, before `setBluetoothScoOn(true)`. Right now this is always routing audio to a wrong channel on many devices. I am using a Nexus 4 running stock Android 5, so I do not consider this to be "device-specific".

moxie0:
If the relevant line is in outgoingringer, does that mean people only experience the "low volume" bug when making outgoing calls?

cascheberg:
No, because it is also used for `playHandshake()` and such, and this is in `onPrepared`.

moxie0:
I see, what a mess.

Ok so you're proposal is that we make this change and forgo automatic bluetooth connectivity, with the hope of eventually working that in somehow?  Have you tested with a bluetooth headset to make sure everything still functions otherwise?

cascheberg:
Yes. (Implementing automatic connectivity the right way might just be a few lines of code.)
I do not have a bluetooth headset at hand, but the stuff that handles the bluetooth call screen control and the `startBluetoothSco()` that triggers it is unchanged.

src/org/thoughtcrime/redphone/RedPhoneService.java
```diff
@@ -293,14 +293,17 @@ private void initializeAudio() {
...
-    audioManager.requestAudioFocus(null, AudioManager.STREAM_VOICE_CALL,
-                                   AudioManager.AUDIOFOCUS_GAIN_TRANSIENT);
+    Log.d(TAG, "request STREAM_VOICE_CALL audio focus");
+    audioManager.requestAudioFocus(null, AudioManager.STREAM_VOICE_CALL, AudioManager.AUDIOFOCUS_GAIN);
   }
 
   private void shutdownAudio() {
-    AudioManager am = (AudioManager) getSystemService(AUDIO_SERVICE);
+    Log.d(TAG, "reset audio mode and abandon focus");
+    AudioUtils.resetConfiguration(this);
```
moxie0: what is this for?

cascheberg: This is to call `setBluetoothScoOn(false)` in case it was enabled by the user. Apparently `stopBluetoothSco()` does not do that.

moxie0: can we just call `setBluetoothScoOn(false)` directly then?

cascheberg: Yes we could. But the only other thing that this does is `setSpeakerphoneOn(false)`, which to me seems to be the right thing to do as well. Any thoughts? Otherwise I'll follow your suggestion and check more closely later on.

cascheberg:
This PR also fixes #4188.

moxie0:
bluetooth headset works as expected too

PR #4886, opened 2015-12-14
Fix call crashes and termination sounds

### LINKED ISSUE #4184: crash when placing a call in airplaine mode
not reproducible every time

```
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime: FATAL EXCEPTION: main
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime: Process: org.thoughtcrime.securesms, PID: 12552
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime: java.lang.IllegalStateException
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime:     at android.media.MediaPlayer._start(Native Method)
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime:     at android.media.MediaPlayer.start(MediaPlayer.java:1219)
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime:     at org.thoughtcrime.redphone.audio.OutgoingRinger.onPrepared(OutgoingRinger.java:146)
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime:     at android.media.MediaPlayer$EventHandler.handleMessage(MediaPlayer.java:2593)
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime:     at android.os.Handler.dispatchMessage(Handler.java:102)
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime:     at android.os.Looper.loop(Looper.java:211)
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime:     at android.app.ActivityThread.main(ActivityThread.java:5373)
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime:     at java.lang.reflect.Method.invoke(Native Method)
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime:     at java.lang.reflect.Method.invoke(Method.java:372)
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime:     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1020)
10-02 11:50:56.927 12552-12552/? E/AndroidRuntime:     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:815)
10-02 11:50:56.934 731-1361/? D/audio_hw_primary: out_set_parameters: enter: usecase(1: low-latency-playback) kvpairs: routing=2
10-02 11:50:56.947 12552-14888/? E/MediaPlayer-JNI: QCMediaPlayer mediaplayer N
```

### LINKED ISSUE #4319: Alert user to call ending
Sometimes, when I'm using Signal to speak with someone, the call will drop unexpectedly. While I expect this to happen sometimes, right now Signal doesn't provide any indication that it has happened (other than the fact it's no longer running). It would be nice if, before shutting down, it played some tone to make it more obvious.

### LINKED ISSUE #5222: Signal crashes sometimes while calling and the person doesn't answer
Hi. Sometimes when I call someone and the person doesn't pick up, the app crashes. No idea why, but here are some logs.

https://gist.github.com/anonymous/c2b1eecfa72eb6a128e7
https://gist.github.com/anonymous/b83542637606292dbdc1
https://gist.github.com/anonymous/c0950a3a70d8f114dc9d

### LINKED ISSUE #5353: OutgoingRinger NPE on network failure
I have:

### Bug description
Got this crash today during the Redphone relay outage:
```
D/RedPhoneService(18728): java.lang.NullPointerException
D/RedPhoneService(18728): 	at org.thoughtcrime.redphone.audio.OutgoingRinger.onPrepared(OutgoingRinger.java:147)
D/RedPhoneService(18728): 	at android.media.MediaPlayer$EventHandler.handleMessage(MediaPlayer.java:1992)
D/RedPhoneService(18728): 	at android.os.Handler.dispatchMessage(Handler.java:99)
D/RedPhoneService(18728): 	at android.os.Looper.loop(Looper.java:153)
D/RedPhoneService(18728): 	at android.app.ActivityThread.main(ActivityThread.java:5299)
```
Happens [here](https://github.com/WhisperSystems/Signal-Android/blob/fa22fb755084b3c2d8064811d6be79ee0b617eac/src/org/thoughtcrime/redphone/audio/OutgoingRinger.java#L147), not sure how that's possible? Might be something for redphone wizard @cascheberg

### Steps to reproduce
- try to place a Signal call during the relay outage (setting phone to airplane mode somehow isn't enough, so I can't reproduce :/)
- get a "network failed" message
- crash

### Device info
 **Android version:** 4.2.2

 **Signal version:** 3.14.2

edit: link to the full log: https://gist.github.com/anonymous/db8a786f76f5400c58aa

cascheberg:
Fixes #4319, fixes #4184, fixes #5222, fixes #5353

### DIFF
--- a/src/org/thoughtcrime/redphone/RedPhoneService.java
+++ b/src/org/thoughtcrime/redphone/RedPhoneService.java
@@ -366,7 +366,9 @@ public class RedPhoneService extends Service implements CallStateListener, CallS
     NotificationBarManager.setCallEnded(this);
 
     incomingRinger.stop();
-    outgoingRinger.stop();
+    if (outgoingRinger.isLooping()) {
+      outgoingRinger.stop();
+    }
 
     if (currentCallManager != null) {
       currentCallManager.terminate();
@@ -498,16 +500,19 @@ public class RedPhoneService extends Service implements CallStateListener, CallS
   }
 
   public void notifyNoSuchUser() {
+    outgoingRinger.playFailure();
     sendMessage(Type.NO_SUCH_USER, getRecipient(), null);
     this.terminate();
   }
 
   public void notifyServerMessage(String message) {
+    outgoingRinger.playFailure();
     sendMessage(Type.SERVER_MESSAGE, getRecipient(), message);
     this.terminate();
   }
 
   public void notifyClientError(String msg) {
+    outgoingRinger.playFailure();
     sendMessage(Type.CLIENT_FAILURE, getRecipient(), msg);
     this.terminate();
   }
--- a/src/org/thoughtcrime/redphone/RedPhoneService.java
+++ b/src/org/thoughtcrime/redphone/RedPhoneService.java
@@ -400,7 +400,7 @@ public class RedPhoneService extends Service implements CallStateListener, CallS
   }
 
   public void notifyBusy() {
-    Log.w("RedPhoneService", "Got busy signal from responder!");
+    Log.w(TAG, "Got busy signal from responder!");
     sendMessage(Type.CALL_BUSY, getRecipient(), null);
 
     outgoingRinger.playBusy();
@@ -432,11 +432,20 @@ public class RedPhoneService extends Service implements CallStateListener, CallS
   }
 
   public void notifyCallDisconnected() {
-    if (state == STATE_RINGING)
-      handleMissedCall(remoteNumber, false);
-
     sendMessage(Type.CALL_DISCONNECTED, getRecipient(), null);
-    this.terminate();
+
+    if (state == STATE_RINGING) {
+      handleMissedCall(remoteNumber, false);
+      this.terminate();
+    } else {
+      outgoingRinger.playBusy();
+      serviceHandler.postDelayed(new Runnable() {
+        @Override
+        public void run() {
+          RedPhoneService.this.terminate();
+        }
+      }, RedPhone.BUSY_SIGNAL_DELAY_FINISH);
+    }
   }
 
   public void notifyHandshakeFailed() {
--- a/src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
+++ b/src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
@@ -125,6 +125,10 @@ public class OutgoingRinger implements MediaPlayer.OnCompletionListener, MediaPl
       }
     }
 
-    mp.start();
+    try {
+      mp.start();
+    } catch (IllegalStateException e) {
+      Log.w(TAG, e);
+    }
   }
 }
--- a/src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
+++ b/src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
@@ -74,14 +74,17 @@ public class OutgoingRinger implements MediaPlayer.OnCompletionListener, MediaPl
   }
 
   private void start( int soundID ) {
-    if( soundID == currentSoundID ) return;
-    currentSoundID = soundID;
-    loopEnabled = true;
-    start();
+    start(soundID, true);
   }
 
-  private void start() {
-    if( mediaPlayer != null ) mediaPlayer.release();
+  private void start(int soundID, boolean loop) {
+    if( soundID == currentSoundID ) return;
+
+    if (mediaPlayer != null) mediaPlayer.release();
+
+    currentSoundID = soundID;
+    loopEnabled = loop;
+
     mediaPlayer = new MediaPlayer();
     mediaPlayer.setAudioStreamType(AudioManager.STREAM_VOICE_CALL);
     mediaPlayer.setOnCompletionListener(this);
@@ -116,9 +119,7 @@ public class OutgoingRinger implements MediaPlayer.OnCompletionListener, MediaPl
   }
 
   private void stop( int soundID ) {
-    currentSoundID = soundID;
-    loopEnabled = false;
-    start();
+    start(soundID, false);
   }
 
   public void onCompletion(MediaPlayer mp) {
--- a/src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
+++ b/src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
@@ -39,14 +39,10 @@ public class OutgoingRinger implements MediaPlayer.OnCompletionListener, MediaPl
 
   private MediaPlayer mediaPlayer;
   private int         currentSoundID;
-  private boolean     loopEnabled;
   private Context     context;
 
   public OutgoingRinger(Context context) {
     this.context = context;
-
-    loopEnabled = true;
-
   }
 
   public void playSonar() {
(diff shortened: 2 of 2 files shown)


src/org/thoughtcrime/redphone/RedPhoneService.java
```diff
@@ -354,6 +354,26 @@ private void startCallCardActivity() {
...
+
+  private void terminateFailure() {
+    outgoingRinger.playFailure();
+    serviceHandler.postDelayed(new Runnable() {
+      @Override
+      public void run() {
+        RedPhoneService.this.terminate();
+      }
+    }, RedPhone.STANDARD_DELAY_FINISH);
+  }
```
moxie0: Doesn't look like most of the callers need this to actually post delayed. I'd prefer not to pull these out into separate methods.

cascheberg: I had tested the existing code but I could never hear the failure sound.
Probably because `outgoingRinger.playFailure()` is asynchronous and `terminate` will call `outgoingRinger.stop()`.

moxie0: let's have outgoingringer return a future then

cascheberg:
@moxie0 I updated the PR to use `Future`. Ready for review comments, but I'd like to do some more testing ;)
The `terminateBusy` is still using `postDelayed` because the sound is looping forever.
I removed several `private` methods in `OutgoingRinger` because they are not quite helpful.
And I fixed bugs in `SettableFuture`.

src/org/thoughtcrime/redphone/RedPhoneService.java
```diff
@@ -357,6 +358,25 @@ private void startCallCardActivity() {
...
+
+  private void terminateFailure() {
+    try {
+      outgoingRinger.playFailure().get();
+    } catch (InterruptedException | ExecutionException e) {
+      Log.w(TAG, e);
+    }
+    this.terminate();
+  }
```
moxie0: I'd still prefer not to have dedicated methods for this

src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
```diff
@@ -37,103 +39,90 @@
...
   private static final String TAG = OutgoingRinger.class.getSimpleName();
 
-  private MediaPlayer mediaPlayer;
-  private int         currentSoundID;
-  private boolean     loopEnabled;
-  private Context     context;
+  private MediaPlayer    mediaPlayer;
+  private int            currentSoundID;
+  private Context        context;
+  private SettableFuture future;
```
moxie0: I don't think this should be an instance variable, there's no need to maintain its state at the instance level and I don't see anything in here that would prevent you from calling play twice on the same instance.

src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
```diff
@@ -37,103 +39,90 @@
...
   }
 
-  public void playBusy() {
-    start(R.raw.redphone_busy);
+  public Future playBusy() {
+    return play(R.raw.redphone_busy, true);
   }
 
-  private void setSound( int soundID ) {
+  private Future play(int soundID, boolean loop) {
```
moxie0: always return ListenableFuture

src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
```diff
@@ -37,103 +39,90 @@
...
-      return;
-    }
-    try {
       mediaPlayer.prepareAsync();
-    } catch (IllegalStateException e) {
-      // TODO Auto-generated catch block
+    } catch (IllegalArgumentException | SecurityException | IllegalStateException | IOException e) {
       Log.w(TAG, e);
-      return;
+      future.setException(e);
```
moxie0: we don't propagate this exception in the synchronous case, so there's no reason to propagate it in the async case

src/org/thoughtcrime/securesms/util/concurrent/SettableFuture.java
```diff
@@ -97,6 +99,7 @@ private void notifyAllListeners() {
     List<Listener<T>> localListeners;
 
     synchronized (this) {
+      notifyAll();
       localListeners = new LinkedList<>(listeners);
     }
```
moxie0: It's very unlikely that I will merge changes to SettableFuture

moxie0:
I think this is on the right track, but all in all I'd like the diff to be much simpler

src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
cascheberg: - I need to return the `Future` in `playFailure`, but I can only set the result in `onCompletion`. How else could I do that?
- One instance of `OutgoingRinger` is used to play multiple sounds, but `shutdown` handles the life cycle of the `MediaPlayer` and `SettableFuture`. Is there a problem with that?

src/org/thoughtcrime/securesms/util/concurrent/SettableFuture.java
cascheberg: - The `SettableFuture.get()` is not working because it will just `wait` forever since it requires a `notify` to wake up. Fixing that is really relevant. This change adds the `notify` on all status changes, but it's ok if you have other ideas.
- The other fixes are additionally required for `cancel` to work.

src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
cascheberg: OK, so then I should add a `ListenableFuture.Listener` instead of calling `.get()`? That would also work around the `SettableFuture` bug.

moxie0: If you're just going to call `get` then there's not really any reason to use a future, since it can just be a blocking call. I'd use the listener since it's async now.

cascheberg:
@moxie0 PR updated. I divided the patch into 10 simple commits. The first ones simplify the `OutgoingRinger` to to have only one `start` and `stop`  method instead of two each. Please have a look at them separately, I made them really small, and the existing code is really not great.
The code now uses the `ListenableFuture.Listener`.
The fix for `SettableFuture` is not required anymore, so it is not included.
All tested.

cascheberg:
I found a much simpler solution: only explicitly stop `OutgoingRinger` when it is looping. `playFailure()` is not looping and can take care of the shutdown on its own.

cascheberg:
Rebased and tested again. This PR is now a bit simpler because of other changes that have been merged in the meantime.

cascheberg:
rebased again

moxie0:
it seems like this PR is trying to do a few different things, and the diff is too large for me right now. I'd prefer to see something that just fixes the crashes.

cascheberg:
I opened two new separate PRs which do not include the refactoring.

Issue #4627, opened 2015-11-20
Deleting audio while playing does not stop the playing


mawed:
Play an audio file.
Mark this file while playing.
Delete it
Still plays until leaving the conversation.

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #4900, opened 2015-12-16
Deleting audio message while playing will stop it

kaladivo:
Fix for #4627.


src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -88,6 +89,7 @@
   private Locale       locale;
   private RecyclerView list;
   private View         loadMoreView;
+  private List<ConversationItem> selectedItems = new ArrayList<>();
```
moxie0: feels very wrong to be managing this state here

moxie0:
closing due to inactivity, please reopen if you revisit thanks!

PR #4907, opened 2015-12-17
Wording: Improve some strings in Privacy preferences

RiseT:
// FREEBIE

Fixes https://github.com/WhisperSystems/Signal-Android/pull/4651

- Change 'Enable lock screen for messages' to **'Lock Signal and message notifications with a passphrase'**
- Change 'Forget passphrase from memory after some interval' to **'Auto-lock Signal after a specified time interval of inactivity'**
- Change 'Timeout passphrase' to **'Inactivity timeout passphrase'**
- Change 'Timeout interval' to **'Inactivity timeout interval'**
- Remove unused string resource *preferences__the_amount_of_time_to_wait_before_forgetting_passphrase*
- Remove unused string resource *preferences__pref_timeout_interval_dialogtitle*

Possibly 'Auto-lock Signal after a specified time**out** interval of inactivity' would be better, not sure. Let me know if this is the case. Perhaps more in line with *preferences__inactivity_timeout_interval* ('Inactivity timeout interval')

I've added that 'inactivity' stuff as I've noticed that I had no clue if the timeout interval is calculated from a) the moment of entering the passphrase or b) the last activity (it's the latter - I've tested it).

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -938,15 +938,14 @@
     <string name="preferences__change_your_passphrase">Change your passphrase</string>
     <string name="preferences__enable_passphrase">Enable passphrase</string>
     <string name="preferences__passphrase_summary">Passphrase %s</string>
-    <string name="preferences__enable_lock_screen_for_messages">Enable lock screen for messages</string>
+    <string name="preferences__lock_signal_and_message_notifications_with_passphrase">Lock Signal and message notifications with passphrase</string>
     <string name="preferences__screen_security">Screen security</string>
     <string name="preferences__screen_security_summary">Screen security %s</string>
     <string name="preferences__disable_screen_security_to_allow_screen_shots">Block screenshots in the recents list and inside the app</string>
-    <string name="preferences__forget_passphrase_from_memory_after_some_interval">Forget passphrase from memory after some interval</string>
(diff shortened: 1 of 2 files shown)


2-4601:
Yeah, the term _'lock screen'_ in _**'Enable lock screen for messages'**_ is very ambiguous. Both Android and iOS use the term _'lock screen'_ for the screen that you see when you press the power button. I've also showed the current string to people (n=5) and asked what they think it refers to and the answer was always the device's lock screen. So I think this is a much needed change. :+1:

More minimal change that would help with the ambiguity would be to just add the word _'Signal'_ in there: _**'Enable**_ _Signal's_ _**lock screen for messages'**_

But I prefer @RiseT's suggestion because it's more explicit.
Or perhaps drop the word _'notificatons'_ and just use:
_**'Lock Signal and messages with passphrase'**_

RiseT:
@2-4601 Yes, I remember back then the German translation team puzzling over that string and what it actually meant. I think it took roughly 10 attempts and some research before the correct translation had been found...

moxie0:
in 3.9.0

2-4601:
@RiseT You forgot this one :smiley:
`ApplicationPreferencesActivity_disable_lock_screen = "Disable lock screen for messages?"`

RiseT:
@2-4601 Dang, will do the next couple of days. Suffering of a two-sided conjunctivitis and it's sort of painful to read at the moment...

PR #4916, opened 2015-12-18
RTL support

### LINKED ISSUE #4714: RTL support appears to be incomplete
Hi,

So I just received an update of Signal with complete translation to Hebrew, and it's WONDERFUL. Thank you for having such a great localization process - not all apps are so smooth.

However, since Hebrew is written right-to-left, as are Arabic, Persian and a few other languages, it requires a bit of extra work to make it properly functional in RTL.

For example, in the Settings menu all the items are aligned to the left, and they should be aligned to the right.

I'm not an Android developer, but as far as I know, since Android 4 or so, most of this can be done by enabling auto-support for RTL somewhere in the app's build definitions (I'm sorry if my language here is imprecise).

agrajaghh:
fixes #4714 

@amire80 helped me testing this, but if there otheres who speak any RTL language, please test this :blush:


amire80:
Thumbs up. Thank you very much!

res/layout/conversation_list_fragment.xml
```diff
@@ -32,7 +32,7 @@
             android:id="@+id/fab"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:layout_gravity="bottom|right"
+            android:layout_gravity="bottom|end"
```
moxie0: isn't end api 14+?

src/org/thoughtcrime/securesms/ConversationListItem.java
```diff
@@ -106,6 +106,9 @@ protected void onFinishInflate() {
     this.thumbnailView           = (ThumbnailView)      findViewById(R.id.thumbnail);
     this.archivedView            = ViewUtil.findById(this, R.id.archived);
     thumbnailView.setClickable(false);
+
+    ViewUtil.setTextViewGravityStart(this.fromView, getContext());
+    ViewUtil.setTextViewGravityStart(this.subjectView, getContext());
```
moxie0: what's this about?

src/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
```diff
@@ -166,31 +166,21 @@ public void onClick(DialogInterface dialog, int which) {
...
     final String onRes               = context.getString(R.string.ApplicationPreferencesActivity_on);
     final String offRes              = context.getString(R.string.ApplicationPreferencesActivity_off);
 
-    if (TextSecurePreferences.isScreenSecurityEnabled(context)) {
-      return context.getString(screenSecurityResId, onRes);
+    if (TextSecurePreferences.isPasswordDisabled(context)) {
+      if (TextSecurePreferences.isScreenSecurityEnabled(context))
+        return context.getString(privacySummaryResId, offRes, onRes);
+      else
+        return context.getString(privacySummaryResId, offRes, offRes);
```
moxie0: braces if multi-line

res/layout/conversation_list_fragment.xml
agrajaghh: from the lint check:

Using `Gravity#LEFT` and `Gravity#RIGHT` can lead to problems when a layout is rendered in locales where text flows from right to left. Use `Gravity#START` and `Gravity#END` instead. Similarly, in XML `gravity` and `layout_gravity` attributes, use `start` rather than `left`.
For XML attributes such as paddingLeft and `layout_marginLeft`, use `paddingStart` and `layout_marginStart`. *NOTE*: If your `minSdkVersion` is less than 17, you should add *both* the older left/right attributes *as well as* the new start/right attributes. On older platforms, where RTL is not supported and the start/right attributes are unknown and therefore ignored, you need the older left/right attributes. There is a separate lint check which catches that type of error.
**(Note: For `Gravity#LEFT` and `Gravity#START`, you can use these constants even when targeting older platforms, because the `start` bitmask is a superset of the `left` bitmask. Therefore, you can use `gravity="start"` rather than `gravity="left|start"`.)"**

src/org/thoughtcrime/securesms/ConversationListItem.java
agrajaghh: have a look at http://stackoverflow.com/questions/34275076/use-locale-ltr-rtl-for-gravity-in-textview

moxie0:
cool can you squash and rebase onto 3.9.0?

agrajaghh:
hmmm.. I just tested it again and noticed maybe a bug. If I text a phone number e.g. "+123456789" to a phone with this branch and RTL language its displayed as "123456789+". Is this normal? Should it be like that?

agrajaghh:
WhatsApp is doing the same btw

amire80:
Yes, it's one of the general curses of RTL input in all operating systems. It's not a bug unique to Signal.

moxie0:
in 3.9.0

PR #4926, opened 2015-12-19
setting to disable notifications for new contacts

### LINKED ISSUE #4572: Notification for new contacts
Signal gives you a notification, when it finds someone from your contacts on the service. It is basically a useful feature. But it needs some user control, I think.

For example I prefer that those notifications are silent and without vibration.

agrajaghh:
also mute these notifications between 23:00 and 09:00

fixes #4572

### DIFF
--- a/src/org/thoughtcrime/securesms/util/DirectoryHelper.java
+++ b/src/org/thoughtcrime/securesms/util/DirectoryHelper.java
@@ -16,7 +16,6 @@ import org.thoughtcrime.securesms.ApplicationContext;
 import org.thoughtcrime.securesms.R;
 import org.thoughtcrime.securesms.crypto.MasterSecret;
 import org.thoughtcrime.securesms.crypto.SessionUtil;
-import org.thoughtcrime.securesms.crypto.storage.TextSecureSessionStore;
 import org.thoughtcrime.securesms.database.DatabaseFactory;
 import org.thoughtcrime.securesms.database.NotInDirectoryException;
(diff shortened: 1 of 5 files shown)


heyimgay:
Can you add an option for a silent notification? So On/Silence/Off

agrajaghh:
I would be okay with it, its up to @moxie0

src/org/thoughtcrime/securesms/util/DirectoryHelper.java
```diff
@@ -217,7 +217,14 @@ private static void notifyNewUsers(@NonNull  Context context,
       if (!SessionUtil.hasSession(context, masterSecret, newUser)) {
         IncomingJoinedMessage message        = new IncomingJoinedMessage(newUser);
         Pair<Long, Long>      smsAndThreadId = DatabaseFactory.getSmsDatabase(context).insertMessageInbox(message);
-        MessageNotifier.updateNotification(context, masterSecret, smsAndThreadId.second);
+        if (TextSecurePreferences.isNewContactsNotificationEnabled(context)) {
```
moxie0: I was thinking we wouldn't do anything at all if this is disabled.

agrajaghh: ah okay, I thought its just about the notification. I'll remove the messages as well..

agrajaghh:
@moxie0 adressed your comment

moxie0:
in 3.9.0

PR #4957, opened 2015-12-24
Enable silent ringtone preference per recipient

### LINKED ISSUE #3957: Allow disable ringtone per conversation
Currently, it is possible to choose a different ringtone per conversation. However, it does not seem to be possible to have _no_ ringtone for a specific conversation, while having some ringtone (the system one, in my case) for all the others: The per-conversation ringtone selection lacks the "None" option.

cascheberg:
Fixes #3957

### DIFF
--- a/res/xml/recipient_preferences.xml
+++ b/res/xml/recipient_preferences.xml
@@ -14,7 +14,7 @@
                         android:key="pref_key_recipient_ringtone"
                         android:title="@string/recipient_preferences__notification_sound"
                         android:ringtoneType="notification"
-                        android:showSilent="false"
+                        android:showSilent="true"
                         android:showDefault="true"
                         android:persistent="false"/>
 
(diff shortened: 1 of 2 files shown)


moxie0:
i think we intentionally left out silent per-recipient ringtones in favor of mute

cascheberg:
Mute is quite different in that it also blocks notifications and does not allow a vibration only mode. (And it's time-limited.) I see no downside in adding this option to the list, but it enables a certain use case.

oittaa:
It's a bit annoying that you have to add ringtone with nothing in it to have "vibration only" conversations. Please consider merging this, there aren't any obvious downsides.

cascheberg:
PR marginally improved; now `// FREEBIE`
Btw, it also fixes that "default" ringtone is not selected properly

cascheberg:
I recently split this PR into two commits for easier review.
(This PR is one of the oldest, while it only changes a few lines.)

moxie0:
in 3.30.0

Issue #4959, opened 2015-12-24
[UX] The text under camera in the attachment menu is not aligned.


mimi89999:
The text under camera in the attachment menu is not aligned. Look at my screenshot.
![screen](https://cloud.githubusercontent.com/assets/8530546/11997462/810e7fba-aa7c-11e5-9025-dd60a6f6bd33.png)
Device info: Samsung Galaxy S4 mini, CM13

McLoo:
Make a multi line issue. 
(French needs even more space than German 😊)

PR #4961, opened 2015-12-25
Fix for #4959

mimi89999:
// FREEBIE
I tested it.


McLoo:
You should always only pull request changes, that you have successfully tested (on different android versions of possible)

mimi89999:
I tested it on android 6.

mimi89999:
@McLoo Could you make this PR?

PR #4963, opened 2015-12-25
remove newlines in text-only group messages

agrajaghh:
with this change (https://github.com/WhisperSystems/Signal-Android/commit/bcf95e50aa38cae2dfe640ea1a1e26cbaa3a84b7#diff-c36f0750d909b343079accc77ce529e2R32) text-only group messages always started with two newlines.

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/OutgoingMediaMessage.java
+++ b/src/org/thoughtcrime/securesms/mms/OutgoingMediaMessage.java
@@ -29,7 +29,7 @@ public class OutgoingMediaMessage {
   public OutgoingMediaMessage(Recipients recipients, SlideDeck slideDeck, String message, long sentTimeMillis, int distributionType)
   {
     this(recipients,
-         TextUtils.isEmpty(message) ? slideDeck.getBody() : slideDeck.getBody() + "\n\n" + message,
+         buildMessage(slideDeck, message),
          slideDeck.asAttachments(),
          sentTimeMillis,
          distributionType);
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/mms/OutgoingMediaMessage.java
```diff
@@ -71,4 +71,15 @@ public long getSentTimeMillis() {
...
+  private String buildMessage(SlideDeck slideDeck, String message) {
+    if (TextUtils.isEmpty(message)) {
+      return slideDeck.getBody();
+    } else {
+      if (TextUtils.isEmpty(slideDeck.getBody())){
+        return message;
+      } else {
+        return slideDeck.getBody() + "\n\n" + message;
+      }
+    }
```
moxie0: I like to avoid nested if conditions if possible.  Can this be expressed as if, else if, else?

src/org/thoughtcrime/securesms/mms/OutgoingMediaMessage.java
```diff
@@ -71,4 +71,15 @@ public long getSentTimeMillis() {
     return sentTimeMillis;
   }
 
+  private String buildMessage(SlideDeck slideDeck, String message) {
```
cascheberg: You could make this a `static` method and call it directly from within the `this(...)` constructor call

agrajaghh:
updated

@moxie0 true, easier to read like that
@cascheberg yep, thats better

PR #4971, opened 2015-12-27
Fix call card photo resolution and placement

### LINKED ISSUE #4245: Low quality contact photo during call
Calling screen doesn't show the full resolution contact photo on Signal 3.0.3

haffenloher:
Screenshots:
Contact with photo [before](https://cloud.githubusercontent.com/assets/264472/12008532/ed94e162-ac41-11e5-8dd9-d8ea4f23ad63.png) / [after](https://cloud.githubusercontent.com/assets/264472/12008534/f49d2e1a-ac41-11e5-8260-f89d916df6a5.png), without photo [before](https://cloud.githubusercontent.com/assets/264472/12008533/ef2f23ca-ac41-11e5-94b8-f72517a7d12e.png) / [after](https://cloud.githubusercontent.com/assets/264472/12008536/f6665848-ac41-11e5-881b-b04fe382663a.png)

I'm now fetching the contact image asynchronously which introduces a small, but noticeable delay as you can see in [this video](https://vid.me/IXyG).

Fixes #4245

// FREEBIE


src/org/thoughtcrime/redphone/ui/CallCard.java
```diff
@@ -36,12 +43,13 @@
...
-  private TextView label;
-  private TextView elapsedTime;
-  private TextView status;
+  private       ImageView photo;
+  private       TextView  name;
+  private       TextView  phoneNumber;
+  private       TextView  label;
+  private       TextView  elapsedTime;
+  private       TextView  status;
+  private final Context   context = getContext();
```
moxie0: this seems weird, i think just call getContext() when needed or at the top of the method, no need for this to be an instance variable

src/org/thoughtcrime/redphone/ui/CallCard.java
```diff
@@ -62,8 +70,24 @@ public void setElapsedTime(String time) {
...
-    this.photo.setImageDrawable(recipient.getContactPhoto().asCallCard(getContext()));
+  private void setPersonInfo(final Recipient recipient) {
+    new AsyncTask<Void, Void, ContactPhoto>() {
+      @Override
+      protected ContactPhoto doInBackground(Void... params) {
+        DisplayMetrics metrics       = new DisplayMetrics();
+        WindowManager  windowManager = (WindowManager) context.getSystemService(Context.WINDOW_SERVICE);
+        Uri            contentUri    = ContactsContract.Contacts.lookupContact(context.getContentResolver(),
+                                                                               recipient.getContactUri());
+        windowManager.getDefaultDisplay().getMetrics(metrics);
```
moxie0: i forget is this api 9 compatible?

moxie0:
it's uncommon to generate an image for only one dpi

haffenloher:
> no need for this to be an instance variable

don't know why I did this again, makes no sense. will fix

> i forget is this api 9 compatible?

Yes, according to the docs the Display & DisplayMetrics stuff is from api 1.
I wasted hours wrestling with different api 9 / api 10 emulators but couldn't even get GCM to work, so no Signal registration :( is there a way to do api 9 testing in an emulator?

> it's uncommon to generate an image for only one dpi

Right. tbh I was happy that I managed to open the .psd file (found somewhere in the git history) with GIMP, change some colors and export it as PNG at full size (512x512 px).
Now GIMP doesn't seem to take the whole thing as a vector graphic, so upscaling makes it a little blurry, which I guess isn't good enough.
Perhaps someone who has Photoshop could easily generate high-res assets [from the files](https://github.com/WhisperSystems/Signal-Android/files/72878/artwork.zip)? If not, I'll try to recreate the image manually at a higher resolution.

agrajaghh:
@haffenloher have a look at: https://github.com/agrajaghh/Signal-Android/tree/4971_artwork/artwork/4971_artwork

I created a .svg and generated the different png files. Tell me if you need different sizes...

haffenloher:
@agrajaghh Fantastic! Thank you =)

I took your .svg and used a base size of 480x480px, so that each image is at least as wide as the maximum phone width for its dpi from https://design.google.com/devices/

agrajaghh:
you might want to use png optimization tools (like OptiPNG or pngcrush) to decrease file size or use the files here: (95KB -> 38KB).

https://github.com/agrajaghh/Signal-Android/tree/4971_artwork/artwork/4971_artwork

altough I'm not sure if gradle is optimizing the png's anyway :smirk:

haffenloher:
I tried pngcrush on them with literally no effect :blush: 
Thanks again! Which settings did you use?

agrajaghh:
yeah I think optiPNG is better than pngcrush, try `optipng -o7 file`

And I just discovered ZopfliPNG which seems to compress them even better :blush: (but it's REALLY slow)

haffenloher:
cool! Tried ZopfliPNG now, my best result has the exact same size as your earlier (optiPNG?) compressed images (39,1 kB), so I'll just stick with those :)

PR #4987, opened 2015-12-28
Release MediaPlayers immediately after stopping ringers

haffenloher:
When receiving multiple subsequent calls, my Android 4.2 MTK device was incorrectly playing the ringtone from where it left off last time.
We do already call `stop()` and the docs say that should be enough, but my phone needs an extra `seekTo(0)`.

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/redphone/audio/IncomingRinger.java
+++ b/src/org/thoughtcrime/redphone/audio/IncomingRinger.java
@@ -56,7 +56,6 @@ public class IncomingRinger {
   public IncomingRinger(Context context) {
     this.context = context.getApplicationContext();
     vibrator = (Vibrator)context.getSystemService(Context.VIBRATOR_SERVICE);
-    player = createPlayer();
   }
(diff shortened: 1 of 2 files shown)


cascheberg:
How about this instead?
```
player.stop();
player.release();
player = null;
```
It seems that the release is currently missing anyway.

haffenloher:
I thought of that, too, but maybe the current code tries to reuse the MediaPlayer for a reason?
On the other hand, the docs do suggest that not calling release() might cause problems

cascheberg:
That is exactly what I read in the docs as well ;)
I don't see why it should reuse it. `OutgoingRinger` is not doing it either.

haffenloher:
@cascheberg Implemented your suggestion now, works better than before. With my initial solution, I could sometimes hear the ringtone being resumed at its old position for the fraction of a second before the "rewind" kicked in.
I guess I'll change the PR title now ;)

PR #4991, opened 2015-12-29
Add BBM Emojies

wizdom13:


moxie0:
thanks but we're just going to stick with signal emoji and android emoji

PR #5008, opened 2015-12-31
Upgrade Gradle to 2.10 and Android plugin to 1.5.0

szpak:
The main reason behind this change was to enable Gradle tarball
checksum verification introduced in Gradle recently. It is a great
companion to already used gradle-witness plugin.

Due to a bug in Android Plugin for Gradle 2.10 was considered
less than 2.2. It was fixed in recently released 1.5.0.

Unfortunately the checksums are not even published by Gradle team
(not to mention digital signing...).

FREEBIE


szpak:
Any chance to merge this PR (or get constructive feedback what should be enhanced)?

moxie0:
thanks but we generally don't bump the version on dependencies unless we have a specific reason

Issue #3828, opened 2015-07-28
Support SMS URIs with no number?


jpsingleton:
I previously commented on a closed issue https://github.com/WhisperSystems/TextSecure/issues/2578 and pull request https://github.com/WhisperSystems/TextSecure/pull/3030 but I'm not sure these will get noticed.

Would it be possible to extend support for SMS URIs to handle links with body text but no number (aka no `path` or `data` variables set in the code)? e.g. `sms:?body=Hi%20there`.

The use case is you have a message but want to choose who to send it to. Like you can with [`mailto:?subject=Hi%20there&body=Hello%20there`](mailto:?subject=Hi%20there&body=Hello%20there) and `whatsapp://send?text=xxx`. Other SMS apps already support this behaviour.

I'm not sure if it's just as simple as not throwing the `URISyntaxException` in `Rfc5724Uri.java` and returning an empty path.

I'm happy to have a go at doing this if it's something people want and has a chance of getting merged.

mcginty:
Yeah, looks like we should let people choose via ShareActivity. That sounds like a reasonable PR idea to me!

jpsingleton:
@mcginty OK, I've got this working and it's a very small change. It's tested on device and all looking good. However, I've been struggling for ages to get the unit tests to run. I think it might be related to issue https://github.com/WhisperSystems/TextSecure/issues/3474. I don't want to check in broken tests but I can't verify they work. Any ideas?

pgaubatz:
Hey @jpsingleton!

Are you planning on submitting a PR on this issue?
I am asking, because I've just stumbled upon this bug... 

Cheers,
Patrick

jpsingleton:
That's the plan but I need pointing to some better documentation on how to run the unit tests so I can be sure I haven't broken anything. Also I need to merge from upstream which is non-trivial as the folder structure has changed with the rename to Signal.

jpsingleton:
@pgaubatz I was going to have another go at applying my changes to a fresh fork but it looks like you've beaten me to it! Good work. It's probably worth asking someone like @mcginty or @moxie0 to get this merged. I'd like to see this in the stable version on play. It's the only thing currently stopping me from recommending Signal over WhatsApp. Perhaps a quick code review would help.

[...]

PR #5017, opened 2016-01-03
Support "sms:" URIs with "body" parameter but no recipient(s)

pgaubatz:
This change allows for URIs like `sms:?body=hello` to be used and resolves issue #3828.

### DIFF
--- a/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
+++ b/src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
@@ -46,24 +46,23 @@ public class Rfc5724Uri {
   private String parsePath() throws URISyntaxException {
     String[] parts = uri.split("\\?")[0].split(":", 2);
 
-    if (parts.length < 2 || parts[1].isEmpty()) throw new URISyntaxException(uri, "invalid path");
-    else                                        return parts[1];
+    if (parts.length < 2) throw new URISyntaxException(uri, "invalid path");
+    else                  return parts[1];
   }
 
   private Map<String, String> parseQueryParams() throws URISyntaxException {
     Map<String, String> queryParams = new HashMap<>();
-    if (uri.split("\\?").length < 2) {
-      return queryParams;
+    if (uri.split("\\?").length > 1) {
+      for (String keyValue : uri.split("\\?")[1].split("&")) {
+        String[] parts = keyValue.split("=");
+
+        if (parts.length == 1) queryParams.put(parts[0], "");
+        else                   queryParams.put(parts[0], URLDecoder.decode(parts[1]));
+      }
     }
 
-    for (String keyValue : uri.split("\\?")[1].split("&")) {
-      String[] parts = keyValue.split("=");
-
-      if (parts.length == 1) queryParams.put(parts[0], "");
-      else                  queryParams.put(parts[0], URLDecoder.decode(parts[1]));
-    }
-
-    return queryParams;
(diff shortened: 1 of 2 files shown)


jpsingleton:
:+1: :shipit:

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -46,24 +46,23 @@ private String parseSchema() throws URISyntaxException {
   private String parsePath() throws URISyntaxException {
     String[] parts = uri.split("\\?")[0].split(":", 2);
 
-    if (parts.length < 2 || parts[1].isEmpty()) throw new URISyntaxException(uri, "invalid path");
-    else                                        return parts[1];
+    if (parts.length < 2) throw new URISyntaxException(uri, "invalid path");
```
samlanning: why remove this check? All it was doing is ensuring that URLs like `"sms:"` were invalid...

samlanning: ah wait i see you now allow URLs with no path and a query string, my bad.

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
```diff
@@ -46,24 +46,23 @@ private String parseSchema() throws URISyntaxException {
...
 
-      if (parts.length == 1) queryParams.put(parts[0], "");
-      else                  queryParams.put(parts[0], URLDecoder.decode(parts[1]));
+        if (parts.length == 1) queryParams.put(parts[0], "");
+        else                   queryParams.put(parts[0], URLDecoder.decode(parts[1]));
+      }
     }
 
-    return queryParams;
+    if (path.isEmpty() && !queryParams.containsKey("body")) throw new URISyntaxException(uri, "missing recipient(s)");
```
samlanning: I'm not personally a fan of using the path parameter here like this, it now adds an explicit call order requirement for these methods in the constructor. I personally think it would probably be an improvement to make these methods static, and pass in the properties that are currently used as method arguments.

pgaubatz:
Hey @samlanning!

Thanks for your comments. I've just pushed another commit that should (hopefully) address your concerns...

Cheers,
Patrick

samlanning:
@pgaubatz perfect =)

src/org/thoughtcrime/securesms/util/Rfc5724Uri.java
moxie0: This is pretty weird. This is a generic URI class, it shouldn't know anything about "body" or recipients or anything like that.

pgaubatz: Hey!

Thanks for commenting!
I surely don't want to argue, but the name of the class lead me to believe that it was designed explicitly for parsing URIs according to RFC 5724. And in fact, this RFC does define "body" and the concept of recipients...

So although i really want to get this change merged, I'm not quite sure how to address your concern... :)

moxie0: huh, well what do you know.

moxie0:
I think it's a little weird that getPath() returns an empty string (?) if no recipient is specified.  That should probably be null, and annotated as @ nullable.  If you follow that up the call change and extend the annotations, I think it's going to require a check before passing into RecipientFactory?

moxie0:
Closing due to inactivity. Please reopen if you revisit!

PR #5028, opened 2016-01-05
Do not force close when the user choose to share location on devices with GmsCore

danielegobbetti:
GmsCore (https://github.com/microg/android_packages_apps_GmsCore) mimics google play services,
but does not implement location.


danielegobbetti:
Not going to give all the details requested to sign the CLA (postal address and phone number???), sorry.

PR #5043, opened 2016-01-08
Add Background color to chat conversations.

wizdom13:
Add the ability to set the background colors to chat conversation
screen.


moxie0:
You can't just unilaterally bump the minsdk version.  Also, if we were doing to do something like this it would more likely be wallpapers per-conversation.

wizdom13:
I am working on the wallpaper too. Background color should be available in parallel to wallpapers because it has alpha transparency over the wallpaper. Like if you wanna make a 30% white color over the background image.

PR #5063, opened 2016-01-13
Improve TransportOption strings, landscape message composition

### LINKED ISSUE #4786: UI issues with message composition in portrait & landscape
(See https://github.com/WhisperSystems/Signal-Android/pull/4755 for initial discussion)

**Issues in portrait & landscape:**
- The current compose hint is "Send Signal message" (or "Send unsecured SMS", "Send unsecured MMS"), which doesn't make much sense. After tapping on the compose field, the user types the message and doesn't send it.

**Issues in landscape:**
- Same problem as in portrait with the compose hint
- Additionally, the Send button uses the very same strings as the compose hint (here it makes sense). But these rather long strings make that button very wide, taking far too much screen estate, i. e. leaving not much room for the actual message content.

**Suggested solution:**
- Separate the compose hint strings from the button text strings
- Use appropriate wording in the compose hint strings
- Wrap the button text strings into 3 lines, so the button gets horizontally smaller and vertically larger, leading to more screen estate for the message content on the left side of the button. New lines ("\n") are already used by some currently existing Signal strings, so shouldn't be anything new for translators.

**Suggested Compose Hint strings (portrait & landscape):**
- Type Signal message
- Type unsecured SMS
- Type unsecured MMS

**Suggested Button Text strings (landscape):**
- Send\nSignal\nmessage
- Send\nunsecured\nSMS
- Send\nunsecured\nMMS

Ideally, the Send button in landscape should be colored signal blue for encrypted and gray for unencrypted messages. But I understand this is not feasible right now.

unrulygnu:
Fixes #4786 

**Before:**
before

**After:**
after

Used "enter" instead of "type" to accommodate alternate input methods (speech-to-text, etc.), as in other apps.  @RiseT, please let me know what you think.

Also made TransportOption strings more consistent with other strings by replacing "insecure" with "unsecured".

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -131,8 +131,8 @@
     <string name="ConversationActivity_this_device_does_not_appear_to_support_dial_actions">This device does not appear to support dial actions.</string>
     <string name="ConversationActivity_leave_group">Leave group?</string>
     <string name="ConversationActivity_are_you_sure_you_want_to_leave_this_group">Are you sure you want to leave this group?</string>
-    <string name="ConversationActivity_transport_insecure_sms">Insecure SMS</string>
-    <string name="ConversationActivity_transport_insecure_mms">Insecure MMS</string>
+    <string name="ConversationActivity_transport_unsecured_sms">Unsecured SMS</string>
+    <string name="ConversationActivity_transport_unsecured_mms">Unsecured MMS</string>
     <string name="ConversationActivity_transport_signal">Signal</string>
     <string name="ConversationActivity_lets_switch_to_signal">Let\'s switch to Signal %1$s</string>
     <string name="ConversationActivity_lets_use_this_to_chat">Let\'s use this to chat: %1$s</string>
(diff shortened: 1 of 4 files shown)


RiseT:
@unrulygnu :+1: Perfect, thanks!

unrulygnu:
Just rebased on current master (72064d8), fixed conflicts.  Tested. Awesome work on dual SIM!

Please let me know if you need anything else on this PR.

2-4601:
@unrulygnu Tested your PR, please have a look at #5294.

unrulygnu:
I'm following up on that issue's thread.  It seems to be unrelated to this PR.

RiseT:
PR probably became obsolete because of https://github.com/WhisperSystems/Signal-Android/commit/fe1cc3d6d1906b3503309b82aaa68c581e929a80...

cascheberg:
I think it's still a good idea to show "Signal message" in the input field as it is now, but specifically label the send button "Send Signal message" (including newlines).

Issue #4611, opened 2015-11-19
Dark theme flaw in "manage linked devices" 


McLoo:
Master at 4314a4b4
![screenshot_2015-11-19-21-13-46](https://cloud.githubusercontent.com/assets/5326916/11283014/c1e5c696-8f02-11e5-8118-06f2256110c3.png)

2-4601:
Hmm... I see this:
![devprov](https://cloud.githubusercontent.com/assets/174176/11283359/f0b12e0c-8f0c-11e5-8e95-174942fbd689.png)
Is the black rounded rectangle supposed to be a camera view (to scan the QR)?

mcginty:
@2-4601 yep, and the screen that @McLoo is what you get after a scan of the QR. Do you just see a black screen? If so, can you post logs?

2-4601:
Ok, will create a new issue with logs an delete my comments from here (I thought first that these are the same issue)

McLoo:
And also for the scanning activity. So bright...  
(killed cam therefore the black preview) 



![screenshot_2015-11-26-00-42-02](https://cloud.githubusercontent.com/assets/5326916/11411960/e4e4fde6-93d6-11e5-82a5-dbb0db17da02.png)

benmayne:
this was my log when I saw it: 

https://gist.github.com/anonymous/7e2a22d1d181aeb5a83e

![2015-12-10 03 58 10](https://cloud.githubusercontent.com/assets/3374110/11706682/b2eea858-9eaf-11e5-85a3-63730eb66743.png)

PR #5066, opened 2016-01-14
Theme background color in device link card

wlritchi:
Fix an issue where the card was almost unreadable on dark theme. The text was themed as expected, to very light grey, but the background of the card remained white.

Before and after:

Before and after comparison

### DIFF
--- a/res/layout/device_add_fragment.xml
+++ b/res/layout/device_add_fragment.xml
@@ -30,7 +30,7 @@
             android:layout_weight="1"
             android:paddingLeft="16dp"
             android:paddingRight="16dp"
-            android:background="@color/gray5"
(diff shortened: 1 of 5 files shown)


McLoo:
@wlritchi if you extended this to the scanning screen, #4611 would be completely fixed :)

wlritchi:
@McLoo d0a65f8 at least makes that dialog readable, though it could probably do with being themed as well. I'll look into it tonight :)

wlritchi:
@McLoo Here's a before and after for the scanning screen using textColorSecondary (light theme is unchanged). Does this look like a good fix for #4611?

before and after

McLoo:
Good enough fix for me 😁

PR #5075, opened 2016-01-15
Include name when sending contact info

cascheberg:
Sending a contacts phone number usually requires to also provide the contacts name.

// FREEBIE


infinite-Joy:
composeText.append(contactData.name + "\n" + contactData.numbers.get(0).number would have been better.
Name and number separated by a new line.
What do you think?
And also will it seamlessly take the name also when I try to save the number in my contacts list?

cascheberg:
No, this is just a quick fix because sending a number without a name is pretty much useless.
See #535 for more advanced contact sending.

moxie0:
I kind of want to hold out for vcard, but we'd have to synchronize with ios

cascheberg:
While vCard support would be great, this is a two lines change that improves the usability of this feature and is ready to be merged.

moxie0:
Sorry, I don't actually think this improves the usability of the feature the way I've seen people currently use it. I think we either want a full attachment format, or just the number.  Thanks!

cascheberg:
OK. (Although I'd be curious to know which use cases you have in mind.)

PR #5082, opened 2016-01-16
Moved Help section to the top of the README.

diederikdehaas:
Lots of 'average users' probably won't scroll past all the 'technical'
information (bug reports, beta, translations, bitcoin, etc) and are
therefor likely to miss the Help section and direct their 'simple'
requests/questions to the wrong place, like issues.

Technical users are far more likely to read the whole thing and should
therefor not be negatively impacted by this change.

This PR tries to address that issue.
It may be useful to add another level 1 type header below it, named "Contributing" for example, to make the distinction more clear. I haven't done it yet, as it's easier to add a commit then to 'remove' it. (Cherry-picking would be another option).

This PR is of course a FREEBIE :wink: 
(If this is the wrong way to indicate this, please point me on how to do it properly)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,18 @@ Currently available on the Play store.
 
 *[![Play Store Badge](https://developer.android.com/images/brand/en_app_rgb_wo_60.png)](https://play.google.com/store/apps/details?id=org.thoughtcrime.securesms)*
 
+Help
+====
+## Support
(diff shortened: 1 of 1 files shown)


diederikdehaas:
I just signed the CLA and provided quite some private data. I did it because I'm confident that **you** respect my privacy, but you may want to clarify that (for others).

PR #5095, opened 2016-01-19
Fix five strings in strings.xml

2-4601:
Fixed
- one ambiguous pronoun
- one remaining reference of ambiguous **'lock screen'** (See: https://github.com/WhisperSystems/Signal-Android/pull/4907#issuecomment-167146108)
- three cases of **'click'** -> **'tap'**

And gets rid of unnecessary line breaks in two strings.

-----
As a side note:
There are still some strings which have **'click'** in their name but all the values use **'tap'** so users would not benefit from changing them and thus I ignored them.

### DIFF
--- a/res/layout/prompt_apn_activity.xml
+++ b/res/layout/prompt_apn_activity.xml
@@ -29,7 +29,7 @@
                       android:layout_width="fill_parent"
                       android:layout_height="wrap_content"
                       android:layout_marginBottom="16dip"
-                      android:text="@string/prompt_mms_activity__to_send_media_and_group_messages_click_ok"/>
+                      android:text="@string/prompt_mms_activity__to_send_media_and_group_messages_tap_ok"/>
                         
             <LinearLayout android:orientation="horizontal"
                           android:gravity="center"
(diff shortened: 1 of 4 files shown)


RiseT:
Sorry for being late (my eyes were a pain in the ass - if that makes any sense).

The dialog already has the question in its title, so I suggest

**"This will permanently unlock Signal and message notifications."**

for the description string, which is in line with Google's design guidelines and all other dialogs in Signal.

2-4601:
Thanks, wasn't familiar with that design guideline. At first I had a string without a question mark but then changed it to a question because the "OK" button was labelled 'Disable'.

RiseT:
While you're at it - and if you don't mind - in

*ConversationItem_this_media_has_been_stored_in_an_encrypted_database_external_viewer_warning*

could you replace **"disk"** with **"storage"**?

Don't know if there are too many phones with an HDD in it... :grinning:

2-4601:
:boom::floppy_disk::boom:

RiseT:
:+1:

Issue #5102, opened 2016-01-22
System doesn't Support multiple mobile Devices (non Desktop)


bamberdd:
I can't use Signal on two mobile Devices at the same time because Signal-App does not allow me to  link the devices via Settings->Devices ...QR-Code scan. 
If the requested feature is realized it would be very useful to give the user the possibility to assign each conntected device with a priority für message delivery.

kaueraal:
I suppose this would be an important feature for later tablet support, as most users of a tablets also own a mobile phone.

x12x:
The qr code you're trying to scan is not for linking the device, but for verifying the other user's identity. It's not your fault if you misunderstood, just poor design choices

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

Issue #4642, opened 2015-11-21
NPE when attempting to link devices


jtracey:
I have a tablet with Signal still installed on it from #4457 that thinks it's still registered, but isn't. When I use my phone to scan the tablet's identity key QR code to try to link the two, I consistently get a NPE on my phone.
debug log: https://gist.github.com/38db7f98d0e4fe2b36c4

bartz1:
I don't think you can link tablets yet. You should try with the Signal extension after properly changing the server URL.

2-4601:
I guess without any in-app instructions this might seem the right way to scan the QR code of the device to be linked.
But the device provisioning needs a different kind of a QR code. Not the slave's public identity key.

Currently it only works with the Chrome plugin which will create a provisioning QR for the Android client. But the plugin is unreleased software and as far as I understand you should not really use it yet or if you decide to use it and something goes wrong OWS does not have the resources to support you.

Back to this issue. Tried to scan a random QR code and it went successfully through all the dialogs:
* Link this device?
 * It will be able to...
* Linking device...

And then crashed. It should definitely fail earlier with non-provisioning QRs.

PR #5111, opened 2016-01-23
Change device add string to refer only to Desktop

2-4601:
It seems that some users (#4642, #5102, https://github.com/WhisperSystems/Signal-Desktop/issues/624) are trying to link Signal with their other mobile devices even though only Desktop beta is supported.

This will change the string in the scanning view to give a nudge in the right direction:
_"Scan the QR code displayed on **the device** to link"_
**-->**
_"Scan the QR code displayed on **Signal Desktop** to link"_

This is a temporary fix so I didn't change the name of the string itself.

// FREEBIE

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -688,7 +688,7 @@
     <string name="country_selection_fragment__search">Search</string>
 
     <!-- device_add_fragment -->
-    <string name="device_add_fragment__scan_the_qr_code_displayed_on_the_device_to_link">Scan the QR code displayed on the device to link</string>
+    <string name="device_add_fragment__scan_the_qr_code_displayed_on_the_device_to_link">Scan the QR code displayed on Signal Desktop to link</string>
 
     <!-- device_link_fragment -->
     <string name="device_link_fragment__link_device">Link device</string>


bmaleck1:
Is there a plan to support multiple mobile devices?

moxie0:
ehhh I hear you, but I want to leave it as-is for now

Cj-Malone:
I just fell into this hole, please re consider @moxie0.

I registered a second device ( Android ) trying to do this, only to realise my main phone ( iPhone, same number ) won't connect anymore. To make my situation worse you can't re register on the iOS app so I'd have to uninstall and reinstall, and to compound the issues you can't back up on iOS and I'm not prepared to lose my data.

Trolldemorted:
@Cj-Malone Signal-Android can re-register (settings->advanced, disable "Signal messages and calls" and enable it again) without loosing your chat history, identity key or sessions. If Signal-iOS does not suppport that, you can try opening an issue in its repo.

If you want test running Signal-Android as a slave, deploy [this](https://github.com/Trolldemorted/Signal-Android/tree/multidevice) branch and report every bug you encounter. I have tested it with other Signal-Android and signal-cli master devices, but due to the lack of iOS devices i cannot test it with an iOS slave.

Cj-Malone:
@Trolldemorted Thanks, I have opened an issue on [Signal-iOS](https://github.com/WhisperSystems/Signal-iOS/issues/1812).

I've currently not got access to any non mobile devices so I can't build anything. It shouldn't be more than a week or two so I'll attempt it then, unless you want to provide an apk. I'm unsure of the internals of Signal, is the slave/master relationship "native", can I use my main (i)phone as a master and my secondary (Android) phone as a slave or would I have to use signal-cli as the master to both phones?

Trolldemorted:
I could provide you an apk, but it would be signed with my key and you will be unable to update/change it yourself, so i recommend building it yourself.

You should™ be able to link it to your iphone master, but since i have never done it i cannot promise it will work - please let me know. Vanilla Android master and my branch as a slave works, though. I don't think Signal-iOS supports being a slave (not sure if there already is an issue for that), otherwise you could just link it to a regular Signal-Android.

moxie0:
@Cj-Malone This issue is about changing the wording to say "desktop," which I decided to leave as-is for now. It sounds like you're saying you want me to reconsider the larger question of allowing non-desktop devices to be siblings? That I have no objection to, it's "on the roadmap."

Cj-Malone:
@moxie0 No I'm asking you to reconsider the wording, currently _it is_ confusing and counter intuitive. "Desktop" seems to be a much better fit, at least until non-desktop devices can be siblings.

Issue #5139, opened 2016-01-27
README.md image link broken


jzbz:
Same issue essentially as https://github.com/WhisperSystems/whispersystems.org/issues/100 except you will likely need to resize and host your own image...

Google has stopped hosting the Google Play badge linked in the README.md file[1]. A new one can be grabbed form their site[2].

[1] https://developer.android.com/images/brand/en_app_rgb_wo_60.png
[2] https://play.google.com/intl/en_us/badges/

2-4601:
Resolved by c5bb606

PR #5147, opened 2016-01-28
Fix google play store badge link in readme.md

theroflcoptr:
Generated HTML from https://play.google.com/intl/en_us/badges/ corrects #5139

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Signal uses your phone's data connection (WiFi/3G/4G) to communicate securely, o
 
 Currently available on the Play store.
 
-*[![Play Store Badge](https://developer.android.com/images/brand/en_app_rgb_wo_60.png)](https://play.google.com/store/apps/details?id=org.thoughtcrime.securesms)*
+<a href="https://play.google.com/store/apps/details?id=org.thoughtcrime.securesms&utm_source=global_co&utm_medium=prtnr&utm_content=Mar2515&utm_campaign=PartBadge&pcampaignid=MKT-Other-global-all-co-prtnr-py-PartBadge-Mar2515-1"><img alt="Get it on Google Play" src="https://play.google.com/intl/en_us/badges/images/generic/en-play-badge.png" height=60px /></a>
 
 ## Contributing Bug reports
 We use GitHub for bug tracking. Please search the existing issues for your bug and create a new one if the issue is not yet tracked!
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,7 +6,7 @@ Signal uses your phone's data connection (WiFi/3G/4G) to communicate securely, o
 
 Currently available on the Play store.
 
-*[![Play Store Badge](https://developer.android.com/images/brand/en_app_rgb_wo_60.png)](https://play.google.com/store/apps/details?id=org.thoughtcrime.securesms)*
+<a href="https://play.google.com/store/apps/details?id=org.thoughtcrime.securesms&utm_source=global_co&utm_medium=prtnr&utm_content=Mar2515&utm_campaign=PartBadge&pcampaignid=MKT-Other-global-all-co-prtnr-py-PartBadge-Mar2515-1"><img alt="Get it on Google Play" src="https://play.google.com/intl/en_us/badges/images/generic/en-play-badge.png" height=36px /></a>
```
2-4601: I think it's preferred not to use HTML in README.md. It's a Markdown file and it should use the Markdown syntax. See https://guides.github.com/features/mastering-markdown/ for details on how to use it (the current line can also be used as base for your changes).

skulumani: I believe the issue is that Markdown doesn't really allow you to set the image size. You have to resort to raw HTML. 

Or save an image at the correct size and link to that instead of Google.

2-4601: > Or save an image at the correct size and link to that instead of Google.

The Bitcoin donation badge is done like that. For consistency I think this should be done like that too.

skulumani: If you go to https://play.google.com/intl/en_us/badges/ you can generate a link to a badge for an app. 

Google seems to add those parameters for it's own purposes?

theroflcoptr:
> I think it's preferred to not use HTML in README.md

True, but github removed the ability to specify image size in their markdown.

> The Bitcoin donation badge is done like that. For consistency I think this should be done like that too.

The bitcoin donation badge isn't another company's branding. If we host the Google badge somewhere else, and Google updates their branding again, then the new badge would have to be re-hosted, and readme.md updated again. The PR as written seems like a lower-maintenance solution.

README.md
theroflcoptr: > I think it's preferred to not use HTML in README.md

True, but github removed the ability to specify image size in their markdown.

> The Bitcoin donation badge is done like that. For consistency I think this should be done like that too.

The bitcoin donation badge isn't another company's branding. If we host the Google badge somewhere else, and Google updates their branding again, then the new badge would have to be re-hosted, and readme.md updated again. The PR as written seems like a lower-maintenance solution.

Issue #5073, opened 2016-01-15
Advanced settings menu UI inconsistency


janimo:
The emoji setting uses a checkbox while the other two boolean settings use toggle buttons.

tehabe:
This isn't all. 

In the Notifications settings, "Vibrate" and "In-conversation notifications" have checkboxes
And in Privacy: Inactivity timeout passphrase

And yes, the emoji setting is still a checkbox.

Tested with Signal 3.12.0 on Android 6.0.1

janimo:
The others are subpreferences, as moxie pointed out in the related PR.

PR #5167, opened 2016-02-01
Use switch instead of checkbox for emoji setting.

janimo:
Material design guidelines recommend on/off switches instead of checkboxes
for toggling single value options.

// FREEBIE

### DIFF
--- a/res/xml/preferences_advanced.xml
+++ b/res/xml/preferences_advanced.xml
@@ -13,7 +13,8 @@
                         android:summary="@string/preferences__pressing_the_enter_key_will_send_text_messages"
                         android:title="@string/preferences__pref_enter_sends_title"/>
 
-    <CheckBoxPreference android:defaultValue="false"
+    <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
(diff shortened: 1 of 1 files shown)


dcherian:
Fixes #5073

eosrei:
Tested and works as designed.

res/xml/preferences_notifications.xml
```diff
@@ -36,7 +37,8 @@
         android:entries="@array/pref_led_blink_pattern_entries"
         android:entryValues="@array/pref_led_blink_pattern_values" />
 
-    <CheckBoxPreference android:key="pref_key_inthread_notifications"
+    <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
+                        android:key="pref_key_inthread_notifications"
                         android:title="@string/preferences__in_conversation_notifications"
                         android:summary="@string/preferences__play_inthread_notifications"
                         android:dependency="pref_key_enable_notifications"
```
moxie0: Both of these should still be check boxes, since they're sub-preferences

moxie0:
in 3.14.0

PR #5169, opened 2016-02-01
Attack vector: added alternative to failed bitmap, otherwise app becomes unusable.

neurotensin:
This could be used by attacker to force use of less secure methods, and possibly
other damage to device, simply by renaming a file to an incorrect mime-type. A bug report was filed (4228) so  more elegant solutions are welcome! This patch adds a "Util.isValidDimensions()" call from Glide to anticipate failure, and prepare a default Bitmap instead of crashing. The default error is left, in case the default BitMap cannot be used.


neurotensin:
https://gist.github.com/anonymous/ea45adee0e0a1d286717

moxie0:
this should fixed in 3.10

PR #5178, opened 2016-02-04
Honor synced end session messages

### LINKED ISSUE #5174: Honor synced end session messages
If Desktop resets a session, the remote party will delete all sessions with our number, including any with the Android client. Thus, when the android client receives a synced sent message with its end session flag set, it too should close all sessions with the destination and render the message as _Secure session reset_.

Currently it renders as a normal message and does not delete sessions.

haffenloher:
Fixes #5174


Trolldemorted:
is there anything wrong with this PR?

moxie0:
in 3.32.0

Issue #4279, opened 2015-10-23
Update emojis


5boro:
Apple has added [a bunch of new emojis with IOS 9.1](http://emojipedia.org/new-ios-9-1/), and users of signal for IOS are able to send them.
![1445607827279](https://cloud.githubusercontent.com/assets/3460479/10694907/b11ae118-7998-11e5-977a-559ff404af6e.jpg)
The intended emoji here was [taco](http://emojipedia.org/taco/)

jeremymasters:
I think we should hold off on signal until emojis are all added...;)
http://www.droid-life.com/2015/10/23/new-emoji-are-coming-to-android-new-emoji-are-coming-to-android/#more-176429

therealmarv:
Next week will be new emojis for Android Nexus phones: https://twitter.com/lockheimer/status/672206241003937792

![image](https://cloud.githubusercontent.com/assets/1050582/11556947/2c88039a-99aa-11e5-8cf1-f839184701a0.png)

JuLx64:
I see an option related to system emojis. It Signal can use its own emojis, why doesn't it support the new ones ?

PR #5185, opened 2016-02-06
Added 🕵&🖕 emoji

Pythonix:
Added 🕵&🖕 emoji


art1fa:
What do you think about adding the whole new iOS 9.1 emoji set? See issue #4279. Would be great!

Pythonix:
I want to work on that yes! But It's really painful to add all the images in photoshop and bring them to the right size. So if you know any programm that makes that easier let me know!

art1fa:
Check [this](http://otsarag.deviantart.com/art/Complete-Emoji-Collection-555211222) out, not sure about the license though. Hope it helps!

eosrei:
@Pythonix You can script the process with the Imagemagick `convert` and `montage` commands.

Pythonix:
@art1fa @eosrei I started working on adding all emojis. But now I'm facing two problems and I think somebody with more knowledge has to help me...
Screenshot
The first problem is that both noses are displayed black but the first one is white.
The second problem is that with the black nose there is the color modifier displayed.
Can you give me any advice where I should start to fix that?

art1fa:
Sorry I didn't look into this. Anyway, I think it's better to move the discussion, close this PR and create a new one once it's finished and tested 🙂 That's how OWS likes it I guess.

Pythonix:
Just re-opened to see if @eosrei can give me any hint. We'll close it later on.

moxie0:
Thanks @Pythonix but please only open PRs that are fully baked, fully tested, and 100% ready to be merged.

eosrei:
@Pythonix I don't think the Android fonts support Unicode modifiers right now. "Platforms without support for emoji modifiers display a missing symbol character (box with a cross, alien, or question mark in it) next to any modified emoji." src: http://emojipedia.org/boy-emoji-modifier-fitzpatrick-type-4/ Example: 👦🏽

Pythonix:
Yep. You're right. But whatsapp for example has its own implementation for that. That should be possible within signal too.

Date: Wed, 10 Feb 2016 01:06:28 -0800

PR #5197, opened 2016-02-09
Enable Gradle distribution checksum verification

szpak:
Gradle tarball checksum verification is a great companion to already
used gradle-witness plugin.

Gradle version has to be upgraded as the feature was introduced in 2.6.
Gradle 2.10+ cannot be used without upgrading Gradle plugin to 1.5.0
(as the earlier versions conside 2.10 as something lower than 2.2).

Unfortunately the checksums are not even published by Gradle team
(not to mention digital signing...).

This PR is similar to #5008, but without Gradle plugin upgrade.

FREEBIE


szpak:
Any comment why it was closed without merge?

PR #5199, opened 2016-02-09
Support for dual-sim SMS/MMS functionality

moxie0:
Allow source selection for sending SMS/MMS, and display the
SIM that received SMS/MMS.

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/InviteActivity.java
+++ b/src/org/thoughtcrime/securesms/InviteActivity.java
@@ -28,13 +28,15 @@ import org.thoughtcrime.securesms.components.ContactFilterToolbar;
 import org.thoughtcrime.securesms.components.ContactFilterToolbar.OnFilterChangedListener;
 import org.thoughtcrime.securesms.crypto.MasterSecret;
 import org.thoughtcrime.securesms.database.DatabaseFactory;
+import org.thoughtcrime.securesms.database.RecipientPreferenceDatabase.RecipientsPreferences;
 import org.thoughtcrime.securesms.recipients.RecipientFactory;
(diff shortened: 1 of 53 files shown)


src/org/thoughtcrime/securesms/InviteActivity.java
```diff
@@ -227,9 +229,14 @@ protected Void doInBackground(String... numbers) {
       if (context == null) return null;
 
       for (String number : numbers) {
-        final Recipients recipients = RecipientFactory.getRecipientsFromString(context, number, false);
-        if (recipients != null && recipients.getPrimaryRecipient() != null) {
-          MessageSender.send(context, masterSecret, new OutgoingTextMessage(recipients, message), -1L, true);
+        Recipients                      recipients  = RecipientFactory.getRecipientsFromString(context, number, false);
```
haffenloher: that large whitespace was probably unintended =)

src/org/thoughtcrime/securesms/service/MmsListener.java
```diff
@@ -63,9 +63,11 @@ public void onReceive(Context context, Intent intent) {
          isRelevant(context, intent)))
     {
       Log.w(TAG, "Relevant!");
+      int      subscriptionId = intent.getExtras().getInt("subscription", -1);
```
haffenloher: same here

2-4601:
I've been running this on a Huawei ICS device. It's not a dual SIM device and actually it doesn't even have a single SIM at the moment so maybe this was quite a useless test. But at least there doesn't seem to be any collateral damage when using just for Signal messages.

Issue #3257, opened 2015-05-25
Support GSM 7-bit alphabet for SMS


athanasio:
Hello,
it would be useful (in order to further promote TextSecure as a full messaging replacement to friends and family) to support the GSM 7-bit alphabet, http://en.wikipedia.org/wiki/GSM_03.38 , which would allow keeping the cost of SMS low in some countries.

moxie0:
Thanks but we have no control over how SMS is encoded or sent.  We pass a string to the OS, and it's up to your Android ROM.

athanasio:
Hello, thank you for your reply; unfortunately your answer is not correct. It's the SMS app's responsibility to convert the text to 7-bit. I have replicated the issue; the default messaging app in Android 5.1 has the option for 7-bit and converts the text accordingly; if set to automatic, then it uses the full alphabet and the text is reduced to 60 characters per message. TextSecure defaults to automatic without the option to enable 7-bit.

Issue #3217, opened 2015-05-21
UTF-16 messages leads to 70 characters  SMS


vincenttschanz:
For some reasons, textsecure on my phone seems to send 16bit encoded SMS. So I can only have 70 characters per SMS instead of 140 in 8bit or 160 in 7bit.

Is the character encoding choosed by TextSecure or is it enforced by android system?

I suspect 16bit encoding is mandatory for emoji, but in the case if I don't use them, could TextSecure switch to 8bit, or 7bit if I don't even use accents? This could cut by two my SMS budget.

moxie0:
We have no control over this.  We hand a string to the OS and it does what it does.

vincenttschanz:
Ok, I supected that, thank you.

Issue #1864, opened 2014-08-22
Feature Request: Option to strip unicode characters


merdox:
Recently I stopped using the stock CyanogenMod SMS app in favor of TextSecure. One thing that bothered me and continues to do so is that there is no option to strip all unicode characters. It would be really usefull to have this option for alphabets that use non-standard characters ( e.g Cyrillic, Greek ). This basically means swapping any non-standard character for its closest counterpart in the default GSM alphabet ( e.g. 'á' => 'a' ) allowing the use of 160 characters per SMS instead of 70. I consider it an essential feature especially if you are not online, as the same text in unicode might charge you more than double than it would on the default GSM alphabet.

I hope you implement this feature. Thanks and keep up the great work! :)

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #5203, opened 2016-02-09
Add Unicode Stripping for SMS

Rawi666:
Adds a functionality for stripping unicode characters for unencrypted sms messages only. Signal messages are always intact.
I grabbed UnicodeFilter class from CyanogenMod and modified it a bit and therefore I left their licence as it was.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -999,6 +999,8 @@
     <string name="preferences__mmsc_password">MMSC Password</string>
     <string name="preferences__sms_delivery_reports">SMS delivery reports</string>
     <string name="preferences__request_a_delivery_report_for_each_sms_message_you_send">Request a delivery report for each SMS message you send</string>
+    <string name="preferences__sms_unicode_stripping">Simple characters only</string>
+    <string name="preferences__strip_all_unicode_characters">Convert special characters in SMS messages you send</string>
     <string name="preferences__automatically_delete_older_messages_once_a_conversation_exceeds_a_specified_length">Automatically delete older messages once a conversation exceeds a specified length</string>
     <string name="preferences__delete_old_messages">Delete old messages</string>
     <string name="preferences__chats">Chats and media</string>
--- a/src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
+++ b/src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
@@ -23,11 +23,14 @@ import org.thoughtcrime.securesms.service.SmsDeliveryListener;
 import org.thoughtcrime.securesms.transport.UndeliverableMessageException;
 import org.thoughtcrime.securesms.util.NumberUtil;
 import org.thoughtcrime.securesms.util.TextSecurePreferences;
+import org.thoughtcrime.securesms.util.UnicodeFilter;
 import org.whispersystems.jobqueue.JobParameters;
 
 import java.util.ArrayList;
 
 public class SmsSendJob extends SendJob {
+
+  private static final long serialVersionUID = 0L;
 
   private static final String TAG = SmsSendJob.class.getSimpleName();
 
@@ -75,6 +78,14 @@ public class SmsSendJob extends SendJob {
     MessageNotifier.notifyMessageDeliveryFailed(context, recipients, threadId);
   }
 
+  private String getMessageBody(String body) {
+    if (TextSecurePreferences.isUnicodeStrippingEnabled(context)) {
+      return UnicodeFilter.filter(body);
+    } else {
+      return body;
+    }
+  }
+
   private void deliver(SmsMessageRecord message)
       throws UndeliverableMessageException
   {
@@ -96,7 +107,7 @@ public class SmsSendJob extends SendJob {
       throw new UndeliverableMessageException("Not a valid SMS destination! " + recipient);
     }
 
-    ArrayList<String> messages                = SmsManager.getDefault().divideMessage(message.getBody().getBody());
+    ArrayList<String> messages                = SmsManager.getDefault().divideMessage(getMessageBody(message.getBody().getBody()));
     ArrayList<PendingIntent> sentIntents      = constructSentIntents(message.getId(), message.getType(), messages, false);
     ArrayList<PendingIntent> deliveredIntents = constructDeliveredIntents(message.getId(), message.getType(), messages);
 
--- a/src/org/thoughtcrime/securesms/util/UnicodeFilter.java
+++ b/src/org/thoughtcrime/securesms/util/UnicodeFilter.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (C) 2012-2016 The CyanogenMod Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.thoughtcrime.securesms.util;
+
+import java.text.Normalizer;
+import java.util.regex.Pattern;
+
+/**
+ * Attempts to substitute characters that cannot be encoded in the limited
+ * GSM 03.38 character set. In many cases this will prevent sending a message
+ * containing characters that would switch the message from 7-bit GSM
+ * encoding (160 char limit) to 16-bit Unicode encoding (70 char limit).
+ */
+public class UnicodeFilter {
+    private static final Pattern DIACRITICS_PATTERN =
+            Pattern.compile("\\p{InCombiningDiacriticalMarks}");
+
+    public static String filter(String source) {
+        StringBuilder output = new StringBuilder();
+        final int sourceLength = source.length();
+
+        for (int i = 0; i < sourceLength; i++) {
+            String s = String.valueOf(source.charAt(i));
+
+            // Try normalizing the character into Unicode NFKD form and
+            // stripping out diacritic mark characters.
+            s = Normalizer.normalize(s, Normalizer.Form.NFKD);
+            s = DIACRITICS_PATTERN.matcher(s).replaceAll("");
+
+            // Special case characters that don't get stripped by the
+            // above technique.
+            s = s.replace("Œ", "OE");
+            s = s.replace("œ", "oe");
+            s = s.replace("Ł", "L");
+            s = s.replace("ł", "l");
+            s = s.replace("Đ", "Dj");
+            s = s.replace("đ", "dj");
+            s = s.replace("Α", "A");
+            s = s.replace("Β", "B");
+            s = s.replace("Ε", "E");
+            s = s.replace("Ζ", "Z");
+            s = s.replace("Η", "H");
+            s = s.replace("Ι", "I");
+            s = s.replace("Κ", "K");
+            s = s.replace("Μ", "M");
+            s = s.replace("Ν", "N");
+            s = s.replace("Ο", "O");
+            s = s.replace("Ρ", "P");
+            s = s.replace("Τ", "T");
+            s = s.replace("Υ", "Y");
+            s = s.replace("Χ", "X");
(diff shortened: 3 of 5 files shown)


moxie0:
why?

Rawi666:
Why I left their licence? Because I'm not sure which one I should put. I can change it if you want

moxie0:
why do you want this functionality?

Rawi666:
Unicode stripping is very important and popular in countries where you use non-ascii characters. If you put at least 1 non-ascii character then sms will be converted to utf16 and you can put max 70 characters in single message. This results that the application displays false information how many messages will be really sent and you are charged for multiple messages from mobile provider. Therefore aware users use this functionality to strip non-ascii characters and pay less for 1 single message

moxie0:
https://github.com/WhisperSystems/Signal-Android/blob/master/CONTRIBUTING.md#development-ideology

I've never heard of this, and nobody has ever asked for this before. Seems unlikely that it's worth the preference?

2-4601:
I recall someone asking this before... Search found this old issue #1864. Still doesn't seem to be a high demand feature.

Rawi666:
All three issues you mentioned #1864, #3217, #3257 relate to this exactly issue ;)
Trust me, this option is **very** needed in languages with non-ascii characters, but you are the boss, it's up to you.
I suspect not everyone knows that there is the discussion on github about this.
Btw. this app was mentioned in the most popular Polish portal about security https://niebezpiecznik.pl/post/ustawa-inwigilacyjna-jak-sie-bronic-przed-podgladaniem-przez-policje-i-sluzby/
In short: our government has just implementet very controversial surveillance law so the demand for this app will raise.
So please be aware that many Polish users will start using this app and here in Poland we need unicode stripping unfortunately. All keyboard suggestions always puts word with our non-ascii chars and that sucks :(

bungabunga:
I live in Slovenia and can confirm what @Rawi666 is saying is true although not as important as it used to be, because here every smartphone user now have a plan with near unlimited amount of SMS messages included.

JavaJens:
@Rawi666 Especially then you should not need to worry, as you should be sending encrypted messages only :wink:

oldium:
I was just looking for an app that could send encrypted messages as well as non-encrypted SMS texts and found out this request. I also think that this feature would be really appreciated for all non-English languages (I am from the Czech Republic) if you really want to use the application for sending also non-encrypted messages via standard SMS delivery channel. I used this feature on my good old Nokia Symbian phone, which I recently replaced with a new Android phone, which now also has Unicode stripping (I switched to Cyanogenmod because of this). The Unicode stripping in my old Symbian phone was even more advanced and changed close-looking matches like á to à, ó to ò and the like. This feature would be beneficial for anybody who wants to save money (70 UTF16 characters vs 160 GSM-7bit characters in single SMS).

Typing with diacritics and removing it afterwards is needed, because all modern system keyboards offer support for writing just with diacritics during auto-completion and word correction. So having this feature is needed for proper internationalization support.

res/values/strings.xml
```diff
@@ -999,6 +999,8 @@
     <string name="preferences__mmsc_password">MMSC Password</string>
     <string name="preferences__sms_delivery_reports">SMS delivery reports</string>
     <string name="preferences__request_a_delivery_report_for_each_sms_message_you_send">Request a delivery report for each SMS message you send</string>
+    <string name="preferences__sms_unicode_stripping">Strip unicode</string>
+    <string name="preferences__strip_all_unicode_characters">Strip all unicode characters</string>
```
moxie0: In Google's Messenger this is called "Simple characters only" and the summary is "Convert special characters in SMS messages you send."  Which I like better.

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -126,6 +126,7 @@
 import org.thoughtcrime.securesms.util.GroupUtil;
```
moxie0: This change shouldn't touch anything in CoversationActivity.  Just the preferences, and SmsSendJob.

res/values/strings.xml
Rawi666: Ok, I'll change it

src/org/thoughtcrime/securesms/ConversationActivity.java
Rawi666: But if you move to a lower layer then with multi-character replacement like Œ->OE, Đ->Dj, etc. the app will display wrong number of characters in conversation activity and potentialy wrong number of messages to be sent. Do you really want me to move to SmsSendJob anyway?

moxie0: Yes, the transformation should happen in SmsSendJob.  If you want to preserve the character count, then create a new subclass of CharacterCalculator and change the base method signature to accept a charsequence rather than an int.

Rawi666: Ok... and wouldn't this cause storing sms message in the database in its original form (with special characters) and only send without special characters? This way the user will see his message in its original form even though the message is sent without special characters. May be a bit confusing as the sender will not see the actual replacement. Or am I wrong?

moxie0: What do other messengers do? Should it translate them as they're typed?

Rawi666: In cyanogen and in go sms pro the translation is done in the view activity so that the translated version is stored, displayed and sent consistently

Rawi666: I removed stripping when saving a message as a draft as this was not intended and could affect Signal messages. The rest is fine as it is as otherwise it will not work as it should

moxie0: Would it be better if the translation happened as you type? That way it's clear what you're sending?

Rawi666: I have never seen such a solution in other sms apps. This may sound like a better solution but I'm not sure if it will work for everyone for the following reasons:
- Different keyboards with different auto-completion algorithms can behave differently if the text is automatically changed as the keyboard changes the text - some keyboards still have the autocompleted word selected, some don't. This would require extensive testing.
- The translation will have to be performed each time a character is entered and that means that phones with slower CPU (or laggy roms like Samsung's...) may introduce some lag, especially when entering long messages (hard to test for me... I have Nexus 5 that is fast enough anyway I suppose). Unicode filtering uses RegExp that I think is CPU consuming in some way
- I'm not sure where would different keyboards place cursor if a user is changing (using autocompletion) text that is not at the end of the text (eg. in the middle) and the text length has changed (and it can be changed due to multi-character replacement that is in Czech language for example).
Even if it worked for me I wouldn't be sure if it worked for everyone else.
I think that all non-English users are already used to translating text just before sending messages and it's the proven solution.

The best solution in my opinion would be to implement such a feature in keyboards itself so that a user may choose in his keyboard settings that for a specific set of applications he doesn't want the keyboard to use any non-English characters, but I think that keyboards don't know what application uses it.
On the other hand that would require from the user some knowledge where to configure such a setting in his specific keyboard...

oldium: Usually (in SMS applications that I met) you can change the character set after you actually write the SMS (or during writing) - but before sending the message. The stored message is the one that was really sent and if you do not send it, the full character set is kept (i.e. for draft messages). The shown character count during typing respects the currently selected character set (the selected conversion).

Having this in the keyboard itself looks flexible (the application would need to tell to the keyboard about the wanted conversions), but this would also remove the option to change the character set during the typing or when you finished the typing.

moxie0:
doesn't feel right to me in the activity, i think it should either go in the edittext or the job

Rawi666:
OK, moved to the job

src/org/thoughtcrime/securesms/util/UnicodeFilter.java
```diff
@@ -0,0 +1,99 @@
...
+ * encoding (160 char limit) to 16-bit Unicode encoding (70 char limit).
+ */
+public class UnicodeFilter {
+    private static final Pattern DIACRITICS_PATTERN =
+            Pattern.compile("\\p{InCombiningDiacriticalMarks}");
+
+    public UnicodeFilter() {
+    }
+
+    public String filter(String source) {
```
moxie0: is there a reason not to make this a static method?

src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
```diff
@@ -75,6 +76,14 @@ public void onCanceled() {
     MessageNotifier.notifyMessageDeliveryFailed(context, recipients, threadId);
   }
 
+  private String getMessageBody(String body) {
+    if (TextSecurePreferences.isUnicodeStrippingEnabled(context)) {
+      return new UnicodeFilter().filter(body);
+    } else {
+      return body;
+    }
+  }
```
moxie0: sadly this class doesn't have a serialization version id, and i'm not willing to break pending jobs for this change, so please make the change such that it doesn't break the current serialization signature

Rawi666: I made UnicodeFilter.filter a static method so it should be ok now. In fact there are no reasons for this method to be instance as there are no instance fields.

src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
```diff
@@ -75,6 +76,14 @@ public void onCanceled() {
     MessageNotifier.notifyMessageDeliveryFailed(context, recipients, threadId);
   }
 
+  private String getMessageBody(String body) {
+    if (TextSecurePreferences.isUnicodeStrippingEnabled(context)) {
+      return UnicodeFilter.filter(body);
+    } else {
+      return body;
+    }
+  }
```
moxie0: still need to preserve the serialization signature

Rawi666: I'm not sure if we are on the same page.
Do you refer to "private static final long serialVersionUID = 0L;" inside UnicodeFilter ?
It's not needed, UnicodeFilter is not being serializer or deserialized and it's working perfectly fine as it is now (already checked).
Or do you refer to something else?

milankni:
Thanks guys for implementing unicode stripping.

Apart from issues mentioned earlier: even that I could type w/o diacritics, such words destroy the keyboard dictionary (not only for the particular user, but also for others because some big-brother companies use crowd sourced dictionaries).

eerozeteen:
What's about this function? Is it working? I didn't see any new options in Signal about "Use 7-bit GSM Alphabet". What should I do to make it possible?

pmav99:
Any updates on this one? Being able to use Signal for both encrypted/unencrypted messages is a really important point and double SMS charges are surely a break dealer for many people

Kamul-PL:
Hello!

I only wanted to admit, that I'm also looking forward for "unicode stripping" option. Are there any news in this topic? I look at all options in preferences on every Signal beta release, but unfortunately can't find this one, really important for me...

milankni:
Hello guys,
is there any progress on this topic (i.e. the possibility to write accented chars but strip (transliterate) those to plain 7-bit ASCII before sending the unencrypted SMS)?
Paying up to three times more for monthly SMS bill does not promote the use of Signal as messaging app replacement...

Rawi666:
I think this feature is not welcomed in Signal and therefore it stuck on pull request

milankni:
I see, it is a pity... While I can live with a separate app for SMS (supporting this feature), the rest of my family does not seem to bother. I'll check other options available if they fit better our use case.

Kamul-PL:
That's a pity that feature that is already done is not available to the users. Even if it is useful for many people and for sure it wouldn't bloat Signal... :/

josepedrocorreia:
Any updates? I just started using Signal and would like to use it as my SMS app too. Being Portuguese, there are a lot of non-ASCII characters in my messages, which limits their length substantially.

iriseecz:
Any updates? Will be implemented in the near future?

aeharake:
> 
> 
> Unicode stripping is very important and popular in countries where you use non-ascii characters. If you put at least 1 non-ascii character then sms will be converted to utf16 and you can put max 70 characters in single message. This results that the application displays false information how many messages will be really sent and you are charged for multiple messages from mobile provider. Therefore aware users use this functionality to strip non-ascii characters and pay less for 1 single message

Could you explain what does it really strip? 
I mean if I type my name in Arabic and send it through signal to my friend, and my friend opens the normal sms app, can they view my name? or will it be only visible through the signal app?

PR #5217, opened 2016-02-11
Exclude call screen from recents

cascheberg:
Opening Signal from recent apps list after a call is flawed.

Steps to reproduce:
- open a Signal contact in Android system contacts app
- click "Signal call" -> call screen shows up, Signal immediately starts the call
- end the call -> Signal disappears to the background
- open Signal from recent apps list

Problem: Signal opens in call screen, starts a new call to the last called contact

Another glitch: opening Signal from recent apps list after ending or rejecting an incoming call is also unusable

// FREEBIE


moxie0:
in 3.12.0

PR #5220, opened 2016-02-12
Improve missed call notifications

cascheberg:
Missed call notifications are always silent. This is great if the phone was actually ringing. But sometimes you completely miss a call because you are offline. A notification is shown once you are back online. In that case the notification should not be silent so that you have a chance to actually notice it.

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/redphone/RedPhoneService.java
+++ b/src/org/thoughtcrime/redphone/RedPhoneService.java
@@ -30,6 +30,7 @@ import android.support.annotation.Nullable;
 import android.telephony.TelephonyManager;
 import android.text.TextUtils;
 import android.util.Log;
+import android.util.Pair;
 
 import org.thoughtcrime.redphone.audio.IncomingRinger;
 import org.thoughtcrime.redphone.audio.OutgoingRinger;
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/redphone/audio/IncomingRinger.java
```diff
@@ -121,6 +121,15 @@ public void stop() {
...
   }
 
+  public boolean isPlaying() {
+    try {
+      if (player != null) return player.isPlaying();
+    } catch (IllegalStateException e) {
+      Log.w(TAG, e);
+    }
+    return false;
+  }
```
moxie0: Is this really necessary, shouldn't STATE_RINGING be enough?

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -101,11 +101,17 @@ public static void notifyMessageDeliveryFailed(Context context, Recipients recip
   }
 
   public static void updateNotification(@NonNull Context context, @Nullable MasterSecret masterSecret) {
+    updateNotification(context, masterSecret, false);
+  }
+
+  public static void updateNotification(@NonNull Context context,
+                                        @Nullable MasterSecret masterSecret,
+                                        boolean signal) {
```
moxie0: If there's one thing we don't need, it's another variation of this method.  Just specify the thread id instead.

src/org/thoughtcrime/redphone/RedPhoneService.java
```diff
@@ -259,7 +259,6 @@ private void handleAnswerCall(Intent intent) {
 
   private void handleDenyCall(Intent intent) {
     state = STATE_IDLE;
-    incomingRinger.stop();
```
moxie0: what is this doing in this change?

cascheberg:
@moxie0 Thanks, PR updated.

Explanation for the now static `signal` arguments instead of `incomingRinger.isPlaying()`:
These exceptions can be thrown while the state is `STATE_RINGING`:
- `SignalingException`, `SessionStaleException`, `LoginFailedException` and (maybe) `RuntimeException`

So while the state is `STATE_RINGING`:
1. `notifyServerFailure` , `notifyLoginFailed` and `notifyCallStale` are only called before the phone actually starts ringing.
2. `notifyCallDisconnected` is only called after the phone started ringing.
3. I don't know about `notifyClientFailure`, but I just used `false` to keep the code simple.

moxie0:
in 3.14.0

PR #5224, opened 2016-02-13
Added JDK as requirement for building

xrazvan:
// FREEBIE


moxie0:
thanks we try not to spell out version numbers, though, since they are going to change

PR #5228, opened 2016-02-15
Use high-res callcard for TransparentContactPhoto

haffenloher:
Apparently I overlooked that one in #4971.

// FREEBIE


moxie0:
why does this one need to be large?

haffenloher:
it may be displayed for a short time as a loading photo when starting a Signal call from the address book. The small version looks pixelated when displayed at full screen width.

moxie0:
but isn't this also what's displayed in the circle avatar while loading? does it still display correctly?

haffenloher:
I think `asDrawable` is responsible for returning the circle avatars. I only changed `asCallCard` which is only used in `redphone.ui.CallCard.setPersonInfo()`.

PR #5237, opened 2016-02-16
Rename single underscore variable names.

### LINKED ISSUE #5236: Remove single '_' identifiers from java source.
Rename the conventional ignored exception variable name in catch clauses where they exist.

i.e. prefer something like `try { ... } catch (TException ignored)  { }` to `try { ... } catch (TException _)  { }`

These result in compile-time warnings under Java 8, and will be compile-time errors in Java 9.

To reproduce:
 - Build Signal-Android using a Java 8 SDK [1]

[1] $ java -version
java version "1.8.0_74"
Java(TM) SE Runtime Environment (build 1.8.0_74-b02)
Java HotSpot(TM) 64-Bit Server VM (build 25.74-b02, mixed mode)

johnspurlock:
These are compiler warnings in Java8, and will be compiler errors
in Java9.

Closes #5236
// FREEBIE


moxie0:
this is from upstream code, which we prefer to leave as-is

PR #5240, opened 2016-02-16
swap statusbar color in dark theme

### LINKED ISSUE #5232: Status bar color in dark theme inconsistency
Status bar color is lighter than the conversation color. That apply on every color.
screenshot_20160215
I dislike this, and it's harder to use in low light.

Edit (copy/paste of a later message in the ticket)
Oh, and on Lollipop you usually get the same color, this is why i opened this bug in the first place.
I don't know what is meant by design, but Lollipop and Marshmallow are not consistent.
So it's your choice to assign the bug to one or the other.

Please consider every other message in the ticket as the main one.

agrajaghh:
fixes #5232

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/color/MaterialColor.java
+++ b/src/org/thoughtcrime/securesms/color/MaterialColor.java
@@ -54,8 +54,8 @@ public enum MaterialColor {
     this.serialized             = serialized;
   }
 
-  MaterialColor(int color500, int color900, int color700, String serialized) {
-    this(color500, color500, color700, color700, color700, color900, serialized);
+  MaterialColor(int lightColor, int darkColor, int semiDarkColor, String serialized) {
+    this(lightColor, lightColor, semiDarkColor, semiDarkColor, semiDarkColor, darkColor, serialized);
   }
 
   public int toConversationColor(@NonNull Context context) {
--- a/src/org/thoughtcrime/securesms/color/MaterialColor.java
+++ b/src/org/thoughtcrime/securesms/color/MaterialColor.java
@@ -54,8 +54,8 @@ public enum MaterialColor {
     this.serialized             = serialized;
   }
 
-  MaterialColor(int lightColor, int darkColor, int statusBarColor, String serialized) {
-    this(lightColor, lightColor, statusBarColor, darkColor, darkColor, statusBarColor, serialized);
+  MaterialColor(int color500, int color900, int color700, String serialized) {
+    this(color500, color500, color700, color700, color700, color900, serialized);
   }
 
   public int toConversationColor(@NonNull Context context) {
--- a/src/org/thoughtcrime/securesms/color/MaterialColor.java
+++ b/src/org/thoughtcrime/securesms/color/MaterialColor.java
@@ -7,26 +7,25 @@ import android.util.TypedValue;
 import org.thoughtcrime.securesms.R;
 
 public enum MaterialColor {
-
-  RED        (R.color.red_400,         R.color.red_900,         R.color.red_700,         "red"),
-  PINK       (R.color.pink_400,        R.color.pink_900,        R.color.pink_700,        "pink"),
-  PURPLE     (R.color.purple_400,      R.color.purple_900,      R.color.purple_700,      "purple"),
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/color/MaterialColor.java
```diff
@@ -54,8 +54,8 @@
     this.serialized             = serialized;
   }
 
-  MaterialColor(int lightColor, int darkColor, int statusBarColor, String serialized) {
-    this(lightColor, lightColor, statusBarColor, darkColor, darkColor, statusBarColor, serialized);
+  MaterialColor(int color500, int color900, int color700, String serialized) {
+    this(color500, color500, color700, color700, color700, color900, serialized);
```
moxie0: not down with the renaming, the class shouldn't be exposing these kinds of internals

agrajaghh:
changed the names

moxie0:
what we're deciding here is that the the status bar color is different in light and dark themes, so there should be two parameters for statusBarColor now -- one for light, one for dark.

agrajaghh:
okay, hope I did it right now...

src/org/thoughtcrime/securesms/color/MaterialColor.java
```diff
@@ -54,8 +53,11 @@
     this.serialized             = serialized;
   }
 
-  MaterialColor(int lightColor, int darkColor, int statusBarColor, String serialized) {
-    this(lightColor, lightColor, statusBarColor, darkColor, darkColor, statusBarColor, serialized);
+  MaterialColor(int lightThemeLightColor, int lightThemeDarkColor,
+                int darkThemeLightColor, int darkThemeDarkColor, String serialized)
```
moxie0: lightColor, darkColor, lightStatusBarColor, darkStatusBarColor

agrajaghh: I wanted to stick to the 7-argument constructor:
conversationColorLight, actionBarColorLight, statusBarColorLight, conversationColorDark, actionBarColorDark, statusBarColorDark

But I can change the order (and names) if you like

agrajaghh:
done

moxie0:
in 3.12 thanks!

RiseT:
@agrajaghh Does this commit swap the colors (status bar <-> action bar) in the recipient preferences as well?

agrajaghh:
@RiseT are you talking about this:?
screenshot_2016-02-19-11-15-04

RiseT:
@agrajaghh Exactly. So I guess this is a "yes" ;-) Just wanted to make sure.

RiseT:
@agrajaghh One off-topic question:
Where do you test you changes? I've found building with Android Studio surprisingly easy, but its Android emulator surprisingly slow (and I mean "slow" in the sense of "almost unusable" - even when using HAMX).
And how do I get beyond Signal's initial registration screen within the emulator? I cannot register the emulated Signal on the Signal server, I guess,,,?

agrajaghh:
@RiseT this sounds like a topic for the new [community forum](https://whispersystems.discoursehosting.net), but to answer your question shortly:
I normally use physical devices (my main phone if I think the change isn't going to break anything, and otherwise old devices)

Also I made good experiences with [genymotion emulator](https://www.genymotion.com/) where you are able to install google apps so you're able to register and use push notifications. (I got a AMD CPU, so I can't use HAMX, but genymotion uses hardware acceleration based on virtualbox so It's quite fast)

RiseT:
@agrajaghh Thanks for the info (and for the hint about the new community forum, of which I wasn't aware of at all. Such a forum - with the possibility of posting images etc. - was sorely needed)!

scienmind:
Although this PR made it more in line with design guidelines and the status bar looks better, The color swap has somewhat reduced the readability of the message bubbles in dark theme (due to brighter background).

I know moxie mentioned he's against darker colors, and is up only for the swap. 
But if more people think that "color swap only" reduces readability, maybe the use of darker colors should be reconsidered.

agrajaghh:
The light theme uses white text as well, and the background colors are way brighter (400 for light, and 700 for dark). I never had a readability problem in the light theme...

PR #5257, opened 2016-02-18
Add issue template

jtracey:
As documented here: https://help.github.com/articles/creating-an-issue-template-for-your-repository/
Github has new issue template functionality. Loosely based on the wiki page linked within the template.

// FREEBIE


agrajaghh:
nice! maybe add this link as well?

https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

2-4601:
I think this clutters the root directory and would be better in the `./github` directory.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
+Please note this is a bug tracker, not a support forum. After you have read the [Submitting useful bug reports guide](https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports), delete this line. Also delete any sections that aren't applicable.
```
2-4601: This line would be better if it was wrapped as a comment within `<!-- -->`

Also because the template will be displayed in edit mode (and many people will never preview) I think all links should be raw to make them easily copy-pasteable. Also preferably on their own separate line.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+
+### Bug description
```
2-4601: I think template list would be nice here, e.g.
```
### Steps to reproduce
- add your
- steps here
- as a list
- using hyphens
```

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+
+### Steps to reproduce
```
2-4601: I think template list would be nice here, e.g.
```
### Steps to reproduce
- add your
- steps here
- as a list
- using hyphens
```

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+
+### Screenshots
```
2-4601: Perhaps a tip here would be helpful:
`<!-- You can drag and drop images here -->`

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+
+### Device info
+ **Model:** 
+ 
+ **Android version:** 
+ 
+ **Signal version:** 
+
```
2-4601: Might be helpful to populate these with examples, such as:
```
### Device info
**Model:** Nexus 4

**Android version:** 5.1.1

**Signal version:** 3.11.1
```

agrajaghh: problem is people might leave them just in, and then we don't know if its real or not. So there should be some fake devices and version numbers I guess

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+### Screenshots
+
+### Device info
+ **Model:** 
+ 
+ **Android version:** 
+ 
+ **Signal version:** 
+ 
+### App state
```
2-4601: This could benefit from check-boxes. For example:
```
### App state
- [ ] Signal messages and calls registered
- [ ] SMS/MMS disabled
- [ ] Dark theme
- [ ] Light theme
```
Hmm... this might also be too much Markdown for some people. I guess another option is just a list. I'm also not sure how relevant the app state is to majority of issues. I suppose it was more important when encrypted SMS was supported?

ISSUE_TEMPLATE.md
2-4601: :+1: Very good point!

ISSUE_TEMPLATE.md
jtracey: Yeah, I actually originally had this as a comment, but decided to make it default to visible as a small shibboleth showing they actually read it. Check-boxes work just about as well though.

jtracey:
I had put it in the root directory since that's where CONTRIBUTING is, but I can see why that should be more visible.

I ended up just removing the "App state" section, since as @2-4601 points out, it's not typically used, and I figure if it is relevant people will include it in the reproducing steps anyway.

PR #5274, opened 2016-02-21
Clarify "SMS import complete" notification text

### LINKED ISSUE #5262: Does the string "Signal system database import" make any sense...?
The recent commit https://github.com/WhisperSystems/Signal-Android/commit/25d7931358658764b99881d8a507ebd806dc53ef uses the string **"Signal system database import is complete."** (I am aware that @haffenloher just made the already existing string translatable and replaced "TextSecure" with "Signal" - so not pointing fingers).

**"Signal system database"** doesn't make any sense to me, i. e. having "Signal" and "system" in the same expression imho contradicts itself.

There are
a) Signal's (encrypted) database and
b) Android's central database for SMS/MMS, which is used by other messaging apps.

What this string means to describe is the process of importing from (b) to (a), i. e. **from the system database to Signal's database**.

So this really should read **"System database import is complete."**, correct?

While I'm at it, I'd rather suggest **"Import of the system's SMS/MMS database has been completed."** (this assumes that this string is the text body of a dialog - more space for more detailed info available - while "Import complete" is the dialog's title) or **"Import of system database completed."** if something shorter is needed.

haffenloher:
#### Contributor checklist

 * Emulator, Android 5.1

----------

#### Description

Fixes #5262

// FREEBIE

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -595,7 +595,7 @@
     <string name="ApplicationMigrationService_import_in_progress">Import in progress</string>
     <string name="ApplicationMigrationService_importing_text_messages">Importing text messages</string>
     <string name="ApplicationMigrationService_import_complete">Import complete</string>
-    <string name="ApplicationMigrationService_Signal_system_database_import_is_complete">Signal system database import is complete.</string>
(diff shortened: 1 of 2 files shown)


moxie0:
in 3.14.0

PR #5285, opened 2016-02-24
Update issue and pull request templates

2-4601:
### Contributor checklist
- [x] I have tested my contribution on these browsers:
 * Firefox 44.0.2
 * Chrome 48.0.2564.116 

----------

### Description
**Issue template**
- add link to the discussion forum to redirect discussion there
- don't encourage people to add raw links for screenshots, they are just an additional hoop compared to inline images and might stop working after some time
- improve structure (the preview below is inaccurate, I've tested this with the actual issue preview and it works as planned)

**PR template**
- add link to an explanation for BitHub rewards (this has caused confusion before)

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,7 +1,7 @@
 <!--
-Please note that this is an issue tracker
+Please note that this is an issue tracker.
 
-For support go to the Support Center:
+For support go to the support center:
 http://support.whispersystems.org/
 or email support@whispersystems.org
 
@@ -22,7 +22,7 @@ I have:
 ### Steps to reproduce
 - add your steps here
 - as a list
-- using hyphens
+- using hyphens as bullet points
 
 ### Screenshots
 <!-- you can drag and drop images here -->
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -38,4 +38,5 @@ I have:
 **Signal version:** 0.0.0
 
 ### Link to debug log
+<!-- immediately after the bug has happened capture a debug log via Signal's advanced settings and paste the link below -->
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -11,7 +11,7 @@ https://whispersystems.discoursehosting.net
 Delete any sections that aren't relevant.
 -->
 
-<!-- This is relevant: mark with x between the [ ] to show you have understood -->
+<!-- This is relevant: replace the empty checkboxes [ ] below with checked ones [x] accordingly -->
 I have:
 - [ ] searched open and closed issues for duplicates
 - [ ] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,38 +1,40 @@
-<!--
-Please note that this is an issue tracker.
+<!-- This is a bug report template. By following the instructions below and filling out the sections with your information, you will help the developers to get all the necessary data to fix your issue.
+You can also preview your report before submitting it. You may remove sections that aren't relevant to your particular case.
 
-For support go to the support center:
+Before we begin, please note that this tracker is meant for issues only.
+
+If you are looking for support, please see our support center instead:
 http://support.whispersystems.org/
 or email support@whispersystems.org
 
-For discussion go to the community forum:
+If you wish to ask questions or start a discussion, please go to the community forum instead:
 https://whispersystems.discoursehosting.net
 
-Delete any sections that aren't relevant.
--->
+Let's begin with a checklist: replace the empty checkboxes [ ] below with checked ones [x] accordingly -->
 
-<!-- This is relevant: replace the empty checkboxes [ ] below with checked ones [x] accordingly -->
 I have:
 - [ ] searched open and closed issues for duplicates
 - [ ] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports
 
-### Bug description
+----------------------------------------
 
+### Bug description
+Describe here the issue that you are experiencing.
 
 ### Steps to reproduce
 - using hyphens as bullet points
 - list the steps
 - that reproduce the bug
 
-**Actual result:** Describe here what happens after you run the above steps (i.e. the buggy behaviour)
-**Expected result:** Describe here what should happen after you run the above steps (i.e. what would be the correct behaviour)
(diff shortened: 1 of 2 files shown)


McLoo:
...and no spaces inside the brackets pls

I have:
- [ x] space to the left
- [x ] space to the right
- [ x ] spaces all over
- [x] left out all the nasty little spacccessss

2-4601:
True, but
`- [] no spaces`
will render as
- [] no spaces

While
`- [ ] space`
renders as
- [ ] space

So the space is there to render an empty checkbox :)

2-4601:
Made the checkbox stuff hopefully more clear now.

cascheberg:
Two more ideas:
* For the issue templates steps to reproduce: explicitly ask for the **actual result** and the **expected result**
* For the PR template description: ask **what** the PR fixes and **how** it has been tested

2-4601:
> For the issue templates steps to reproduce: explicitly ask for the **actual result** and the **expected result**

I've also thought about adding actual and expected results before because they seem to be repeated in many bug report guides. But because I couldn't think of a simple or non-clunky way to add them I just let it be. The original [Submitting useful bug reports](https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports/f79faa4a2d6c16c96324e958391142e6309b6507) (by mcginty) didn't have them and I can't remember if there has ever been a huge need for them at this repository. But I'm not against adding them. Where and how would you add them?

Here's one example (but feels kinda clunky to me):

----------

### Steps to reproduce
- add your steps here
- as a list
- using hyphens as bullet points
- **actual result:** final point being the actual result or buggy behaviour

**Expected result**: describe what you expected to happen here

----------

> For the PR template description: ask **what** the PR fixes

Isn't _"**what** the PR fixes"_ already answered in the commit message (which GitHub copies to the title and end of the description)? I guess we could add a checklist item for writing good commit messages (like Signal Desktop has :)

> and **how** it has been tested

This would probably go after the checklists (i.e. below the **Description** heading). Or how did you envision this?

Thanks for the comments :+1:

2-4601:
> ask **what** the PR fixes and **how** it has been tested

Addressed those.
Didn't add actual and expected results because couldn't figure out a nice way to add them.

cascheberg:
Great, I like that it is all in a comment. That way you do not have to remove the text first.
How about this then:

---
### Steps to reproduce
< !--
Add instructions in the form of an ordered list on how to reproduce the issue.
Also include the actual result that you see, as well as the result that you would have expected instead.
-->
1.
2.
3.

---

2-4601:
> Add instructions in the form of an ordered list on how to reproduce the issue.

I've deliberately chosen **unordered** lists because ordered lists are too cumbersome :)
If you have made a ten step ordered list, and you want to add an additional first step you'll need to change ten numbers! One solution is to use Markdown's ordered list syntax where all the numbers are the same, i.e
1.
1.
1.
But that's a bit too unintuitive and an extra hoop (learn Markdown) for the majority of users who report bugs. For pull requests the ordered list syntax might fly because the crowd is different and it's expected that they know or can grasp the syntax immediately because they are accustomed to the subject.
So this is why unordered list makes more sense. It's how everyone is used to making lists in their daily lives. And if you want to add a step in the middle you don't have to change the following steps.

> Also include the actual result that you see, as well as the result that you would have expected instead.

Heh, I like that because it makes the reporter figure out how to express it ;)
But I guess that's a bit _do as we say, not as we do_ kinda thing. I mean if we say what the reporter should do, shouldn't we also be able to tell them how to do it and give out the example format? But what is a good format for that?

cascheberg:
Unordered list is fine also ;)

Not sure what you mean regarding the rest. I think a reporter would be able to find a suitable format for how to describe results (maybe even put the text about the "ordered / unordered list" in a way that it is more of a suggestion).

---

Add instructions on how to reproduce the issue, for example in the form of a list using hyphens as bullet points.
Additionally in a suitable manner describe the actual result that you see, as well as the result that you would have expected instead.

2-4601:
I meant that an example which the reporter can simply just follow and replace is more convenient than a description or suggestion of what the reporter should do. I'm looking for a basic template format for it.

2-4601:
Here's what I came up with (still not that satisfying :):

----------

### Steps to reproduce
- using hyphens as bullet points
- add the steps 
- that reproduce the bug

**Actual result:** describe what happens when you run the above steps (i.e. the buggy behaviour)
**Expected result:** describe what should happen when you run the above steps (i.e. what would be the correct behaviour)

----------

Adding @agrajaghh, @haffenloher also to the participants (and anyone else who's blessed their retinas with lots of bug reports at this repository :cold_sweat:). Do you think adding actual and expected results are needed or is it enough to have a bug description and a simple list of steps to reproduce?

agrajaghh:
looks good to me like that! Some people are going to write bad bug reports anyay (and even delete the template)...  

> Adding @agrajaghh, @haffenloher also to the participants (and anyone else who's blessed their retinas with lots of bug reports at this repository :cold_sweat:)

Especially your bug reports are gold :blush: Although I have the feeling they were better in the past :stuck_out_tongue_winking_eye:

2-4601:
Added.

> Although I have the feeling they were better in the past :stuck_out_tongue_winking_eye: 

Truth. I paid for someone else to do them before but have since run out of money :)

cascheberg:
@2-4601 
> Added

Missing newline before "Expected result"?

Another idea regarding debug log: add a comment like "should be created right after the issue occurred"

2-4601:
> Missing newline before "Expected result"?

Are you referring to the fact that the commit preview shows the actual and expected result in one line? If so that preview is inaccurate.

> Another idea regarding debug log: add a comment like "should be created right after the issue occurred"

It's in the _"How to submit useful bug reports"_ article but yes it's a good idea to repeat it here.

`<!-- immediately after the bug has happened capture a debug log via Signal's advanced settings and paste the link below -->`

McLoo:
+ unlock screen
+ pseudo underline 
```<!-- __immediately__  after the bug has happened capture a debug log via Signal's advanced settings or unlock screen and paste the link below -->```

2-4601:
`__immediately__` is a bit too nerdy don't you think? :neckbeard: :smile: 
All caps `IMMEDIATELY` looks more friendly (if emphasizing is needed)

Hmm... what's this _unlock screen_ thing?

McLoo:
I didn't want to shout 😰
and it seems I dropped my vocabulary somewhere 
screenshot_2016-03-11-07-34-53-01

2-4601:
Oh yes, of course that screen! Well, I think that's somewhat a rare case so maybe leave it out of the issue template comment. But it definitely should be in the wiki article (it now is), thanks!

2-4601:
Debug log comment is now in there. No emphasizing style used for now, good middle ground between nerdy underscores and shouting :)

PR #5286, opened 2016-02-24
add notification stacking to display message text properly on wearables

jonhester:
### First time contributor checklist

- [x] I have read [how to contribute](https://github.com/WhisperSystems/Signal-Android/blob/master/CONTRIBUTING.md) to this project
- [x] I have signed the [Contributor Licence Agreement](https://whispersystems.org/cla/)

### Contributor checklist

- [x] I am following the [Code Style Guidelines](https://github.com/WhisperSystems/Signal-Android/wiki/Code-Style-Guidelines)
- [x] I have tested my contribution on these devices:
 * Nexus 6P, Android 6.0.1
 * Pebble Time, v3.9.2
- [x] My contribution is fully baked and ready to be merged as is
- [x] I have made the choice whether I want the Bithub reward or not by omitting or adding the word `FREEBIE` in my commit message

----------

### Description

Just switched from Telegram and love Signal, but hated that notifications didn't work properly on my wearable, so I decided to take a stab at it. I implemented my fix based on [Stacking Notifications](http://developer.android.com/training/wearables/notifications/stacks.html).

- Summary notification always uses the same id
- the individual stacked notifications use an `AtomicInteger` to generate unique notification IDs
- all intents had to be changed to so that they are distinct ([see here](http://developer.android.com/intl/pt-br/reference/android/app/PendingIntent.html#getActivity(android.content.Context, int, android.content.Intent, int))), so that a reply sent from a wearable will go to the right person
- the `markAsRead` intent had to be modified so that it could mark individual messages as read from a wearable notification, as well as marking all as read with the already existing behavior
- tested with a Pebble, but it uses Android Wear, so it effectively works the same way interacting with notifications
- fix for #2525

Issue #2525, opened 2015-02-25
1 card per message in Android Wear


maxauvy:
When I receive a message in TS, a notification is sent to the smartwatch, with the content of the message.

If I receive another SMS from the same contact, the content of the second message is added to the first one, so that's OK.

But if I receive another message from a different contact, TS shows something like "X messages", with only the name of each contact. So I can't read the content on the watch, I have to take out my phone and open TS.

Is it possible to add a feature with something like 1 notification / card per contact ?

If I can help, don't hesitate :)

makubi:
Hi,

@dr03lf and I are currently working on this.

z0l2k:
Is this still being worked on? Its the main bug preventing me from using Signal as my main sms app.

agrajaghh:
@z0l2k Unfortunately bumping or +1's don't help fixing this issue (they just create extra email notifications for the devs). Please only post if you have new information that will help to solve the issue. Thanks

lolstarz:
I'm happy to take this up, since it affects me too :). I propose a notification preference to disable the "summary notification" -- does that sound acceptable?

2-4601:
@lolstarz Thanks for your let's fix this attitude :)
You might want to check out #4488 and see why that's on hold.
Also https://github.com/WhisperSystems/Signal-Android/blob/master/CONTRIBUTING.md has some info on contributions.

lolstarz:
Thanks for the link, I have been searching through issues for dupilicates/related tickets but not having much success. I see and appreciate the "don't add more options" strategy in the contrib.md. I will use the code from that PR for now on my device and see if in anything changes wrt strategy to fix the issue.

jpph:
Please notice the smssecure which is forked from TextSecure doesn't have this issue. You might want to check how they fixed this. For now I have replaced ts with smssecure as default sms app. Works very well with my pebble

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #5288, opened 2016-02-24
add notification stacking to display message text properly on wearables

jonhester:

### Contributor checklist

 * Nexus 6P, Android 6.0.1
 * Pebble Time, v3.9.2

----------

### Description

Just switched from Telegram and love Signal, but hated that notifications didn't work properly on my wearable, so I decided to take a stab at it. I implemented my fix based on [Stacking Notifications](http://developer.android.com/training/wearables/notifications/stacks.html).

- Summary notification always uses the same id
- the individual stacked notifications use a static variable to generate unique notification IDs
- all intents had to be changed to so that they are distinct ([see here](http://developer.android.com/intl/pt-br/reference/android/app/PendingIntent.html#getActivity(android.content.Context, int, android.content.Intent, int))), so that a reply sent from a wearable will go to the right person
- the `markAsRead` intent had to be modified so that it could mark individual messages as read from a wearable notification, as well as marking all as read with the already existing behavior
- tested with a Pebble, but it uses Android Wear, so it effectively works the same way interacting with notifications
- revised #5286. Guess it maybe wasn't the greatest idea to submit a pull request hours after I should be in bed. All is well now.
- fix for #2525 

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -199,7 +199,7 @@ public class MessageNotifier {
         }
 
         if (notificationState.hasMultipleThreads()) {
-          sendMultipleThreadNotification(context, notificationState, signal, item);
+          sendMultipleThreadNotification(context, notificationState, signal);
         } else {
           sendSingleThreadNotification(context, masterSecret, notificationState, signal, item, true);
         }
@@ -273,16 +273,15 @@ public class MessageNotifier {
 
   private static void sendMultipleThreadNotification(@NonNull  Context context,
                                                      @NonNull  NotificationState notificationState,
-                                                     boolean signal,
-                                                     NotificationItem notificationItem)
+                                                     boolean signal)
   {
     MultipleRecipientNotificationBuilder builder       = new MultipleRecipientNotificationBuilder(context, TextSecurePreferences.getNotificationPrivacy(context));
     List<NotificationItem>               notifications = notificationState.getNotifications();
 
     builder.setMessageCount(notificationState.getMessageCount(), notificationState.getThreadCount());
-    builder.setMostRecentSender(notificationItem.getIndividualRecipient());
+    builder.setMostRecentSender(notifications.get(0).getIndividualRecipient());
 
-    long timestamp = notificationItem.getTimestamp();
+    long timestamp = notifications.get(0).getTimestamp();
     if (timestamp != 0) builder.setWhen(timestamp);
 
     builder.addActions(notificationState.getMarkAsReadIntent(context, SUMMARY_NOTIFICATION_ID, null));
@@ -296,7 +295,7 @@ public class MessageNotifier {
 
     if (signal) {
       builder.setAlarms(notificationState.getRingtone(), notificationState.getVibrate());
-      builder.setTicker(notificationItem.getText());
+      builder.setTicker(notifications.get(0).getText());
     }
 
     ((NotificationManager)context.getSystemService(Context.NOTIFICATION_SERVICE))
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -244,7 +244,7 @@ public class MessageNotifier {
     if (signal) {
       builder.setAlarms(notificationState.getRingtone(), notificationState.getVibrate());
       builder.setTicker(notifications.get(0).getIndividualRecipient(),
-          notifications.get(0).getText());
+                        notifications.get(0).getText());
     }
 
     ((NotificationManager)context.getSystemService(Context.NOTIFICATION_SERVICE))
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -60,6 +60,7 @@ import org.whispersystems.textsecure.api.messages.TextSecureEnvelope;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import me.leolin.shortcutbadger.ShortcutBadger;
 
@@ -75,11 +76,15 @@ public class MessageNotifier {
 
   private static final String TAG = MessageNotifier.class.getSimpleName();
 
-  public static final int NOTIFICATION_ID = 1338;
+  private static AtomicInteger notificationCounter = new AtomicInteger(0);
+
+  public static final int SUMMARY_NOTIFICATION_ID = 1338;
 
   private volatile static long visibleThread = -1;
 
   public static final String EXTRA_VOICE_REPLY = "extra_voice_reply";
+
+  public final static String GROUP_KEY_MESSAGES = "group_key_messages";
 
   public static void setVisibleThread(long threadId) {
     visibleThread = threadId;
@@ -169,7 +174,7 @@ public class MessageNotifier {
           (pushCursor == null || pushCursor.isAfterLast()))
       {
         ((NotificationManager)context.getSystemService(Context.NOTIFICATION_SERVICE))
-          .cancel(NOTIFICATION_ID);
+          .cancel(SUMMARY_NOTIFICATION_ID);
         updateBadge(context, 0);
         clearReminder(context);
         return;
@@ -181,10 +186,10 @@ public class MessageNotifier {
         appendPushNotificationState(context, notificationState, pushCursor);
       }
 
-      if (notificationState.hasMultipleThreads()) {
+      sendSingleThreadNotification(context, masterSecret, notificationState, signal);
+
+      if (notificationState.getMessageCount() > 1) {
         sendMultipleThreadNotification(context, notificationState, signal);
-      } else {
-        sendSingleThreadNotification(context, masterSecret, notificationState, signal);
       }
 
       updateBadge(context, notificationState.getMessageCount());
@@ -203,9 +208,11 @@ public class MessageNotifier {
                                                    @NonNull  NotificationState notificationState,
                                                    boolean signal)
   {
+    int notificationId = notificationCounter.getAndIncrement();
+
     if (notificationState.getNotifications().isEmpty()) {
       ((NotificationManager)context.getSystemService(Context.NOTIFICATION_SERVICE))
-          .cancel(NOTIFICATION_ID);
+          .cancelAll();
       return;
     }
 
@@ -223,9 +230,9 @@ public class MessageNotifier {
     if (timestamp != 0) builder.setWhen(timestamp);
 
     builder.addActions(masterSecret,
-                       notificationState.getMarkAsReadIntent(context),
-                       notificationState.getQuickReplyIntent(context, notifications.get(0).getRecipients()),
-                       notificationState.getWearableReplyIntent(context, notifications.get(0).getRecipients()));
+                       notificationState.getMarkAsReadIntent(context, notificationId, notifications.get(0).getThreadId()),
+                       notificationState.getQuickReplyIntent(context, notifications.get(0).getRecipients(), notificationId),
+                       notificationState.getWearableReplyIntent(context, notifications.get(0).getRecipients(), notificationId));
 
     ListIterator<NotificationItem> iterator = notifications.listIterator(notifications.size());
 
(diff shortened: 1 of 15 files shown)


src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -252,27 +259,28 @@ private static void sendMultipleThreadNotification(@NonNull  Context context,
...
 
-    builder.addActions(notificationState.getMarkAsReadIntent(context));
+    builder.addActions(notificationState.getMarkAsReadIntent(context, SUMMARY_NOTIFICATION_ID, null));
 
-    ListIterator<NotificationItem> iterator = notifications.listIterator(notifications.size());
+    ListIterator<NotificationItem> iterator = notifications.listIterator();
 
-    while(iterator.hasPrevious()) {
-      NotificationItem item = iterator.previous();
+    while(iterator.hasNext()) {
```
jonhester: So the messages are shown in reverse chronological order. Most recent first, which is more logical for a messaging app

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -237,11 +244,11 @@ private static void sendSingleThreadNotification(@NonNull  Context context,
     if (signal) {
       builder.setAlarms(notificationState.getRingtone(), notificationState.getVibrate());
       builder.setTicker(notifications.get(0).getIndividualRecipient(),
-                        notifications.get(0).getText());
+          notifications.get(0).getText());
```
agrajaghh: unintended align change?

jonhester: Yep, I'll fix. So used to just working in a text editor I am apparently terrible when using an IDE.

McLoo: It's definitely the IDE that behaves terrible. Not you (or me :grimacing:)

jonhester: Thanks for pointing out my mistake.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -75,12 +76,16 @@
 
   private static final String TAG = MessageNotifier.class.getSimpleName();
 
-  public static final int NOTIFICATION_ID = 1338;
+  private static AtomicInteger notificationCounter = new AtomicInteger(0);
```
moxie0: Why is this an AtomicInteger? Is there something that needs to interact with this off the UI thread?

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -181,10 +186,10 @@ private static void updateNotification(@NonNull  Context context,
         appendPushNotificationState(context, notificationState, pushCursor);
       }
 
-      if (notificationState.hasMultipleThreads()) {
+      sendSingleThreadNotification(context, masterSecret, notificationState, signal);
```
moxie0: For better or worse, this class is currently idempotent.  Changing the behavior this way breaks that, which a lot of other interactions depend on a very subtle ways.  This change would need to be done in a way that preserves that contract.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
moxie0: Please confine PRs to one logical change.  If you'd like to change the way the summary notification is ordered, that should go in a separate PR, but for the record that's not something we'd accept.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
jonhester: Just an easy way to get unique IDs for the single notifications. The existing behavior only uses one notification id for single thread and multiple thread notifications. Using separate notifications is the only way to stack messages so it works properly on wearables.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
jonhester: Any ideas on this? Or examples of where this could cause an issue?

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
jonhester: Makes sense. I can revert that change so the summary is exactly the same as before. Only reason I went down that path at all was that stacking the notifications loses the message text that was currently there if there is only a single thread, so it's still visible on popping up.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
moxie0: I get that you want a counter, why is Atomic?

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
moxie0: I'm not sure what you mean, it just needs to be idempotent.  Every time this code is called, it needs to tear down and create notifications that reflect the current state of the database.  It can be called 100 times in a row with the exact same message state, or it could be called once after 10 messages have been read or unread.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
jonhester: Good point. No reason it can't just be an int. I'm a developer by trade, but while I have done some Java development, I am not a Java developer. I'll make that change too. I'll update the pull request when I finish work today.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
jonhester: @moxie0 Thanks, that answers my question. But that's the entire problem. Makes total sense when there is only one notification, but you need to have the stack. If you recreate all the notifications from the notification state then you would issue all the notifications again. I guess what I'm saying is that I don't see how it's possible to do achieve while implementing this: http://developer.android.com/training/wearables/notifications/stacks.html.

jonhester: Only thing I can think of off the top of my head, is keeping track of the notification status (has it been issued already) and id in the notification state. And then checking against each one to see if it has already been issued or not.

jonhester:
@moxie0 Thoughts? 

@moxie0, took a crack at persisting whether or not notifications had already been displayed. I'm doing this by saving the NotificationState to a static variable when a new one is built. This lets me compare the notifications in the old state with those in the state and set `true` for any that have already been displayed. This way it doesn't need to be persisted to the messages db.

I'm also using the Single Thread notification as the groupSummary when there is only a single thread so notifications should show up the same as they did before.

image

image

image

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -75,12 +75,18 @@
 
   private static final String TAG = MessageNotifier.class.getSimpleName();
 
-  public static final int NOTIFICATION_ID = 1338;
+  public static final int SUMMARY_NOTIFICATION_ID = 1338;
+
+  private static int notificationCounter = 2000;
```
jonhester: Start at 2000 instead of 0 since it doesn't take much more than 1000 message until we get a conflict with `1338`.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -181,10 +187,22 @@ private static void updateNotification(@NonNull  Context context,
...
 
-      if (notificationState.hasMultipleThreads()) {
-        sendMultipleThreadNotification(context, notificationState, signal);
-      } else {
-        sendSingleThreadNotification(context, masterSecret, notificationState, signal);
+      List<NotificationItem> notifications = notificationState.getNotifications();
+
+      ListIterator<NotificationItem> iterator = notifications.listIterator();
+
+      while(iterator.hasNext()) {
```
jonhester: Since it is now multiple notifications instead of just one that is getting replaced, iterate through all the notifications to see what needs to happen

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -181,10 +187,22 @@ private static void updateNotification(@NonNull  Context context,
...
+        NotificationItem item = iterator.next();
+
+        if (!item.isNotified()) {
+          sendSingleThreadNotification(context, masterSecret, notificationState, signal, item, false);
```
jonhester: Haven't yet displayed a notification to the user for this item so let's do it

jonhester: This always needs to be sent once and only once so it displays on works properly on the wearable

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -181,10 +187,22 @@ private static void updateNotification(@NonNull  Context context,
...
+        }
+
+        if (notificationState.hasMultipleThreads()) {
```
jonhester: Send a groupSummary notification to ensure it displays on the device

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -181,10 +187,22 @@ private static void updateNotification(@NonNull  Context context,
...
+          sendMultipleThreadNotification(context, notificationState, signal, item);
```
jonhester: multiple threads so send a multipleThreads notification. This is all that will display on the device. Even if a single notification was just sent for an item that had not yet been displayed to the user, this still will be the only notification shown on the device.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -181,10 +187,22 @@ private static void updateNotification(@NonNull  Context context,
...
+        } else {
+          sendSingleThreadNotification(context, masterSecret, notificationState, signal, item, true);
```
jonhester: Same as multiple threads block, even if even if we just sent a single message notification to the user this groupSummary message is all that they will see.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -201,63 +219,73 @@ private static void updateNotification(@NonNull  Context context,
   private static void sendSingleThreadNotification(@NonNull  Context context,
                                                    @Nullable MasterSecret masterSecret,
                                                    @NonNull  NotificationState notificationState,
-                                                   boolean signal)
+                                                   boolean signal,
+                                                   NotificationItem notificationItem,
+                                                   boolean summary)
   {
+    int notificationId = (summary) ? SUMMARY_NOTIFICATION_ID : notificationCounter++;
```
jonhester: Make sure we use `SUMMARY_NOTIFICATION_ID` constant if this is a summary message so we don't end up with duplicate summary notifications displayed on the device.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
jonhester: Pass in the NotificationItem to make sure the correct one is display. Added to ensure idempotency so a non summary notification can only be sent one time per message.

jonhester:
I realized that not only did I not need to pass the notification into the multiThread builder, but doing so caused me the wrong information to be set for ticker, time, and most recent text. Taken care of in 2579f77

jonhester:
Wearable will now only show the most recent message text instead of all messages in a thread as it does on the phone. Also disabled alarm/ticker on non-summary messages since they are only displayed on the wearable and should not make a sound or ticker on the phone.

src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
moxie0: Please no comments: https://github.com/WhisperSystems/Signal-Android/wiki/Code-Style-Guidelines#comments

If you need to comment something, either here or in the code directly, it should be a sign that the code itself is unclear.

moxie0:
This looks like it depends on the app staying in memory for as long as the notifications are displayed?

@jonhester it looks like this might be you first experience developing for Android or with Java.  I'd recommend that you start with some smaller tasks to get a feel for the project, how Android works, and what the normal Java conventions are.  This is kind of a complicated thing to take on as your first PR to a project, and I think it's unlikely to get merged before you've spent some more time around the code.

jonhester:
@moxie0 Correct, it would have to stay in memory for this to work. Ideally it would be persisted in the message store, but it's not worth the time for me to go down that path if it isn't something that is going to be eventually merged. 

The current notification system in the app makes this update more complex than would be ideal. The app is not useful to me without this feature since I only use my watch for notifications. Basically, I did this for purely selfish reasons and am happy to just run my fork until this feature gets implemented. However, I'm also happy to update this based on feedback or for someone else to take over. Otherwise, feel free to close.

Correct, it would have to stay in memory for this to work. Ideally it would be persisted in the message store, but it's not worth the time for me to go down that path if it isn't something that is going to be eventually merged.

Finally, decided to mark this as freebie.

jonhester:
DB persistence and switched to NotifcationManagerCompat to make notification stacking play nice with older Android versions.

jonhester:
I'm gonna pull this. I'm not happy with the way this is implemented. If I figure out a better way, then I will revisit this issue.

PR #5320, opened 2016-03-06
Don't disable the proximity sensor in horizontal orientation

### LINKED ISSUE #4813: Proximity sensor disabled in horizontal orientation during signal calls
During signal call, if the phone is placed horizontally on it's back (i.e. flat on the desk with screen facing up), then the proximity sensor will not turn off the screen.

If the screen was already off before reaching this position, it'll remain off, and will be turned on while sensor clears, but if covered again, screen will not turn off.

This causes accidental clicks if you call while laying on the couch for example.
Original android dialer listens to the proximity sensor in all orientations, and I think this is the correct behavior to implement.

Tested with several devices, Signal 3.6.1.

This may be one of the reasons for: #4434 ,if you prefer, I can close this and add this info there.
(also may be relevant: #4519 and it's PR: #4522)

haffenloher:
### Contributor checklist

 * Mediatek 6589M phone with Android 4.2.2 

----------

### Description

Fixes #4813

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/redphone/call/LockManager.java
+++ b/src/org/thoughtcrime/redphone/call/LockManager.java
@@ -20,12 +20,7 @@ public class LockManager {
   private final PowerManager.WakeLock        partialLock;
   private final WifiManager.WifiLock         wifiLock;
   private final ProximityLock                proximityLock;
-
-  private final AccelerometerListener accelerometerListener;
-  private final boolean               wifiLockEnforced;
-
-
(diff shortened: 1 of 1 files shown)


moxie0:
I didn't write this originally, but I remember there was some incredibly convoluted device-specific reason why this had to be like it is.

haffenloher:
Okay. If someone still knows which devices this was for, we could maybe limit it to those?
But it's of course no urgent problem.

PR #5341, opened 2016-03-14
allow "panic button" apps to lock

eighthave:

### Contributor checklist
 * Samsung Galaxy Note 2, Android 4.4.4
 * Asus ZenFone2, Android 4.4.2 (x86)
 * its [included in SMSSecure](https://github.com/SMSSecure/SMSSecure/pulls?q=is%3Apr+is%3Aclosed+author%3Aeighthave) official releases, so its been tested on a bunch of devices there

----------

### Description

As discussed back in October face to face with @moxie0 and @mcginty, here is the "panic button" support.  We've been developing "panic button" features in our apps for years, and now we are working to make a more flexible and open ecosystem of apps that can both trigger other apps, or respond to triggers.  This has now become a partnership effort called [Panic Initiative](https://github.com/PanicInitiative) that has been spun out of Amnesty International's work in conjunction with iilab, Guardian Project and Security First.

Since Signal already includes a "lock" feature, it is a natural panic responder app. This setup allows for easy configuration of a single trigger action (e.g. button press) which can then trigger multiple apps. The first commit adds support for receiving the panic trigger. The second commit wipes Signal from the Recent Apps after responding to the panic trigger.

You can see an example of this user experience in this video:
https://www.youtube.com/watch?v=mS1gstS6YS8

The first panic button app that sends this kind of trigger is called Ripple, and it is available here:
https://github.com/guardianproject/ripple
https://play.google.com/store/apps/details?id=info.guardianproject.ripple

Amnesty Panic Button supports this in the beta version (v1.4.1) right now:
https://panicbutton.io/
https://play.google.com/store/apps/details?id=org.iilab.pb

More info on the panic kit work here:
https://dev.guardianproject.info/projects/panic/wiki

The source repo for the included TrustedIntents library is here:
https://github.com/guardianproject/TrustedIntents

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -373,6 +373,9 @@
         </intent-filter>
     </activity>
 
+    <activity android:name=".ExitActivity"
+              android:theme="@android:style/Theme.NoDisplay" />
+
     <service android:enabled="true" android:name="org.thoughtcrime.redphone.RedPhoneService"/>
 
     <service android:enabled="true" android:name=".service.ApplicationMigrationService"/>
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -363,6 +363,16 @@
 
     <activity android:name="com.soundcloud.android.crop.CropImageActivity" />
 
+    <!-- this can never have launchMode singleTask or singleInstance! -->
+    <activity android:name=".PanicResponderActivity"
+              android:noHistory="true"
+              android:theme="@android:style/Theme.NoDisplay">
+        <intent-filter>
+            <action android:name="info.guardianproject.panic.action.TRIGGER" />
+            <category android:name="android.intent.category.DEFAULT" />
+        </intent-filter>
+    </activity>
+
     <service android:enabled="true" android:name="org.thoughtcrime.redphone.RedPhoneService"/>
 
     <service android:enabled="true" android:name=".service.ApplicationMigrationService"/>
--- a/src/org/thoughtcrime/securesms/PanicResponderActivity.java
+++ b/src/org/thoughtcrime/securesms/PanicResponderActivity.java
@@ -62,6 +62,7 @@ public class PanicResponderActivity extends Activity {
             && !TextSecurePreferences.isPasswordDisabled(this)
             && PANIC_TRIGGER_ACTION.equals(intent.getAction())) {
       handleClearPassphrase();
+      ExitActivity.exitAndRemoveFromRecentApps(this);
     }
 
     if (Build.VERSION.SDK_INT >= 21) {
--- a/src/org/thoughtcrime/securesms/PanicResponderActivity.java
+++ b/src/org/thoughtcrime/securesms/PanicResponderActivity.java
@@ -1,0 +1,79 @@
+/**
+ * Copyright (C) 2016 Open Whisper Systems
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.thoughtcrime.securesms;
+
+import android.app.Activity;
+import android.content.Intent;
+import android.os.Build;
+import android.os.Bundle;
+
+import org.iilab.IilabEngineeringRSA2048Pin;
+import org.thoughtcrime.securesms.service.KeyCachingService;
+import org.thoughtcrime.securesms.util.TextSecurePreferences;
+
+import info.guardianproject.GuardianProjectRSA4096;
+import info.guardianproject.trustedintents.TrustedIntents;
+
+/**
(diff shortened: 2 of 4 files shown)


AndroidManifest.xml
```diff
@@ -363,6 +363,19 @@
 
     <activity android:name="com.soundcloud.android.crop.CropImageActivity" />
 
+    <!-- this can never have launchMode singleTask or singleInstance! -->
+    <activity android:name=".PanicResponderActivity"
```
moxie0: why is this an activity?

AndroidManifest.xml
```diff
@@ -363,6 +363,19 @@
 
     <activity android:name="com.soundcloud.android.crop.CropImageActivity" />
 
+    <!-- this can never have launchMode singleTask or singleInstance! -->
```
moxie0: no comments please

src/org/thoughtcrime/securesms/PanicResponderActivity.java
```diff
@@ -0,0 +1,77 @@
...
+import org.thoughtcrime.securesms.util.TextSecurePreferences;
+
+import info.guardianproject.GuardianProjectRSA4096;
+import info.guardianproject.trustedintents.TrustedIntents;
+
+/**
+ * Respond to a PanicKit trigger Intent by locking the app.  PanicKit provides a
+ * common framework for creating "panic button" apps that can trigger actions
+ * in "panic responder" apps.  In this case, the response is to lock the app,
+ * if it has been configured to do so.
```
moxie0: Where is this configured?

src/org/thoughtcrime/securesms/PanicResponderActivity.java
```diff
@@ -0,0 +1,77 @@
...
+  public static final String PANIC_TRIGGER_ACTION = "info.guardianproject.panic.action.TRIGGER";
+
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+
+    TrustedIntents trustedIntents = TrustedIntents.get(this);
+    // Guardian Project Ripple (info.guardianproject.ripple)
+    trustedIntents.addTrustedSigner(GuardianProjectRSA4096.class);
+    // Amnesty International's Panic Button, made by iilab.org (org.iilab.pb)
```
moxie0: no comments

src/org/thoughtcrime/securesms/PanicResponderActivity.java
```diff
@@ -0,0 +1,77 @@
...
+  private static final String TAG = PanicResponderActivity.class.getSimpleName();
+
+  public static final String PANIC_TRIGGER_ACTION = "info.guardianproject.panic.action.TRIGGER";
+
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+
+    TrustedIntents trustedIntents = TrustedIntents.get(this);
+    // Guardian Project Ripple (info.guardianproject.ripple)
```
moxie0: no comments

AndroidManifest.xml
eighthave: I think this comment is in accordance with the contributing guidelines since this is an non-intuitive breakage that can't be documented in the XML itself.  If the launch mode of this activity is changed, it'll silently stop working, because [getCallingActivity()](https://developer.android.com/reference/android/app/Activity.html#getCallingActivity%28%29) will return null because the two Activities will be in different tasks .  But if you don't want it, I'll remove it.

moxie0: even with the comment there, I don't understand. why is this an activity at all?

AndroidManifest.xml
eighthave: TrustedIntents verifies the signing key of the calling Activity to enforce which apps can send Signal the trigger Intent without having to resort to Android permissions and all those related issues. In order for Android to provide the calling Activity, some conditions must be met:
* the Intent must have been sent using `Activity.startActivityForResult()`
* [Activity.getCallingActivity()](https://developer.android.com/reference/android/app/Activity.html#getCallingActivity%28%29) is the only provided method to fetch the app/Activity that sent the Intent (besides `Activity.getCallingPackage()`, which has issues)
* the receiving Activity must have a launchMode of `standard` or `singleTop` 

It is possible to implement this in a Service or a BroadcastReceiver, but then it is not possible to fetch the packageName of the app that sent the Intent without resorting to hacks in the sending process.  Then you're relying on the sender to do stuff properly in order to verify who the sender is.

src/org/thoughtcrime/securesms/PanicResponderActivity.java
eighthave: This relies on the existing lock preference in Signal.  If the user has not set a passphrase, then the panic trigger intent does nothing.

eighthave: see below in the code: `&& !TextSecurePreferences.isPasswordDisabled(this) `

src/org/thoughtcrime/securesms/PanicResponderActivity.java
eighthave: moved into the class javadoc statement

src/org/thoughtcrime/securesms/PanicResponderActivity.java
eighthave: moved into the class javadoc statement

moxie0:
Thanks but I think this is too hackey for us. I'd consider a three line broadcast receiver if you want to go down that road.

eighthave:
I'm fine with a broadcast receiver here, as long as you're OK with an unverified sender.  I added that bit based on feedback from our discussions back in October.  It seems to me that the last thing a malware author would do would be to harrass Signal users by locking the app, so I don't see really any risk in allowing a BroadcastReceiver to lock Signal.

PR #5358, opened 2016-03-19
Fix call termination sounds

### LINKED ISSUE #4319: Alert user to call ending

cascheberg:
### Contributor checklist

  - HTC Sensation XE

---
### Description

Fixes #4319.

Was previously part of PR #4886.

### DIFF
--- a/src/org/thoughtcrime/redphone/RedPhoneService.java
+++ b/src/org/thoughtcrime/redphone/RedPhoneService.java
@@ -216,7 +216,6 @@ public class RedPhoneService extends Service implements CallStateListener, CallS
 
     NotificationBarManager.setCallInProgress(this, NotificationBarManager.TYPE_OUTGOING_RINGING, recipient);
     DatabaseFactory.getSmsDatabase(this).insertOutgoingCall(remoteNumber);
-
   }
 
   private void handleBusyCall(Intent intent) {
@@ -400,7 +399,7 @@ public class RedPhoneService extends Service implements CallStateListener, CallS
   }
 
   public void notifyBusy() {
-    Log.w("RedPhoneService", "Got busy signal from responder!");
+    Log.w(TAG, "Got busy signal from responder!");
     sendMessage(Type.CALL_BUSY, getRecipient(), null);
 
     outgoingRinger.playBusy();
@@ -432,10 +431,13 @@ public class RedPhoneService extends Service implements CallStateListener, CallS
   }
 
   public void notifyCallDisconnected() {
-    if (state == STATE_RINGING)
-      handleMissedCall(remoteNumber, false);
-
     sendMessage(Type.CALL_DISCONNECTED, getRecipient(), null);
+
+    if (state == STATE_RINGING) {
+      handleMissedCall(remoteNumber, false);
+    } else {
+      outgoingRinger.playDisconnected();
+    }
     this.terminate();
   }
 
@@ -489,16 +491,19 @@ public class RedPhoneService extends Service implements CallStateListener, CallS
   }
 
   public void notifyNoSuchUser() {
+    outgoingRinger.playFailure();
     sendMessage(Type.NO_SUCH_USER, getRecipient(), null);
     this.terminate();
   }
 
   public void notifyServerMessage(String message) {
+    outgoingRinger.playFailure();
     sendMessage(Type.SERVER_MESSAGE, getRecipient(), message);
     this.terminate();
   }
 
   public void notifyClientError(String msg) {
+    outgoingRinger.playFailure();
     sendMessage(Type.CLIENT_FAILURE, getRecipient(), msg);
     this.terminate();
   }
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/redphone/RedPhoneService.java
```diff
@@ -366,7 +366,9 @@ private synchronized void terminate() {
     NotificationBarManager.setCallEnded(this);
 
     incomingRinger.stop();
-    outgoingRinger.stop();
+    if (outgoingRinger.isLooping()) {
```
moxie0: why does it have to be looping to call stop?

src/org/thoughtcrime/redphone/RedPhoneService.java
```diff
@@ -432,11 +434,20 @@ public void notifyConnectingtoInitiator() {
...
-      handleMissedCall(remoteNumber, false);
-
     sendMessage(Type.CALL_DISCONNECTED, getRecipient(), null);
-    this.terminate();
+
+    if (state == STATE_RINGING) {
+      handleMissedCall(remoteNumber, false);
+      this.terminate();
+    } else {
+      outgoingRinger.playBusy();
```
moxie0: why play busy here?

cascheberg: The sound is to inform the user that the other party has hung up. Or do you think that the sound should be different?

src/org/thoughtcrime/redphone/RedPhoneService.java
cascheberg: The problem is that `playFailure()` and then `terminate()` can stop the player before the sound was played, since it's async. The idea is to only stop looping players here. Non-looping players are handled automatically in `onCompletion`.

moxie0: Is there ever a non-looping sound that we want to stop immediately? Maybe we should encapsulate this logic into the player itself.

src/org/thoughtcrime/redphone/RedPhoneService.java
moxie0: Yeah that should probably be a different sound. Just whatever a normal Android call plays.

src/org/thoughtcrime/redphone/RedPhoneService.java
cascheberg: Can be done, I was considering that as well.

moxie0: Because we still need to release it, even if we're not stopping it, right?

src/org/thoughtcrime/redphone/RedPhoneService.java
cascheberg: German landline phones have a sound similar to the current busy sound, while my Android phone has a single sound much more similar to the failure sound.
Do you know where I can download sounds that could be used in this project? I could not find a reference for the currently used sounds.

moxie0: You should be able to get whatever sound Android is playing out of AOSP.

cascheberg:
I checked AOSP but could not find any hang up sounds. I then read that the `ToneGenerator` is used for that, which is a pretty interesting class. Using that would require some more changes though.

@moxie would you consider merging this as-is and wait for a follow-up to be discussed? I think using the failure sound for now is not all that bad.

So I updated the PR to:
* handle looping player stuff inside `OutgoingRinger`
* also play 'failure' sound at the end of the call

src/org/thoughtcrime/redphone/audio/OutgoingRinger.java
```diff
@@ -121,8 +129,9 @@ private void stop( int soundID ) {
   }
 
   public void onCompletion(MediaPlayer mp) {
-    //mediaPlayer.release();
-    //mediaPlayer = null;
+    if (mp.isLooping()) return;
```
moxie0: what is this for?

cascheberg: This method is also being called for looping sounds after each iteration. But since we want them to loop, we should only release non-looping players.
(I don't know if that behaviour actually follows the official specs...)

jsalatiel:
what's missing to this get merged ? Apparently this fixes the "talking alone" when call is dropped.

moxie0:
@cascheberg I apologize for taking so long to get to this. The beta calling system is going to replace the redphone system soon, though, and it has end call notification sounds.  I'm going to close this since I look forward to deleting the entire package soon, let me know if you think there's something that needs to be done for the webrtc side though.

PR #5392, opened 2016-03-29
Stop double-setting Linkifying.

### LINKED ISSUE #4874: Links won't open in new tabs
Since version 3.7.2, it is impossible to open multiple links from a conversation in separate browser tabs, as every link will be opened in the same tab and so, overwrite the previously opened link. I am using Firefox for Android. It does not seem to be a Firefox bug, as with other apps the browser behaves as expected (it is opening a new tab for every link)

### LINKED ISSUE #5356: Signal should open a new tab for each clicked link
I have:
- [x ] searched open and closed issues for duplicates
- [ x] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

### Bug description
Android apps can choose whether their links open a new tab or overwrite any previous tab the app had opened. Signal should so the former, but does the latter.

For example, if a friend sends a link to a newspaper article about how awesome end to end encryption is, and then while I'm reading it I get a message from a restaurant asking me to click a link to confirm my reservation, doing so will overwrite the tab where I was reading the article, requiring me to go back to that conversation, click it again, and then find my place again. This is annoying, and sometimes hard to anticipate (who remembers how they opened a tab four hours ago?)

### Steps to reproduce
- click a link in a message you received
- use the app switcher button to switch back to signal
- click a different link in a message you received

### Device info

 **Model:** Google/LG Nexus 5X

 **Android version:** 6.0.1

 **Signal version:** 3.14.2

### Solution
You need to set EXTRA_CREATE_NEW_TAB. see https://developer.android.com/reference/android/provider/Browser.html#EXTRA_CREATE_NEW_TAB for details. I guess if some people find the current behavior preferable, this could be a configuration option, though I really think this is almost always unexpected behavior.

I could send a pull request, but:
- I don't have an appropriate dev environment set up, and so would have to do so without testing, which makes me nervous.
- I can't actually find where Signal is calling out to the browser.

jkdf2:

### Contributor checklist

 * Nexus 5 (hammerhead), Android 4.4.4 (Cyanogenmod 11)

----------

### Description

The conversation_item_body is already Linkified by XML magic in
conversation_item_sent.xml and conversation_item_received.xml by the
entries android:linksClickable="true" and android:autoLink="all".
Not setting it twice:
Fixes #5356
Fixes #4874
May affect #3873

Tested that phone numbers, web URLs, and email addresses all appear formatted and clickable as expected, and that URLs clicked multiple times each opened a new tab rather than overwriting one another.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -27,7 +27,6 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v7.app.AlertDialog;
 import android.text.TextUtils;
-import android.text.util.Linkify;
 import android.util.AttributeSet;
 import android.util.Log;
 import android.view.View;
@@ -241,7 +240,6 @@ public class ConversationItem extends LinearLayout
     mediaThumbnail.setFocusable(!shouldInterceptClicks(messageRecord) && batchSelected.isEmpty());
     mediaThumbnail.setClickable(!shouldInterceptClicks(messageRecord) && batchSelected.isEmpty());
     mediaThumbnail.setLongClickable(batchSelected.isEmpty());
-    bodyText.setAutoLinkMask(batchSelected.isEmpty() ? Linkify.ALL : 0);
   }
 
   private boolean isCaptionlessMms(MessageRecord messageRecord) {


cascheberg:
I tested this PR and sadly I can **not** confirm that it fixes the mentioned bug on my test device (CM 12.1 - Android 5.1.1).
However, the PR changes the linkify behavior during batch selection. But then again, it's still easy to batch select a link-only message.

jkdf2:
Thanks @cascheberg for testing my PR and for pointing out my mistakes with such grace.

Well, this is embarrassing. Turns out my particular browser (Lightning Browser) opens up the URLs in a new tab both before and after my commit. My commit does not fix the issue in Chrome. And, like you said, my commit botches the batch selection.

I clearly need to do some more code reading before I come back.

PR #5397, opened 2016-03-29
Override startActivity to add CREATE_NEW_TAB extra

### LINKED ISSUE #4874: Links won't open in new tabs

### LINKED ISSUE #5356: Signal should open a new tab for each clicked link

jkdf2:
### Contributor checklist

 * Nexus 5 (hammerhead), Android 4.4.4 (Cyanogenmod 11)

----------

### Description

The TextView uses the URLSpan onClick to create the intent when a user
clicks on Linkified URLs, with an extra of Browser.EXTRA_APPLICATION_ID.
This results in the browser window being re-used for any future URLs. By
overriding onClick in the Activity and adding
Browser.EXTRA_CREATE_NEW_TAB to any Browser requests, we prevent the
window re-use.

[Further reading](https://www.reddit.com/r/androiddev/comments/4ci1en/peculiar_behavior_of_autolink/).

Fixes #4874
Fixes #5356

Tested by making sure links open new windows in Chrome and ensured consistent behavior of telephone numbers, email addresses, plain text, and long-presses.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -371,11 +371,11 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
 
   @Override
   public void startActivity(Intent intent) {
-     if (intent.getStringExtra(Browser.EXTRA_APPLICATION_ID) != null) {
-        Log.d(TAG, "Adding EXTRA_CREATE_NEW_TAB to Browser intent.");
-        intent.putExtra(Browser.EXTRA_CREATE_NEW_TAB, true);
-     }
-     super.startActivity(intent);
+    if (intent.getStringExtra(Browser.EXTRA_APPLICATION_ID) != null) {
+      Log.d(TAG, "Adding EXTRA_CREATE_NEW_TAB to Browser intent.");
+      intent.putExtra(Browser.EXTRA_CREATE_NEW_TAB, true);
+    }
+    super.startActivity(intent);
   }
 
   @Override
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -372,8 +372,8 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   @Override
   public void startActivity(Intent intent) {
     if (intent.getStringExtra(Browser.EXTRA_APPLICATION_ID) != null) {
-      Log.d(TAG, "Adding EXTRA_CREATE_NEW_TAB to Browser intent.");
-      intent.putExtra(Browser.EXTRA_CREATE_NEW_TAB, true);
+      Log.d(TAG, "Starting the Browser without replacing current tab.");
+      intent.removeExtra(Browser.EXTRA_APPLICATION_ID);
     }
     super.startActivity(intent);
   }
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -33,6 +33,7 @@ import android.os.AsyncTask;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Vibrator;
+import android.provider.Browser;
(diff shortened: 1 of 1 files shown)


haffenloher:
> The TextView uses the URLSpan onClick to create the intent when a user clicks on Linkified URLs, with an extra of Browser.EXTRA_CREATE_NEW_TAB.

Do you mean `Browser.EXTRA_APPLICATION_ID`? ;) 
Also please use two space indents!

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -369,6 +370,15 @@ public void onActivityResult(final int reqCode, int resultCode, Intent data) {
   }
 
   @Override
+  public void startActivity(Intent intent) {
+     if (intent.getStringExtra(Browser.EXTRA_APPLICATION_ID) != null) {
+        Log.d(TAG, "Adding EXTRA_CREATE_NEW_TAB to Browser intent.");
```
haffenloher: I think if you're using `DEBUG` as the log level you could as well remove the log statement entirely.

jkdf2:
Updated the commit log to state EXTRA_APPLICATION_ID :+1:  Thanks for the catch.

I also pushed a commit with corrected spacing.

2-4601:
[Android docs](https://developer.android.com/reference/android/provider/Browser.html#EXTRA_CREATE_NEW_TAB) say that `EXTRA_CREATE_NEW_TAB` was "Added in API level 12"

What happens with Signal installs on <12 API levels?

haffenloher:
> Android docs say that EXTRA_CREATE_NEW_TAB was "Added in API level 12"

Good point. How about just unsetting `EXTRA_APPLICATION_ID`? Sounds like it should work on both old and new API levels.

Also @jkdf2 have you seen my [other comment](https://github.com/WhisperSystems/Signal-Android/pull/5397#discussion_r57820356)? Why do you want to keep this log statement?

jkdf2:
Implemented and tested the suggestion of just removing the Browser.EXTRA_APPLICATION_ID. Looks good!

I prefer not to remove the Log.d because it may assist developers in the future who may want to modify the Linkify behavior. It took me a long time to track down the wizardry of the XML attribute - I was adding log statements to all of the clickListeners trying to figure out how the clicks on URLs were getting registered, and not figuring it out. You can see what I mean in #5356 where @laurencebgood said "I can't actually figure out where Signal is calling out to the browser." I think the Log statement would help clear that up for future developers and, as you pointed out, the Log.d level will not affect the Release build.

PR #5398, opened 2016-03-30
Provide a fallback for building without git

pR0Ps:
### Contributor checklist
 * Virtual Nexus 5, Android 6.0

----------

### Description
This PR changes the behavior of the `getLastCommitTimestamp` function to fall back to the current time in cases where the `git log` command fails (for whatever reason). This allows people to build the app from the zip files provided on the releases page.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -250,14 +250,19 @@ tasks.whenTaskAdded { task ->
 }
 
 def getLastCommitTimestamp() {
-    new ByteArrayOutputStream().withStream { os ->
-        def result = exec {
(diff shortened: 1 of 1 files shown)


2-4601:
`getLastCommitTimestamp()` was added for reproducible builds. I think there should at least be a warning that reproducible builds will fail if Git is not available or the directory is not a Git repository.

moxie0:
Sorry we will no longer support builds outside of git. I would accept a PR that fails with a more graceful message, but not a fallback.

PR #5408, opened 2016-04-02
Update appearance of ImportExportActivity

### LINKED ISSUE #5387: Tab not highlighted in import/export view (light theme)
### Bug description

In the import/export view with the light theme in use the selected tab is not highlighted.
### Steps to reproduce
- Open Signal (light theme)
- Press the three dots in the top right corner
- Choose "Import / export".
- (optional) click on a tab ("import" or "export")
- Notice that no tab is highlighted.

**Actual result:** No tab highlighted (like with the dark theme)
**Expected result:** Tab highlighted (like with the dark theme)
### Screenshots

how it should look like (dark theme here)
after_dark-backup
### Device info

**Device:** Fairphone 2
**Android version:** 5.1.1
**Signal version:** 3.14.2

FeuRenard:
### Contributor checklist
  - Sony Xperia U, Android 4.4.4 (screenshots)
  - AVD: Nexus 5, Android 6.0

---
### Description

I replaced the deprecated ActionBar TabBar code with the recommended TabLayout scheme in the ImportExportActivity. Then I tried to adjust the appearance to match the old one. Therefore I added colors for both themes. To get rid of the border between the ActionBar and the tabs, I had to change quite much code, because API levels <21 only recognize 'elevation' from theme and not from the activity's .java or .xml.

All of the above appears as a "side effect" of trying to fix #5387.

Fixes #5387
### Screenshots

tabs-light tabs-dark

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -129,6 +129,7 @@
               android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"/>
 
     <activity android:name=".ImportExportActivity"
+              android:theme="@style/TextSecure.LightFlatActionBarTheme"
               android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"/>
 
     <activity android:name=".InviteActivity"
--- a/res/layout/import_export_activity.xml
+++ b/res/layout/import_export_activity.xml
@@ -13,7 +13,7 @@
 
     <android.support.v4.view.ViewPager
         android:id="@+id/import_export_pager"
-        android:layout_width="wrap_content"
+        android:layout_width="match_parent"
         android:layout_height="match_parent"
         android:layout_gravity="center_horizontal" />
 
--- a/res/layout/import_export_activity.xml
+++ b/res/layout/import_export_activity.xml
@@ -11,8 +11,8 @@
         android:layout_height="wrap_content"
         android:background="?import_export_tab_background_color"
         android:elevation="2dp"
-        app:tabTextColor="?import_export_tab_text_color"
-        app:tabSelectedTextColor="?import_export_tab_selected_text_color"
+        app:tabTextColor="#BFffffff"
+        app:tabSelectedTextColor="@color/white"
         app:tabIndicatorColor="?import_export_tab_indicator_color" />
 
     <android.support.v4.view.ViewPager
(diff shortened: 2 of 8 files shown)


moxie0:
This is a really big diff for a very minor change in appearance on an infrequently accessed activity.

FeuRenard:
- The [ActionBar.Tab](https://developer.android.com/reference/android/app/ActionBar.Tab.html) code is deprecated. So with the first commit I performed a refactoring to replace ActionBar.Tab with a [TabLayout](https://developer.android.com/reference/android/support/design/widget/TabLayout.html). This refactoring even saves some lines of code (Switch to TabLayout + Remove all ActionBar TabBar styles)
- In the light theme there is absolutely no tab highlighting at the moment. This is solved by the refactoring and the following coloring.

nice side effects
- the tabs are not only highlighted by a bar a the bottom but also the label show a highlighting effect
- the tab highlight bar gets versatile and a nice animation is created when switching between import and export.

The last screenshot shows the effect of leaving out the 4th commit. This would reduce the diff significantly.

### Screenshots
before/after dark (4.4.4)
highlighting-dark-api19
before/after light (4.4.4)
highlighting-light-api19
after light (6.0)
highlighting-light-api23
after, without commit "Remove ActionBar border in ImportExportActivity" (4.4.4)
highlighting-light-api19-withborder

FeuRenard:
I slightly reduced the diff with b761ac2.

FeuRenard:
I have some more ideas to lower the diff. I'll reopen this PR when they are implemented.

FeuRenard:
Since #5675 was declined and another idea did not work I have no clue how to further reduce the diff without using the deprecated code. Feel free to officially close and reject this PR if you don't like it.

moxie0:
Making a change to a theme is always going to make me pause, because I have no way of knowing what side effects that causes. For instance, these make a change to some tab attribute in a theme that is used by the Giphy activity, which is an activity with tabs in it. Is there a way to not change the existing theme?

FeuRenard:
Thanks for pushing me in the right direction! Now I color this activity the same way Giphy activity is colored. Results vary a bit due to different themes (giphy uses noactionbar theme). I just added `colorPrimary` to the dark theme ([should only affect the action bar](https://developer.android.com/training/material/theme.html#ColorPalette)) because this controls the background color of the tabs. Without this change the tabs are lighter than the action bar.

You may ask: Why do the Giphy tabs have the same color as the action bar? Because the Giphy action bar is colored too light either.

moxie0:
@FeuRenard I really really hate to do this to you, but while I was looking at your change I realized that it's pretty weird to have tabs here at all.  The "export" tab had a single little view in it, so I just switched it to two cards on one screen.

PR #5412, opened 2016-04-04
Adjust NoActionBar themes' style

### LINKED ISSUE #5239: Inconsistencies in rendering colors on Marshmallow vs. Lollipop
While first thinking of a strange design decision (I'm on Marshmallow), I've just noticed that using the **light theme**, the **background colors** of the **conversation list** vs. the **contact list** are are different ones on Marshmallow vs. Lollipop.

**Marshmallow, light theme**
Conversation list: EEEEEE
Contact list: FAFAFA

**Lollipop, light theme**
Conversation list: EEEEEE
Contact list: EEEEEE

On the other hand, with the **dark theme**, the background colors of the conversation list and the contact list - while respectively using different colors (which IS a strange design decision) - **are identical on both platforms**.

**Marshmallow & Lollipop, dark theme**
Conversation list: 000000
Contact list: 303030

The background colors of the conversation list and the contact list should be identical per theme imho.

This is in addition to the button text & switch coloring issues mentioned in #5227, which partially behave differently on Marshmallow & Lollipop as well.
There might be more coloring inconsistencies between Marshmallow & Lollipop, but these are the ones I've discovered so far.
Colors have been measured from screenshots with paint.net

### LINKED ISSUE #5383: Conversation and contact search field design is inconsistent
### Bug description

The designs of the search fields are inconsistent for the conversation list search and the contact search. Especially look at:
- back button color,
- input field text size/shape and
- (dark theme only) the top bar color
### Steps to reproduce
- go to conversation list
- tap the magnifier icon
- have a look at the search field (top bar)
- tap the pencil icon
- have a look at the search field (top bar)
- tap the back button
- ...

**Actual result:** The search fields differ in design. (see bug description)
**Expected result:** The search fields should have the same design.
### Device info

**Device:** Fairphone 2
**Android version:** 5.1.1
**Signal version:** 3.14.2

FeuRenard:

### Contributor checklist
 * Sony Xperia U, Android 4.4.4 (screenshots)
 * AVD: Nexus 5, Android 5.1
 * AVD: Nexus 5, Android 6.0

----------

### Description
The changes affect those activities, which use NoActionBar themes. These are the `NewConversationActivity` (contact search) and the `RecipientPreferenceActivity`.

Fixes #5383
Fixes #5239 (in case @RiseT agrees)

### Screenshots
contact search
contacts-light
contacts-dark


RiseT:
Looks fine to me!

While I don't know anything about Java nor what part of themes.xml corresponds to which UI element, one (potentially meaningless) observation:

In the file
`res/values/themes.xml`

the code block
`<style name="TextSecure.DarkIntroTheme" parent="@style/Theme.AppCompat">`

contains
`<item name="android:windowBackground">@color/black</item>`

while the code block
`<style name="TextSecure.LightIntroTheme" parent="@style/Theme.AppCompat.Light">`

doesn't contain any line specifying the background color

like e. g.
`<item name="android:windowBackground">@color/gray5</item>`

while otherwise having exactly the same code structure.


I don't know where exactly "IntroTheme" is being used and if this could cause potential problems like the ones this PR fixes. I'm just seeing a break of a pattern here... :smile:

FeuRenard:
The pattern is fixed in #5424 :wink:.

moxie0:
And before/after for RecipientPreferencesActivity?

FeuRenard:
updated first post with an additional before/after screenshot

moxie0:
I get changing the font size on the contact search bar, but is there a reason to make dark theme even darker?

FeuRenard:
After this PR the background colors all over the app would be consistent. E.g. only the two activities above use a lighter dark background.
#5239 tried to point this out. Additionally there were inconsistencies noticed between Marshmallow and Lollipop that arose from not setting colors explicitly. This is also resolved here.

RiseT:
@moxie0 I agree that in the longer run, the contrast of the dark theme actually should get lowered. See https://github.com/WhisperSystems/Signal-Android/issues/5470 regarding this topic.

But my issue and  @FeuRenard 's PR are about bringing consistency to theme coloring as a start, as it currently is all over the place on different activity screens and different Android versions.

Once this is done and thus there is  a solid foundation, one can fine-tune the coloring (i. e. lower the contrast imho).

res/values/themes.xml
```diff
@@ -8,7 +8,9 @@
         <item name="actionBarTabBarStyle">@style/TextSecure.LightActionBar.TabBar</item>
         <item name="colorPrimary">@color/textsecure_primary</item>
         <item name="colorPrimaryDark">@color/textsecure_primary_dark</item>
-        <item name="colorAccent">@color/textsecure_primary_dark</item>
+        <item name="colorAccent">@color/signal_primary</item>
+        <item name="android:colorControlActivated">@color/signal_primary</item>
+        <item name="android:windowBackground">@color/gray5</item>
```
moxie0: Why are colors changing in light theme?

RiseT:
@moxie0 There are currently various shades of white in the light theme. See my issue https://github.com/WhisperSystems/Signal-Android/issues/5239 for the exact hex codes.

res/values/themes.xml
FeuRenard: - `colorAccent` sets the color of the mute toggle for API<21.
- `android:colorControlActivated` sets the color of the mute toggle for API>=21. Before the darker blue is used, probably because of `colorPrimaryDark`. After the lighter blue is used.
- `android:windowBackground` is set to achieve consistency among the backgrounds in the app

moxie0: hard for me to follow all of that, but is there a way to do this without making changes to light theme? i'm much less concerned about dark theme changes

FeuRenard: Sorry for not explaining this clearly. In light theme the preference toggles all over the settings area are colored with `signal_primary`. In the recipient preferences the toggle is colored with `signal_primary_dark`. This is the reason why I changed these colors in light theme. But this could be done in a separate PR if it's an issue for somebody.

However I still would like to change the background color to the one that is used everywhere in Signal's light theme: `gray5`.

FeuRenard:
This PR doesn't touch light theme colors anymore except for the background color which changes from white to `gray5` (which is used everywhere in Signal).

RiseT:
@FeuRenard The color of the dark theme background for the safety number screen (recipient preferences) still differs from the rest of the app's backgrounds. I think that screen got redesigned after you created this PR.

FeuRenard:
That color is set here: https://github.com/WhisperSystems/Signal-Android/blob/d24fb8c855ef6be18a6c56e90f381e969dee8aaa/res/layout/verify_display_fragment.xml#L12
So it's not linked to the NoActionBar theme.

RiseT:
Dumb question: It's not linked to the NoActionBar theme. But shouldn't it be linked to it...?

FeuRenard:
You mean it should use the theme's background? Because `VerifyIdentityActivity` has a classic ActionBar and therefore uses the default theme, not NoActionBar. Maybe it's a design deecision to use another grey here to increase readability.

RiseT:
> Because VerifyIdentityActivity has a classic ActionBar and therefore uses the default theme, not NoActionBar.

I see. Thanks.

> Maybe it's a design deecision to use another grey here to increase readability.

Bad design decision imho. While I'm generally all for a grayish background (see https://github.com/WhisperSystems/Signal-Android/issues/5470), the background color should be exactly the same throughout the app. At the very least, it should not vary in minor shades (a completely different color is a different topic).

PR #5419, opened 2016-04-06
Enabled building on windows using the current system time as build_ti…

bottiger:
[See issue 5418 for details](https://github.com/WhisperSystems/Signal-Android/issues/5418)

-----------------------
EDIT: Okay, I clearly **did not** read "how to contribute to this project" before making this pull request. However, I have now.

First time contributor checklist
=======================

✓ I have read how to contribute to this project
✓ I have signed the Contributor Licence Agreement

Contributor checklist
=======================

!!! I am following the Code Style Guidelines **There are no guidelines for the gradle files**
✓ I have tested my contribution on these devices:
        Windows 7
✓ My contribution is fully baked and ready to be merged as is
✕ I have made the choice whether I want the BitHub reward or not by omitting or adding the word FREEBIE in my commit message. **I do not need any reward**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -163,7 +163,11 @@ android {
         minSdkVersion 9
         targetSdkVersion 22
 
-        buildConfigField "long", "BUILD_TIMESTAMP", getLastCommitTimestamp() + "L"
+        def build_timestamp = System.currentTimeMillis();
+        if (!org.gradle.internal.os.OperatingSystem.current().windows) {
+            build_timestamp = getLastCommitTimestamp();
+        }
+        buildConfigField "long", "BUILD_TIMESTAMP", build_timestamp + "L"
         buildConfigField "String", "TEXTSECURE_URL", "\"https://textsecure-service.whispersystems.org\""
         buildConfigField "String", "USER_AGENT", "\"OWA\""
         buildConfigField "String", "REDPHONE_MASTER_URL", "\"https://redphone-master.whispersystems.org\""


2-4601:
> ✕ I have made the choice whether I want the BitHub reward or not by omitting or adding the word FREEBIE in my commit message. **I do not need any reward**

Then you should add `FREEBIE` in the commit message of your first commit...

But this won't be accepted because:
https://github.com/WhisperSystems/Signal-Android/pull/5398#issuecomment-203235677
> **moxie0 wrote:**
> Sorry we will no longer support builds outside of git. I would accept a PR that fails with a more graceful message, but not a fallback.

If you like, you can open a new PR that supports git builds on Windows (and possibly fails more gracefully when there's no git). Thanks for contributing!

bottiger:
@2-4601 The biggest problem with supporting git on windows is that git will not be in the system path as default, and since all tools are working with the assumption that it's not there, no one puts it there. 

You could look for it in the default install path, which is a reasonable guess, but are you ready to pollute your source code with a hard coded path to: "C:\Program Files\Git\bin\git.exe" ?

2-4601:
Can you make it a setting for Gradle?

bottiger:
@2-4601 I don't know anything about Gradle settings, sorry.

bottiger:
@2-4601 That said, you could just include these 3 lines in the build.gradle to make it fail more gracefully. I'm not going to make a pull request with it, but I think other future windows users would be glad to have it here.

        if (org.gradle.internal.os.OperatingSystem.current().windows) {
            throw new GradleException('For building on windows, please see this issue: https://github.com/WhisperSystems/Signal-Android/issues/5418')
        }

(I have tested it)

2-4601:
Or instruct the user to add Git in Windows' system path. A quick search tells that there are also other tools failing on this.

bottiger:
@2-4601 This is obviously very personal, but I prefer not to modify my path on Windows.
On windows I feel it's a little bit dirty and prefer not to do it, I cannot make a logical argument of why it's so different on linux, 

That said, I'm not here to start a long argument :) I'm sure you can pick the solution you think fits Singal best.

Issue #965, opened 2014-03-03
Mark new/unread messages in conversation


jocelynthode:
In whatsapp, when you open a group chat for example, you see just above the new messages that were posted a little banner "New messages /unread messages". I find this feature really helpful to not be lost in a conversation.

Stuie:
It looks like the problem here is that the `ThreadRecord` class only has a binary representation of whether a given conversation thread is "read" or not.

If this read state was persisted on `MessageRecord` instead (need to double-check that it's not already there) then it would be simpler to add in a header in the right place. It would also be possible to flag how many new messages are in a conversation on the `ConversationListActivity` instead of just making the message and date texts bold.

The work required is fairly significant.

1. Add a `read` flag to `MessageRecord` (DB update and schema change)
1. Remove `read` flag from `ThreadRecord` and add foreign key lookup for `MessageRecord`s where required (maintaining both may lead to bugs, but would also keep the query faster)
1. Implement a new header similar to the "Today" and "Yesterday" date headers that exist
1. Integrate the header with the `ConversationAdapter` to get the header in the right place

FeuRenard:
Fixed by d9b42c4

PR #5422, opened 2016-04-07
Mark unread messages in conversation

cascheberg:
### Contributor checklist

---
### Description

Indicator for unread messages, see #965.

A bar on the right side of the conversation indicates unread messages. The bar covers the full  message height and uses the senders color.

Messages are marked as read when the conversation is left, a message is sent by the user, or when the conversation is scrolled to the bottom.

**Screenshot:**
unread_new

### DIFF
--- a/res/layout/conversation_item_received.xml
+++ b/res/layout/conversation_item_received.xml
@@ -3,8 +3,7 @@
         android:id="@+id/conversation_item"
         android:layout_width="fill_parent"
         android:layout_height="wrap_content"
-        android:paddingRight="10dip"
-        android:orientation="vertical"
+        android:orientation="horizontal"
         android:background="@drawable/conversation_item_background"
         android:focusable="true"
         android:nextFocusLeft="@+id/container"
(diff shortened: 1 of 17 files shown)


FeuRenard:
I tested this on my Sony Xperia U (KitKat 4.4.4) and it works in most cases as expected. The cases, where it behave a little unexpected, are messages arriving while the conversation is open.

- A new message has a blue dot, that disappears with the arrival of the next message (that shows a dot instead). I consider having all new messages with the dot more useful.
- Also some arriving messages in the open conversation did not get a blue dot after arrival. But I was not able to reproduce this behavior.

Additionally all those dots feel redundant. One marker above the oldest unread message would be great, but more difficult I guess.

cascheberg:
Thanks, feedback is very welcome.
I can confirm your results. The PR does not change the logic that controls when messages are marked as read. I did not look into that (yet), so I am not sure if it is complex to do or if it would introduce performance problems.
I was considering a continuous bar instead of dots. I can try to get that working, but it seems to require bigger changes to the layout.

cascheberg:
I think changing the background color is also a very nice and simple solution. But I'll probably wait for some "official" feedback before I invest more time to change this in any direction.

cascheberg:
Well, I was curious, so I modified the logic which marks threads as read. And I think it's quite neat now.

cascheberg:
I updated this PR and I think it really looks much nicer now.

> A bar on the right side of the conversation indicates unread messages. The bar covers the full message height and uses the senders color.

> Messages are marked as read when the conversation is left, a message is sent by the user, or when the conversation is scrolled to the bottom.

> (Sreenshot updated)

Please test it!

2-4601:
Does this introduce database changes that could cause issues if switching to a version without this feature?

cascheberg:
No, it does not cause issues. It does not alter the database layout.

haffenloher:
Hey @cascheberg :) Tested this today, for some reason I don't see the bars on Android 4.2.

2-4601:
I can confirm that I don't see the colored bars on my Android 4.0.4. device.
On Marshmallow they work great.

cascheberg:
Thanks for testing. I found a workaround for that Android bug. I have successfully tested it in the emulator on Android 5.0, 4.4, 4.1 and 2.3.3.

2-4601:
Works now on my 4.0.4 device. :+1: 

Some things I've noticed after using this for a couple of weeks:
All the unread marks are cleared
- if I open a picture message
- if I open a link

cascheberg:
@2-4601 I moved the code to automatically `markThreadAsRead()` from `onPause` to `onDestroy`. That solves the issue you reported.

2-4601:
Thanks, it works now.

I noticed one more issue today. When someone leaves a group there is no coloured bar next to it.
Don't know if other similar status messages have the same issue (at least in the screenshot in the first post it appears that group renames are working).

cascheberg:
@2-4601 I can't reproduce your issue on Android 5 or Android 4.4. And I can't test it on older AVD versions because I can't get Google Play Services to work.
Group leave messages use the same layout as group updates and incoming calls.

2-4601:
On my 4.0.4 device group status messages do have colours.

On my Marshmallow device they don't have colours:
screenshot_20160712-195852
Nexus 4, Android 6.0.1

cascheberg:
@2-4601 I figured it out, it is an existing bug in Signal: group updates are always saved as read, unless you set Signal as your default SMS app. Therefore they don't have the unread marker.
That also means that you are not notified when a group was created or updated. Can you confirm that?
I will see if I can open a separate PR to fix that.

2-4601:
Nice catch!

> That also means that you are not notified when a group was created or updated. Can you confirm that?

Yes, I can confirm this with 6.0.1 when SMS is disabled: colours don't work and there are no notifications for group status messages. But if I enable SMS, colours work and notifications show up properly.

But for some reason 4.0.4 has colours and notifications regardless if SMS is enabled or disabled.

haffenloher:
Hey again! Have been using this on Android 4.2 for a while now, I think I like it :) One thing: I noticed that messages _do_ get marked as read when exiting the conversation using the back button, but not when using the home button or tapping a link. Not sure how it was intended, but I feel like they should get marked as read in all three cases.

2-4601:
> but not when ... tapping a link

I think it was changed because I commented earlier that opening a link clears all of the read marks.
For example, if you got 20 new messages and the 10th was a link and you opened it, all read marks were gone when you returned. Now they stay unread.
Don't know if it would be possible to just dismiss the 10 marks from the link upwards and leave the bottom 10 unread.

haffenloher:
@2-4601 hmm, I'm not completely convinced but I get your point. This still leaves us with the question whether the home button should clear the unread state?

From a UX perspective I think most of these little problems would go away if it was possible to treat the thread's unread state (with the corresponding Android notification) and the individual posts' unread marks separately. It always irritates me a little that when I open a thread, the notification in Android's notification bar stays there and the phone's LED is still blinking furiously at me while I read the conversation ;) I'd prefer if the notification got cleared immediately on thread opening (like it was before this patch). But I fear there are good implementation reasons for why @cascheberg chose not to do it like this :\

moxie0:
Seems like this is functionality we should definitely have, but I think I'd prefer the WhatsApp style marker instead of this

Trolldemorted:
was the group-updates-are-marked-as-unread bug tracked in a seperate issue?

2-4601:
@Trolldemorted do you mean #5575?

PR #5425, opened 2016-04-07
Fix spelling mistakes

1ace:
Spotted using https://github.com/lucasdemarchi/codespell


1ace:
Sorry, apparently I'm quite thick: `src/ws/com/google/android/mms/pdu/EncodedStringValue.java` that name doesn't quite look like it originated here.
Taking this change out.

moxie0:
thanks but we're spelling mistakes in comments are fine

Issue #5454, opened 2016-04-21
Crash on startup - sender changed number?


edent:
I have:
- [x] searched open and closed issues for duplicates
- [x] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

----------------------------------------

### Bug description
As soon as I open Signal I get "Unfortunately Signal has stopped."

I *think* this is due to a Textsecure contact changing her phone number.

### Steps to reproduce
- I was able to send & receive SMS & Signal messages this morning.
- A contact has swapped a SIM and has a new number (signal didn't ask her to re-register)
- I received a heads-up notification of a message from her this morning.
- Tried to open Signal and got the "Unfortunately Signal has stopped." error.

I know correlation does not imply causation - but I can't think of anything else which has changed.

I don't want to uninstall / wipe data as a I don't want to lose my messages.

Wiping the cache and restarting the device did not fix the problem.

**Actual result:** Instant crash. Cannot open, take bug reports, nor back up.

**Expected result:** Should open - perhaps isolating the errant message.

### Screenshots

http://imgur.com/Pcorrb6

### Device info

**Device:** Motorola Nexus 6
**Android version:** 6.0.1 (CyanogenMod 13)
**Signal version:** 3.15.2

### Link to debug log

Can't open the app, here's the CatLog

http://imgur.com/Pcorrb6

cascheberg:
The debug log seems to contain helpful information, but obviously not much of it can be read. It would be great if you could export the content.

---

If your contact wants to be found on Signal under her new number, she should **un-register** her current account and then re-register, see http://support.whispersystems.org/hc/en-us/articles/212535818-What-do-I-do-if-I-get-a-new-number-What-if-I-am-switching-to-a-new-phone-

edent:
Here's the export from catlog - let me know if you need more.

She's currently porting her number, so the regular one should be back in a few days. Hope that'll stop the crashes! ☺️

```
04-21 08:58:15.566 W/JobConsumer(3640): 	at org.whispersystems.jobqueue.JobConsumer.runJob(JobConsumer.java:76)
04-21 08:58:15.566 W/JobConsumer(3640): 	at org.whispersystems.jobqueue.JobConsumer.run(JobConsumer.java:46)
04-21 08:58:15.567 E/AndroidRuntime(3640): 	at org.whispersystems.jobqueue.JobConsumer.runJob(JobConsumer.java:76)
[...]

PR #5460, opened 2016-04-24
Prevent contact number NPE

cascheberg:
### Contributor checklist

----------

### Description
I can't reproduce the crash in #5454, but this should fix it: if the returned contact phone number is `null`, fall through to return default `RecipientDetails`.
It's still not clear to me which result is actually returned.
I have some "numberless" SMS in my stock messaging app, and there I can't even add the sender to my contacts.

//FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/recipients/RecipientProvider.java
+++ b/src/org/thoughtcrime/securesms/recipients/RecipientProvider.java
@@ -138,13 +138,18 @@ public class RecipientProvider {
 
     try {
       if (cursor != null && cursor.moveToFirst()) {
-        Uri          contactUri   = Contacts.getLookupUri(cursor.getLong(2), cursor.getString(1));
-        String       name         = cursor.getString(3).equals(cursor.getString(0)) ? null : cursor.getString(0);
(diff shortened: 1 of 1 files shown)


glems2:
@cascheberg Great, thank you for the fix!
Any idea on who might merge this and when? Or what might be in the way of doing so?
I'm currently up in the air about Signal.

siloportem:
Got also an SMS that caused the crash described in #5454, this fixes it for me. After compiling my own Signal apk with this patch applied, I am able to access all my messages again including the SMS that caused the crash. Thank you, hope this will be merged into the main-branch soon.

edent:
Is there any way I can apply this to an existing installation? My messages are trapped due to  #5454 :-(

Or, any details on when this will be merged & released?

2-4601:
If you have compiled Signal yourself or you have root you can do that. But it's better to ask help for those at https://whispersystems.discoursehosting.net/

smarek:
Thank you, I now did verify, this PR is correct solution for problems discussed in #5454

Issue #1036, opened 2014-03-05
LED notification doesn't work


donjoe0:
I had the LED set to blue and normal blinking pattern in TextSecure v2.0.3 and upon receiving messages from service "numbers" (like my mobile carrier and my bank) I saw no LED notification whatsoever. I checked this with the screen still off because the Motorola Droid 3 only lights the notification LED when the screen is off, but there was no LED action (though the ringtone and vibration had worked just fine).

rhodey:
@donjoe0 @Martuh if you could provide a debug log documenting this behavior that'd be a great help, thanks!

donjoe0:
I'm not using the Droid 3 anymore and my Asus Padfone 2 has no notification LED.

jimbo0848:
https://gist.github.com/557de22561ccae235218

agrajaghh:
@jimbo0848 there is nothing in that log, what are you experiencing?

jimbo0848:
Feb 24, 07:20

I have a LGG3 D850 AT&T the Led notification is not working at all. I tried selecting all colors to see if that was the problem. Nothing seem to work to get Led notifications to work. Really need those so when phone is in silent mode you are able to see that you received a text or incoming call. Absolutely love the program! Would like to see this fixed in a future update. Believe I am not complaining one bit but I am a believer in if you don't know it is broken than how can you fix it. Keep up the great work!!!!!! Thank you for keeping private private!!!!

2-4601:
#5189 2016/02/07:
>The LED used to work fine with text messages prior to the installation of Signal. It still works correctly with WhatsApp. Signal seems to be the only app where the LED is malfunctioning.
>
>Here's my debug log : https://gist.github.com/anonymous/a4338e57a0ee8ddfc0a3

p4nci:
@devs
maybe similar issue(???):
https://github.com/SMSSecure/SMSSecure/commit/d1afcf1e08ee3e1d9caad078c729255e6ca067bb

Pazu:
I reported this same issue a few months ago using an older version of Signal on my Sony Ericsson Xperia Mini ST15i, but at some point the issue was resolved. I'm not sure when. Now, I'm using Signal v3.20.4 and the notification LED works as expected.

FeuRenard:
https://github.com/WhisperSystems/Signal-Android/commit/bd2fc82360c2a2fbd5e92194074104d465b8bf19 possibly resolved this issue.

[...]

PR #5466, opened 2016-04-28
Always enable LED notification

cascheberg:
### Contributor checklist

- [X] I have tested my contribution

----------

### Description
Fixes that missed call notifications do not enable LED. Might also help with #1036.
Without this, every silent notification update (`!signal` / `updateNotification(context, masterSecret)`) will turn off the LED.

// FREEBIE


moxie0:
This is a really large diff that's difficult for me to follow, are all of these changes necessary for fixing the issue at hand?

cascheberg:
The PR consists of two commits. The first one is not necessary, but I think it's a nice and simple cleanup, basically text replacement:
* introduce variable `notification` instead of `notifications.get(0)`
* introduce variable `recipients` instead of `notifications.get(0).getRecipients()`
* introduce `MessageNotifier.cancelNotification` method instead of `context.getSystemService[...]`

The second commit is the actual patch, which allows to call `setLed` separately instead of handling everything in `setAlarms`.

I was looking into this code while working on the patch, so I like to add some simple cleanups.

moxie0:
In general I'd like PRs to be as small as possible. Can you resubmit this with only the necessary changes?  Thanks.

PR #5486, opened 2016-05-08
Simplify updateNotification variants

cascheberg:
### Contributor checklist

----------

### Description
Simplify the `updateNotification` variants. There is only one very special caller with `includePushDatabase = true`, so that case can be moved to a dedicated method.

It is in `PushDecryptJob`:
* `masterSecret = null`
* `includePushDatabase = true`
* `threadId = -2`, is not a valid thread
* by default `signal = true`

As a result, there are now "only" three variants of `updateNotification(...)` instead of four.

// FREEBIE


moxie0:
Thanks for looking at this, but unless this fixes a specific problem, I typically don't like to accept PRs that are general refactoring. Given the limitations of automated testing on Android, it's a lot of work to code review, so I'm generally looking for the smallest number of lines necessary to get a specific objective accomplished cleanly. Thanks again.

Issue #4637, opened 2015-11-20
Can't link devices / scan QR


bartz1:
I am stuck at the step saying "scan the qr code displayed...". I can see the qr code in the camera, but it won't do anything. I'm assuming there should be a moment when the phone goes like "okay I saw that's a qr code", but nothing happens. I think the problem may be that the camera is a bit out of focus.
Samsung galaxy s3, signal 3.5

bartz1:
I can confirm the problem is that the camera not being in focus. I have the same problem when using the quick direct capture (can't click to focus).
I came to this conclusion after taking a screenshot and zooming -- it linked the device

2-4601:
You might already be aware of this pr #3767 but marking it here as related.

bartz1:
I don't know if the code that the quick camera uses is the same as the one used by the qr scanner (I haven't looked at the code to be honest) but it seems to be the case.

ndevln:
As in #4804 autofocus is not working when QR Code scanning. I have the same setup:
Signal version: 3.6.1
Android version: 5.1.1 (Cyanogenmod 12.1) snapshot
Device: Samsung Galaxy S2 ("i9100")

ndevln:
Saving the QR-Code png and zoomind doesn't work for me. Even when the QR Code is sharp on the screen, it is not recognized. Also when using Signal 3.7.0, I can scan the Code via an external App and it opens Signal, but in Signal I have to rescan.

moxie0:
@ndevln does it make a difference if you try rotating the device to portrait or landscape?

HansHerbert:
No difference for me

Signal version: 3.6.1
Android version: 5.1.1 (Cyanogenmod 12.1)
Device: Samsung Galaxy S2 ("i9100")

ndevln:
No makes no difference. Also moving the QR Code in from all sides makes no difference (in portrait and landscape mode). Also I can't take pictures with my back camera in the messaging view (but i can see the preview). My front camera works in the quick camera mode (autofocus and capturing pictures). I think using the front camera for the qr code scan would work, but it's still only a workaround.

lolonaut:
Same situation here, Signal doesn't focus and the QR code won't get recognized (zoomed screenshot or switching between landscape/portrait doesn't help).

Signal version: 3.6.1
Android: 5.1.1 (Cyanogenmod 12.1)
Device: Galaxy S2 (i9100)

JStyle21:
[...]

PR #5538, opened 2016-06-18
Workaround to add tsdevice:/? string when QR codes won't work

f41c0r:

### Contributor checklist

- [x] I am following the [Code Style Guidelines](https://github.com/WhisperSystems/Signal-Android/wiki/Code-Style-Guidelines) (To the best of my knowledge :grimacing:)
 * LG Nexus 5, Android 4.4.4
 * Samsung Galaxy S3, Replicant/Android 4.2.2

----------

### Description

So I don't know if this is mergeable for UX/philosophical reasons so Moxie you'll have to make that call, but I thought since I had already written the code as a jank hack just to get Signal-Desktop to pair with my personal device, I thought I would offer it as a potential solution to #4637. 

This commit adds an EditText and Button that will allow a user to manually type the QR code text (`tsdevice:/?blahblahblah`) into the EditText and then submit the device that way. The UX for this may not be optimal, and I ended up using an `setOnFocusChangeListener` on the EditText because it makes it a nightmare to type because the soft keyboard and screen takes too much space while you're trying to type if you don't hide the camera preview.

I have tested this on two devices with two separate phone numbers and it allowed me to pair signal-desktop to the device both times. Here is a screenshot:

screen

Thanks!

### DIFF
--- a/res/layout/device_add_fragment.xml
+++ b/res/layout/device_add_fragment.xml
@@ -20,6 +20,7 @@
 
         <org.thoughtcrime.securesms.components.ShapeScrim
             android:layout_weight="1"
+            android:id="@+id/camera_frame"
             android:layout_width="match_parent"
             android:layout_height="match_parent"/>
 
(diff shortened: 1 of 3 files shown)


moxie0:
thanks for submitting but you're right, this isn't really a ux we want to expose

f41c0r:
Yeah, that's what I figured, I just thought since I already had the code, I'd offer it to you guys if you wanted it.

Vinnl:
How about as a context menu option when long pressing the camera view, perhaps even only visible when holding a certain button (e.g. volume up)? Or are there also security risks or maintainability issues related with exposing this in any way?

straussmaus:
> thanks for submitting but you're right, this isn't really a ux we want to expose

@moxie0 Could you briefly explain why?  This is a perfect solution to a problem I am having.  I don't want to buy a new phone just for an unscratched camera that I will probably never use.  Maybe in 5 years your reasons have been updated?  Thank you.

vince-from-nice:
I have just bought a new Xiaomi Redmi Note 11 and Signal crashes when it opens the cam. 
So I have no other choice than using Molly, thanks for the link !

SeSodesa:
This feature is certainly needed, as the camera of my phone just decided to stop working altogether, but the phone is just fine otherwise. I am not going to replace the phone just because of that, or because Signal does not support other device registration methods than QR-codes, but I do miss the possibility of typing messages on my laptop instead of a touch screen.

Issue #5364, opened 2016-03-21
Enhancement: Append timestamp to export filename


Fmstrat:
I have:
- [X] searched open and closed issues for duplicates
- [X] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports
### Enhancement description

Right now, if you run a Plain Text Export, it exports to a file named `SignalPlaintextBackup.xml`

Since every time you backup, it uses the same filename, a user has to rename the file if they archive them off in a folder. It also makes it easy to accidentally overwrite an old backup when testing.

This request is to append a formatted timestamp to the filename, thus `SignalPlaintextBackup-YYYYMMDDHHSSmm.xml`, or `SignalPlaintextBackup-EPOCH.xml` or any other combination that would ensure unique filenames.

Suggest to include this with feature request: https://github.com/WhisperSystems/Signal-Android/issues/2791

Would likely require alterations here: https://github.com/WhisperSystems/Signal-Android/blob/3d651ef29a2b270ac291946ba6f0eda3db43fc6e/src/org/thoughtcrime/securesms/database/PlaintextBackupExporter.java#L15

And alterations to import code to select a file.
### Steps to reproduce

Just a plain text backup.

 **Signal version:** 3.14.2

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #5542, opened 2016-06-20
Added Timestamp to PlainTextBackup

char32:
### Contributor checklist

- [x] I have tested my contribution on these devices: Nexus 4 ( Android 5.0 )

-------

Like requested in #5364 this patch adds a timestamp to the backup.
The file will be named like SignalPlainTextBackup_20160620201010.xml.
So yyyyMMddhhmmss as format.
To test it is just necessary to do a simple PlainTextBackup.

### DIFF
--- a/src/org/thoughtcrime/securesms/database/PlaintextBackupExporter.java
+++ b/src/org/thoughtcrime/securesms/database/PlaintextBackupExporter.java
@@ -9,10 +9,13 @@ import org.thoughtcrime.securesms.database.model.SmsMessageRecord;
 
 import java.io.File;
 import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 
 public class PlaintextBackupExporter {
 
-  private static final String FILENAME = "SignalPlaintextBackup.xml";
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/database/PlaintextBackupExporter.java
```diff
@@ -9,10 +9,13 @@
...
 import java.io.File;
 import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 
 public class PlaintextBackupExporter {
 
-  private static final String FILENAME = "SignalPlaintextBackup.xml";
+  private static final String TIMESTAMP = new SimpleDateFormat("yyyyMMddhhmmss").format(new Date(System.currentTimeMillis()));
+  private static final String FILENAME = "SignalPlaintextBackup_.xml" + TIMESTAMP;
```
jeremymasters: Would it be better to have the timestamp after the underscore (and before the .) instead of smooshed to the file suffix?

char32: @jeremymasters You are right!
I looked at it with ls and did not notice it.

I will discard this pull request and create a new one, so we do not need to squash commits, ok?

char32:
will reupload

PR #5575, opened 2016-07-15
Fix group update message notifications, mark as unread

cascheberg:
### Contributor checklist

----------

### Description
Users are not notified when someone creates or updates a group. That is because group messages are usually not saved as unread. It only works if you set Signal as your default SMS app, or on Android versions before Kitkat where you can't set it as default app.

cc @2-4601 

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/database/SmsDatabase.java
+++ b/src/org/thoughtcrime/securesms/database/SmsDatabase.java
@@ -490,7 +490,7 @@ public class SmsDatabase extends MessagingDatabase {
     }
 
     boolean    unread     = org.thoughtcrime.securesms.util.Util.isDefaultSmsProvider(context) ||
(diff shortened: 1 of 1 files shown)


2-4601:
Tested on Nexus 4, 6.0.1.
Works :)

moxie0:
I think we're intentionally trying not to notify users on these actions. Where does the desire to issue a notification come from?

cascheberg:
Sometimes you might want to defer discussing things in a group until a certain third person was added to the group by someone else (or even until someone created the group), so the notification would be relevant. Or you just want to know that someone renamed a group so that you can still identify it...

As I pointed out, a big amount of users already receive these notifications, such as those that set Signal as default SMS app. And this type of notification usually does not pop up frequently.

The inconsistency was noticed while testing #5422.

moxie0:
in 3.18.0

PR #5577, opened 2016-07-19
Ignore fast sender clock

cascheberg:
### Contributor checklist

- [X] I have tested my contribution

----------

### Description
Steps to reproduce the issue:
- take phone A and set date/time some hours (or days) into the future
- send message to phone B
- check phone B:

1. The message from A will stay on top of the conversations list even if newer messages are received, until B replies to A, or B reaches the date/time of A.
2. In the conversation with A, the message will show a wrong relative timestamp, it will show "now" for quite a while.

This PR reduces the annoying effect of the fast sender clock by checking if the message was received before it was sent. In that case it is referencing the time the message was received, which is based on B's local clock.

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/database/model/MessageRecord.java
+++ b/src/org/thoughtcrime/securesms/database/model/MessageRecord.java
@@ -119,8 +119,10 @@ public abstract class MessageRecord extends DisplayRecord {
   }
 
   public long getTimestamp() {
-    if (isPush()) return getDateSent();
-    else          return getDateReceived();
+    if (isPush() && getDateSent() < getDateReceived()) {
+      return getDateSent();
+    }
+    return getDateReceived();
   }
 
   public boolean isForcedSms() {


tge-linee:
It would be good if Signal Desktop got the same update, otherwise the two apps would show the same conversation differently.

moxie0:
I'm worried that the lack of consistency might be worse. What are the situations you're encountering where the sender has a bad clock?

tge-linee:
With the one I enoutered the sender had his desktop set to the wrong time (UTC one hour off) and to the wrong time zone (to one which was one hour off) so that he didn't realise that he had a bad clock. All his messages had a timestamp that said "Now" for an hour and the conversation with him would stay at the top of the conversations list even though new messages in other conversations arrived.

cascheberg:
I can confirm what @tge-linee wrote. This might not happen often, but the resulting situation is also lacking consistency in a very noticeable way. Note that you can still see the original dates in the message details.

moxie0:
I see, yeah desktop hmm. Alright that makes sense, but I think sent > received might be a little too strict, maybe add some slop in there.

cascheberg:
I can add a threshold, but honestly I don't see an advantage in that yet. If someones clock is off by only 5 minutes, it will still say "sent now" for 5 minutes. Then it will say "sent 10 minutes ago" although it was sent 15 minutes ago and so on, which is quite a lot in an actively ongoing conversation.

Screenshot:
clock

moxie0:
in 3.18.0

PR #5580, opened 2016-07-20
Add shadow under compose view on scroll

### LINKED ISSUE #5098: Shadow above the message text box when texts are covered
This is how Signal currently looks when texts are covered:

This is what Google messenger does when texts are covered:

As you can see, a shadow appears on the bottom of the message area on Google messenger, and not on Signal.

This bottom shadow serves two purposes: to visually communicate the presence of later messages and to provide consistency with the shadow on the action bar. It would be great if this could be added to Signal.

mateoeh:
### Contributor checklist
 * Samsung Galaxy S5, Android 6.0.1 (CyanogenMod 13)
 * AVD Nexus 5X, Android 6.0.1
 * AVD Nexus One, Android 2.3

----------

### Description

Adds a shadow under the compose view (like the one in Google Messenger) that divides it from the conversation list when scrolling.

Fixes #5098

screenshot_20160719-202632

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -88,6 +88,7 @@ public class ConversationFragment extends Fragment
   private Locale       locale;
   private RecyclerView list;
   private View         loadMoreView;
+  private View         composeDivider;
 
   @Override
   public void onCreate(Bundle icicle) {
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -104,6 +105,32 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle bu
...
+          } else if (Build.VERSION.SDK_INT > Build.VERSION_CODES.HONEYCOMB) {
+            composeDivider.setAlpha(isAtBottom() ? 0 : 1);
+          } else {
+            composeDivider.setVisibility(isAtBottom() ? View.INVISIBLE : View.VISIBLE);
+          }
+
+          wasAtBottom = isAtBottom();
+        }
+      }
+    });
```
moxie0: Let's not make this anonymous if we're going to maintaining instance state, that way we can contain the helper method as well.

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -104,6 +105,32 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle bu
     list.setHasFixedSize(false);
     list.setLayoutManager(layoutManager);
 
+    composeDivider = view.findViewById(R.id.compose_divider);
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.HONEYCOMB) {
+      composeDivider.setVisibility(View.INVISIBLE);
+    }
```
moxie0: why is this necessary?

moxie0:
lookin good

haffenloher:
Isn't there already a very subtle shadow beneath the action bar at the top? Compared to that, I feel like the shadow you added above the compose text field is too strong.

src/org/thoughtcrime/securesms/ConversationFragment.java
mateoeh: The `alpha` attribute on the divider view does nothing in API 10 and below; `visibility` is only manipulated in API 10 and below to avoid redundancy.

mateoeh:
@haffenloher Good point, I'll be sure to change that. Closing and resubmitting later

PR #5615, opened 2016-08-14
Fix ConversationItem recipients listener

### LINKED ISSUE #4420: Halloween easter egg: Muting notifications in group chat messes up avatars & names

cascheberg:
### Contributor checklist

----------

### Description
Fixes #4420, which was introduced by 7838e8fbe0ad992407529be81d48cf9f686dc61d, as @jarteaga said.

The idea of that commit was to correctly color outgoing audio message controls, i.e. according to the conversation partners color.

But the `recipient` of outgoing audio messages is "Anonymous", so the `conversationRecipients` were passed in and used for that. And it is only used for that.

In group conversations, the `conversationRecipients` and its `getPrimaryRecipient()` is a group recipient (i.e. group name and photo), so it must not be used to update a single conversation items color, contact name and photo, which is exactly what is happening in the newly registered `onModified` listener (again as @jarteaga has already pointed out).

_The diff is not so nice, I have only created a new method `setAudioViewTint`._

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -38,10 +38,10 @@ import android.widget.LinearLayout;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import org.thoughtcrime.securesms.components.AlertView;
 import org.thoughtcrime.securesms.components.AudioView;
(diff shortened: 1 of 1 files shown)


moxie0:
in 3.18.0

PR #5628, opened 2016-08-23
Support for disappearing messages

moxie0:
Support for disappearing messages


nrizzio:
Looks pretty good so far. Current bugs I noticed:

 - When **you** enable disappearing messages, the generated notice that indicates that you have done so will disappear from **your** phone, but not from the recipient's phone.
 - When a message expiration notice is the last message in a conversation, the conversation list summary is "Media Message". It should say "(You/Beta) set disappearing message time to [time]".
signal4
 - It's impossible to delete expiring message notices. This is more of a design thing, but I think it would be better to leave absolutely no indication that messages were automatically deleted, wheverer possible.

Other than that, keep up the great work! :clap:

-----
To discuss this feature, please do it [here](https://whispersystems.discoursehosting.net/t/auto-deleting-messages-becoming-a-real-thing/473), not on GitHub. Thanks!

moxie0:
Thanks, those are great catches.  I'm not sure I understand what you're saying about the call button, do you mean you're not seeing it in the actionbar anymore?

FeuRenard:
It depends on the screen size whether the call button is shown in the action bar or the action bar menu. In my case I can see it directly on AVD Nexus 5X (5.2 inches) but not on my tiny Sony Xperia U (3.5 inches). On the tiny phone I have to tap the three dots to see a "Signal call" entry.

moxie0:
thanks, will update showInActionBar to always

FeuRenard:
@nrizzio wrote:
> It's impossible to delete expiring message notices. This is more of a design thing, but I think it would be better to leave absolutely no indication that messages were automatically deleted, wherever possible.

This is not a problem on the sender's phone. The notice (like the messages) disappears after the set time. So the notice will disappear before the associated messages will do. But you are right for the "off" setting. This will be visible forever, I think. And that's a hint that the feature was used.

As @nrizzio already covered many valid points, I have to be very nit-picking to add further issues :wink: 

- When you set the timer multiple times without sending messages in between, you can end up in a confusing conversation screen like shown below (it's also confusing if you filter out Emu's conversation updates). Maybe the expiration notice can disappear earlier, when a new timer is set?
device-2016-08-23-235442
- I can imagine a scenario, where multiple messages are shown in a conversation, but they have different expiration times. Set 1 week, after a little less than one week you send a message and the next moment the notice about this 1 week disappears. Then you set a new shorter time (e.g. 1 hour). After a little less than this hour you send another message. In the next moment the notice that you set 1 hour disappears. Now you can change the timer again. Then you cannot say exactly, how long a certain message will be displayed. At least in the message details the user should be able to find this information.
- A tooltip could be displayed when you long press the expiration timer button. So the user can see what he is going to do, when he presses it. ("set disappearing messages time"?)
- A phone running Signal without this PR shows an empty message when the timer is set on the other phone. Maybe this friction during migration is unavoidable.

That's it for now.

moxie0:
I think all of @nrizzio 's issues should be taken care of now.

Two things remain:

1) Deleting timer update messages.

I'd prefer not to add an action for the former, since we generally don't provide actions for these types of updates.  They expire, which I think is enough, except @FeuRenard points out that "off" never expires.

Maybe we have a default expiration for the "off" updates as well? 1 minute?

2) Concurrent timer update messages can be confusing.

I kind of like the idea of removing any former timer update messages whenever you receive a new one.  Can't really think of any other way to make that less confusing, other than not having them expire at all.  Any other ideas?

nrizzio:
> 2) Concurrent timer update messages can be confusing.
I kind of like the idea of removing any former timer update messages whenever you receive a new one. Can't really think of any other way to make that less confusing, other than not having them expire at all. Any other ideas?

In my opinion, that would make the most sense. It doesn't hide the fact that a message is going to expire, because all expiring messages show an hourglass. And e5b030b added the ability for the user to see the amount of time an individual message has until it expires, so no information is truly being lost by the removal of the previous timer update messages.

> Maybe we have a default expiration for the "off" updates as well? 1 minute?

Yeah, all "off" updates have to expire, or else it indicates that expiring messages were previously transmitted. I think one minute is reasonable.

-----
One possible issue that I just thought about was the situation where a long backlog of messages could expire before the user had the chance to read them. Suppose that you are in a group, and 100 messages were sent with the expiration time of 5 seconds, but you haven't opened the conversation yet. Can we ensure that a) the message view will start where you last left off so the messages can be read in chronological order, and b) the countdown timer will start only as each message comes into view of your screen?

-----
This is lower priority as I think it's an edge case that only @FeuRenard and I will experience, but after I `git pull`ed (jumping from 58e22db to 3d3b1d4) and installed the APK on my test phones, Signal crashes on startup. I'm not sure if this is a cause for concern, but I'm mentioning it here just in case. If I have to delete the app from the test phones it's no big deal.
```
08-23 20:08:25.192 7811-7811/org.thoughtcrime.securesms E/AndroidRuntime: FATAL EXCEPTION: main
                                                                          Process: org.thoughtcrime.securesms, PID: 7811
                                                                          java.lang.IllegalArgumentException: column 'expires_in' does not exist
                                                                              at android.database.AbstractCursor.getColumnIndexOrThrow(AbstractCursor.java:303)
                                                                              at org.thoughtcrime.securesms.database.ThreadDatabase$Reader.getCurrent(ThreadDatabase.java:578)
                                                                              at org.thoughtcrime.securesms.ConversationListAdapter.getThreadRecord(ConversationListAdapter.java:169)
                                                                              at org.thoughtcrime.securesms.ConversationListAdapter.getItemViewType(ConversationListAdapter.java:159)
                                                                              at org.thoughtcrime.securesms.database.CursorRecyclerViewAdapter.getItemViewType(CursorRecyclerViewAdapter.java:158)
                                                                              at android.support.v7.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:4373)
                                                                              at android.support.v7.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:4326)
                                                                              at android.support.v7.widget.LinearLayoutManager$LayoutState.next(LinearLayoutManager.java:1955)
                                                                              at android.support.v7.widget.LinearLayoutManager.layoutChunk(LinearLayoutManager.java:1364)
                                                                              at android.support.v7.widget.LinearLayoutManager.fill(LinearLayoutManager.java:1327)
                                                                              at android.support.v7.widget.LinearLayoutManager.onLayoutChildren(LinearLayoutManager.java:556)
                                                                              at android.support.v7.widget.RecyclerView.dispatchLayout(RecyclerView.java:2713)
                                                                              at android.support.v7.widget.RecyclerView.onLayout(RecyclerView.java:3011)
                                                                              at android.view.View.layout(View.java:14834)
                                                                              at android.view.ViewGroup.layout(ViewGroup.java:4631)
                                                                              at android.widget.LinearLayout.setChildFrame(LinearLayout.java:1671)
                                                                              at android.widget.LinearLayout.layoutVertical(LinearLayout.java:1525)
                                                                              at android.widget.LinearLayout.onLayout(LinearLayout.java:1434)
                                                                              at android.view.View.layout(View.java:14834)
                                                                              at android.view.ViewGroup.layout(ViewGroup.java:4631)
                                                                              at android.support.design.widget.CoordinatorLayout.layoutChild(CoordinatorLayout.java:1009)
                                                                              at android.support.design.widget.CoordinatorLayout.onLayoutChild(CoordinatorLayout.java:719)
                                                                              at android.support.design.widget.CoordinatorLayout.onLayout(CoordinatorLayout.java:733)
                                                                              at android.view.View.layout(View.java:14834)
                                                                              at android.view.ViewGroup.layout(ViewGroup.java:4631)
                                                                              at android.widget.FrameLayout.layoutChildren(FrameLayout.java:453)
                                                                              at android.widget.FrameLayout.onLayout(FrameLayout.java:388)
                                                                              at android.view.View.layout(View.java:14834)
                                                                              at android.view.ViewGroup.layout(ViewGroup.java:4631)
                                                                              at android.support.v7.internal.widget.ActionBarOverlayLayout.onLayout(ActionBarOverlayLayout.java:493)
                                                                              at android.view.View.layout(View.java:14834)
                                                                              at android.view.ViewGroup.layout(ViewGroup.java:4631)
                                                                              at android.widget.FrameLayout.layoutChildren(FrameLayout.java:453)
                                                                              at android.widget.FrameLayout.onLayout(FrameLayout.java:388)
                                                                              at android.view.View.layout(View.java:14834)
                                                                              at android.view.ViewGroup.layout(ViewGroup.java:4631)
                                                                              at android.widget.LinearLayout.setChildFrame(LinearLayout.java:1671)
                                                                              at android.widget.LinearLayout.layoutVertical(LinearLayout.java:1525)
                                                                              at android.widget.LinearLayout.onLayout(LinearLayout.java:1434)
                                                                              at android.view.View.layout(View.java:14834)
                                                                              at android.view.ViewGroup.layout(ViewGroup.java:4631)
                                                                              at android.widget.FrameLayout.layoutChildren(FrameLayout.java:453)
                                                                              at android.widget.FrameLayout.onLayout(FrameLayout.java:388)
                                                                              at android.view.View.layout(View.java:14834)
                                                                              at android.view.ViewGroup.layout(ViewGroup.java:4631)
                                                                              at android.view.ViewRootImpl.performLayout(ViewRootImpl.java:1983)
                                                                              at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:1740)
                                                                              at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:996)
                                                                              at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:5600)
                                                                              at android.view.Choreographer$CallbackRecord.run(Choreographer.java:761)
                                                                              at android.view.Choreographer.doCallbacks(Choreographer.java:574)
                                                                              at android.view.Choreographer.doFrame(Choreographer.java:544)
                                                                              at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:747)
                                                                              at android.os.Handler.handleCallback(Handler.java:733)
                                                                              at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                          	at android.os.L

```

nrizzio:
Some more things I noticed, this time dealing with attachments:
- Open photos do not close when they expire.
- The message deletion timer starts when the attachment has been "read", but it should be when the attachment has completed downloading.
- Attachment save dialogues do not disappear when its corresponding attachment does, which causes an error (not a crash however) if the user presses "yes".
- Audio files will keep playing after they have disappeared. (Not sure if this really matters, but mentioning it anyway.)

FeuRenard:
> 1) Deleting timer update messages.

It's ok to have "off" disappear after 1 minute.

> 2) Concurrent timer update messages can be confusing.
I kind of like the idea of removing any former timer update messages whenever you receive a new one. Can't really think of any other way to make that less confusing, other than not having them expire at all. Any other ideas?

I have another solution in mind: Let the notice disappear when the following conditions are both met:

- no more message with the notice's expiration time is shown below the notice `&&`
- a new expiration time was set

With this solution an expiring message would never be shown without the associated timer notice and we get rid of "overridden" timer notices that were not followed by messages.

And still a tooltip for the timer button would be nice, especially to explain the feature when it is fresh. I doubt that most people read release notes.

FeuRenard:
Another design suggestion concerning the timer button in the action bar:
Only show it, when an expiration time is set. So when the timer is set off the next time you open the conversation the timer setting is hidden in the menu. But when a time is set it should reappear in the action bar.
I am thinking of users, that don't want to use this feature (many, I think). They "off" symbol is always shown to them and maybe on small displays a name is not shown in full because of this button. I know that it is not a very consistent design, but a way to move the button out of the way for people who not plan to use this feature.

agrajaghh:
> The message deletion timer starts when the attachment has been "read", but it should be when the attachment has completed downloading.

What about the auto download setting? Wouldn't this lead to potentially deleting the message before the user has actually seen it?

nrizzio:
> Wouldn't this lead to potentially deleting the message before the user has actually seen it?

That's what happens currently. But you're correct, I didn't notice that I inverted the problem.
The countdown should begin only when *both* the attachment is completely downloaded and the attachment is read.

moxie0:
Updated behavior:

1) Media message timer starts after it has both been downloaded and seen.

2) Expiration notifications no longer expire.  Just didn't feel quite right, so I've changed it to allow for manual deletion of the notices (along with group updates and call notices as well).

3) When the expiration timer is disabled for a conversation, the action gets moved into the overflow menu instead of the actionbar.

moxie0:
in 3.18.0 (but disabled)

nrizzio:
Have the bugs (other than the second bullet) documented [in my comment here](https://github.com/WhisperSystems/Signal-Android/pull/5628#issuecomment-241935571) been fixed? cc @FeuRenard

moxie0:
The second one has been fixed (timer starts after attachment has been seen after download completes).

bungabunga:
Sorry for this rather philosophical post, I am aware that you don't like discussions here and I promise I'll keep my mouth shut after this one (that really bothers me).

Can Automatically disepperaing messages be disabled on receiver's device?

I feel a big dislike towards disappearing messages that can not be disabled on receiver's device. I think that nobody from the outside should interfere with the stuff on my device. The sender should think twice whether to send or not to send the message beforehead but when the message left her device and enter mine, she shouldn't have the control of it anymore. She could only ask me to delete the message myself and I will do it if I feel so because her message is not within her jurisdiction anymore.

I see messages that disappear on my device without my consent as intrusion.

cascheberg:
@moxie0 Outgoing timer update messages are still disappearing from the senders phone in group conversations. Single conversations are fine. Tested with Signal 3.19.1.

PR #5635, opened 2016-08-25
Support for updated fingerprint format

moxie0:


res/values/strings.xml
```diff
@@ -539,16 +539,10 @@
...
-    <string name="VerifyIdentityActivity_warning_the_scanned_key_does_not_match_please_check_the_fingerprint_text_carefully">WARNING, the scanned key DOES NOT match! Please check the fingerprint text carefully.</string>
-    <string name="VerifyIdentityActivity_not_verified_exclamation">NOT Verified!</string>
-    <string name="VerifyIdentityActivity_their_key_is_correct_it_is_also_necessary_to_verify_your_key_with_them_as_well">Their key is correct. It is also necessary to verify your key with them as well.</string>
-    <string name="VerifyIdentityActivity_verified_exclamation">Verified!</string>
-    <string name="VerifyIdentityActivity_you_don_t_have_an_identity_key_exclamation">You don\'t have an identity key!</string>
+    <string name="VerifyIdentityActivity_your_contact_is_running_an_old_version_of_signal">Your contact is running an old version of Signal, please ask them to update before verifying security numbers.</string>
+    <string name="VerifyIdentityActivity_you_re_attempting_to_verify_security_numbers_with">You\'re attempting to verify security numbers with %s, but scanned %s instead.</string>
```
nrizzio: This line causes the following build error:
> Multiple substitutions specified in non-positional format; did you mean to add the formatted="false" attribute?

I'm assuming it means that `%s` shouldn't be in the same string twice?

moxie0: whoops got it

moxie0:
Superseded by #5646

PR #5653, opened 2016-09-01
Use ACTION_GET_CONTENT for attachment picking on all platforms

moxie0:
@unrulygnu This is more what I had in mind for #3154 when I was asking for a single public method that returns a future.  It seems like we should always return a future when the input is a stream, since the Uri shouldn't be usable until the persistence task is complete.

It's been a while since I looked at the mechanisms of all this, so I could be wrong, but let me know if this meets your needs.  Thanks!


unrulygnu:
@moxie0: Sorry for the delay, just got back into town.  Will review this week. Thanks for the update and welcome back.

PR #5662, opened 2016-09-03
Add shadow under compose view on scroll (v2)

### LINKED ISSUE #5098: Shadow above the message text box when texts are covered

### LINKED ISSUE #5908: labling the unlabled buttons in Signal for Android
Hi.
I am Pradip Sikdar, a visually challenged android intermediate user. I have started using Signal Private Messanger for Android recently and most of the part, I am quite happy with it.
As you know, we use screen reader technology to use our android or IOS devices. So, if there is a menu or button which is not properly tagged or labled, the screen reader doesn't recognize it.
For example, let's say there is a play button. Visually there is a sign or graphics which represents that, it is a play button. But if the button is not properly tagged or labled as "Play" button, our screen reader only says "Button Unlabled". So, we cannot recognize what the button is.
So, I am here by requesting to the developer of this nice application, to lable all the buttons of this application. For example, when I am clicking on the "toggle attachment" button, I am getting some additional buttons which are not labled. So, I cannot recognize them.
Thanks in advance from all the blind community.
Pradip

mateoeh:
### Contributor checklist
 * Samsung Galaxy S5, Android 6.0.1 (CyanogenMod 13)
 * AVD Nexus 5X, Android 6.0
 * AVD Nexus One, Android 2.3

----------

### Description

Adds a shadow separating the compose view from the message list when scrolled.

Fixes #5098 

device-2016-09-03-134758

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -33,6 +33,7 @@ import android.support.v7.view.ActionMode;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
 import android.support.v7.widget.RecyclerView.ItemAnimator.ItemAnimatorFinishedListener;
+import android.support.v7.widget.RecyclerView.OnScrollListener;
 import android.text.ClipboardManager;
 import android.text.TextUtils;
 import android.util.Log;
@@ -78,6 +79,7 @@ public class ConversationFragment extends Fragment
 
   private final ActionModeCallback actionModeCallback     = new ActionModeCallback();
   private final ItemClickListener  selectionClickListener = new ConversationFragmentItemClickListener();
+  private final OnScrollListener   scrollListener         = new ConversationScrollListener();
 
   private ConversationFragmentListener listener;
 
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -383,6 +393,36 @@ public void onLoaderReset(Loader<Cursor> arg0) {
     void setThreadId(long threadId);
   }
 
+  private class ConversationScrollListener extends OnScrollListener {
+    boolean wasAtBottom = true;
```
moxie0: why is this package level access?

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -104,6 +107,13 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle bu
     list.setHasFixedSize(false);
     list.setLayoutManager(layoutManager);
 
+    composeDivider = view.findViewById(R.id.compose_divider);
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.HONEYCOMB) {
```
moxie0: What's this check about?

PR #5591, opened 2016-07-31
Fix unlink device endless spinning wheel

FeuRenard:
### First time contributor checklist
- [x] I have read [how to contribute](https://github.com/WhisperSystems/Signal-Android/blob/master/CONTRIBUTING.md) to this project
- [x] I have signed the [Contributor Licence Agreement](https://whispersystems.org/cla/)

### Contributor checklist
- [x] I am following the [Code Style Guidelines](https://github.com/WhisperSystems/Signal-Android/wiki/Code-Style-Guidelines)
- [x] I have tested my contribution on these devices:
 * Sony Xperia U, Android 4.4.4 (CyanogenMod)
- [x] My contribution is fully baked and ready to be merged as is
- [x] I have made the choice whether I want the [BitHub reward](https://github.com/WhisperSystems/Signal-Android/wiki/BitHub-Rewards) or not by omitting or adding the word `FREEBIE` in my commit message

----------

### Description
Fixes #4641

Tested with one and two linked devices (in the latter case both devices were unlinked successively). After each unlink the device list is shown again and the wheel of confusion does not spin forever.

//FREEBIE

PR #5682, opened 2016-09-14
Fix DeviceListLoader behavior on init and restart

### LINKED ISSUE #4641: Unlinking device: endless Wheels of Confusion
1. Unlink a device
2. The Wheel of Confusion spins till heat death

black sabbath vol 4 wheels of confusion

This is just a visual problem. I.e. the device gets actually unlinked.

Tested on 3.5.0 (though the wheels have been spinning since summer... was just too afraid to report it before :)
* Nexus 4, stock 5.1.1
* Huawei U8800, 4.0.4

Edit:
Also tested when there are more than one device linked. Same issue.

FeuRenard:
### Contributor checklist
 * Sony Xperia U, Android 4.4.4

----------

### Description
Our `DeviceListLoader` does not start a loader upon calling `initLoader(...)` or `restartLoader(...)`. It has to be forced to do this with `forceLoad()`. This is a [known issue](https://code.google.com/p/android/issues/detail?id=14944) and further discussed on [stackoverflow](http://stackoverflow.com/questions/10524667/android-asynctaskloader-doesnt-start-loadinbackground).

The reason seems to be, that a loader is created upon init and restart (I tested this), but it is not started/loaded. After creation `onStartLoading()` will be called. This is implemented without functionality in `android.support.v4.content.Loader`. We have to provide the functionality by overriding this method. That's what I do with this PR.

In my implementation simply `forceLoad()` is called inside `onStartLoading()`. In the above discussions there are some checks suggested before doing this `forceLoad()`. But I think everytime we init or restart the loader we want fresh data to be fetched. So from my standpoint the proposed checks are unnecessary in our case.

Replaces #5591
Fixes #4641
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/DeviceListFragment.java
+++ b/src/org/thoughtcrime/securesms/DeviceListFragment.java
@@ -77,7 +77,7 @@ public class DeviceListFragment extends ListFragment
   @Override
   public void onActivityCreated(Bundle bundle) {
     super.onActivityCreated(bundle);
-    getLoaderManager().initLoader(0, null, this).forceLoad();
+    getLoaderManager().initLoader(0, null, this);
     getListView().setOnItemClickListener(this);
   }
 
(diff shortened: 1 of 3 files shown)


moxie0:
Why not use the AsyncLoader abstract class linked from that issue?

FeuRenard:
I included AsyncLoader from this issue:
https://code.google.com/p/android/issues/detail?id=14944

I had to slightly adapt it:

- use `android.support.v4.content.AsyncTaskLoader` instead of `android.content.AsyncTaskLoader`
- add package statement

I tested it, seems to work fine.

moxie0:
in 3.20.0

PR #5693, opened 2016-09-17
fingerprint scanning fixups

michaelkirk:
Really wanted to see this working with the iOS branch.

I don't really do Java, so feel free to throw this away.


moxie0:
in 3.20.0

PR #5731, opened 2016-09-29
Fix Dockerfile build image

spk:
### Contributor checklist

  - Docker version 1.12.1, build 23cf638

---
### Description
- Use official java:openjdk-8-jdk instead of ubuntu
- Check Android SDK download with sha256
- Other minor fixes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,38 @@
-FROM ubuntu:17.04
+FROM openjdk:8-jdk
 
-RUN dpkg --add-architecture i386 && \
-    apt-get update -y && \
-    apt-get install -y software-properties-common && \
-    apt-get update -y && \
-    apt-get install -y libc6:i386=2.24-9ubuntu2.2 libncurses5:i386=6.0+20160625-1ubuntu1 libstdc++6:i386=6.3.0-12ubuntu2 lib32z1=1:1.2.11.dfsg-0ubuntu1 wget openjdk-8-jdk=8u131-b11-2ubuntu1.17.04.3 git unzip && \
-    rm -rf /var/lib/apt/lists/* && \
-    apt-get autoremove -y && \
-    apt-get clean
+ENV DEBIAN_FRONTEND=noninteractive \
+    DEBIAN_PRIORITY=critical \
+    DEBCONF_NOWARNINGS=yes
+
+RUN set -ex \
+        && dpkg --add-architecture i386 \
+        && apt-get update -qqy \
+        && apt-get -qqy --no-install-recommends install \
+            software-properties-common \
+            libc6:i386=2.24-11+deb9u1 \
+            libncurses5:i386=6.0+20161126-1+deb9u1 \
+            libstdc++6:i386=6.3.0-18+deb9u1 \
+            lib32z1=1:1.2.8.dfsg-5 \
+            wget git unzip \
+        && apt-get purge -y \
+        && apt-get autoremove -y \
+        && apt-get clean \
+        && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 ENV ANDROID_SDK_FILENAME android-sdk_r24.4.1-linux.tgz
+ENV ANDROID_SDK_SHA256 e16917ad685c1563ccbc5dd782930ee1a700a1b6a6fd3e44b83ac694650435e9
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,26 +1,34 @@
...
+    DEBIAN_PRIORITY=critical \
+    DEBCONF_NOWARNINGS=yes
+
+RUN set -ex \
+        && dpkg --add-architecture i386 \
+        && apt-get update -qqy \
+        && apt-get -qqy --no-install-recommends install \
+            software-properties-common \
+            libc6:i386 libncurses5:i386 libstdc++6:i386 \
+            lib32z1 wget git unzip \
```
moxie0: Without pinning the packages, these are not guaranteed to be the same every time

spk: If the version is updated/removed the the build will fail, is there a reason for that @moxie0 ? Why not thrusting last version from Debian repository ?

2-4601: The primary goal for the docker image is reproducible builds. See: https://whispersystems.org/blog/reproducible-android/

moxie0: My understanding is that debian maintains a package archive of every package they've ever published, right?

spk: Yes they do on http://snapshot.debian.org/ but its just in case of the package is removed and you really need the package version.
If a package have a security issue the version is removed, and so we need to update the Dockerfile every time.

spk: I don't trust pre-build image when I cannot check the [Dockerfile](https://hub.docker.com/r/whispersystems/signal-android/) and when the image is not an automatic build.
That's why I wanted to build it myself.
Anyway if its a requirement for you to pin the versions; I will do it

moxie0: For this to be reproducible, the binaries need to be exactly the same. If it's not true that debian keeps an archive of every version of every package they've ever distributed, then we'll have to figure something else out.

moxie0:
looks like this still isn't pinning the debian version?

spk:
@moxie0 its done on https://github.com/WhisperSystems/Signal-Android/pull/5731/commits/c1e3bff31d33081c0304c625ab4d63cf980940a6
I think it will broke soon but there is other improvements like checking `sha256sum` that would be nice to merge

spk:
Rebased, is there other issues to address ?

xmikos:
I have just tried this modified Dockerfile, but build still fails:

    E: Version '2.19-18+deb8u6' for 'libc6:i386' was not found

spk:
I've updated the Debian package version @xmikos 
For me it is not useful to pin the Debian version, because Debian version is about `debian/` files changes so not related with the binary; but change was asked by @moxie0 in https://github.com/WhisperSystems/Signal-Android/pull/5731#issuecomment-263333120

xmikos:
@spk Build is now working. But looking at `Dockerfile`, you are using `java:openjdk-8-jdk` as base image. But Docker [java repository](https://hub.docker.com/_/java/) is deprecated:

    DEPRECATED
    This image is officially deprecated in favor of the openjdk image, and will receive no 
    further updates after 2016-12-31 (Dec 31, 2016). Please adjust your usage accordingly.

Instead of it, Docker [openjdk repository](https://hub.docker.com/_/openjdk/) should be used.

xmikos:
I have found several other issues:

1. too old `ANDROID_BUILD_TOOLS_VERSION` (Dockerfile in official Signal-Android repository specifies 25.0.0 and Signal build fails with old version specified in your pull request)

2. final APK (built with `./gradlew clean assemblePlayRelease`) is not reproducible. It differs from Signal 4.2.4 downloaded from Google Play (looking at disassembled classes.dex file, there are small differences with line numbers, etc.)

3. there is no NDK specified in Dockerfile so you can't rebuild native shared libraries with `ndk-build`. Using prebuilt native shared libraries defeats whole purpose of reproducible builds (there can be hidden native malicious code)

spk:
Thanks @xmikos I did not notice the deprecation on java
Updated on 467e8713eff62d7a3baaddac6d20472a8a5658db
For the `ANDROID_BUILD_TOOLS_VERSION` I've just rebased the branch

ale5000-git:
Any update about this?

PanderMusubi:
Hope this will get reviewed again or, even better, merged soon.

spk:
https://github.com/signalapp/Signal-Android/tree/master/reproducible-builds

Issue #672, opened 2014-02-25
Lower compression for sent pictures


shmartens:
It would be nice if pictures which are sent over the data channel aren't so highly compressed.

its4everyone:
This is definitely an issue for me as well. Textra is currently my default texting application due to this setting alone. It renders any pictures I send/receive nearly useless, especially if there is some important/pertinent text contained within.

dkasak:
I was under the impression (from reading this thread) that the compression was changed to cap the images at 1280x960, but I still see larger images downsized to 768xY when sent through TextSecure with the latest version on both phones. Why is this happening?

mcginty:
If you're on a device with a limited amount of memory, images are sent with a max of 768 in either dimension. Otherwise, the max is 1280.

dkasak:
Oh. What is considered a limited amount of memory?

mcginty:
If you're on KitKat or higher, Android has a flag to alert us that the device we're running on is in a low memory class. Otherwise, we decide you're low memory if the "memory class" (read: the max heap size for our app) is <= 64.

BlacklightShining:
@mcginty 64…what? 64 MiB? That sounds pretty low, even for a phone. Even so, normal stills captured from the camera shouldn't be more than, what, 4 or 5 MiB? And if there _isn't_ a limited amount of memory, why have the cap at all?

agrajaghh:
@BlacklightShining a single 24bit image with 8 megapixels has an uncompressed size of over 20MB...

64MB is not the device memory, its the maxmimum available memory for a single app (afaik)

mcginty:
what @agrajaghh said

BlacklightShining:
@agrajaghh When would you ever have to deal with an uncompressed image, though? Don't pretty much all devices immediately compress pictures with JPEG?

mcginty:
@BlacklightShining images are uncompressed bitmaps when loaded into memory.

BlacklightShining:
@mcginty Ahh.

Dyras:
@mcginty So what exactly qualifies as "limited amount of memory"? I have a Galaxy Nexus with 1 gig memory and every time I receive a picture it's 1280x720. The picture is 119 kilobytes and that seems awfully low, and the picture itself is just really ugly.  Any chance you could knock it up to 1920x1080?

However it's entirely possible that my eyes are broken. Or maybe I'm too spoiled by my full HD screen. Here's the picture I'm talking about.
[...]

PR #5765, opened 2016-10-09
Better panorama resizing

### LINKED ISSUE #5574: Panoramas suffer considerably from image compression
I have:
- [x] searched open and closed issues for duplicates : "panorama"

---
### Bug description

Sending a Panorama does crop the size / compress it to an unviewable picture. neither thumbnail nor picture when u open/save has a correct quality.
Funny thing is, that i got a panorama send from IOS, and quality was perfect after receiving.
### Steps to reproduce
- send attachment panorama picture
- see the quality / crop ...

**Actual result:** signal cropping or resizing the quality way to heavy.
**Expected result:** receiver or sender should see the sent panorama in full quality
### Screenshots

fsdfsdfsd
### Device info

**Device:** samsung i9305
**Android version:** 6.0.1
**Signal version:** 3.16.0
### Link to debug log

https://gist.github.com/anonymous/3c649e2e6a1658ecc7a0c9b542b04f63

haffenloher:
### Contributor checklist

- [:camel:] I am following the [Code Style Guidelines](https://github.com/WhisperSystems/Signal-Android/wiki/Code-Style-Guidelines)
- [:camel:] I have tested my contribution on these devices:
 * Emulator, Android 5.1
- [:camel:] My contribution is fully baked and ready to be merged as is
- [:camel:] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the `Fixes #1234` [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)
- [:camel:] I have made the choice whether I want the [BitHub reward](https://github.com/WhisperSystems/Signal-Android/wiki/BitHub-Rewards) or not by omitting or adding the word `FREEBIE` in the commit message of my first commit

----------

### Description

We're downscaling images to avoid OOM errors. The amount of memory needed is proportional to the amount of pixels. This PR replaces the edge length constraint with a pixel count constraint to achieve better results for exotic aspect ratios (e.g. with panoramic pictures).

To convert the existing edge length limits into pixel count limits, I (arbitrarily) chose an aspect ratio of 4:3, so the memory usage and the resulting image size for a 4:3 image should be the same before and after applying this PR (while 16:9 images should turn out a bit larger and panoramic images much larger than before).

Fixes #5574
Related #672

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/MediaConstraints.java
+++ b/src/org/thoughtcrime/securesms/mms/MediaConstraints.java
@@ -26,8 +26,7 @@ public abstract class MediaConstraints {
   public static MediaConstraints MMS_CONSTRAINTS  = new MmsMediaConstraints();
   public static MediaConstraints PUSH_CONSTRAINTS = new PushMediaConstraints();
 
-  public abstract int getImageMaxWidth(Context context);
-  public abstract int getImageMaxHeight(Context context);
+  public abstract int getImageMaxPixels(Context context);
   public abstract int getImageMaxSize();
 
   public abstract int getGifMaxSize();
@@ -53,8 +52,8 @@ public abstract class MediaConstraints {
     try {
       InputStream is = PartAuthority.getAttachmentStream(context, masterSecret, uri);
       Pair<Integer, Integer> dimensions = BitmapUtil.getDimensions(is);
-      return dimensions.first  > 0 && dimensions.first  <= getImageMaxWidth(context) &&
-             dimensions.second > 0 && dimensions.second <= getImageMaxHeight(context);
+      return dimensions.first > 0 && dimensions.second > 0 &&
+             dimensions.first * dimensions.second <= getImageMaxPixels(context);
     } catch (BitmapDecodingException e) {
       throw new IOException(e);
     }
--- a/src/org/thoughtcrime/securesms/mms/MmsMediaConstraints.java
+++ b/src/org/thoughtcrime/securesms/mms/MmsMediaConstraints.java
@@ -5,18 +5,13 @@ import android.content.Context;
 import org.thoughtcrime.securesms.util.Util;
 
 public class MmsMediaConstraints extends MediaConstraints {
-  private static final int MAX_IMAGE_DIMEN_LOWMEM = 768;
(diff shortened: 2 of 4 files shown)


mcginty:
I haven't tested this, but it's worth mentioning that I used to run into issues on many phones when an image had any dimension greater than 2048 pixels which is why I had it enforcing a max dimension instead of max pixel count (the limit is due to Android's OpenGL settings having a max texture size of 2048x2048). If you haven't yet, I'd test this against edge-case images on phones that aren't super top of the line.

See https://stackoverflow.com/questions/4527695/how-can-i-find-the-maximum-texture-size-for-different-phones for how to work out your phone's max texture size, but IIRC 2048 was a safe value.

haffenloher:
ugh :( will reopen as soon as I get around to investigate the stuff you mentioned. thanks!

moxie0:
@mcginty Shouldn't glide take care of all of this at load time? Did we start downscaling to prevent OOM and performance issues before glide, or because glide wasn't doing this correctly?

haffenloher:
Yes, from what I read, glide should by default load a perfectly sized image for the view (using downsampling). Maybe we're overriding this default though (as it would make zoomed-in images look ugly)? Haven't found the time to test this yet.

mcginty:
@moxie0 Don't remember perfectly, but I think the issue came from loading the full image in the preview where you could zoom. You should be able to support larger than 2048-w/h images if you change the image preview to something like https://github.com/davemorrissey/subsampling-scale-image-view which is smarter about what parts and scales of the image it loads in memory.

moxie0:
@mcginty Nice, thanks for the link. I guess the other consideration though is that if a device with a ton of memory transmits a 40MB image, the receiving device still has to load the whole thing into memory in order to scale it down. Unless glide has figured out some kind of stream-based scaling?

haffenloher:
@moxie0 by default, glide uses BitmapFactory's [`inSampleSize`](https://developer.android.com/reference/android/graphics/BitmapFactory.Options.html#inSampleSize) option to subsample images while loading them into memory, see their [`Downsampler`](https://github.com/bumptech/glide/blob/master/library/src/main/java/com/bumptech/glide/load/resource/bitmap/Downsampler.java) class.

PR #5788, opened 2016-10-18
Giphy integration

moxie0:
Giphy integration


FeuRenard:
I tested on Sony Xperia U, Android 4.4.4, 3.5 inch screen, 512 MB RAM.
Scrolling through the list of GIFS gives me the following logcat output while Signal freezes:
https://gist.github.com/FeuRenard/1908709f7f46f531fe589e25550933ca
The lines at the end repeat very quickly until Signal dies. But I cannot reproduce this. Just happened during the first ~3 trials.

Furthermore this:
device-2016-10-19-193832

The search field allows line breaks when entering a search. That's not very intuitive, newlines should be disabled.

FeuRenard:
Two other issues, not very reproducible. Described them on gist.
https://gist.github.com/FeuRenard/9f23a28bdf221fc3025e607548e551c1
https://gist.github.com/FeuRenard/ba074faaa9e677c31574d6efc895c686

This PR seems to be very memory consuming, at least sometimes.

moxie0:
Yeah looks like this doesn't work great on low-end devices, thanks for testing. If we can trust the memclass, I can change this to not actually load gifs into memory on low memclass devices. The gif/sticker browser would only show stills until you select one, which would populate the attachment preview with the actual gif.

FeuRenard:
I don't think that this is necessary, at least in my case. Memory near 100mb just happened to me the first three times I tried the browser. During the trials afterwards garbage collection seemed to be working as intended and just about 30-40mb RAM were in use by Signal. Something occured during those first trials that later didn't.

And these "two other issues" I reported won't be affected by your suggested change if I get this right.

moxie0:
in 3.21.0

Issue #3990, opened 2015-08-27
TextSecure crashes when passphrase locked and something is shared to it


paride:
This is easy to reproduce:
1. Configure TextSecure so it's locked with a (non cached) passphrase;
2. Share something to it with the Android "Share to" function, for example an image from the Photos app;
3. TextSecure opens and asks for the passphrase. Enter it.
4. Press Unlock. TextSecure crashes.

Debug log:

https://gist.github.com/anonymous/32aae2e2352b96d4bf40

```
E/AndroidRuntime( 5665): FATAL EXCEPTION: main
E/AndroidRuntime( 5665): Process: org.thoughtcrime.securesms, PID: 5665
E/AndroidRuntime( 5665): java.lang.SecurityException: Uid 10099 does not have permission to uri 0 @ content://com.google.android.apps.photos.contentprovider/-1/1/content%3A%2F%2Fmedia%2Fexternal%2Fimages%2Fmedia%2F3651/ACTUAL/1707585164
E/AndroidRuntime( 5665):    at android.os.Parcel.readException(Parcel.java:1546)
E/AndroidRuntime( 5665):    at android.os.Parcel.readException(Parcel.java:1499)
E/AndroidRuntime( 5665):    at android.app.ActivityManagerProxy.startActivity(ActivityManagerNative.java:2430)
E/AndroidRuntime( 5665):    at android.app.Instrumentation.execStartActivity(Instrumentation.java:1496)
E/AndroidRuntime( 5665):    at android.app.Activity.startActivityForResult(Activity.java:3745)
E/AndroidRuntime( 5665):    at android.app.Activity.startActivityForResult(Activity.java:3706)
E/AndroidRuntime( 5665):    at android.support.v4.app.FragmentActivity.startActivityForResult(FragmentActivity.java:820)
E/AndroidRuntime( 5665):    at android.app.Activity.startActivity(Activity.java:4016)
E/AndroidRuntime( 5665):    at android.app.Activity.startActivity(Activity.java:3984)
E/AndroidRuntime( 5665):    at org.thoughtcrime.securesms.PassphraseActivity$1.onServiceConnected(PassphraseActivity.java:60)
E/AndroidRuntime( 5665):    at android.app.LoadedApk$ServiceDispatcher.doConnected(LoadedApk.java:1208)
E/AndroidRuntime( 5665):    at android.app.LoadedApk$ServiceDispatcher$RunConnection.run(LoadedApk.java:1225)
[...]

PR #5805, opened 2016-10-23
Improve #3990

Oishikatta:

### Contributor checklist

  - Virtual Device Nexus 5X, Android 7.0

---
### Description

By catching the crash Signal remains unlocked. The user can quickly
retry sharing from the previous app and it will work, rather than crashing again until the user returns to the home screen and unlocks Signal.

Since this issue occurs rarely (passphrase used, Signal not unlocked, media shared to Signal), I believe the simplest thing that could be called a fix might be to send a Toast informing the user that they should try sharing again.

This commit is just catching the crash, not sending a Toast. Comments appreciated, I'm interested in solving this issue and contributing more to Signal.

// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/PassphraseActivity.java
+++ b/src/org/thoughtcrime/securesms/PassphraseActivity.java
@@ -21,6 +21,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.ServiceConnection;
 import android.os.IBinder;
+import android.util.Log;
 
 import org.thoughtcrime.securesms.crypto.MasterSecret;
 import org.thoughtcrime.securesms.service.KeyCachingService;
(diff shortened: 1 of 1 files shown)


moxie0:
Apologies, I'm not following. Can you restate what issue this is trying to solve? Thanks!

Oishikatta:
Sure,

The issue occurs under the following conditions:

1.  User has Signal on Android configured with a passphrase
2.  User tries to "Share" media to Signal while Signal is not unlocked

The issue occurs because of Android's temporary permissions system for sharing (permissions expire when the receiving activity ends, which is the passphrase activity for this issue). I wasn't able to find any method to maintain/pass the permission to the normal sharing activity after the passphrase activity. One solution would be to make a local copy of the media, but that seems unreasonable for something like this.

The issue:

1.  User uses the "Share" menu from another app (e.g. Photos) to "share" media to Signal
2.  User enters their passphrase to unlock Signal
3.  Signal crashes. (Access permission to shared media expired with passphrase activity)
4.  At this point, a typical user might try sharing again. Because Signal crashed, it will be locked again and crash again as many times as they try sharing.

This pull request:

1.  Instead of crashing, Signal will return the user to the previous activity (without completing the intended Sharing). E.g. if the user was sharing from the Photos app, they will be returned to the Photos app.
2.  Because it doesn't crash, Signal will remain unlocked so that the user can then try sharing again and it will work normally without crashing.

I would suggest possibly also displaying a toast to let the user know they can/should retry sharing -- and I can easily add this to the PR if wanted.

moxie0:
thanks, in 3.22.0

PR #5811, opened 2016-10-26
add fastscroller to ConversationFragment

zoff99:
### Contributor checklist

---
### Description

adds a FastScroller to ConversationFragment,
and fix the color of that new horizontal seperator (depending on the actual background color)
-> android:id="@+id/compose_divider"


moxie0:
Thanks but this looks like you're checking another library's entire source tree into our source tree? We can't do things like that, if it doesn't have a maven artifact we can't use it.

PR #5814, opened 2016-10-27
Giphy: "No results found." --> "Nothing found"

RiseT:

### Contributor checklist

  - Device A, Android X.Y.Z
  - Device B, Android Z.Y
  - Virtual device W, Android Y.Y.Z

---
### Description

Maybe it's just me and I'm splitting hairs, but I don't think that "No results found." makes much sense.

In such a case, there is a "result" of the search, and that result is the fact that no "matches" were found.

Thus, I'm changing the string to

**"No matches found."**

Edit: Changed to **"Nothing found"**

// FREEBIE

### DIFF
--- a/res/layout/giphy_fragment.xml
+++ b/res/layout/giphy_fragment.xml
@@ -17,7 +17,7 @@
                  android:indeterminate="true"/>
 
     <TextView android:id="@+id/no_results"
-              android:text="@string/giphy_fragment__no_matches"
+              android:text="@string/giphy_fragment__nothing_found"
               android:gravity="center"
               android:layout_width="match_parent"
               android:layout_height="wrap_content"
--- a/res/layout/giphy_fragment.xml
+++ b/res/layout/giphy_fragment.xml
@@ -17,7 +17,7 @@
                  android:indeterminate="true"/>
 
     <TextView android:id="@+id/no_results"
-              android:text="@string/giphy_fragment__no_matches_found"
+              android:text="@string/giphy_fragment__no_matches"
               android:gravity="center"
               android:layout_width="match_parent"
               android:layout_height="wrap_content"
--- a/res/layout/giphy_fragment.xml
+++ b/res/layout/giphy_fragment.xml
@@ -17,7 +17,7 @@
                  android:indeterminate="true"/>
 
     <TextView android:id="@+id/no_results"
-              android:text="@string/giphy_fragment__no_results_found"
+              android:text="@string/giphy_fragment__no_matches_found"
               android:gravity="center"
               android:layout_width="match_parent"
               android:layout_height="wrap_content"
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -763,7 +763,7 @@
     <string name="giphy_activity_toolbar__search_gifs_and_stickers">Search GIFs and stickers</string>
 
     <!-- giphy_fragment -->
-    <string name="giphy_fragment__no_results_found">No results found.</string>
+    <string name="giphy_fragment__no_matches_found">No matches found.</string>
 
 
     <!-- log_submit_activity -->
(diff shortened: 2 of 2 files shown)


jeremymasters:
How about just "No matches"?

RiseT:
@jeremymasters 
Very well. Without a period at the end, correct?

jeremymasters:
Yeah, sans period IMO.

moxie0:
Why do you like matches more than results? Maybe I'm out of touch, but in english I think people usually refer to "search results" rather than "search matches."

TimesEnemy:
Just to jump into this, the phrase, "No results found" does present a negative feeling to the user.  It starts with a negative word, which seems to be unavoidable in this situation.  Then it reinforces the negativity by saying, "No results."  People want results.  By using this phrase it is associating "No results" with the application, which is not ideal.  And to add "found," to "No results found," then stirs the possible feeling of being lost.

By using the phrase, "No matches" it is obviously stating that this is a no fire zone, and matches are not allowed here.  It is also moving the ownership of the query to the user by implying they are responsible for the lack of "matches" because they typed in something that does not match existing data.  Silly user.  It is also smaller, saving precious bits.

I realize it may seem I am being lighthearted in this post, and I am, however, these are valid concepts to be considered ... or not.

jeremymasters:
I think when we took Computer Science, this was the goal we had in mind...finding the correct phrasing for strings. :stuck_out_tongue_winking_eye:  
"Results" or "matches" works for me.
"No GIPHY Results" maybe?

RiseT:
@moxie0 Ugh. 😄 Difficult to put into words... Of course you're correct when you say that people usually refer to "search results". I'll try - lengthy explanations following. Actually *too* lengthy for such a short string... It wasn't my intention to make a big thing out of this - just a tiny bit of polishing. If you want to leave it in its original form, no problem for me.

- It's not about the actual language. There's a 1:1 equivalent for "search results" in German as well. "No results found" sounds somewhat weird to me in English *and* in German.
- As stated above, it's mostly a logical thing: 'In such a case, there is a "result" of the search, and that result is the fact that no "matches" were found.' This is what instantly stood out for me when reading that string for the first time.
- Besides of this, I'm slightly unsure what exactly makes that specific sentence sound strange. Is it the omission of "search" ("No [search] results found"), is it the negation of the sentence, or a combination of both.

I've now looked at Google's search engine and how it handles these situations, and interestingly, they seem to avoid a negation of "results" when nothing is found (don't know if they did this on purpose, though):

"About 1,010,000 **results**"
https://www.google.com/ncr#q=open+whisper+systems

"Your search - jgfdskljsdlkf - **did not match** any documents."
https://www.google.com/ncr#q=jgfdskljsdlkf

RiseT:
@TimesEnemy If that string isn't a factor in the user's suicide, a severe depression is the very least to expect...

TimesEnemy:
Yeah, I went out there on that last post....

From a logical perspective, it is acceptable to say, "No results found."  This is a very common statement, and the meaning/message is properly transferred to the user.

"No matches" has fewer characters.

Google technically calls searches a query, hence the "q" character within their URLs.  The word "query" is a better word, but has not been as accepted by the masses as the word, "search."  So, though technically Google calls it a query, publicly Google uses the term, "search."  This is most likely due to marketing and trying to own the term, "search," but either way they use this word.  When reading Google blogs, or Help files, or most of their texts that are intended for public consumption, they use the phrase, "search results."  The few times I do see Google using the word "matches" it is in statements similar to, "the results are unlikely to match...."  Also, whenever Google does use the "No matches" phrase, it seems to be in error and a source of amusement.

In English, "search results" flows and makes sense, "search matches" does not.  In the phrase, "No results found" the term "search" is implied.  I am not sure why it may sound odd to some.

[https://www.seroundtable.com/google-no-matches-found-answer-20682.html](https://www.seroundtable.com/google-no-matches-found-answer-20682.html)

And as @RiseT mentioned, right now Google uses, "Your search - [querystringgoeshere] - did not match any documents."

moxie0:
Maybe "Nothing found" ?

moxie0:
thanks for the attention to detail, in 3.22.0

Issue #4844, opened 2015-12-09
Self-sent messages get synced from Desktop to Android but not the other way


mawed:
Using the Signal App it is possible to send messages to the phone, but messages from the phone do not get sent, as they appear in-app as received directly.

2-4601:
Also with groups.
- on Android create a group with your cat
- your cat leaves the group
- on Android send a message to the group (you are the only member)
- Desktop does not receive the message

mawed:
Group behaviour also described in #3090

moxie0:
whoops

gulp21:
I only did a quick test, but removing the ``!forceSms && isSelfSend`` case in ``sendMediaMessage`` and ``sendTextMessage`` seems to fix this issue (ie. messages I send from Android to myself appear in the Desktop app). I’m not sure, though, whether it has some bad implications which I don’t see.

[Possible use cases for this feature: forward messages which contain important information to myself to make them easier to find; send notes to myself such that they are available on all synced devices]

simplestuf:
PS: As I don't have a smartphone, my wife and I were hoping to use this 'feature' to communicate with each other using her phone and my desktop.

Trolldemorted:
@simplestuf you can use [signal-cli](https://github.com/AsamK/signal-cli) as your root device and link signal-desktop instances to it. I did the same since i do not own a smartphone too.

But this issue looks easily solved, maybe @gulp21 can create a pull request?

simplestuf:
@Trolldemorted Thanks: signal-cli seems to be working for us.

Trolldemorted:
I created PR #5816 after having tested @gulp21's proposal for several days.

It would be great to get this fixed at last.

FeuRenard:
Related #4147

2-4601:
>Related #4147

I believe the view regarding self-sending in #4147 has changed, see https://github.com/WhisperSystems/Signal-Android/issues/3090#issuecomment-163483522

MavrikD:
Is the reverse feature being worked on? (mobile to desktop)

Trolldemorted:
I am currently not actively working on it, but these two PRs are awaiting review:

https://github.com/WhisperSystems/Signal-Android/pull/6209
https://github.com/WhisperSystems/Signal-Desktop/pull/959

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #5822, opened 2016-10-29
Fixes messages sent to self

Trolldemorted:

### Contributor checklist

  - Device A, Android X.Y.Z
  - Device B, Android Z.Y
  - Virtual device W, Android Y.Y.Z

---
### Description

Messages sent to self are now sent to all linked devices. Credits to gulp21 and effigies.
Fixes: #4844

// FREEBIE

I am very sorry for spamming you with notifications, but i messed up my local repository so bad that github does not allow me to open a new PR and does not notice pushes to the old one.

Nevertheless, checking for slavedevices looks like the optimal solution to me. Will test more thoroughly at work on monday, but this should do the trick.

### DIFF
--- a/src/org/thoughtcrime/securesms/sms/MessageSender.java
+++ b/src/org/thoughtcrime/securesms/sms/MessageSender.java
@@ -286,6 +286,10 @@ public class MessageSender {
       return false;
     }
 
+    if (TextSecurePreferences.isMultiDevice(context)) {
+      return false;
+    }
+
     return Util.isOwnNumber(context, recipients.getPrimaryRecipient().getNumber());
   }
 


moxie0:
Hey thanks @Trolldemorted but it's going to be way more complicated than this. The self-send check is in there for necessary reasons, so disabling it will just break self-send all together. A device can't have an encrypted session with itself, it would be both Alice and Bob, but is only designed to store state for one of them.

Trolldemorted:
@moxie0, are you really sure it is more complicated than this? This is precisely what Signal-Desktop and signal-cli are doing, and the api obviously supports this. A device can't have an encrpted session with itself, but linked Signal-Desktops/signal-clis are not the same device, after all.

I tested this PR now extensively a nexus4, a moto g, and a moto x, and it has been working flawlessly on each of them ever since. Could you have a second look?

Logs show no exceptions or anomalies, as you can see here:
http://pastebin.com/Xw5vEgP8

moxie0:
The change would have to not send messages to your own device id, only linked device ids.

Trolldemorted:
that is exactly what is [happening right now](https://github.com/WhisperSystems/libsignal-service-java/blob/master/java/src/main/java/org/whispersystems/signalservice/api/SignalServiceMessageSender.java#L404), isn't it?

A message to the device 1 is added to the `OutgoingPushMessageList` if and only if the recipient does not equal your local address. Should the recipient be equal to yourself, only messages to the subdevices are added.

Trolldemorted:
@moxie0 btw, if it would open a session with itself, we should see a `MismatchedDevicesException` in the logs. I am really really sure it does not.

PR #5841, opened 2016-11-06
Lower memory consumption of Giphy browser

### LINKED ISSUE #5831: Giphy browser crashes
I have:

### Bug description
While scrolling down the GIF list, after ≈ 10 pages Signal crashes. But you have to wait for all GIFs to load completely.

### Steps to reproduce
- open conversation
- go to GIPHY browser
- select the two column GIF view in the top right corner
- wait until all GIFs loaded (then all of them are animated)
- scroll down ≈1 page
- wait until all GIFs loaded
- scroll down ≈1 page
- ...

**Actual result:** Signal crashes. Sometimes with Android alert dialog that Signal crashed, sometimes without it.
**Expected result:** Signal doesn't crash.

### Device info
**Device:** Fairphone 2 (2GB RAM)
**Android version:** 5.1.1
**Signal version:** 3.21.3

### Link to debug log
https://gist.github.com/anonymous/3d7e7d0e928153575b443a8d8514b8d2

FeuRenard:
### Contributor checklist

 * AVD Nexus 5X, Android 6.0
 * AVD Nexus 5X, Android 5.1

----------

### Description
First of all: showing multiple GIFs is very memory consuming. On Android they are loaded into memory as bitmap and this needs much memory (and causes e.g. this oblong discussion #672).

Nevertheless we want to do some caching and keep GIFs we once downloaded for a while. RecyclerView attempts to accomplish caching for us. It holds views to be reused later (scroll upwards). Then the view showing the GIF can be recycled (loaded from memory cache). But doing this in memory with the bitmaps is not a good idea as they are rather large. We should perform another way of caching.

There is also Glide which we use to load the GIFs into the RecyclerView. It also offers to cache images (the GIFs itself, I assume, not the large bitmaps). Glide offers to do this caching on disk instead of in memory. This alone should be sufficient for our purpose.

tl;dr
So **this PR marks the GIF views as not recyclable**. So RecyclerView doesn't cache them in memory for later reuse. But the GIFs are still on disk so that they don't have to be downloaded every time we scroll by them. This caching was already implemented before.

Memory consumption (with all GIFs loaded/animated):

- before: 95-150MB
- after: 70-100MB, sometimes up to 110MB because of a lazy GC, I suppose

Fixes #5831

### DIFF
--- a/src/org/thoughtcrime/securesms/giph/ui/GiphyAdapter.java
+++ b/src/org/thoughtcrime/securesms/giph/ui/GiphyAdapter.java
@@ -46,7 +46,6 @@ public class GiphyAdapter extends RecyclerView.Adapter<GiphyAdapter.GiphyViewHol
 
     GiphyViewHolder(View view) {
       super(view);
-      super.setIsRecyclable(false);
       thumbnail   = ViewUtil.findById(view, R.id.thumbnail);
       gifProgress = ViewUtil.findById(view, R.id.gif_progress);
       thumbnail.setOnClickListener(this);
@@ -155,6 +154,12 @@ public class GiphyAdapter extends RecyclerView.Adapter<GiphyAdapter.GiphyViewHol
            .listener(holder)
            .into(holder.thumbnail);
     }
+  }
+
+  @Override
+  public void onViewRecycled(GiphyViewHolder holder) {
+    super.onViewRecycled(holder);
+    Glide.clear(holder.thumbnail);
   }
 
   @Override
--- a/src/org/thoughtcrime/securesms/giph/ui/GiphyAdapter.java
+++ b/src/org/thoughtcrime/securesms/giph/ui/GiphyAdapter.java
@@ -46,6 +46,7 @@ public class GiphyAdapter extends RecyclerView.Adapter<GiphyAdapter.GiphyViewHol
 
     GiphyViewHolder(View view) {
       super(view);
+      super.setIsRecyclable(false);
       thumbnail   = ViewUtil.findById(view, R.id.thumbnail);
       gifProgress = ViewUtil.findById(view, R.id.gif_progress);
       thumbnail.setOnClickListener(this);


moxie0:
Thanks for looking into this. It seems like you're saying that excess memory consumption is coming from references maintained by RecyclerView's cache of unused items.

The purpose of RecyclerView caching isn't to cache the GIFs (it doesn't), but simply to cache/reuse the view objects as the list scrolls.  Each item is being "rebound" every time it gets recycled, so there is no GIF caching (you're correct that Glide does that for us).

Instead of turning off view recycling, let's try freeing up the memory from cached views:

`````
@Override
public void onViewRecycled(GiphyViewHolder viewHolder) {
super.onViewRecycled(viewHolder);
Glide.clear(viewHolder.thumbnail);
}
`````

Can you check how that works?

FeuRenard:
> It seems like you're saying that excess memory consumption is coming from references maintained by RecyclerView's cache of unused items.

I saw bitmaps referencing `RecyclerView.RecycledViewPool`. That led me to the conclusion that there is the memory going to.
I'll try your code.

FeuRenard:
Memory consumption of the proposed solution:

- 95-120MB, but I also reached 160MB one time

So my solution seems to be more effective (in terms of releasing memory) but your solution is the correct one (looking at the [documentation](https://developer.android.com/reference/android/support/v7/widget/RecyclerView.Adapter.html#onViewRecycled(VH))).

> If an item view has large or expensive data bound to it such as large bitmaps, [onViewRecycled(...)] may be a good place to release those resources.

Further approaches:

- add `holder.thumbnail.setImageDrawable(null)` to `onViewRecycled(...)`: does seem to have small to no effect
- add `.skipMemoryCache(true)` to `Glide` when setting the GIF: does seem to have no effect
- set `this.recyclerView.setItemViewCacheSize(0);` in `GiphyFragment`: saves some MB, but no more than 10MB

I haven't any time left today. I will be able to play around some more, if I find some time this week.

FeuRenard:
Why is the correct approach @moxie0 proposed consuming more memory than my initial & "false" approach?
Turns out: It only cleans up the bitmaps in the RecyclerView cache (RecycledViewPool). But there is also a recycling cache (pool) within Glide. But as Glide's author points out, we shouldn't mess with Glide's pool. We should trust it:

> From the images it looks like you're observing the normal pooling. Try to get a heap dump / memory usage count after calling `Glide.get(context).clearMemory()`, that should free up space. But it shouldn't be called in production as pooling is why Glide is so fast. If you think about it if an app doesn't use the memory it's given then it's wasting it. It's much easier to grab an already allocated ~1M Bitmap than create a new one. If you try Glide on a lower end phone you should see that the pool is smaller to accommodate for the constraints.
[source: https://github.com/bumptech/glide/issues/1104#issuecomment-205009181, very similar problem to ours]

I tried the suggested method call `Glide.get(context).clearMemory()` ("dirty tweak"), and with this called on every recycling we have the same memory consumption as with my initial approach. But we shouldn't do this.

summary of memory consumptions:

- before: 95-150MB
- my initial approach: 70-100MB
- @moxie0's correct suggestion: 95-120MB
- @moxie0's correct suggestion, with a dirty tweak: 70-100MB

Unfortunately I didn't accomplish to setup an AVD with a low memclass to test if Glide's pool really "accomodate[s] for the constraints". This remains to be done.

moxie0:
in 3.22.0 thanks for all the research!

PR #5877, opened 2016-11-20
If Signal isn't enabled for receiving SMS, don't allow sending SMS

moxie0:


moxie0:
in 3.23.0

p4nci:
@moxie0 
I can still send a SMS (this is true only for Signal users, change mode to 'Insecure SMS' and send SMS). Is/was this intended?
(Signal v3.23.0, SMS Off)

moxie0:
yeah that's fine thanks

PR #5905, opened 2016-11-30
Fixes messages sent to self (attempt 3)

Trolldemorted:

### Contributor checklist

 * Nexus 4, Android 6.0.1
 * Motorola whatever, Android 6.0.1

----------

### Description

Messages sent to self are now sent to all linked devices. Credits to gulp21 and effigies.
Fixes: #4844

Third time's a charm! Since i cannot reopen the old PR because reasons, so i open this one and go more into detail on why it works. @moxie0's objection was that this would cause Signal-Android to attempt to create sessions with itself.

However, when creating the `OutgoingPushMessageList` the messagesender [does the following](https://github.com/WhisperSystems/libsignal-service-java/blob/master/java/src/main/java/org/whispersystems/signalservice/api/SignalServiceMessageSender.java#L402):
```
List<OutgoingPushMessage> messages = new LinkedList<>();

if (!recipient.equals(localAddress)) {
  messages.add(getEncryptedMessage(socket, recipient, SignalServiceAddress.DEFAULT_DEVICE_ID, plaintext, legacy, silent));
}

for (int deviceId : store.getSubDeviceSessions(recipient.getNumber())) {
  messages.add(getEncryptedMessage(socket, recipient, deviceId, plaintext, legacy, silent));
}

return new OutgoingPushMessageList(recipient.getNumber(), timestamp, recipient.getRelay().orNull(), messages);
```

Currently, Signal-Android always is the master device which always has the default device-id (1).
As you see, libsignal adds messages to all slave devices (since they are in the store's subdevicesessions).
A message to the master device is added if **and only if** `!recipient.equals(localAddress)` is true, which is obviously not true if we send a message to ourself. So if we do that, only messages to our slaves are added, and we never attempt to create a session with ourselves.

I have tested this commit on my own and on several of my friends' phones, and it is working fine.
If you are worried that in the future Signal-Android's device-id may not be 1, do not worry, [here ](https://github.com/Turasa/libsignal-service-java/blob/a8a61996627b0fcebd881d189dc4da83a64d4ef2/java/src/main/java/org/whispersystems/signalservice/api/SignalServiceMessageSender.java#L434)is a fix for that which you could merge or incorporate.

### DIFF
--- a/src/org/thoughtcrime/securesms/sms/MessageSender.java
+++ b/src/org/thoughtcrime/securesms/sms/MessageSender.java
@@ -286,6 +286,10 @@ public class MessageSender {
       return false;
     }
 
+    if (TextSecurePreferences.isMultiDevice(context)) {
+      return false;
+    }
+
     return Util.isOwnNumber(context, recipients.getPrimaryRecipient().getNumber());
   }
 


moxie0:
doesn't this mean that if you have a linked device you would no longer "receive" the message you sent to yourself, see a delivery receipt, etc?

Trolldemorted:
The message will be sent to all linked devices as a normal outgoing message. Signal-Android saves it as outgoing and displays it accordingly:

signaldroid


Signal-Desktop saves it as incoming and also displays it fine:

selfmessages


Sending a sync-sent message to the slave devices is not necessary, as they receive the original one, and sending a sync-sent message would only duplicate the content.
Signal-Desktop unfortunately does send one as you see: The content is displayed *once* on the sending Signal-Desktop and *twice* on every other linked device (other linkedSignal-Desktop instances as well).

I think that's hideous, my PR to [stop the sync-sent messages](https://github.com/WhisperSystems/Signal-Desktop/pull/959) on Signal-Desktop is waiting for liliakai's review.

moxie0:
I think what we're going for is the normal unliked self-send view: the outgoing message, and then the incoming version of the same message.  Just as if you sent something to yourself and received what you sent.

Trolldemorted:
Are you really really sure you want that?

I can enable sync-sent messages for Signal-Android, but that would require changes to libsignal-service-java. Sync messages are currently not sent when the recipient is you, because then the [Signal-Server tells Signal-Android to not to sync it](https://github.com/WhisperSystems/Signal-Server/blob/917eaa50fbace93fde31b184adf2895d29245f63/src/main/java/org/whispersystems/textsecuregcm/controllers/MessageController.java#L107) and [it complies](https://github.com/WhisperSystems/libsignal-service-java/blob/master/java/src/main/java/org/whispersystems/signalservice/api/SignalServiceMessageSender.java#L119). Furthermore, since the sync message cannot be sent to Signal-Android itself, we'd have to fake one in the local storage.

If you *really* prefer it that way, i could open up a PR for libsignal that addresses that (Signal-Desktop would need some work as well - it does not do that either, the sending S-D displays the message only once, since it does not receive its own sync message). But i have to say that in my humble opinion duplicating selfmessages will do nothing but confuse and obstruct the user. I am using selfmessages frequently to send sensitive information (like usernames, passwords, IPs, you name it) between two Signal-Desktop instances, and believe me, the sync-sent messages grind my gears every day (hence i created a PR that supresses them).

I sincerely dislike the ratio of gain and effort on implementing that. Is there any way we can convince you that one message should only be displayed once, no matter who the recipient is? I honestly have never seen a messenger that does that, and never seen a human that would prefer that.

mawed:
I also use the sync Desktop-Android and I like it, that desktop only keeps the sent message. On android I delete the sent-synced message all the time as it is harder to find stuff. When sending media the storage is a second thing considering if the fie will be there twice.

My wish would be to get the sent-sync on the sending client and the received-message on all other clients, so I could distinguish on which device I sent the message.

Trolldemorted:
@mawed your own outgoing messages on the sending device are not from sync-sent messages (that would be unnecessary, since they just save the outgoing message).

What you want is a normal message to the sibling devices, and no sent-sync messages, and that is what my PR is doing: The sender saves the outgoing message, the receivers save the incoming message. Every client has a correct view on what happened, no message was duplicated, and nobody needs sync messages.

fsck:
I agree with @mawed that the information from which device the message was send would be nice to have. Also I appreciate the declobbered look of the deduplicated conversation.

moxie0:
The reason the android client works the way it does right now is because people would often install Signal, immediately send themselves a message, and conclude that Signal doesn't work when they didn't get a message "back."  This was an ongoing problem, which is why there is so much code to deal with it.  It would have been much easier just to display the outgoing message without mirroring it as an incoming message as well.

I don't particularly care one way or another, I just don't want people concluding that Signal doesn't work or filing issues when they don't get a message "back."  So if you want to change the behavior, you'd have to also change people's expectations, perhaps by changing the activity so that it's branded as "notes" or something.

Trolldemorted:
wouldn't immediately marking these messages as "received a delivery receipt" suffice?

mawed:
@moxie0 What about 2 cases:
-If a user has no linked devices, keep it as it is.
-If a user has linked devices, use the new process.

With this every user gets a received message (and nobody thinks, it's broken)

Trolldemorted:
@mawed again, this is precisely what my PR is doing^^

mawed:
@Trolldemorted Oh sorry, I did not look into the code, I thought it would just never show the "received"-message on Android.

moxie0:
@mawed Consistency is key. We need to pick a way that we want this to work, make it work that way consistently within the Android app, and then apply that same consistency to the other clients.

In the past, people have felt that "Signal is broken" if they don't get a message "back" when they send one to themselves.  We need to either meet that expectation, or change the UX so that people don't have that expectation to begin with.  Showing a delivery receipt for a message you didn't "receive" is just as likely to make people think delivery receipts are broken.  I think we either need to rebrand self-send as a "notes" kind of thing, or ensure that messages you send come "back".

Trolldemorted:
I will go with anything that does not duplicate messages (and delivers them successfully to my S-D instances). 👍 

What kind of UX do you suggest? Removing yourself from the contact list (if present) and adding a "Notes"-item to the top?

Delivery receipts should be shown when there are no linked devices, or when at least one of them has sent one. Do you agree? (that would be consistent with foreign delivery receipts)
This would require you to [enable them again](https://github.com/WhisperSystems/Signal-Desktop/issues/988#issuecomment-264510031), though. And could you update the Signal-Server sourcecode on github? I spent half of this day digging through the sources of libsignal, Signal-Desktop and SIgnal-Server to figure out where my precious delivery receipts have gone!

moxie0:
I think some kind of "notes" or "memo" thing. You'd have to mock it up in photoshop or something and post a proposal.

Trolldemorted:
how would you like to access your notes, then? Would a fixed entry in the contacts list be ok?

Since the access should be comparable to the one on Signal-Desktop and we do not want to burrow it behind menus, we need some kind of entry in the conversation list activity, so adding something™ to the conversation list would be both easy to implement and easy to grasp and use.

Trolldemorted:
After some thinking, i think we should keep the Memo inside the ConversationList, so that it will move to the bottom if not used. Less work for us, no changes to the UI behaviour for the user, no vertical space at the top wasted for features the user might not use.

notes

I propose these UX changes:
- add Memo/Notebook/Whatever on installation, "send" hello world message with brief explanation, don't care whether the user itself is in his/her contacts
- show "Memo" (instead of the name i gave myself in my contacts) and hide number in the conversation header
- send messages directly (non-sync) to linked devices if available
- display message once as outgoing otherwise
- display outgoing memo messages on the right, and incoming ones on the left
- display confirmed delivery if either no slaves linked or a delivery receipt was received
- disable/hide the redphone button for Memo (until multidevices are around or S-D has redphone support)
- disable/hide the reset secure session button
- disable the on-click menu that opens when you click on an avatar in the contact list
- remove "verify safety numbers" and "block" from the conversation settings menu
- (optional) a custom avatar


What do you think?

moxie0:
why would there ever be an incoming memo message?

Trolldemorted:
from a linked device, of course!

mawed:
Even if the whole process will be changed on the long run.
Would it be possible in the short run to really send the messages instead of just copying, so the messages can be received on desktop?

moxie0:
I don't think it would make sense to have sent messages from the desktop appear as "received" messages in the app. That'd be pretty inconsistent with the way the rest of the app functions, where individual physical clients are treated as one logical entity.

Trolldemorted:
You are right, it is inconsistent with the rest of the app indeed, i will try to display all self-received messages as outgoing (all should be ok since there is no reason any message from your own account should be displayed as outgoing).

And while we are at it, what was the rationale behind sync-sent messages? Can we quit sending them altogether and send every (i.e. also non-self messages) message directly to all siblings instead, if the receiving agent displays it as outgoing?

moxie0:
@Trolldemorted the sibling needs to know who to display the outgoing message to.

Closing this for now, please reopen when you've adjusted thanks!

Trolldemorted:
is there any scenario where the sibling will *not* display the message as outgoing?

Issue #5445, opened 2016-04-17
Please support sending stickers


rbalint:
I have:
- [x] searched open and closed issues for duplicates
- [x] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

---
### Bug description

Users can't send stickers to each other using Signal. This is a deal breaker for some users preventing switching to Signal from other messenger apps such as Viber.

Adding a sticker store may also help funding the project.
### Screenshots

![screen-shot-2013-04-16-at-10 57 46-am1](https://cloud.githubusercontent.com/assets/1520835/14586175/55a420be-0490-11e6-915c-d7844febe1de.png)
Ths screenshots are copied from here:
http://www.adweek.com/socialtimes/facebook-updates-iphone-app-with-stickers-and-chat-heads-ipad-gets-new-design/292460

rhetr:
some open source discussions about stickers:

https://wiki.tox.chat/developers/feature_ideas/stickers
https://github.com/qTox/qTox/issues/402
https://github.com/irungentoo/toxcore/issues/1106
https://github.com/Gargron/xmpp-web/issues/3
https://www.stickerstelegram.com/
http://telegramhub.net/

Fnatte79:
I agree, it is a deal breaker for many.
For time being I use both Telegram and Signal but still have a few friends who not seems to wanna change to Signal just because it doesn't support stickers.

hermannsblum:
*push*
I find this THE dealbraker when trying to convince people to switch from telegram to signal.

nrizzio:
This has been implemented in 69d0242574cc68b068afe3a00861e574b176f08b (Giphy support).

2-4601:
Is the Stickers tab in the Giphy browser the same as the sticker feature requested in this issue? I have really no idea... :thinking:

rbalint:
Thanks for improving Signal, Giphy support is indeed a related addition.
I think to claim this case resolved Signal needs a more user friendly method of sending ''stickers' and on the receiver side the sticker should not be in a bounding box rectangle because it is less nice than Viber's solution.
I also tried sending PNG-s with alpha background and only a black box is shown in the conversation.

nrizzio:
Fixed by 1b44bdc (for sure!)

2-4601:
[...]

PR #5943, opened 2016-12-10
Support for stickers and scribbles

moxie0:


2-4601:
Fixes #5445.

Is this still WIP? Tried this but I can't find where to browse the stickers or how to activate the scribble mode?
drawer

moxie0:
@2-4601 If you "attach" an image, there should be a little "edit" pencil in the bottom right corner that you can tap on. That takes you to an activity you can add stickers, text, and draw on.

2-4601:
Oh, I see it now.

#### A bug with the pencil color display:
1. Open scribble mode
1. Tap the text input icon (Tt) and notice that the color is red
1. Keyboard opens, dismiss it with the back button
1. Tap the brush icon (notice that the first tap doesn't select it but the color of the status bar changes?)
1. Tap the brush icon again (it gets selected now) and notice that the color is red
1. Change the color from the color slider to purple

**Actual result:** The pencil icon is still red
**Expected result:** The pencil icon should be purple

moxie0:
@2-4601 good catches! should be fixed

FeuRenard:
Yes, fixed: brush icon shows color change & brush icon doesn't have to be hit twice.

On my real device (3.5 inch display) the color slider is sometimes hard to hit with my finger. Works 90% fine after some practice :wink:.

2-4601:
The brush color switching issue is solved for me now too. :+1: 

This latest commit introduced something strange. Now every time I start the scribble mode the keyboard is open by default. When I switch back to a build without this latest commit the keyboard is not open when the scribble mode starts. This happens only with my Nvidia K1 6.0.1 device. On my 5.1.1 Lollipop device the keyboard is never open by default. I'm using the Google keyboards on both devices.

One feature I would like to see is white in the color slider. I've been drawing stuff and I really miss that there's no white color. From UX point of view I would expect it to be at the bottom of the slider because black is on top.

moxie0:
@2-4601 Thanks I'll look into the keyboard thing, think I know what to do.  

To include white, we'd have to change the "text" and "brush" (and "emoji" for consistency) icons to have a slight drop shadow or outline. Otherwise it'd be white on white. 

There's no way to tint drawables like that (that I know of), so we'd need to create modified versions of the actual png actual assets.  I'm afraid I might not have it in me at the moment, but if someone with photoshop skills does it and submits a PR against this branch I'll add white.  =)

moxie0:
@2-4601 Can you test that most recent change and see if the keyboard is still open on that device by default?

2-4601:
The keyboard issue is fixed now :+1: 

I see adding white isn't as simple as I thought. Well, I'll just have to settle for light yellow to draw snow for now. I hope it doesn't look too much like piss :D

moxie0:
@2-4601 there's a snowflake sticker!

moxie0:
in 3.25.0

PR #5962, opened 2016-12-17
Remove the hard dependency on Google Play Services

eightbitkid:

### Contributor checklist
 * Nexus 5X (running build NMF26O of CopperheadOS), Android 7.1.1
 * Blu Diamond M, Android 6.0

----------

### Description

Taken from one of the endless discussions that have already taken place on this point @moxie0 has said:

 > If the only thing that the remaining people here want out of LibreSignal is a websocket-only solution and gmscore isn't an option for whatever reason, I would consider a clean, well written, and well tested PR for websocket-only support in Signal. I expect it to have high battery consumption and an unreliable user experience, but would be fine with it if it comes with a warning and only runs in the absence of play services.

This pull request is my attempt at fulfilling that offer. I stress that I'm not concerned with removing Google from Signal completely, I just want to run it on my phone. My friends are not interested in computers and I'd rather have an app that looks after them, than try an teach them something they don't care about.

#### What does this pull request do?

In short this series of patches:

 * Creates a new method `isRegistered` to determin if a user is registered, and reserves `isPushRegistered` for checks about receiving push notifications;

 * Moves the warning that takes place just before an attempted registration without Google Play Services to the start of that activity, and allows the user to continue once they have acknowledged it; and

 * Informs the server that a device without Google Play Services is unable to receive voice calls because it'll never get the notification.

 * Provides the option for the user to disable battery optimisations for Signal (but only when Google Services are not available).

These changes mean that __a device with Play Services installed functions identically to how it does right now__, while a device without those services gains the ability use the app.

#### Why is "_My contribution is fully baked and ready to be merged as is_" left unticked?

While this pull request has been rebased and organised so it applies cleanly to the current mainline it is relient on [a patch to a libsignal-service-java][2]. Once this patch is accepted this request can update its build script.

#### Results of testing

I've been running these changes on CopperheadOS for several days and the app has performed as expected. With the exception of the loss of voice communication there have been no issues.

Testing on a platform with Google's binaries installed has been shorter but all of the functionality has been tested: there is no warning of missing packages, messages send and receive correctly (both to single users, and to groups), encrypted calls function as normal.

Both platforms register, and unregister from Signal's servers correctly.

I'm happy with this implementation. Messages are received in a timely fashion, battery life is nowhere near as bad as I thought it would be, and the app copes completely with the loss of Internet access (both dropping a 4G connection wandering around town, and turning mobile data/Wi-Fi off).

I did run into Android's doze/sleep functionality once: I left my phone not charging over night. I had no messages in the morning until I checked Signal manually, then they all came through. However, now that the option exists to combat this (in a small way) I don't see this as much of a problem.

Battery life is still good. I still get three days of my phone with no charging between.

   [1]: https://github.com/LibreSignal/LibreSignal/issues/37#issuecomment-226646872
   [2]: https://github.com/WhisperSystems/libsignal-service-java/pull/28

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -85,6 +85,9 @@
     <uses-permission android:name="android.permission.READ_LOGS"/>
     <uses-permission android:name="android.permission.WRITE_CALL_LOG"/>
 
+    <!-- Required network connection if Google Play is missing -->
+    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS"/>
+
     <permission android:name="org.thoughtcrime.securesms.permission.C2D_MESSAGE"
                 android:protectionLevel="signature" />
     <uses-permission android:name="org.thoughtcrime.securesms.permission.C2D_MESSAGE" />
--- a/src/org/thoughtcrime/securesms/RegistrationActivity.java
+++ b/src/org/thoughtcrime/securesms/RegistrationActivity.java
@@ -67,6 +67,16 @@ public class RegistrationActivity extends BaseActionBarActivity {
   }
 
   @Override
+  public void onResume() {
+    super.onResume();
+
+    if (!(GooglePlayServicesUtil.isGooglePlayServicesAvailable(this) == ConnectionResult.SUCCESS)) {
+      Dialogs.showAlertDialog(this, getString(R.string.RegistrationActivity_unsupported),
+                              getString(R.string.RegistrationActivity_sorry_this_device_is_not_supported_for_data_messaging));
+    }
+  }
+
+  @Override
   protected void onActivityResult(int requestCode, int resultCode, Intent data) {
     if (requestCode == PICK_COUNTRY && resultCode == RESULT_OK && data != null) {
       this.countryCode.setText(data.getIntExtra("country_code", 1)+"");
@@ -208,18 +218,6 @@ public class RegistrationActivity extends BaseActionBarActivity {
                              getString(R.string.RegistrationActivity_invalid_number),
                              String.format(getString(R.string.RegistrationActivity_the_number_you_specified_s_is_invalid),
                                            e164number));
-        return;
-      }
-
-      int gcmStatus = GooglePlayServicesUtil.isGooglePlayServicesAvailable(self);
-
-      if (gcmStatus != ConnectionResult.SUCCESS) {
-        if (GooglePlayServicesUtil.isUserRecoverableError(gcmStatus)) {
-          GooglePlayServicesUtil.getErrorDialog(gcmStatus, self, 9000).show();
-        } else {
-          Dialogs.showAlertDialog(self, getString(R.string.RegistrationActivity_unsupported),
-                                  getString(R.string.RegistrationActivity_sorry_this_device_is_not_supported_for_data_messaging));
-        }
         return;
       }
 
--- a/src/org/thoughtcrime/securesms/jobs/RefreshAttributesJob.java
+++ b/src/org/thoughtcrime/securesms/jobs/RefreshAttributesJob.java
@@ -3,6 +3,9 @@ package org.thoughtcrime.securesms.jobs;
 import android.content.Context;
 import android.util.Log;
 
+import com.google.android.gms.common.ConnectionResult;
+import com.google.android.gms.common.GooglePlayServicesUtil;
+
 import org.thoughtcrime.redphone.signaling.RedPhoneAccountAttributes;
 import org.thoughtcrime.redphone.signaling.RedPhoneAccountManager;
 import org.thoughtcrime.securesms.dependencies.InjectableType;
@@ -42,10 +45,14 @@ public class RefreshAttributesJob extends ContextJob implements InjectableType {
     String gcmRegistrationId = TextSecurePreferences.getGcmRegistrationId(context);
     int    registrationId    = TextSecurePreferences.getLocalRegistrationId(context);
 
-    String token = textSecureAccountManager.getAccountVerificationToken();
+    if (GooglePlayServicesUtil.isGooglePlayServicesAvailable(context) == ConnectionResult.SUCCESS) {
+      String token = textSecureAccountManager.getAccountVerificationToken();
 
-    redPhoneAccountManager.createAccount(token, new RedPhoneAccountAttributes(signalingKey, gcmRegistrationId));
-    textSecureAccountManager.setAccountAttributes(signalingKey, registrationId, true);
+      redPhoneAccountManager.createAccount(token, new RedPhoneAccountAttributes(signalingKey, gcmRegistrationId));
+      textSecureAccountManager.setAccountAttributes(signalingKey, registrationId, true, false);
+    } else {
+      textSecureAccountManager.setAccountAttributes(signalingKey, registrationId, false, true);
+    }
   }
 
   @Override
--- a/src/org/thoughtcrime/securesms/service/RegistrationService.java
+++ b/src/org/thoughtcrime/securesms/service/RegistrationService.java
@@ -246,21 +246,24 @@ public class RegistrationService extends Service {
 
     setState(new RegistrationState(RegistrationState.STATE_GCM_REGISTERING, number));
 
-    String gcmRegistrationId = GoogleCloudMessaging.getInstance(this).register(GcmRefreshJob.REGISTRATION_ID);
-    accountManager.setGcmId(Optional.of(gcmRegistrationId));
-
-    TextSecurePreferences.setGcmRegistrationId(this, gcmRegistrationId);
     TextSecurePreferences.setWebsocketRegistered(this, true);
 
     DatabaseFactory.getIdentityDatabase(this).saveIdentity(self.getRecipientId(), identityKey.getPublicKey());
     DirectoryHelper.refreshDirectory(this, accountManager, number);
 
-    RedPhoneAccountManager redPhoneAccountManager = new RedPhoneAccountManager(BuildConfig.REDPHONE_MASTER_URL,
-                                                                               new RedPhoneTrustStore(this),
-                                                                               number, password);
+    if (GooglePlayServicesUtil.isGooglePlayServicesAvailable(this) == ConnectionResult.SUCCESS) {
+      String gcmRegistrationId = GoogleCloudMessaging.getInstance(this).register(GcmRefreshJob.REGISTRATION_ID);
+      accountManager.setGcmId(Optional.of(gcmRegistrationId));
 
-    String verificationToken = accountManager.getAccountVerificationToken();
-    redPhoneAccountManager.createAccount(verificationToken, new RedPhoneAccountAttributes(signalingKey, gcmRegistrationId));
+      TextSecurePreferences.setGcmRegistrationId(this, gcmRegistrationId);
+
+      RedPhoneAccountManager redPhoneAccountManager = new RedPhoneAccountManager(BuildConfig.REDPHONE_MASTER_URL,
+                                                                                 new RedPhoneTrustStore(this),
+                                                                                 number, password);
+
+      String verificationToken = accountManager.getAccountVerificationToken();
+      redPhoneAccountManager.createAccount(verificationToken, new RedPhoneAccountAttributes(signalingKey, gcmRegistrationId));
+    }
 
     DirectoryRefreshListener.schedule(this);
     RotateSignedPreKeyListener.schedule(this);
--- a/src/org/thoughtcrime/securesms/service/RegistrationService.java
+++ b/src/org/thoughtcrime/securesms/service/RegistrationService.java
@@ -194,8 +194,9 @@ public class RegistrationService extends Service {
     }
 
     try {
-      String password     = Util.getSecret(18);
-      String signalingKey = Util.getSecret(52);
+      String  password     = Util.getSecret(18);
+      String  signalingKey = Util.getSecret(52);
+      boolean hasServices  = GooglePlayServicesUtil.isGooglePlayServicesAvailable(this) == ConnectionResult.SUCCESS;
 
       initializeChallengeListener();
 
@@ -205,7 +206,7 @@ public class RegistrationService extends Service {
 
       setState(new RegistrationState(RegistrationState.STATE_VERIFYING, number));
       String challenge = waitForChallenge();
-      accountManager.verifyAccountWithCode(challenge, signalingKey, registrationId, true);
+      accountManager.verifyAccountWithCode(challenge, signalingKey, registrationId, hasServices, !hasServices);
 
       handleCommonRegistration(accountManager, number, password, signalingKey);
       markAsVerified(number, password, signalingKey);
--- a/src/org/thoughtcrime/securesms/service/RegistrationService.java
+++ b/src/org/thoughtcrime/securesms/service/RegistrationService.java
@@ -291,12 +291,14 @@ public class RegistrationService extends Service {
 
     if (verifying) {
       TextSecurePreferences.setPushRegistered(this, false);
+      TextSecurePreferences.setRegistered(this, false);
     }
   }
 
   private void markAsVerified(String number, String password, String signalingKey) {
     TextSecurePreferences.setVerifying(this, false);
     TextSecurePreferences.setPushRegistered(this, true);
+    TextSecurePreferences.setRegistered(this, true);
     TextSecurePreferences.setLocalNumber(this, number);
     TextSecurePreferences.setPushServerPassword(this, password);
     TextSecurePreferences.setSignalingKey(this, signalingKey);
--- a/src/org/thoughtcrime/securesms/service/RegistrationService.java
+++ b/src/org/thoughtcrime/securesms/service/RegistrationService.java
@@ -10,6 +10,8 @@ import android.os.Handler;
 import android.os.IBinder;
 import android.util.Log;
 
+import com.google.android.gms.common.ConnectionResult;
+import com.google.android.gms.common.GooglePlayServicesUtil;
 import com.google.android.gms.gcm.GoogleCloudMessaging;
 
 import org.thoughtcrime.redphone.signaling.RedPhoneAccountAttributes;
@@ -297,7 +299,11 @@ public class RegistrationService extends Service {
 
   private void markAsVerified(String number, String password, String signalingKey) {
     TextSecurePreferences.setVerifying(this, false);
-    TextSecurePreferences.setPushRegistered(this, true);
+
+    if (GooglePlayServicesUtil.isGooglePlayServicesAvailable(this) == ConnectionResult.SUCCESS) {
+      TextSecurePreferences.setPushRegistered(this, true);
+    }
+
     TextSecurePreferences.setRegistered(this, true);
     TextSecurePreferences.setLocalNumber(this, number);
     TextSecurePreferences.setPushServerPassword(this, password);
--- a/src/org/thoughtcrime/securesms/util/TextSecurePreferences.java
+++ b/src/org/thoughtcrime/securesms/util/TextSecurePreferences.java
@@ -65,6 +65,7 @@ public class TextSecurePreferences {
   private static final String THREAD_TRIM_ENABLED              = "pref_trim_threads";
   private static final String LOCAL_NUMBER_PREF                = "pref_local_number";
   private static final String VERIFYING_STATE_PREF             = "pref_verifying";
+  public  static final String REGISTERED_PREF                  = "pref_registered";
   public  static final String REGISTERED_GCM_PREF              = "pref_gcm_registered";
   private static final String GCM_PASSWORD_PREF                = "pref_gcm_password";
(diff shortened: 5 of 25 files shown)


src/org/thoughtcrime/securesms/util/TextSecurePreferences.java
```diff
@@ -400,11 +408,23 @@ public static void setVerifying(Context context, boolean verifying) {
...
+  public static boolean isRegistered(Context context) {
+    return getBooleanPreference(context, REGISTERED_PREF, false);
+  }
+
+  public static void setRegistered(Context context, boolean registered) {
+    setBooleanPreference(context, REGISTERED_PREF, registered);
+
+    if (isGooglePlayServicesAvailable(context)) {
+      setPushRegistered(context, registered);
+    }
```
moxie0: No side effects please

src/org/thoughtcrime/securesms/util/TextSecurePreferences.java
```diff
@@ -400,11 +408,23 @@ public static void setVerifying(Context context, boolean verifying) {
     setBooleanPreference(context, VERIFYING_STATE_PREF, verifying);
   }
 
+  public static boolean isRegistered(Context context) {
+    return getBooleanPreference(context, REGISTERED_PREF, false);
+  }
```
moxie0: Looks like this means that all existing users are unregistered?

src/org/thoughtcrime/securesms/service/RegistrationService.java
```diff
@@ -192,8 +192,12 @@ private void handleSmsRegistrationIntent(Intent intent) {
     }
 
     try {
-      String password     = Util.getSecret(18);
-      String signalingKey = Util.getSecret(52);
+      String  password      = Util.getSecret(18);
+      String  signalingKey  = Util.getSecret(52);
+      boolean hasServices   = TextSecurePreferences.isGooglePlayServicesAvailable(this);
+      boolean hasVoice      = hasServices;
+      boolean fetchMessages = !hasServices;
```
moxie0: Please don't define these here if they're only used once

src/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -64,6 +64,11 @@ public void onCreate(Bundle icicle) {
     initializeResources();
     initializeSpinner();
     initializeNumber();
+
+    if (!TextSecurePreferences.isGooglePlayServicesAvailable(this)) {
+      Dialogs.showAlertDialog(this, getString(R.string.RegistrationActivity_unsupported),
+                              getString(R.string.RegistrationActivity_sorry_this_device_is_not_supported_for_data_messaging));
```
moxie0: This seems like the wrong string

src/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -64,6 +64,11 @@ public void onCreate(Bundle icicle) {
...
+    }
```
moxie0: This shouldn't be happening in onCreate

thestinger:
Note that if this is relying on maintaining a connection in the background it's going to run into App Standby / Doze. There will be periods of time where it's forced to be idle and can't receive messages. To avoid that, it needs to ask for a battery optimization exception like Conversations does when GCM is unavailable. It can trigger the prompt after the warning about Play Services being missing and it's fine if the user doesn't want to grant an exception but they should be given an explicit choice.

thestinger:
Some documentation: https://developer.android.com/training/monitoring-device-state/doze-standby.html.

It's enabled on CopperheadOS so you should run into it, but it won't happen when the device is plugged in without `dumpsys battery unplug` and it will still *work* but with a delay.

eightbitkid:
> This change will also have to include support for calling, it won't make sense to people if there are some Signal users they randomly can't call.

@moxie0: To be clear, at the minute the app just takes you to whatever application you use to make phone calls and the icon loses the little lock; it's not like others can just call you but it rings off the hook. Do want to implement calling though.

 > There will be periods of time where it's forced to be idle and can't receive messages. To avoid that, it needs to ask for a battery optimization exception...

@thestinger: I don't think I've run into this. Maybe I just check my phone to much to let it drop into a deeper sleep. Thanks for the heads up though! I shall do some research.

In the mean time I've manually enabled this to see how it affects battery life.

src/org/thoughtcrime/securesms/util/TextSecurePreferences.java
eightbitkid: Damn, you're right. I've spent to much time registering/unregistering between debug sessions.

src/org/thoughtcrime/securesms/service/RegistrationService.java
eightbitkid: Fixed with eightbitkid/Signal-Android@6d72b3f323c79db478177ea2bc2f4f821001c5dc.

src/org/thoughtcrime/securesms/RegistrationActivity.java
eightbitkid: The patch eightbitkid/Signal-Android@377443669181340112f89c7c031988f693af943e moves this to `onResume()` as a temporary fix.

src/org/thoughtcrime/securesms/util/TextSecurePreferences.java
```diff
@@ -97,6 +101,10 @@
   private static final String MULTI_DEVICE_PROVISIONED_PREF    = "pref_multi_device";
   public  static final String DIRECT_CAPTURE_CAMERA_ID         = "pref_direct_capture_camera_id";
 
+  public static boolean isGooglePlayServicesAvailable(Context context) {
+    return GooglePlayServicesUtil.isGooglePlayServicesAvailable(context) == ConnectionResult.SUCCESS;
+  }
```
moxie0: The class is for accessing preferences, so if it's not a preference accessor it belongs elsewhere. Seems like a util method to me.

src/org/thoughtcrime/securesms/util/TextSecurePreferences.java
moxie0: Sorry, no side effects is a pretty basic requirement for us.

moxie0:
> @moxie0: To be clear, if the user you're trying to call isn't voice registered, at the minute the app just takes you to whatever application you use to make phone calls and the icon loses the little lock; it's not like others can just call you via Signal but it rings off the hook. Do want to implement calling though.

That's a holdover from when TS and RP were two different apps, and is in the process of being removed. It's not currently a reliable mechanism.

src/org/thoughtcrime/securesms/util/TextSecurePreferences.java
eightbitkid: This change will be removed during a rebase later.

eightbitkid:
Looking into the request to Android to for Signal to exist outside of the sleep/doze functionality (kinda, that's a simplified statement) brings up and issue with Google Play. It's actually listed on the page @thestinger links too:

 >  __Note:__ Google Play policies prohibit apps from requesting direct exemption from Power Management features in Android 6.0+ (Doze and App Standby) unless the core function of the app is adversely affected.

Several people have mentioned having their apps revoked from the store with no warning (admittedly this is just hearsay from the Internet). Given that Signal is only officially distributed via the store is this a potential risk? Did the team working on Conversations do anything special to make sure their app is okay?

thestinger:
It shouldn't be done in a way that it ever asks when Play Services isn't present. That's a given. It needs to ask when GCM is not available. Nothing special needs to be done.

eightbitkid:
Awesome sauce. Just figure I'd check; I've never published so I'm not sure what the exact procedure is.

eightbitkid:
@thestinger: Battery optimisation is implemented as a reminder pop-up thing; like the 'use as default texting app' message in commit eightbitkid/Signal-Android@6fcce22af3bc333576d12578a315291a2c58188d. Thought this made the most sense for consistency. Appreciate the assist on that one.

thestinger:
Is that checking for Play Services to know when it shouldn't show it though? I don't see it there.

eightbitkid:
The check is via `!TextSecurePreferences.isPushRegistered(context)`. If we're push registered then we're getting notifications then we've got Play Services installed; if we're not we haven't.

eightbitkid:
As something stable for devices without Google I'd consider this complete.

TextSecure and RedPhone look like they handle network connections very differently (obviously) and I see no way to change the library that bolts RedPhone onto TextSecure into something that will keep listening to the server. I suspect there isn't one because a Cloud Messaging identifier is built directly into the manager.

eightbitkid:
Rebased onto release v2.25.2. The most recent update to [libsignal-service-java](https://github.com/WhisperSystems/libsignal-service-java) seems to not have been pushed so I can't update my changes to that library.

eightbitkid:
@moxie0: I am of the mind now that I cannot support voice communications in this patch. The RedPhone server doesn't appear to be open-source (from my searching, if it is and you are able, can you throw it my way?) and thus I can't make changes to only one end and expect anything to work.

Would you consider accepting this pull request without voice support? If not I will close this because I don't see a way to provide it. Information I need is missing.

Thanks.

moxie0:
@eightbitkid I think the best thing to do is wait for the webrtc_video branch. Shouldn't be too long now, just waiting on iOS.

res/values-ar/strings.xml
```diff
@@ -467,7 +467,6 @@
   <string name="RegistrationActivity_invalid_number">الرقم غير صحيح</string>
   <string name="RegistrationActivity_the_number_you_specified_s_is_invalid">الرقم الذي حددته (%s) غير صحيح.</string>
   <string name="RegistrationActivity_unsupported">غير مدعوم</string>
-  <string name="RegistrationActivity_sorry_this_device_is_not_supported_for_data_messaging">عذرا، هذا المحمول لا يدعم المراسلات عبر بيانات شبكة الجوال. الهواتف العاملة بأنظمة أقدم من أندرويد 4.0 يجب أن يكون لديها حساب جوجل مسجل. الهواتف العاملة بنظام أندرويد 4.0 أو الأعلى لا تطلب حساب جوجل ولكن يجب أن يكون مثبت بها تطبيق سوق بلاي.</string>
```
moxie0: Please never modify the translations

AndroidManifest.xml
```diff
@@ -85,6 +85,9 @@
     <uses-permission android:name="android.permission.READ_LOGS"/>
     <uses-permission android:name="android.permission.WRITE_CALL_LOG"/>
 
+    <!-- Required network connection if Google Play is missing -->
+    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS"/>
```
moxie0: What is the security level on this permission? If it requires all Signal users to manually approve before updating, it's unlikely that we'd take that hit.

src/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -67,6 +67,16 @@ public void onCreate(Bundle icicle) {
   }
 
   @Override
+  public void onResume() {
+    super.onResume();
+
+    if (!(GooglePlayServicesUtil.isGooglePlayServicesAvailable(this) == ConnectionResult.SUCCESS)) {
```
moxie0: First, !=.  But this is also not the correct way to handle this result.  There are many possible result codes, some of which require different handling than displaying this dialog.

src/org/thoughtcrime/securesms/jobs/RefreshAttributesJob.java
```diff
@@ -42,10 +45,14 @@ public void onRun() throws IOException {
     String gcmRegistrationId = TextSecurePreferences.getGcmRegistrationId(context);
     int    registrationId    = TextSecurePreferences.getLocalRegistrationId(context);
 
-    String token = textSecureAccountManager.getAccountVerificationToken();
+    if (GooglePlayServicesUtil.isGooglePlayServicesAvailable(context) == ConnectionResult.SUCCESS) {
```
moxie0: not the correct way to handle this

AndroidManifest.xml
thestinger: It's a normal permission rather than a dangerous one since it only allows requesting an exception that the user can choose to approve or deny.

thestinger: I guess they did it this way so apps could adopt it painlessly before migrating to API level 23 or later where dangerous permissions already get requested automatically and don't block updates. They added both in 6.x but doing this one as a normal permission was friendlier to existing apps.

eightbitkid:
Just so it doesn't look like I've wandered off: I'm not working on this until the tech is around for me to provide encrypted calls. I have other stuff to do rather than running in place. I'll continue to rebase but they'll be no new work.

moxie0:
This has a ways to go still, so it'd be faster to do them in parallel, but up to you. If you don't want to keep working on it, please close and reopen when you're ready to be involved.  Thanks!

thestinger:
@eightbitkid Could drop the translations though, which would probably make it easier to maintain.

ByteHamster:
It probably won't get merged with translation changes anyway. Translations are only managed using Transifex

thestinger:
@ByteHamster  Yeah, that was one of the review comments, so it makes sense to drop them even if the intention is only to keep rebasing it.

thestinger:
@moxie0 I can rework this to address the issues you pointed out, but could https://github.com/WhisperSystems/libsignal-service-java/pull/28 be merged early? It would make it easier to work on this if it didn't require out-of-tree changes to actually build / test it.

nrizzio:
@thestinger [this](https://whispersystems.discoursehosting.net/t/building-and-testing-the-webrtc-video-branch/644?u=nrizzio) might help you test changes in libsignal without having Moxie actually upload them to Maven.
If you're still having trouble with it, feel free to open a topic in that forum and I'll help you with it.

thestinger:
I know how to build it, but I think it would make sense to merge the library change early. Otherwise it has to keep being rebased along with the hacks to build the out-of-tree library, and having the change merged won't impact Signal users.

thestinger:
BTW, there's at least one seemingly non-GCM related isPushRegistered that wasn't converted in these changes. That isn't the approach @moxie0 wants anyway but I think it's a problem regardless of the chosen style.

eightbitkid:
I've dropped most of the translations for now. It means that the warning that appears on start up is wrong; not sure how Transifex works and how I would create a string.

The lack of attention on that pull request, along with [comments on a similar request](https://github.com/WhisperSystems/libsignal-service-java/pull/5), is part of the reason I've slowed on this. There's no point doing the work if the fundamental dependency isn't going to be accepted.

I'm more than happy to start over when the 'webrtc_video' topic branch is merged but I'm not working off non-mainline work. Until then all this work is publicly available.

thestinger:
AFAIK translations happen after changes land, not before. It's not meant to be done in pull requests adding features.

eightbitkid:
@thestinger: Does that include brand new strings too?

thestinger:
Yeah, I think translations are always maintained out-of-band.

eightbitkid:
Hm. Okay. Thanks for clarifying!

haffenloher:
> I've dropped most of the translations for now. It means that the warning that appears on start up is wrong

@eightbitkid I'm not sure what you did (doesn't seem like you did it in a new commit?), but maybe there's a misunderstanding. Adding and removing strings to and from `res/values/strings.xml` is fine. There's just no need to touch the translated files as they're generated by Transifex.

thestinger:
You update the reference English data as part of code changes (or as the only changes) and then the translations to other languages are done out-of-band. The development process for Android, Signal, most apps and really most software projects is English-only and then translations are dealt with separately. Only people doing translations and releases need to worry about it. It just means not touching anything in the language suffixed resource directories.

pbeckmann:
I would really love this feature. So pls don't abound it :smile:

thomasmarkiewicz:
I would love this feature as well.  I am trying to build it and test on my device but can't figure out how to include the patched libservice-signal-java in Signal-Android.  Any tips?

thestinger:
@thomasmarkiewicz Clone https://github.com/copperhead/Noise, clone https://github.com/copperhead/libsignal-service-java inside it and build that. I moved some more calls over to isRegistered since they seemed to be missing, other than that it should match this. I had already dropped the translations. I plan on adjusting it to address @moxie0's concerns and it should also make it easier to maintain. It would be nice if the library change could land and then if these changes could MOSTLY land just without a final change actually letting it work. It makes sense to do work incrementally IMO and if this is really not being rejected but rather just delayed then it makes sense for the code to land once it conforms to the standards, which is straightforward.

thomasmarkiewicz:
@thestinger That worked for me.  I have it running on my Note4 + Cynogenmod 14.1 w/o Google and successfully exchanged some msgs with iOS Signal app.  Good enough for now and looking forward to these changes being merged.  Great work, thanks!

moxie0:
Everyone, please remember that this is not a discussion forum. Please do not post here unless you are contributing to the development of the PR.

@eightbitkid if you don't want to keep working on this, can you please close it and reopen when you're ready.

thestinger:
@eightbitkid I can submit a PR from a branch in the Noise repository and you'll still get credit for being the one to get the ball rolling, i.e. if there's a bounty you can have it. I have to maintain this anyway and you're just ending up doing rebasing that I already had to do for Noise. I really don't mind adjusting it to use a setting for GCM and avoiding the isPushRegistered changes as requested. It will be easier to maintain that way and that's always my focus with downstream changes (minimal, non-invasive changes) so it's nice when that aligns with what's needed to upstream it. So don't be scared to close it, I'll just open up a new PR right away.

eightbitkid:
@haffenloher: My changes to translation files where in a separate commit; I dropped it during the last rebase. Thanks for jumping in and explaining though. I understand the process now!

@thestinger: If you're happy to take over the management of the actual pull request, then I'm happy for that to happen (I confess I'm actually using Noise). I'll close my request on the library too; makes sense for you to be in control of both parts (assuming it won't be accepted...)

I'll keep an eye out for when the rest of the functionality is included and be back to help if it's needed. Definitely want this upstream.

moshpirit:
So, is Signal Google-free now?

thestinger:
It uses Google Cloud Messaging if it's present but can now be used without it.

maincui:
Hi all, 

two years later, it seems Signal-Android still relies on GCM/FCM, without it when the app is invisible to user, it cant receive messages/call alerts. After some research in the code, it did have a WebSocketConnection but the connection halts when the app is invisible. Any help to code off GCM/FCM?

gumb00:
>... without it when the app is invisible to user, it cant receive messages/call alerts.

Without GCM/FCM installed you should see a notification saying "Background connection enabled". Otherwise you are probably doing something wrong.

thestinger:
If you've blocked the notification, it won't work.

maincui:
@gumb00 @thestinger yes, notification is allowed and shown, but still there is no new message notification shown to user when invisible, and will retrieve all undelivered messages when back to visible. Is this still in development or possibly a bug? the code branch/version are master/4.45.2.

Besides, it seems there is some problem with RealtimeSleepTimer, I dont know whether it's the cause of undelivered message. below is the log.

`2019-08-22 11:16:30.159 25982-26218/org.thoughtcrime.securesms W/WebSocketConnection: java.lang.IllegalArgumentException: Receiver not registered: org.whispersystems.signalservice.api.util.RealtimeSleepTimer$AlarmReceiver@1ea378f
        at android.app.LoadedApk.forgetReceiverDispatcher(LoadedApk.java:1199)
        at android.app.ContextImpl.unregisterReceiver(ContextImpl.java:1449)
        at android.content.ContextWrapper.unregisterReceiver(ContextWrapper.java:645)
        at org.whispersystems.signalservice.api.util.RealtimeSleepTimer.sleep(RealtimeSleepTimer.java:51)
        at org.whispersystems.signalservice.internal.websocket.WebSocketConnection$KeepAliveSender.run(WebSocketConnection.java:323)`

Be-ing:
I did not experience that issue when I used Signal with the WebSocket connection. However, now I use [LineageOS with microG](https://lineage.microg.org/) so I can use the GCM push notifications and Signal uses much less battery.

maincui:
using Profiler in Android Studio to see what

> I did not experience that issue when I used Signal with the WebSocket connection.
Please help me how do you specify Signal what to use, especially how to make it totally rely on WebSocket, GCM/FCM free?

haffenloher:
Hey @maincui, [this is not a support forum](https://github.com/signalapp/Signal-Android/blob/master/CONTRIBUTING.md#github-is-not-a-generic-discussion-forum). Looks like you already found the community forum since you posted there as well. Please continue the discussion in your community forum thread.

maincui:
@haffenloher , sorry for the inconvenience. I searched at both sides and tried to find whether the feature is done or the problem is fixed. Although this issue is closed, but it has quite a few participants, maybe more attentions could be drawn here. sorry again, I just want to know the up to date state of the non-GCM/FCM solution or the FCM-first-WebSocket-fallback solution.

PR #5984, opened 2016-12-22
Update crowdfunding link

echo743:
Second attempt, because I forgot to add "FREEBIE" to the last commit message in my previous pull request.

// FREEBIE


moxie0:
in 3.26.0

PR #6086, opened 2017-01-23
Add scroll to bottom button in conversation view

### LINKED ISSUE #5651: Add shortcut to jump to end of conversation
I have:

---
### Feature Request

When you are in a conversation and scrolled a bit away from the most recent message (upwards direction) a shortcut to the bottom of the conversation could be displayed near the scroll bar. This would make it easier to return to the most recent messages after reading some old messages.
### Screenshot (WhatsApp)

screenshot_2016-08-31-22-12-07

Stuie:
Added a FloatingActionButton to the conversation_fragment that appears and disappears using the same logic as the existing compose divider: appear when the conversation list is scrolled away from the bottom, disappear when the list is scrolled to the bottom.

Fixes #5651 

### Contributor checklist

 * LG Nexus 4, Android 5.0.2

----------

### Description
Added FloatingActionButton to the ConversationFragment to allow quick scrolling to the bottom of the conversation when not already at the bottom.

Re-used an existing icon and a method for smooth scrolling.

### DIFF
--- a/res/layout/conversation_fragment.xml
+++ b/res/layout/conversation_fragment.xml
@@ -21,4 +21,16 @@
           android:alpha="0"
           android:visibility="invisible" />
 
+    <android.support.design.widget.FloatingActionButton
+        android:id="@+id/scroll_to_bottom_button"
+        android:visibility="invisible"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginEnd="@dimen/scroll_to_bottom_button_margin_end"
+        android:layout_marginRight="@dimen/scroll_to_bottom_button_margin_end"
+        android:layout_marginBottom="@dimen/scroll_to_bottom_button_margin_bottom"
+        android:layout_gravity="bottom|end"
+        android:contentDescription="@string/conversation_fragment__scroll_to_the_bottom_content_description"
+        android:src="@drawable/ic_keyboard_arrow_down_white_36dp" />
+
 </FrameLayout>
--- a/res/values/dimens.xml
+++ b/res/values/dimens.xml
@@ -66,4 +66,7 @@
     <dimen name="onboarding_subtitle_size">20sp</dimen>
 
     <dimen name="scribble_stroke_size">3dp</dimen>
+
+    <dimen name="scroll_to_bottom_button_margin_end">10dp</dimen>
+    <dimen name="scroll_to_bottom_button_margin_bottom">10dp</dimen>
 </resources>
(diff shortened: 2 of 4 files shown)


ByteHamster:
Could you please post a screenshot? In my opinion, the button should be quite small. Not as big as a fab usually is (that would be distracting). Maybe the button can even be semi-transparent and gray 

Anyway, nice to see that someone adds the button :)

Stuie:
Here's a screenshot of what I have. It would be useful to have some design requirements for this.

Personally I think it could do with being a bit smaller than the attachment button, but should probably match the color.

Is there a design specification for the app in general?

screenshot_2017-01-23-14-47-26

scienmind:
Definitely should be smaller. 
Also I'd argue that it should be "lighter" than the send/attachment button (or darker in dark theme), to be less "in your face" than the send/attachment button. 

Making it smaller and semi transparent should do, i guess.

> Is there a design specification for the app in general?

I think general advice would be to go with google's material design guidelines

Stuie:
Google's material design guidelines were surely considered when they came up with the FloatingActionButton that uses the app's defined primary and secondary colors. It also has a shadow, unlike the attachment button.

I've messed around with the colors a bit, and I must admit that I'm so unfamiliar with the way the themes and styles work in Signal that I can't seem to get the right values for light and dark themes.

Specifying `app:backgroundTint` and `app:rippleColor` in `conversation_fragment.xml` will obviously work for one theme, but as soon as you change to the other it looks awful.

I've tried creating an entry in `styles.xml`, but I hit runtime exceptions.

Does anyone have some quick advice on how to correctly define the colors for a button if it's not as simple as setting the `android:background`?

FeuRenard:
You could try using another fab. Search for "melnykov" in this repository. That one should be easier to color.

motatuc:
@Stuie, thank you for your contribution.

A humble question though: 

Wouldn't it be better to have fast scrolling instead of another button?
If I want to jump to the bottom of a conversation I usually leave and open the message.
If fast scrolling would be available we could:
1) prevent another button that clutters the UI
2) a solution for scrolling in both directions fast and more selectively (not only fast to the end or start of a conversation but rather then to a specific point).

What do you think @Stuie, @ByteHamster, @FeuRenard, @scw, @scienmind?

Stuie:
I think that fast scrolling might be difficult to do. I haven't looked into it very much, but not all of the message history is loaded up front, so you don't necessarily have an accurate position for the scroll bar.

Once you reached the top it would keep loading more and more messages from the past, which would be quite awkward.

That's perhaps partially why WhatsApp and Telegram don't  do it either.

motatuc:
I see. Sounds a bit tricky then. 

But doesn't Signal loads the full history in only one additional step?
Wouldn't it be possible to, lets say, fast scroll for the messages that get already loaded by standard. Once you press "see full conversation" the fast scroll position would be recalculated?

Issue #4778, opened 2015-12-01
[UI] Scroll bar changes its size in conversation view


art1fa:
When scrolling through a conversation, the scroll bar changes continuously its size.
Expected behavior: Constant size of scroll bar, just like in the message view

GIF:

![animated](https://cloud.githubusercontent.com/assets/11131859/11514149/495adbac-9877-11e5-8e7f-2a372f468cff.gif)

Device info: Nexus 5, OmniROM 5.1, TextSecure 3.6.0

heyimgay:
I can confirm this. It is also not the case that old messages get unloaded and thus the scrollbar becomes larger. This becomes obvious if you look at the scrollbar of a longer conversation.

cascheberg:
It seems that the length of the scrollbar increases with the number of message items that are currently visible on the screen. So it is shorter if you are seeing images or multi-line messages.

unrulygnu:
https://www.quora.com/Why-hasnt-Google-fixed-the-scrollbar-length-problem-in-Android

iambmelt:
Did a quick peek to investigate this.

Relevant class / layout
[`ConversationFragment.java`](https://github.com/WhisperSystems/Signal-Android/blob/master/src/org/thoughtcrime/securesms/ConversationFragment.java) | [`conversation_fragment.xml`](https://github.com/WhisperSystems/Signal-Android/blob/master/res/layout/conversation_fragment.xml)

At a glance, it doesn't look like [AbsListView#setSmoothScrollbarEnabled(boolean)](http://developer.android.com/reference/android/widget/AbsListView.html#setSmoothScrollbarEnabled(boolean)) will work because this layout is using [RecyclerView](https://github.com/WhisperSystems/Signal-Android/blob/master/res/layout/conversation_fragment.xml#L8).

Looks like this may have to wait until RecyclerView is enhanced per [this issue](https://code.google.com/p/android/issues/detail?id=78513) to include `setSmoothScrollbarEnabled` or some equivalent.

unrulygnu:
`setSmoothScrollbarEnabled(false)` makes the scrollbar movement twitchy and distracting, probably not the desired user experience.

And if you look at [the source for `LinearLayoutManager#computeScrollExtent`](https://android.googlesource.com/platform/frameworks/support/+/master/v7/recyclerview/src/android/support/v7/widget/LinearLayoutManager.java), you'll see that the scrollbar thumb height still varies, despite what the docs say.  

[...]

PR #6105, opened 2017-01-26
Add fast scroller to conversation view

### LINKED ISSUE #6095: Calendrical fast scrolling for conversation view
I have:

### Bug description
With the introduction of (sticky) date headers, it would make sense to add fast scrolling to the conversation view (like the alphabetical fast scrolling in Signal's contact list, but with dates instead of alphabetical characters).

haffenloher:
### Contributor checklist

 * AVD, Android 7.1.1
 * Stupid MT6589 phone, Android 4.2.2

----------

### Description

Fast scroller showing the same text as the new sticky message headers.

screenshot from 2017-01-26 17-39-56
screenshot from 2017-01-26 17-40-15

Fixed some glitches I encountered with the fast scrollbar along the way.
I verified that the contact selection list's fast scroll bubble looks unchanged with this PR.

I had to move the new "scroll to bottom" button to the left because it interfered with the scroll bar handle.

Fixes #6095
Currently only enabled for convos with >20 messages - somewhat related to #4778


Stuie:
I think moving the scroll to bottom button to the left makes it unusable for single-hand use cases for right-handed people (the majority).

If we don't care about that then fine. :)

RiseT:
Don't know if my hands have some strange anatomy, but I (right-handed) actually find it easier on a Nexus 5X to reach to the left side with my thumb than to the right side...

haffenloher:
Problem is, with the old positioning, you'd accidentally activate the scroll bar handle all the time when wanting to press the button (if the handle is near the bottom). Other ideas welcome!

RiseT:
I feel that the right side is slightly crowded anyway.

What you could potentially do is moving the floating bottom a couple of pixels to the right (e. g. center it with the avatar pictures). But this changes things only slightly.

Don't know if horizontally centered (relative to the display) floating botton would make any sense design-wise. Just a random thought.

Stuie:
Having it towards the bottom probably helps to make it reachable for most people, but there are definitely much larger phones than the Nexus 5X.

As a left-handed user I encounter issues interacting with buttons over on the right, above a certain height. On my old Nexus 6 it was quite annoying to tap the scroll to bottom button in WhatsApp/Telegram, but I assume they picked that side for a reason.

haffenloher:
@RiseT centered looks good to me:

screenshot from 2017-01-26 21-15-36

RiseT:
Perfect. Uncrowded area and easily reachable for both right- and left-handed users.

RiseT:
Somehow off-topic: Would Signal benefit from a left-hand mode which mirrors the user interface...? Just tossing this in as an idea.

Stuie:
Android in general would benefit from left-handed support, implemented similarly to right-to-leftlanguage support, perhaps.

I quite like the button being in the middle...

p4nci:
hmmm 🤔 , in the center we have already date headers, so it's a little bit "overpopulated" there (???). Further when rotating in the landscape mode the button will change position and might not be optimal.

IMO:
```
IF possible to implement:
   let the scroll button as is (right bottom position)
   set scroll button to HIDDEN if "fast scroller" in use,
   otherwise set to VISIBLE
ELSE
   remove the scroll button (still possible to jump to the bottom with "fast scroller")
```
I think there is no need for both buttons (scroll + fast scroll) at the same time

RiseT:
The sticky/floating date header is on the top of the display, while the "scroll to bottom" button is on the bottom.
How can this be overpopulated?

On the right side, there are already elements for quick camera, voice messages, attachments, phone calls, the 3-dot-menu and the scroll bar (including fast scroll).

haffenloher:
> set scroll button to HIDDEN if "fast scroller" in use

@p4nci see above - the problem occurs when the scroll bar handle is next to the button. The button then becomes almost impossible to hit with a finger (the scroll bar gets mistakenly activated instead).
Also keep in mind that the button is hidden by default. It appears after you scroll up a bit (after ~5 messages).

p4nci:
> The button then becomes almost impossible to hit with a finger
> (the scroll bar gets mistakenly activated instead).

that's a good point. IMO in this case it's better to remove the scroll button (see above - ELSE) => the "fast scroller" is still sufficient for scrolling and "jumping" to the bottom

> The sticky/floating date header is on the top of the display,
> while the "scroll to bottom" button is on the bottom.
> How can this be overpopulated?

I know, but there are two centered "buttons"(top: date header bubble + bottom: scroll button, which shows up/disappears async.) and that is too twitchy (="overpopulated" 😉) IMO

AlfonsoMuskedunder:
> I know, but there are two centered "buttons"(top: date header bubble + bottom: scroll button, which shows up/disappears async.) and that is too twitchy (="overpopulated" :wink:) IMO

@p4nci Apparently the date header is already invisible when using the fast scroller. Only when scrolling "regularly" will it appear. So that's not a source for clutter.

I agree though that the scroll-down button might be a little redundant once the fast scroller is activated, independent of where it's placed.

motatuc:
>I agree though that the scroll-down button might be a little redundant once the fast scroller is activated, independent of where it's placed.

I agree as well. The question is if we still need the scroll-down button once we have fast-scrolling...

motatuc:
Tested fast scrolling and it works nice. The size of the bubble is a bit intense though... However the scoll-to-bottom button is now somehow in the middle of the screen. Is there already a decision made about which of the two features to use?

moxie0:
hmm i don't really love the idea of having the scroll-to-bottom button in the center of the screen

haffenloher:
@moxie0 okay, on the left it is

p4nci:
I still believe that the best solution is without scroll-to-bottom button

motatuc:
I would also vote for the remove of the scroll-to-bottom button if fast scrolling is accepted.
Two reasons: a) Both features basically provide the same feature (fast scrolling) and b) fast-scroller is faster and works in both directions.

RiseT:
All I can say is that personally, I don't tend to use those scroll-to-bottom buttons. But that's just me. Don't know about others or the "average user".

Stuie:
I use search to scroll back to a specific message. I generally don't need to go to a specific date, but rather a specific message, and then one button takes me to current.

Personally I believe this should be a data-driven decision, but I'm not sure what requirements @moxie0 has for analytics/metrics collection, what with the inherent privacy concerns of this project.

Ideally we would split test both features to determine the right option.

Stanzilla:
Agree with @motatuc here, also the PR was originally about fast scrolling, if anything a scroll to bottom button should be a separate PR.

zoff99:
when is this going to be merged? i already made a PR for fastscroller which was dismissed.
it's crazy that this is taking so long.

haffenloher:
Rebased. Fixes for the fast scroller glitches I found are now in #6149.

hklene:
The screenshots really look great - Is it still possible to resurrect this or will it have to be rewritten from scratch?

Issue #4859, opened 2015-12-10
Some group messages fail to sync to Desktop


lorenzhs:
There is one particular group I'm in in which **no** messages I write on my Android phone sync to Signal-Desktop. As far as I can tell only this group has the issue. I also get an error when sending, but it doesn't show any particular problem. When I click on the error symbol in the message, there isn't an error next to any contact. When I go into the message detail view via long-press -> info, then it shows the "resend" button next to _every_ contact. Bit strange, really.

This group contains a user who unregistered from Signal.

adb logcat from when I pressed the send button until sending was completed and I got the error message on the phone: https://4z2.de/signal_faillog_ipahqu2Shoh6.txt (manually cleaned, because it's from adb logcat - I only anonymized the phone numbers)

haffenloher:
> When I go into the message detail view via long-press -> info, then it shows the "resend" button next to every contact.

That's probably due to #3582, so not related to your issue :)

lorenzhs:
Yes, thanks. I just thought that maybe this inconsistent behaviour was caused by the bug, potentially helping to track it down, so I mentioned it. Apparently it's not ;)

liliakai:
Can confirm, sending to a group with an unregistered user will not sync that message to Desktop.

https://gist.github.com/anonymous/71a25c30a471db5dad2fd3627b951e8a

haffenloher:
Just in case it helps anyone: this happens because the [call to `sendMessage()`](https://github.com/WhisperSystems/libsignal-service-java/blob/a1c93edc40e3ce201fa4e70339a3e6f4e197e319/java/src/main/java/org/whispersystems/signalservice/api/SignalServiceMessageSender.java#L155) in libsignal-service throws `EncapsulatedExceptions` so that [the following lines](https://github.com/WhisperSystems/libsignal-service-java/blob/a1c93edc40e3ce201fa4e70339a3e6f4e197e319/java/src/main/java/org/whispersystems/signalservice/api/SignalServiceMessageSender.java#L158-161) don't get executed.

Trolldemorted:
working on it, gonna test it tomorrow. Let's see if we can get this merged before the issue gets 1year old :)

Trolldemorted:
[PR is open](https://github.com/WhisperSystems/libsignal-service-java/pull/27)!

@moxie0 i understand that you are reluctant merge PRs into libsignal, but in this case it has to be done.

[...]

PR #6175, opened 2017-02-05
Remove unregistered group members

### LINKED ISSUE #5891: Message reports Failed to send for all group members, but is received by all group members
I have:
- [X] searched open and closed issues for duplicates
- [X] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

### Bug description
When sending messages in a group chat sometimes the messages reports Failed to send for all the group members. But they are received correctly by everyone. Other users in the group are experiencing the same issue sometimes and then start hitting the resend button. And then #2639 will spam the group with the same messages.

### Steps to reproduce
- send message in group chat

**Actual result:** message reports Failed to send for all group members
**Expected result:** message should report received by all group members

### Device info

**Device:** motorola XT1021
**Android version:** 4.4.4
**Signal version:** 3.23.0

### Link to debug log
https://gist.github.com/71d04edaca4110cbf470c53fbfb12fd4

### LINKED ISSUE #6003: identical message displayed multiple times
I have:

### Bug description

When sending a Signal message (not SMS) from a  heavily overloaded (WIFI) network the client attempts to send the message multiple times (expected) and eventually displays a red "!" ("Not delivered. Tap for details").
Apparently the message went through nonetheless and not just once but multiple times and is displayed at all recipients multiple times (group chat).

We observed this problem within multiple group chats and was triggered by distinct senders.
(I personally observed it in two distinct group chats, friends also confirmed this in other group chats.)

Note: the sender did _not_ press any of the "RESEND" buttons.

### Steps to reproduce
- reproducing the exact network failure rates/environment is not really possible, but here are the simple steps

- sender: try to send a message 
while the 3 dots show that the client attempts to still send 
- recipient: displays the same message (by the same sender) multiple times 

**Actual result:** Message is displayed multiple time (between 6 and 15 times in our cases)
**Expected result:** The recipient detects that this is the identical message (via a message id, replay protection?)

### Device info
**Sender's Signal version:** 3.25.3  and others
**Recipient's Signal version:** 3.25.3  and others

haffenloher:
### Contributor checklist

 * Android 6.0 emulator

----------

### Description
Locally remove unregistered users from group membership lists and add a message saying "so and so has left the group." as described in [this comment](https://github.com/WhisperSystems/Signal-Android/issues/5891#issuecomment-263163151).

Fixes #5891
Fixes #6003

### DIFF
--- a/src/org/thoughtcrime/securesms/groups/GroupMessageProcessor.java
+++ b/src/org/thoughtcrime/securesms/groups/GroupMessageProcessor.java
@@ -6,8 +6,6 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.util.Log;
 
-import com.google.protobuf.ByteString;
-
 import org.thoughtcrime.securesms.ApplicationContext;
 import org.thoughtcrime.securesms.crypto.MasterSecretUnion;
 import org.thoughtcrime.securesms.database.DatabaseFactory;
@@ -40,7 +38,6 @@ import java.util.Set;
 import ws.com.google.android.mms.MmsException;
 
 import static org.thoughtcrime.securesms.database.GroupDatabase.GroupRecord;
-import static org.whispersystems.signalservice.internal.push.SignalServiceProtos.AttachmentPointer;
 import static org.whispersystems.signalservice.internal.push.SignalServiceProtos.GroupContext;
 
 public class GroupMessageProcessor {
@@ -85,7 +82,7 @@ public class GroupMessageProcessor {
   {
     GroupDatabase        database = DatabaseFactory.getGroupDatabase(context);
     byte[]               id       = group.getGroupId();
-    GroupContext.Builder builder  = createGroupContext(group);
+    GroupContext.Builder builder  = GroupUtil.createGroupContext(group);
     builder.setType(GroupContext.Type.UPDATE);
 
     SignalServiceAttachment avatar = group.getAvatar().orNull();
@@ -117,7 +114,7 @@ public class GroupMessageProcessor {
     Set<String> missingMembers = new HashSet<>(recordMembers);
     missingMembers.removeAll(messageMembers);
 
-    GroupContext.Builder builder = createGroupContext(group);
+    GroupContext.Builder builder = GroupUtil.createGroupContext(group);
     builder.setType(GroupContext.Type.UPDATE);
 
     if (addedMembers.size() > 0) {
@@ -173,7 +170,7 @@ public class GroupMessageProcessor {
     byte[]        id       = group.getGroupId();
     List<String>  members  = record.getMembers();
 
-    GroupContext.Builder builder = createGroupContext(group);
+    GroupContext.Builder builder = GroupUtil.createGroupContext(group);
     builder.setType(GroupContext.Type.QUIT);
 
     if (members.contains(envelope.getSource())) {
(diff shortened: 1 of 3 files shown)


Trolldemorted:
If you remove a member from a group locally, all subsequent group updates from your client will be ignored, at least on Signal-Desktop clients.

Can we change that behaviour and support kicking by excluding from group update?

exploide:
Since this got already merged in Signal-Desktop, I guess it is more important to work towards a consistent behavior now.

Currently this is happening: Signal-Desktop removed two unregistered numbers locally, an actual user reinstalled Android, my phone answered to the group info request and hence the two unregistered numbers are back in Signal-Desktop :-P

Trolldemorted:
When all devices of a group except one notice a member is unregistered and remove him/her, the member re-registers and the last device is coming online, it will still still have the member in the group, but the others do not.

If that previously dormant device is sending a message, the re-registered device(s) will send a group info request to all devices of the sender, receive potentially diverging group infos, and since the group info response is synced to siblings the group will be awfully async:

- the re-registered devices will think they are in the group
- the dormant device which woke up (and all its siblings) will think the re-registered devices are in the group
- all others will not think so

Leading to the re-registered user receiving messages from the dormant device's owner only. Now think of multiple dormant devices, multiple unregistered users and the amount of confusion this will bring upon non tech/signal-savvy users - imho this is *way* worse than the previous behaviour.

moxie0:
I just spoke with @liliakai and we decided to revert the change on desktop.  Locally removing users can cause lots of synchronization issues (as mentioned here), and we want the behavior to be that no error is shown but that the user is still considered a group member.  I think that is what happens on Android now (correct me if I'm wrong), and if that wasn't the case on Desktop then we should resolve whatever bug was occurring there too.

Trolldemorted:
@moxie0 Android currently sends them until you run into rate limits for pre-keys of an unregistered user, which happens rather frequently: https://github.com/WhisperSystems/Signal-Android/issues/6003#issuecomment-271451242, https://github.com/WhisperSystems/libsignal-service-java/pull/31

Also every message is not synced to sibling devices: https://github.com/WhisperSystems/Signal-Android/issues/4859, https://github.com/WhisperSystems/libsignal-service-java/pull/27

haffenloher:
> no error is shown but that the user is still considered a group member

Yep, that's what is happening on Android. Still, after sending ~4 messages in quick succdession, the rate limit for fetching prekeys for the unregistered user kicks in. Due to this,

- messages to some users *are lost* if they're behind the unregistered member in the group membership list (because a RateLimitException aborts the sending process)
- messages to other members are sent five times (the duplicates are now ignored on Android)

My other attempt to work around these issues was https://github.com/WhisperSystems/libsignal-service-java/pull/31 (i.e. ignore the RateLimitExceptions and continue sending to the rest of the group).

haffenloher:
not completely sure about this, but I think messages actually still get marked as failed on RateLimitExceptions, without ability to resend? #5891 seems to confirm this

Trolldemorted:
@haffenloher yeah, every exception except a NetworkFailureException is just ignored: https://github.com/WhisperSystems/Signal-Android/blob/9148b7da5f4604dae1106d3932bfa7d3bfa63ad4/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java#L317

moxie0:
I see what you're saying now, the rate limit check is applied before the user existence check when fetching prekeys. I've done a server deploy which reverses that, so this should no longer be an issue and should not require any client-side changes.

haffenloher:
Okay, cool! That's indeed a simpler solution =)

Trolldemorted:
:+1:

Nevertheless, #4859 is still present - can't we send sync messages immediately?

wizardofid:
:-o   I hoped this would lead to some kind of a solution for one of my repeatedly occuring problems:

groupmember: 'Hi Wiz, i've got a new cellphone and a new number. Please add it to our group and remove the old mumber.'
me: 'I cannot remove your old number. Please simply use the 3-dot-menu to leave the group.'
groupmember: 'Sorry, contract ended and i don't have access to this number any longer.'

Often people also don't de-register. But i think there's a timeout in the server that de-registers numbers after 6 months without contact.
Nevertheless this leads to a bunch of zombie-accounts in our groups. After 6 months the old number can and will be recycled by the telco for another user. If this new user registers for Signal this will lead to a stranger becoming member of our group.
In Germany there's a bad habit to get new phones by signing new contracts, often with a different provider. So most numbers are assigned only for about 2 years to a specific user.

moxie0:
@wizardofid I agree this is a problem, we're thinking about ways to reengineer the whole groups thing

Trolldemorted:
@moxie0 how would you like single point-of-truth-groups with one owner device and a logical clock? Non-admin clients could send a group update proposal to the owner, which accepts/declines/autoaccepts the proposal.

This would prevent a group from updating when the owner device is offline or went south, but i think that is better than the status quo, where groups go async quite frequently (e.g. if your group contains an unregistered user, any group update you send will not arrive at your slaves due to #4859).

PR #6231, opened 2017-02-15
Added persistent verification information...

Ofenhed:
...depending on the peers public key. This will show the "verified" animation when clicking "Verify safety number" if that contact has been verified using qr code. It can not be set arbitrarily by the user. If the peer changes their key it will no longer show as verified.

// FREEBIE

### Contributor checklist

 * Nexus 6P, Android 7.1.1
 * Nexus S, Android 4.4.4

----------

This will make the manual verification persistent in the sense that a user can later go back and check whether a user has been verified or not. I did a similar commit earlier, and I got rejected, but that was a way less elegant solution with a lot more trust in the technical competence of the user involved. If the impact to the source code is acceptable, this will add to the user experience without subtracting.


Trolldemorted:
can we get an icon/indicator on the thread list, or in the thread view?

It would be nice to have an overview of who is already verified and who is not, i think.

Ofenhed:
@Trolldemorted This would definitely allow it, but since a similar PR was denied I've focused on keeping this as subtle as possible.

snaggen:
This is a great feature. I have less than 20 friends on Signal i chat with, and I have verified a few of them. But even with low numbers like this, it is hard to remember who is verified and who isn't. So some kind of "verified" mark is really needed. 
I think that there should be some indication about who is verified, but I also agree that it should be very subtle and not clutter anything. 
However, then there is the situation where you have a conversation with someone and the key changes, because they might have reinstalled signal or something. Then you have the situation were you have some old messages with a verified key, and then there are new messages using a new key. So is the verified state connected with the user or the key? Since it would be good to be able to differentiate the messages with the verified key, from the messages with the non-verified key. 

Currently messages have a lock indicating secure/insecure, so I guess a third version of this lock indicating "secure and verified" would probably be a nice and subtle indication.

Then the send button is also used to differentiate between send secure/insecure, so having another "secure and verified" version of that might be good.

halb9:
Here is my view. 


1. Encryption is good but meaningless if you don't have authentication. 
2. More authentication is better than less authentication. Not only for the single user but the network as a whole. 
3. Average users probably do not know what authentication is and why it is important. 

Until we figure out how to solve 4. in a good way, give people who already care about authentication the option to do it in such a way where the status of authentication is displayed to the user. And at the same time make sure the average user is not confused. 

My proposal: 

We already have the "Verify safety numbers" section. 
Adding in there a bit more of info for authentication shouldn't hurt. You're in there because you know what you're looking for anyways. Or at least are curious. 

See my awesome GIMP Skillz to get an idea of what I'm talking about. 
verify01
verify02
verify03

p4nci:
there was already a topic about this at the unofficial community forum:
[Remember verified safety numbers](https://whispersystems.discoursehosting.net/t/remember-verified-safety-numbers/696)

Ofenhed:
I've noticed the conflict in DatabaseFactory, I'll try to make time to fix it tonight.

p4nci:
seems like this PR now is obsolete. See [Add support for SN verification](https://github.com/WhisperSystems/Signal-Android/commit/76c28cfa7a12b386ce5f62316247e843ceef3290)

Issue #4424, opened 2015-11-06
playing audio should hold a wake lock


mcginty:
if you play audio that's longer than the screen timeout on your phone, it abruptly stops because onPause happens.

psyfx:
Bug is still present in 3.11.1
Audio stops immediately after the screen goes dark. (a few seconds)

moxie0:
@psyfx Please do not bump issues: https://github.com/WhisperSystems/Signal-Android/blob/master/CONTRIBUTING.md

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #6253, opened 2017-02-20
Keep screen on during audio playback

FeuRenard:
### Contributor checklist

 * Sony Xperia U, Android 4.4.4

----------

### Description
#4424 demands a wake lock for audio attachment playback. But a wake lock alone won't do the trick. A CPU in awake state will not be helpful if audio playback is stopped in `onPause()` (which is triggered when the screen turns off and should definitely include stopping audio playback). So I decided to implement a simple workaround that keeps the screen on during audio playback.

WhatsApp for example continues its audio playback even when the screen turned off. That's the behaviour we want for the future, I think. This PR is just an offer to slightly improve the situation.

Improves #4424
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/audio/AudioSlidePlayer.java
+++ b/src/org/thoughtcrime/securesms/audio/AudioSlidePlayer.java
@@ -1,5 +1,6 @@
 package org.thoughtcrime.securesms.audio;
 
+import android.app.Activity;
 import android.content.Context;
 import android.media.AudioManager;
 import android.media.MediaPlayer;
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/audio/AudioSlidePlayer.java
```diff
@@ -89,6 +91,7 @@ public void onPrepared(MediaPlayer mp) {
 
         notifyOnStart();
         progressEventHandler.sendEmptyMessage(0);
+        ((Activity) context).getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
```
moxie0: hmm i don't really like this cast

FeuRenard:
I added a check before each cast. And I added a wake lock release at two sites I missed at first.

moxie0:
Using instanceof is often a bad sign; it wasn't the unsafe cast that bothered me as much as the fact that a cast is happening. my intution is that this might be a case for an actual wakelock rather than using window flags

FeuRenard:
Ok, so this approach is not suitable.

PR #6260, opened 2017-02-20
Fixes #6211 contact custom label name instead of phone number when av…

### LINKED ISSUE #6211: [Feature Request] Add Contact's custom string to Signal app
My contacts are organized by client, and the phone numbers listed as 'so & so's cell'.

Is it possible to add the 'custom' string under the Name of the Contact in Signal? That way I can easily tell who I'm chatting with.

I don't know a better way to organize lots of numbers with Google. It's been rather clunky usually, but this way I can quickly find people.

Bilb:
…ailable

### Contributor checklist

 * Nexus 5, Android 7.1.1

----------

### Description

Enable a feature asked by some people, and that I find useful too. Shows the "custom label" set in the contact field as a subtitle in the conversation view, if available. This is an easy one, before going deeper in bugs resolution/features addition.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationTitleView.java
+++ b/src/org/thoughtcrime/securesms/ConversationTitleView.java
@@ -67,7 +67,10 @@ public class ConversationTitleView extends LinearLayout {
         this.subtitle.setVisibility(View.GONE);
       } else {
         this.title.setText(recipient.getName());
-        this.subtitle.setText(recipient.getNumber());
+        if(recipient.getCustomLabel() != null)
+          this.subtitle.setText(recipient.getCustomLabel());
+        else
+          this.subtitle.setText(recipient.getNumber());
         this.subtitle.setVisibility(View.VISIBLE);
       }
     } else {
--- a/src/org/thoughtcrime/securesms/ConversationTitleView.java
+++ b/src/org/thoughtcrime/securesms/ConversationTitleView.java
@@ -68,7 +68,7 @@ public class ConversationTitleView extends LinearLayout {
       } else {
         this.title.setText(recipient.getName());
 
-        if(recipient.getCustomLabel() != null) {
+        if (recipient.getCustomLabel() != null) {
           this.subtitle.setText(recipient.getCustomLabel());
         }
         else {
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/ConversationTitleView.java
```diff
@@ -67,7 +67,10 @@ private void setRecipientTitle(Recipient recipient) {
         this.subtitle.setVisibility(View.GONE);
       } else {
         this.title.setText(recipient.getName());
-        this.subtitle.setText(recipient.getNumber());
+        if(recipient.getCustomLabel() != null)
```
moxie0: "if ("

moxie0: please use braces when if conditions span multiple lines

Bilb:
Well, currently I am using the getCustomLabel() method, so I only use the custom label set by the user if defined. It means the alias like "Cellphone, Work, Home, ..." will not be used, unless set as custom label. You can check in the contact app : you can choose "Home, Work, etc" or "custom" and this last one is the one I use. 

In fact, using all available aliases would be nice, but we would need something to show the phone number associated with it. A tap on the alias would be nice to toggle alias/phone number, but currently it is used to show the preferences of conversation activity (already available via the menu button).

What do you think about the toggle label/phone number idea?

src/org/thoughtcrime/securesms/ConversationTitleView.java
```diff
@@ -67,7 +67,14 @@ private void setRecipientTitle(Recipient recipient) {
         this.subtitle.setVisibility(View.GONE);
       } else {
         this.title.setText(recipient.getName());
-        this.subtitle.setText(recipient.getNumber());
+
+        if(recipient.getCustomLabel() != null) {
```
p4nci: just few lines above you have an example for if-else: single line else, "if (" with whitespace
[Code Style Guidelines](https://github.com/WhisperSystems/Signal-Android/wiki/Code-Style-Guidelines)

moxie0:
Thanks @Bilb maybe I'm not sure what the problem we're trying to solve is? In general I'm not opposed to the idea of having a label if we can find a place that doesn't look too busy, but I don't really want to replace the number at this time.  A toggle might be fine, but as you say there's already a tap action there.

moxie0:
in 4.6.0

PR #6322, opened 2017-03-01
Fix occasional crash on Android 7 direct reply

mammuth:
### Contributor checklist

 * Nexus 4, Android 7.1
 * Galaxy S5, LineageOS Android 7.1

----------

### Description

The use of the RecipientFactory in the RemoteReplyReceiver was done in the main thread. This is a problem when the recipient's data isn't cached because then the ContactPhotoFactory would perform Glide actions on the main thread, which calls the exception of the issue.

I just moved the use of the RecipientFactory inside the AsyncTask we use later; this is also the place where we need the recipients the first time.

Direct reply still works as intended after this change.

Related Issue: https://github.com/WhisperSystems/Signal-Android/issues/6185


moxie0:
in 3.31.0 thanks

Issue #6307, opened 2017-02-26
Send again doesn't show spinning element after pressing


piratenpanda:
I have:
- [x] searched open and closed issues for duplicates
- [x] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

----------------------------------------

### Bug description
If a message fails to send, the red exclamation mark takes one to the details view where the send again button is located. In earlier Signal versions the button disappeared after pressing it and some spinning element or dancing dots appeared, signalling that something is happening. Since a few versions now nothing happens visually after pressing even though Signal is trying to resend the message. This is confusing and leads to pressing the button again and again because nothing seems to be happening.

### Steps to reproduce
- Send a message that fails to send for whatever reason
- click for details
- press resend button

**Actual result:** 

Nothing signals the user that something is happening, button stays active all the time

**Expected result:**

Deactivate the button and signal that something is happening

### Device info
**Device:** Huawei Nexus 6p
**Android version:** 7.1.1
**Signal version:** 3.29.6

piratenpanda:
It still shows the resend button again when leaving the details view and entering it again before the message gets resent/another warning is shown

iafisher:
This is still an issue for me on 4.53.6. After a message fails to send and I press "Resend", there is no indication either in the conversation view or the detail view that Signal is attempting to resend the message, and the "Resend" button is still visible. As piratenpanda noted, this can lead users to hitting "Resend" multiple times.

Could this issue be reopened?

cody-signal:
This should be fixed by this commit https://github.com/signalapp/Signal-Android/commit/fb600e98291951efbab37ee6033624db67ffa78b. Please test again and let us know.

PR #6347, opened 2017-03-09
Add action indicator when resending a message.

greenjoe:

- [X] I have read [how to contribute](https://github.com/WhisperSystems/Signal-Android/blob/master/CONTRIBUTING.md) to this project

### Contributor checklist

 * Nexus 5x, Android 7.1.1

----------

### Description

When sending of message fails, user can go into details of that message and click RESEND button to retry. As issue #6307 mentions, after clicking that button nothing really indicates on the screen that message is being resent in the background.

This pull request hides the RESEND button when resending asynchronous task is started, as well as hides the "error_description" TextView and replaces it with the "action_description" TextView which says "Resending...". That should provide a nice indicator that clicking the RESEND button actually had an effect.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -880,6 +880,7 @@
     <!-- message_recipients_list_item -->
     <string name="message_recipients_list_item__verify">VERIFY</string>
     <string name="message_recipients_list_item__resend">RESEND</string>
+    <string name="message_recipients_list_item__resending">Resending&#8230;</string>
 
     <!-- GroupUtil -->
     <plurals name="GroupUtil_joined_the_group">
--- a/res/layout/message_recipient_list_item.xml
+++ b/res/layout/message_recipient_list_item.xml
@@ -56,6 +56,13 @@
                       tools:visibility="visible"
                       tools:text="New identity" />
 
+            <TextView android:id="@+id/action_description"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:visibility="gone"
+                tools:visibility="visible"
+                tools:text="action" />
+
         </LinearLayout>
 
         <Button android:id="@+id/conflict_button"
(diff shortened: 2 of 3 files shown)


mammuth:
Cool! 👍 
What happens if the sending fails another time? Becomes the resend-button visible again and the indicator (actionDescription) is removed?

greenjoe:
@mammuth Hello, I've tested the case when sending fails multiple times as well as the case when it fails and then succeeds and it seems to work fine.

There is no explicit code to make the resend-button visible again and hide the actionDescription, but I believe that every time resending is finished the whole `MessageRecipientListItem ` view is created again from scratch. Behavior that is now present in the master branch, which is to just make RESEND button disabled, also probably relies on the fact that the view will be created again (there is no code to re-enable the button).

I'm not an android developer by trade so I'm not 100% sure that my understanding is correct, but it looks like 
- `MessageDetailsActivity` creates instance of `MessageDetailsLoader` that loads message details 
- `MessageDetailsActivity.onLoadFinished` is called every time the resending task finishes and executes `MessageRecipientAsyncTask`
- That task calls `MessageDetailsActivity.updateRecipients` which creates a new adapter ` recipientsList.setAdapter(new MessageDetailsRecipientAdapter(...));` and the adapter in turn creates new instances of `MessageRecipientListItem` class. 

Please correct me if I'm wrong :-)

res/values/strings.xml
```diff
@@ -880,6 +880,7 @@
     <!-- message_recipients_list_item -->
     <string name="message_recipients_list_item__verify">VERIFY</string>
     <string name="message_recipients_list_item__resend">RESEND</string>
+    <string name="message_recipients_list_item__resending">Resending&#8230;</string>
```
McLoo: IRRC we moved to ... instad of using &#8230; (just 3 string left)

greenjoe: Well Android Studio suggested using the entity instead of three dots and I checked if the entity is used anywhere in the file - it was, so I used it too. Now I see that there are indeed more usages of three dots than of the entity. I can change it if it is important, but I believe it may be better to just do find and replace in one commit and change all the instances instead.

moxie0:
Did you try changing the button label and seeing how that looks instead of adding a new view that you swap?

PR #6334, opened 2017-03-09
Speex decoder is vulnerable

klemensbaum:
### Bug description
The speex decoder used by Signal is vulnerable.

A malicious user can initiate a voice call with a victim, then send bad speex frames to crash the victim's Signal app. It may also be possible to achieve remote code execution.

### Steps to reproduce
- Write a minimal program replicating Signal's usage of libspeex. You can use this one: https://gist.github.com/klemensbaum/fb7c41c981ac43a5abbb3d8fb99d3902
- Obtain some sample audio and convert it to raw speex frames. Save each frame as a separate file in a testcases directory.
- *(optional)* Prune redundant test cases with `afl-tmin`
- Run `afl-fuzz` over it: `afl-fuzz -i testcases -o findings -- ./speex-fuzz @@`

**Actual result:** `afl-fuzz` finds a lot of crashes
**Expected result:** `afl-fuzz` doesn't find any crashes

### Example crashes

Some example crashes found during a short `afl-fuzz` run: https://gist.github.com/klemensbaum/3df0260fe459c10a5adb2fd569b41dda

### DIFF
--- a/jni/redphone/AudioCodec.cpp
+++ b/jni/redphone/AudioCodec.cpp
@@ -67,6 +67,11 @@ int AudioCodec::init() {
   __android_log_print(ANDROID_LOG_WARN, TAG, "Encoding frame size: %d", enc_frame_size);
   __android_log_print(ANDROID_LOG_WARN, TAG, "Decoding frame size: %d", dec_frame_size);
 
+  if (dec_frame_size != SPEEX_FRAME_SIZE) {
+    __android_log_print(ANDROID_LOG_ERROR, TAG, "Decoded frame size does not match speex frame size\n");
+    return -1;
+  }
+
   speex_bits_init(&enc_bits);
   speex_bits_init(&dec_bits);
 
@@ -103,16 +108,16 @@ int AudioCodec::encode(short *rawData, char* encodedData, int maxEncodedDataLen)
 }
 
 int AudioCodec::decode(char* encodedData, int encodedDataLen, short *rawData) {
-  int rawDataOffset = 0;
-
   speex_bits_read_from(&dec_bits, encodedData, encodedDataLen);
 
-  while (speex_decode_int(dec, &dec_bits, rawData + rawDataOffset) == 0) { // TODO bounds?
-    WebRtcAecm_BufferFarend(aecm, rawData + rawDataOffset, dec_frame_size);
-    rawDataOffset += dec_frame_size;
+  // Output at most a single frame
+  // XXX: This assumes that dec_frame_size <= webrtc::NetEqImpl::kMaxFrameSize
+  if (speex_decode_int(dec, &dec_bits, rawData) == 0) {
+    WebRtcAecm_BufferFarend(aecm, rawData, dec_frame_size);
+    return dec_frame_size;
   }
 
-  return rawDataOffset;
+  return 0;
 }
 
 int AudioCodec::conceal(int frames, short *rawData) {
--- a/jni/redphone/AudioCodec.h
+++ b/jni/redphone/AudioCodec.h
@@ -36,6 +36,7 @@ public:
 
   int init();
   int encode(short *rawData, char* encodedData, int encodedDataLen);
+  // rawData must be able to hold at least SPEEX_FRAME_SIZE samples
   int decode(char* encodedData, int encodedDataLen, short* rawData);
   int conceal(int frames, short *rawData);
 


klemensbaum:
I haven't been able to test this tentative fix because `ndk-build` complains about missing third-party dependencies:

> Module redphone-audio depends on undefined modules: cutils stlport dl

The commit would have to be amended by someone who can build the jni libs, as these seem to be versioned within the git repository. It is not enough to rebuild only the Java parts with `./gradlew build`!

moxie0:
I haven't looked at this unit test yet, but that comment is a reference left over from pre-webrtc, when this code used to be responsible for managing the memory it is decoding into (check the git history). Once we switched to using webrtc components, that is no longer this code's responsibility, and as you can see we are not even given length information of the output buffer in the callback.

If the calling code (webrtc) is not effectively managing memory, then that's a bug in webrtc, and we should ask for a fix upstream. I would like to see a unit test that wraps the webrtc code, though, first. This doesn't appear to. In either case, RCE does not seem "very likely" here.

klemensbaum:
> If the calling code (webrtc) is not effectively managing memory, then that's a bug in webrtc, and we should ask for a fix upstream.

There is no way for the calling code to manage memory properly when the called function writes an unbounded number of bytes into the output buffer.

Anyhow, the interface has already been adjusted upstream by adding a `max_decoded_bytes` parameter to `AudioDecoder::Decode` (WebRTC BUG=4361, [relevant diff](https://chromium.googlesource.com/external/webrtc/+/7f7d7e3427cc70e1b8b050283ef031e28c83699a%5E%21/#F1))

> In either case, RCE does not seem "very likely" here.

May I ask how you come to this conclusion? If the encoded data sent by the attacker expands to more than `webrtc::NetEqImpl::kMaxFrameSize` (currently `2880`) samples after decoding, attacker-controlled data is written directly over the current stack frame (and later the calling function's stack frame, etc).

Android prior to 4.1 doesn't even provide full ASLR (see https://copperhead.co/blog/2015/05/11/aslr-android-zygote). It's not too far-fetched that a determined attacker could come up with a speex stream that decodes into a useful ROP chain.

erikd:
> There is no way for the calling code to manage memory properly when the called function writes an unbounded number of bytes into the output buffer.

The problem with the code you posted on the [Speex issue](https://github.com/xiph/speex/issues/4) was that the code was calling `speex_decode_int` in a while loop and that the packets you provided originally contained a single frame, but after AFL (American Fuzzy Lop) mutated them they ended up containing multiple frames.

The invariant that a single output frame is no larger than `webrtc::NetEqImpl::kMaxFrameSize` should still hold (fuzzing with my modified program in that Speex issue should confirm that).

The problem is still that `speex_decode_int` is being called in a while loop and there is no check for how many output frames are generated.

klemensbaum:
> The problem is still that `speex_decode_int` is being called in a while loop and there is no check for how many output frames are generated.

Yeah, and exactly this is the issue in the Signal code: https://github.com/WhisperSystems/Signal-Android/blob/c7c411b3f354bbd5930e0c0ef9778831ecc5f329/jni/redphone/AudioCodec.cpp#L110

There should just not be a loop there, period. WebRTC's `AudioDecoder` interface expects `Decode` to output a single frame.

moxie0:
> There should just not be a loop there, period. WebRTC's AudioDecoder interface expects Decode to output a single frame.

If you compile this code and listen to the audio (@klemensbaum you have to run ndk-build from the jni dir), it sounds pretty fucked up (like audio frames are missing). So I don't think that's true. Intuitively, if the calling code is only expecting a single frame, why pass in the full length to the callback for decoding?

erikd:
@moxie0 I can only speak with confidence about the test program that @klemensbaum provided to me.

That code assumes that the while loop will only output a single frame of speex decoded data and that assumption is *incorrect*. It is my understanding is that a single call to `speex_decode_int` will produce at most `kMaxFrameSize` output values. If you call it in a while loop it may produce more output values.

To me, it looks like the code in [Signal](https://github.com/WhisperSystems/Signal-Android/blob/c7c411b3f354bbd5930e0c0ef9778831ecc5f329/jni/redphone/AudioCodec.cpp#L110) suffers from exactly the same problem as @klemensbaum 's test prgram.

moxie0:
> @moxie0 I can only speak with confidence about the test program that @klemensbaum provided to me.

The test program is not wrapping the code that is actually managing the memory in Signal's case, so it's hard for me to tell whether there is an issue or not given the complexity of webrtc.  The webrtc API is expecting us to decode the full input, and we are simply given an output without any bounds information, so that part is pretty clearly up to the calling webrtc code.

If there is indeed an issue, I would like to fix it, but I don't immediately see how there's any bounds checking that we could do given the way that the API is structured. 

It sounds like @klemensbaum is saying that the upstream interface has been changed to include bounds information, so we could pull in that update if it turns out this is a problem.  Or, alternatively, if it is clear that the calling code is always allocating an output buffer of a given size, we could break the abstraction layer and simply hardcode that.

klemensbaum:
> @klemensbaum you have to run ndk-build from the jni dir

With ndk-r14 I had to set `APP_ALLOW_MISSING_DEPS=true` because the `Android.mk` files in `webrtc/`
are specifying system libraries in `LOCAL_SHARED_LIBRARIES`.

> If you compile this code and listen to the audio [...], it sounds pretty fucked up (like audio frames are missing). So I don't think that's true. Intuitively, if the calling code is only expecting a single frame, why pass in the full length to the callback for decoding?

Interesting. I haven't read through the entire code on the WebRTC side, but it seems like there are actually two different situations in which it calls `AudioDecoder::Decode`:
* [Internal comfort noise generator mode](https://github.com/WhisperSystems/Signal-Android/blob/c7c411b3f354bbd5930e0c0ef9778831ecc5f329/jni/webrtc/modules/audio_coding/neteq/neteq_impl.cc#L1594), where it expects only a single frame to be output.
* [To decode an entire packet](https://chromium.googlesource.com/external/webrtc/+/8602c429bec0e267cf06a977f71fb7ff9fcb85ec/webrtc/modules/audio_coding/neteq/neteq_impl.cc#1493), it calls the audio decoder in a loop.

> It sounds like @klemensbaum is saying that the upstream interface has been changed to include bounds information, so we could pull in that update if it turns out this is a problem.

That's correct.

> Or, alternatively, if it is clear that the calling code is always allocating an output buffer of a given size, we could break the abstraction layer and simply hardcode that.

Unfortunately it isn't as simple as that, as `Decode` is called in a loop (once for each packet) with a different position into the output buffer each time.

We could edit the `webrtc/` code to pass the remaining length along, until it's possible to migrate to a newer version.

moxie0:
Normally I wouldn't modify upstream code this way, but this entire code path is going to be deleted in the next release anyway, so if there is a problem this should be the most expedient path until then. Thanks everyone!

PR #6585, opened 2017-04-27
Move avatars to toolbar in individual convos

### LINKED ISSUE #6008: Remove Contact Pic Preceding Messages From Within Individual Conversations
I have:

### Feature Request
Could it be possible to remove the contact pic preceding every message (chat bubble) in a conversation in order to reduce the clutter and also make things consistent with the iOS and Desktop versions?

### Detailed Description
While the presence of contact pics before messages is helpful in group conversations (but still overwhelming when multiple messages are sent by the same person in succession as mentioned in #964), they are extremely redundant and make the screen more busy in one-to-one conversations. 
The colour of the chat bubbles is already an excellent method of distinguishing between messages. 

While doing something along the lines of other messaging apps (QKSMS, Hangouts, Facebook Messenger or Google Messenger) is what #964 sought by requesting that successive messages from the same contact be bundled, with the contact pic appearing only once for one message bundle, I am suggesting going the WhatsApp route and doing away with contact pics entirely, at least in individual conversation where they add very little, especially in excess.

Doing so would have the added benefit of making the Android app consistent with the iOS and Desktop apps where the contact pic doesn't appear within conversations. The desktop app instead displays the contact pic in the action bar which could be one alternate location for displaying the contact pic. 

Also, hopefully there should be no database issues in implementing this change like was mentioned in that other issue and that might expedite its implementation as it would be a purely UI move?

### Screenshots

Here are the iOS and desktop versions respectively:
screen322x572

unnamed

Here is how WhatsApp handles things presently:

unnamed

The contact pic is relegated to the action bar and messages bubbles only carry the name of the contact. Of course, even the name isn't present in case of individual conversations but I couldn't get a screenshot of that on the Play Store.

mateoeh:
// FREEBIE

### Contributor checklist

 * OnePlus 3T, Android 7.1.1
 * AVDs, Android 7.1.1/Android 4.4.2

----------

### Description

A picture is worth a thousand words:

|Before|After|
|--|--|
|screenshot_1493258661|screenshot_1493258613|

Fixes #6008

### DIFF
--- a/res/layout/conversation_title_view.xml
+++ b/res/layout/conversation_title_view.xml
@@ -1,34 +1,53 @@
 <?xml version="1.0" encoding="utf-8"?>
 <org.thoughtcrime.securesms.ConversationTitleView
               xmlns:android="http://schemas.android.com/apk/res/android"
+              xmlns:app="http://schemas.android.com/apk/res-auto"
               xmlns:tools="http://schemas.android.com/tools"
-              android:orientation="vertical"
+              android:orientation="horizontal"
               android:layout_width="match_parent"
               android:layout_height="match_parent"
               android:layout_gravity="center_vertical"
               android:gravity="center_vertical">
 
-    <org.thoughtcrime.securesms.components.emoji.EmojiTextView
-              android:id="@+id/title"
-              android:layout_width="wrap_content"
-              android:layout_height="wrap_content"
-              android:singleLine="true"
-              android:ellipsize="end"
-              android:transitionName="recipient_name"
-              android:drawablePadding="5dp"
-              android:gravity="center_vertical"
-              android:layout_gravity="center_vertical"
-              style="@style/TextSecure.TitleTextStyle"
-              tools:ignore="UnusedAttribute"/>
+    <org.thoughtcrime.securesms.components.AvatarImageView
+        android:id="@+id/contact_photo"
+        android:foreground="@drawable/contact_photo_background"
+        android:layout_width="32dp"
+        android:layout_height="32dp"
+        android:layout_marginRight="8dp"
+        android:layout_marginEnd="8dp"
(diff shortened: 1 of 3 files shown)


paride:
It seems to me that a lot of space is wasted between the "back" arrow and the avatar. Do we really need that arrow? The "back" (software) button does the same thing.

RiseT:
I'm in favor of that arrow, but agree that there's a lot of space wasted between the arrow and the avatar. Is this spacing defined by some Material guideline?

mateoeh:
Yes, the extra padding is supposed to be there to increase the touch target of the back button.

rainerzufall:
What I personally find odd, is the extended empty space in the green bubble (to the right side of the text in the after screenshot). Is this intentional? It seems that the bubble, after removing the avatar, could display more words per line. Like the white bubble already does.

RiseT:
@rainerzufall I guess that's because the first words of each line are too long in this example so they cannot get "unwrapped".

paride:
@mateoeh Still, I think the spacing around the arrow is a bit odd. For example I think it would look better if the space on the right was equal to the space on the left. Even like this it could be a problem on low resolution devices. Maybe it's possible to make the arrow disappear if the avatar contact's name can't be displayed completely?

moxie0:
Thanks but I don't want to make this change at this time. We'll revisit after some future changes are in.

Lokesh-Krishna:
@moxie0 Could you please share what other changes are required for this to get merged so that @mateoeh or somebody else could work on them to expedite this?

OriginalFannibal:
@moxie0 Are the changes in v4.9 (which shows up as major internal database changes on the Play Store changelog) the ones you were referring to? If so then can this be reconsidered now?

Issue #6620, opened 2017-05-09
Signal connects to the OWS server even when unregistered


paride:
I have:
- [x] searched open and closed issues for duplicates
- [x] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

----------------------------------------

### Bug description
Even after unregistering (disabling "Signal messages and calls" from the advanced preferences) Signal (websocket) keeps an open connection to the OWS server. The client uses the credentials that were valid before unregistering, and the server accepts them: they are still valid. This means two things:

- There is a bug in the (websocket) client, which keeps connecting even when it shouldn't. This is quite bad, as no contact to the server is expected when the push functionality is disabled.
- The server doesn't really delete the account when the push functionality is disabled. This is even worse, as the users expects any trace of their account to disappear after unregistering. Even commercial, non privacy-centric services promise this.

### Steps to reproduce
- Register to Signal
- Unregister
- Monitor the app's connections.
- If running an instance of Signal-Server: check the server log.

### Device info

**Signal version:** 4.5.3-website

paride:
The official support pages explicitly write about deleting the account, e.g.:

https://support.whispersystems.org/hc/en-us/articles/214160088-How-do-I-unregister-or-delete-my-account

but even unregistering from https://signal.org/signal/unregister/ doesn't really delete the account, it just sets `supportsSms = false`.

ddiver:
This is quite grave, Signal calling home when unregistered, leaking medadata (when I'm connected, from where, with which IP addresses...) with no reason at all. Please fix!

paride:
My understanding goes like this:
1. The client should not connect to the server when push messages are disabled. This is a bug, I think it affects only websocket clients and should be easy to fix.
1. Disabling "Signal messages and calls" is not really meant for unregistering, but rather for disabling push messages *on that specific device*. And this works as intended (modulo point 1).
1. A real unregistration option is missing, it's needed and I think it should be implemented. I think that the possibility to completely delete my own account is really important and I generally mistrust services which do not allow me to.
1. The same goes for https://signal.org/signal/unregister/: it should really delete the account.

ddiver:
Was my message censored? I wrote it before the last reply from @paride. Not nice!!

sigenc:
@ddiver Nobody does any censoring here. sometimes there are problems with posting them.

sigenc:
[...]

PR #6627, opened 2017-05-11
Don't connect to server if !isPushRegistered

paride:
Partial fix for #6620.

// FREEBIE


sigenc:
THX

paride:
@sigenc I'm not completely happy with this fix, as while it does prevent new connections to the server, it doesn't close the existing one. Unfortunately I don't see an easy way to do so. Let's see what @moxie0 thinks about it.

sigenc:
@paride If force close the app with your commit, it cuts the existing connection and prevents a new one. Or am i wrong about this?

paride:
@sigenc You are right, and I noticed that even if you don't force quit the connection dies anyway after some time and it's not opened again. Still it's not the best behavior.

moxie0:
in 4.7.0

PR #6632, opened 2017-05-12
Allow swipe gestures in media preview

FeuRenard:
### Contributor checklist

- [x] I am following the [Code Style Guidelines](https://github.com/WhisperSystems/Signal-Android/wiki/Code-Style-Guidelines)
- [x] I have tested my contribution on these devices:
 * Sony Xperia U, Android 4.4.4 (CyanogenMod)
 * AVD Nexus 5X, Android 7.1.1
- [x] My contribution is fully baked and ready to be merged as is
- [x] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the `Fixes #1234` [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)
- [x] I have made the choice whether I want the [BitHub reward](https://github.com/WhisperSystems/Signal-Android/wiki/BitHub-Rewards) or not by omitting or adding the word `FREEBIE` in the commit message of my first commit

----------

### Description
Please read the commit messages to see, which changes I applied.

The following problems remain:

- Raise offscreen page limit of the viewPager to enable a smoother paging experience? By default the viewed page plus one page to the right and one to the left is instantiated (offscreen page limit = 1). If you use a slow phone and swipe swiftly, you will reach a page whose media are not yet instantiated. You will see a black page which will display media with some latency. We can try to solve this issue by raising the offscreen page limit to 2. In worst case you are still using a slow phone and 5 GIFs are now instantiated (animated). I observed that the animations get slow in this situation. This could be solved by stopping GIF animations when GIFs are offscreen. If you want this behaviour, I will implement it.
- Do we want to be able to switch pages when an image is zoomed in? I implemented this feature in #5801, but decided to keep it out of this PR to keep the diff low.
- On Android versions prior to Marshmallow no video thumbnails are created. In media preview when swiping to a video the placeholder will be shown. It has to be tapped to start video playback. To make this clear, we should show the play overlay above the placeholder. This should go in a separate PR, too. -> #6635
- The video thumbnails maybe should have the same size as the video and don't fill the whole screen.

Sadly I'm not able to test this contribution with the platforms that use the old video player, which are Android versions 4.0.1 downwards (see #6603). I would be happy, if someone tested this PR using such a device :)

Fixes #2355 
Supersedes #5801

PR #6635, opened 2017-05-13
Show play overlay for video thumbnail placeholder

FeuRenard:
### Contributor checklist

 * Sony Xperia U, Android 4.4.4 (CyanogenMod)
 * AVD Nexus 5X, Android 7.1

----------

### Description
Showing the play overlay above the video thumbnail placeholder should highlight, that the video can be played by clicking it.

### Screenshots
Android 4.4.4 KitKat
device-2017-05-13-152730
device-2017-05-13-152820
device-2017-05-13-155931

Android 7.1.1 Nougat
bildschirmfoto von 2017-05-13 15-44-33
bildschirmfoto von 2017-05-13 15-45-03

### DIFF
--- a/res/values/attrs.xml
+++ b/res/values/attrs.xml
@@ -148,6 +148,7 @@
 
     <declare-styleable name="ThumbnailView">
         <attr name="backgroundColorHint" format="color" />
+        <attr name="thumbnail_play_overlay_tint" format="color" />
     </declare-styleable>
 
     <declare-styleable name="DeliveryStatusView">
--- a/res/values/attrs.xml
+++ b/res/values/attrs.xml
@@ -119,7 +119,7 @@
 
     <attr name="app_protect_timeout_picker_color" format="reference"/>
 
-    <attr name="thumbnail_play_overlay_tint" format="reference|color" />
+    <attr name="placeholder_play_overlay_tint" format="reference|color" />
 
     <declare-styleable name="CustomDefaultPreference">
         <attr name="custom_pref_toggle" format="string"/>
--- a/res/values/attrs.xml
+++ b/res/values/attrs.xml
@@ -119,6 +119,8 @@
 
     <attr name="app_protect_timeout_picker_color" format="reference"/>
 
+    <attr name="thumbnail_play_overlay_tint" format="reference|color" />
+
     <declare-styleable name="CustomDefaultPreference">
         <attr name="custom_pref_toggle" format="string"/>
     </declare-styleable>
@@ -148,7 +150,6 @@
 
     <declare-styleable name="ThumbnailView">
         <attr name="backgroundColorHint" format="color" />
-        <attr name="thumbnail_play_overlay_tint" format="color" />
     </declare-styleable>
 
     <declare-styleable name="DeliveryStatusView">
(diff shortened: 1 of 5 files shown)


res/values/attrs.xml
```diff
@@ -142,6 +142,7 @@
 
     <declare-styleable name="ThumbnailView">
         <attr name="backgroundColorHint" format="color" />
+        <attr name="thumbnail_play_overlay_tint" format="color" />
```
moxie0: We usually define attributes this way if they are "owned" by the widget. For instance, if in specifying the widget via XML, we want to pass along some value to the code. This doesn't seem to be the case here from what I can tell, it's more of a theme thing right?

res/values/colors.xml
```diff
@@ -55,4 +55,6 @@
 
     <color name="sticker_selected_color">#8cf437</color>
     <color name="transparent">#00FFFFFF</color>
+
+    <color name="thumbnail_play_overlay_tint">#77ffffff</color>
```
moxie0: Is there a reason this is a color but also an attribute that's defined differently in themes?

src/org/thoughtcrime/securesms/components/ThumbnailView.java
```diff
@@ -107,9 +109,14 @@ public void setImageResource(@NonNull MasterSecret masterSecret, @NonNull Slide
       getTransferControls().setVisibility(View.GONE);
     }
 
-    if (slide.getThumbnailUri() != null && slide.hasPlayOverlay() &&
-        (slide.getTransferState() == AttachmentDatabase.TRANSFER_PROGRESS_DONE || isPreview))
-    {
+    if ((slide.getThumbnailUri() != null || slide.hasPlaceholder()) &&
+        slide.hasPlayOverlay()                                      &&
+        (slide.getTransferState() == AttachmentDatabase.TRANSFER_PROGRESS_DONE || isPreview)) {
+      if (slide.getThumbnailUri() != null) {
```
moxie0: Is the objective to only tint the placeholder according to theme, but to always tint the overlay to a thumbnail with the same color regardless of theme?

FeuRenard: Exactly that.
White with the current tinting mode seems to be the best for thumbnail overlays. But the white one won't be visible over the dark theme placeholder.

res/values/colors.xml
FeuRenard: See below.

res/values/attrs.xml
FeuRenard: You are right, this attribute is misplaced here. I will move it.

src/org/thoughtcrime/securesms/components/ThumbnailView.java
moxie0: I think I get it, can we name those two things differently? It's kind of confusing to me that there's an attribute that's themed as well as a static color that's not themed, but they're both called the same thing.

FeuRenard: I renamed the tint color for the placeholder (which changes with the theme) to `placeholder_play_overlay_tint`.

FeuRenard:
I no longer need this PR for #6632 and I don't think this is very important. Moreover I consider the proposed overlay redundant information. We need the play overlay to distinguish between an image (no overlay) and a video thumbnail (overlay). But when there is a video placeholder image then we already know it's a video. So I see no longer a need for this overlay.

Issue #6640, opened 2017-05-14
Emoji interpretation depends on emoji set pref


cascheberg:
I have:
- [x] searched open and closed issues for duplicates
- [x] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

----------------------------------------

### Bug description
Different emoji settings potentially lead to different communication experiences for sender and receiver, see screenshots below for a random example. I suggest to remove the preference to use system emoji.

The screenshots below show the exact same message on the same phone.

**Actual result:** Emoji interpretaton depends on user pref
**Expected result:** Emoji depiction should essentially be identical

### Screenshots

![emojiset](https://cloud.githubusercontent.com/assets/3845150/26033430/a1ca2d70-38ac-11e7-8ebe-61a4d76aa3e7.png)
*Built-in emoji set*

![system](https://cloud.githubusercontent.com/assets/3845150/26033431/a3db6b2e-38ac-11e7-9028-7b4bbb8d6db6.png)
*System emoji set*

### Device info
**Android version:** Android 5.1.1
**Signal version:** 4.6.0

rainerzufall:
On Marshmallow:
![on android 6](https://cloud.githubusercontent.com/assets/12666598/26033945/1060708e-38b5-11e7-9559-58501cf23a05.png)
**Device info**
Android version: Android 6.0.1
Signal version: 4.6.0

I agree Emoji depiction should essentially be identical. Maybe instead of removing the option to use system emoji completely, it could only be enabled for the Android version which can depict the most current emoji set Signal uses. Recently Signal updated to Unicode 9 / Emoji 4. AFAIK only Android 7 does support all of these, therefore system emoji could/should only be optional for Android 7.

moxie0:
We started with built-in emoji only when we introduced emoji, but people feel very strongly about this, so we introduced the option to use the system emoji after much demand. I agree that it's not optimal, but I don't think I'm prepared to receive death threats over removing this option again. =(

FeuRenard:
We could add a line of text to the description of the setting. Like "Your friends may not see the same emoji like you."

fencepost-error:
Unfortunately this is still an issue today, and the change in #6659 was never made, the only text is "Disable Signal's built-in emoji support", no mention that this can break emojis.

PR #6659, opened 2017-05-18
Warn about disabling built-in emoji support

cascheberg:
### Contributor checklist

----------

### Description
Warn user about disabling Signal's built-in emoji because it can lead to different interpretations.

See #6640, cc @FeuRenard 

// FREEBIE

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -1192,8 +1192,8 @@
     <string name="preferences_chats__when_roaming">When roaming</string>
     <string name="preferences_chats__media_auto_download">Media auto-download</string>
     <string name="preferences_chats__message_trimming">Message trimming</string>
-    <string name="preferences_advanced__use_system_emoji">Use system emoji</string>
-    <string name="preferences_advanced__disable_signal_built_in_emoji_support">Disable Signal\'s built-in emoji support</string>
+    <string name="preferences_advanced__disable_built_in_emoji">Disable built-in emoji</string>
+    <string name="preferences_advanced__disable_signal_built_in_emoji_support">Disable Signal\'s built-in emoji support and use system emoji instead. This might result in different emoji for you and your contacts.</string>
     <string name="preferences_advanced__video_calling_beta">Video calling beta</string>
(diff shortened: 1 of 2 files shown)


FeuRenard:
I didn't think about a dialog, because of the [material guidelines](https://material.io/guidelines/components/dialogs.html) saying:

> Use dialogs sparingly because they are interruptive.

But its use here could be justified as my suggested approach to just add a sentence in the description of the setting could be easily overlooked.

Dyras:
I think there should be a description below the checkbox that says something like
"Enabling system emojis might result in the person you are chatting with not seeing the exact same emojis as you do"
Otherwise I'd be fine with a dialog box, I think consistency is important enough to justify an extra dialog box

moxie0:
I'd prefer not to have a pop-up dialog

cascheberg:
PR updated, it's now a textual change only. I'm open for suggestions regarding better wording.

RichJeanes:
Is this really necessary? If you use Signal to send an SMS message to a person with a different platform (or they are using the Chrome app), they won't see the same thing as you no matter what this setting is on. Google, Microsoft, Apple, and Samsung all have their own emoji sets and are all platforms that these messages can appear on (Windows and macOS via Chrome App). I can't say that I've ever seen this kind of warning on any messaging system.

Dyras:
@RichJeanes I think it would be a neat addition because not everyone realizes that the emojis that you see may not be what the other person sees.

moxie0:
I think I agree this isn't necessary. Whether this settings is enabled or disabled, people might see different emoji.

fencepost-error:
I'd say this is necessary, I've just spent some time debugging a problem where family member's emoji were showing as emojibake, it took way too long to figure out, without direct access to the devices, that some of them were using system emoji.  If the setting can't be removed then it should at least warn that, when enabled, some people you communicate with are going to get emojibake.

A better option would be to display build-in emoji, or something other than emojibake, if the given emoji can't be rendered.

PR #6669, opened 2017-05-22
Keep screen off after end of audio playback

### LINKED ISSUE #6654: Screen turns on, despite still being next to the ear (voice and audio messages)
I have:

### Bug description
When listening to an audio or voice message ends, the screen turns on and the touchscreen accepts input, despite the phone still being next to the ear, after listening to an audio or voice message. This leads to unwanted calling and other actions, the user never intended to do.

### Steps to reproduce
- Start listening to an audio or voice message.
- Hold the phone next to the ear, so that the screen turns off and input from the touchscreen is ignored.
- Wait for the audio or voice message to end.

**Actual result:** The screen turns back on and input from the touchscreen is no longer ignored.
**Expected result:** The screen stays off and input is ignored, as long as the phone is next to the ear.

### Device info
**Device:** Moto X Play
**Android version:** 6.0.1
**Signal version:** 4.6.0

### Link to debug log
https://gist.github.com/anonymous/d495bc418ede007bf7b5f563f446ea85

FeuRenard:
### Contributor checklist

 * AVD Nexus 5X, Android 7.1.1
 * Sony Xperia U, Android 4.4.4 (CyanogenMod)

----------

### Description
Would be nice if some tested this on a physical Lollipop+ device. The emulator behaves a bit oddly when its screen is turned off (it just freezes and accepts input). But the logs indicate the correct behaviour :wink:.

Fixes #6654
// FREEBIE

### DIFF
--- a/src/org/thoughtcrime/securesms/audio/AudioSlidePlayer.java
+++ b/src/org/thoughtcrime/securesms/audio/AudioSlidePlayer.java
@@ -134,7 +134,7 @@ public class AudioSlidePlayer implements SensorEventListener {
 
           sensorManager.unregisterListener(AudioSlidePlayer.this);
 
-          if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP && wakeLock.isHeld()) {
+          if (wakeLock != null && wakeLock.isHeld()) {
             wakeLock.release(PowerManager.RELEASE_FLAG_WAIT_FOR_NO_PROXIMITY);
           }
         }
@@ -161,7 +161,7 @@ public class AudioSlidePlayer implements SensorEventListener {
 
           sensorManager.unregisterListener(AudioSlidePlayer.this);
 
-          if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP && wakeLock.isHeld()) {
+          if (wakeLock != null && wakeLock.isHeld()) {
             wakeLock.release(PowerManager.RELEASE_FLAG_WAIT_FOR_NO_PROXIMITY);
           }
         }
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/audio/AudioSlidePlayer.java
```diff
@@ -133,7 +133,10 @@ public void onCompletion(MediaPlayer mp) {
           }
 
           sensorManager.unregisterListener(AudioSlidePlayer.this);
-          if (wakeLock != null && wakeLock.isHeld()) wakeLock.release();
+
+          if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP && wakeLock.isHeld()) {
```
moxie0: These version checks feel strange to me. The wakelock should only exist if it's api 21+?

FeuRenard: Yes, the wakelock we use is only available with API 21+:
https://developer.android.com/reference/android/os/PowerManager.html#PROXIMITY_SCREEN_OFF_WAKE_LOCK
Otherwise we already set it null. The functionality is only available in Signal for Lollipop+:
https://github.com/WhisperSystems/Signal-Android/blob/master/src/org/thoughtcrime/securesms/audio/AudioSlidePlayer.java#L79

I also would like to keep the null check, but [`WakeLock.release(flags int)`](https://developer.android.com/reference/android/os/PowerManager.WakeLock.html#release(int)) was introduced with API 21. So this version check is necessary. And the version check makes the null check redundant. Before we used [`WakeLock.release()`](https://developer.android.com/reference/android/os/PowerManager.WakeLock.html#release()) (without parameters/flags) which is as old as Android itself.

moxie0: I don't see why the version check is necessary? It can't be initialized if it's a lower version. The way it's written implies, to me, that it's possible for the wakelock to have been allocated on a lower version, and in that case it is just never released.

FeuRenard: Sorry, I assumed this wouldn't compile:
bildschirmfoto von 2017-05-25 22-29-24

Issue #5146, opened 2016-01-28
SMS cut off at 2000 characters


asvetlovsky:
When sending an unsecured SMS greater than 2000 characters, it appear cut-off in my sent view. The receiver confirmed that the entire text has gone through successfully. 

It seems like Signal cuts off displaying SMS messages at about 2000 character, even though it's possible to send longer ones.

liliakai:
[Evidence suggests](https://github.com/WhisperSystems/Signal-Desktop/issues/401) this is also true for Signal messages, not just SMS. :open_mouth:

agrajaghh:
@liliakai true, I just tested this.

Sending push messages from android is possible with > 2000 characters. The message gets fully trasmitted and shown at Signal-Desktop. But just the first 2000 characters are displayed on android (sending & receiving side)

agrajaghh:
What is the reason for this 2000 characters limit? Can't be DOS protection, since attachments are way bigger anyway, no? 

I'm not sure how to proceed, but I would suggest to raise this limit to at least 20.000 for push messages, since we have desktop clients now as well, and it's an imaginable usecase to copy & paste longer texts I guess.

Since we can't control other SMS apps, I guess incoming SMS should always be displayed at full length, no?

McLoo:
@agrajaghh wasn't  there an issue regarding (decryption) speed for larger messages when scrolling through convos?!

agrajaghh:
@McLoo that rings a bell, couldn't find it at the moment though :smirk: 
but I found #3658 and #3362 (duplicates) :grin:

lolstarz:
What's the verdict on this? 20k character limit for displaying sms? Or 20k characters for outbound messages, display full text for inbound SMS as @agrajaghh suggested? 20k limit for everything seems reasonable, I don't think SMS is protocol over which I want to be receiving >20k character messages.

liliakai:
>What is the reason for this 2000 characters limit?

From what I gather, there are some performance issues and crashes associated with rendering huge text views in Android.

getsadzeg:
@liliakai , So you say it's android fault of rendering huge text and we can't solve it?

If it's true, I think this issue should be closed.

asvetlovsky:
Other SMS apps, like Hangouts, are able to render more than 2000 characters
just fine.

getsadzeg:
Okay, so it's not Android's fault.

[...]

PR #6692, opened 2017-05-30
Inform user that message is too long to be displayed.

greenjoe:
### Contributor checklist

 * LG Nexus 5x, Android 7.1.2

----------

### Description
Currently text messages are silently cut when they are longer than 2000 characters (as mentioned in issue #5146). It can be really annoying as users may not even notice that part of message is not shown if the message is not cut half-sentence. I had this problem multiple times when my contact copied a longer text to me using Signal Desktop and I tried to read it on Android.

Resolving the issue properly probably requires a lot of work (like adding a way to display full message body or splitting large messages into smaller parts) and I believe it will take some time, especially considering that the issue mentioned is already over a year old. In the meantime I think it would be fair to inform users that part of message is not displayed so that they can resolve it with their contacts somehow (like using another communication channel or splitting the message manually).

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -425,6 +425,7 @@
     <string name="MessageRecord_you_marked_your_safety_number_with_s_verified_from_another_device">You marked your safety number with %s verified from another device</string>
     <string name="MessageRecord_you_marked_your_safety_number_with_s_unverified">You marked your safety number with %s unverified</string>
     <string name="MessageRecord_you_marked_your_safety_number_with_s_unverified_from_another_device">You marked your safety number with %s unverified from another device</string>
+    <string name="MessageRecord_message_too_long">Message is too long to be displayed completely.</string>
 
 
     <!-- PassphraseChangeActivity -->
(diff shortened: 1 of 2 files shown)


res/values/strings.xml
```diff
@@ -421,6 +421,7 @@
     <string name="MessageRecord_you_set_disappearing_message_time_to_s">You set disappearing message time to %1$s.</string>
     <string name="MessageRecord_s_set_disappearing_message_time_to_s">%1$s set disappearing message time to %2$s.</string>
     <string name="MessageRecord_your_safety_number_with_s_has_changed">Your safety number with %s has changed.</string>
+    <string name="MessageRecord_message_too_long">Message is too long to be displayed.</string>
```
greenjoe: Added "completely". I think that displaying number of truncated characters would just clutter the view.

greenjoe:
Rebased onto master and resolved conflicts. 
When testing I've noticed that Signal-Desktop now displays a nice warning "Android clients will only receive the first 2000 characters of this message" - wonderful! :)

Issue #751, opened 2014-02-26
Adapt theme to light conditions


merkste:
Under low light conditions the light design is too bright and puts a strain on the eyes. Thus, a nice thing to have would be an option to switch between light and dark theme automagically using the light sensor.

| Theme |  |
| :-- | :-: |
| light theme | ( ) |
| dark theme | ( ) |
| depending on light condition | (o) |

NorthwestBased:
IMO, the problem of screens being too bright isn't a problem with textsecure, but a problem with phones in general, and shouldn't be fixed locally, but globally for a whole phone. Maybe something like this?

https://play.google.com/store/people/details?id=110556507151732274846

@tx3eh8IUD1 Android 4.2.2 (and I'd assume more recent versions) has a font size setting to globally increase the size of text. Maybe the font in the situation you bring up is too small, but I'm not sure that we should reinvent the wheel when Android has built-in fixes for those who have these problems. I don't believe that textsecure is the only app to use small text.

manhole11:
Yes please! This would be so useful, I currently keep swapping between light and dark theme

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #6756, opened 2017-06-21
Automatic dark mode based on ambient light

fercarcedo:
// FREEBIE

### Contributor checklist
 * Nexus 5, Android 7.1.2
 * Virtual device Pixel, Android 7.0.0
 * Virtual device Pixel XL, Android O
 * Virtual device Nexus 5X, Android 4.2

----------

### Description
This pull request enables Signal to automatically change to dark theme based on the current ambient light condition, like many other apps do.
I have tested this pull request in both emulators (setting manually the number of lux), as well as on my Nexus 5, going from outdoors to a dark room indoors, and in the night outside.

### DIFF
--- a/src/org/thoughtcrime/securesms/util/ActivityUtil.java
+++ b/src/org/thoughtcrime/securesms/util/ActivityUtil.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2017 Fernando Garcia Alvarez
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thoughtcrime.securesms.util;
+
+import android.annotation.TargetApi;
+import android.app.Activity;
+import android.content.Intent;
+import android.os.Build;
+
+/**
+ * Class used to perform actions
+ * on Activities in a compatible way
+ *
+ * @author fercarcedo
+ */
+public class ActivityUtil {
+  public static void recreateActivity(Activity activity) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
+      recreateHoneycomb(activity);
+    } else {
+      recreatePreHoneycomb(activity);
+    }
+  }
+
+  @TargetApi(Build.VERSION_CODES.HONEYCOMB)
+  private static void recreateHoneycomb(Activity activity) {
+    activity.recreate();
+  }
+
+  private static void recreatePreHoneycomb(Activity activity) {
+    Intent intent = activity.getIntent();
+    intent.addFlags(Intent.FLAG_ACTIVITY_NO_ANIMATION);
+    activity.finish();
+    activity.overridePendingTransition(0, 0);
+    activity.startActivity(intent);
+    activity.overridePendingTransition(0, 0);
+  }
+}
--- a/src/org/thoughtcrime/securesms/util/ActivityUtil.java
+++ b/src/org/thoughtcrime/securesms/util/ActivityUtil.java
@@ -13,24 +13,13 @@ import android.os.Build;
  */
 public class ActivityUtil {
   public static void recreateActivity(Activity activity) {
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
-      recreateHoneycomb(activity);
-    } else {
-      recreatePreHoneycomb(activity);
+    if (activity != null) {
+      Intent intent = activity.getIntent();
+      intent.addFlags(Intent.FLAG_ACTIVITY_NO_ANIMATION);
+      activity.finish();
+      activity.overridePendingTransition(0, 0);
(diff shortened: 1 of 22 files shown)


src/org/thoughtcrime/securesms/util/AutoDarkModeManager.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+public class AutoDarkModeManager implements SensorEventListener {
+  private static final float MAX_LUX_FOR_DARK_THEME = 10;
+  private static boolean showDarkTheme;
+  private static String lastActivityName;
+
+  private Activity activity;
+
+  public AutoDarkModeManager(Activity activity) {
+    this.activity = activity;
```
moxie0: think this is a memory leak waiting to happen

src/org/thoughtcrime/securesms/util/AutoDarkModeManager.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  public static boolean shouldShowDarkTheme() {
+    return showDarkTheme;
+  }
+
+  public void listenForCurrentActivityIfNecessary() {
+    if (!activity.getClass().getSimpleName().equals(lastActivityName)) {
```
moxie0: this feels a little gross on two levels, first using the textual representation of a class name to track activity rather than identity, and then having potentially multiple instances of this class but only one listening

src/org/thoughtcrime/securesms/util/AutoDarkModeManager.java
```diff
@@ -0,0 +1,82 @@
...
+  @Override
+  public void onSensorChanged(SensorEvent event) {
+    lastActivityName = activity.getClass().getSimpleName();
+
+    float lux = event.values[0];
+    boolean previousShowDarkTheme = showDarkTheme;
+    showDarkTheme = lux <= MAX_LUX_FOR_DARK_THEME;
+    if (previousShowDarkTheme != showDarkTheme)
+      ActivityUtil.recreateActivity(activity);
+    stopListening();
```
moxie0: this feels racey to me

src/org/thoughtcrime/securesms/util/ActivityUtil.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (C) 2017 Fernando Garcia Alvarez
```
moxie0: This would need to be copyright OWS with GPLv3 headers for me to merge

fercarcedo:
I have fixed the problems you mentioned. Please let me now of any problems.
Thank you.

src/org/thoughtcrime/securesms/util/AutoDarkModeManager.java
```diff
@@ -0,0 +1,72 @@
...
+
+  private synchronized void parseAmbientLightValue(SensorEvent event) {
+    luxValueFound = true;
+    float lux = event.values[0];
+    boolean previousShowDarkTheme = showDarkTheme;
+    showDarkTheme = lux <= MAX_LUX_FOR_DARK_THEME;
+    Activity currentActivity = activity.get();
+    if (previousShowDarkTheme != showDarkTheme && currentActivity != null)
+      ActivityUtil.recreateActivity(currentActivity);
+    stopListening();
```
moxie0: It looks like this is the only place you ever unregister a listener. It looks like that means that every time you enter an activity, you get a new instance of this class and a new listener. They stack up forever, unless there's a sensor event.

On the other hand, as soon as a sensor change comes in, whether it's below the threshold or not, it gets unregistered.  From what I can understand, this behavior is both a memory leak and breaks the functionality?

src/org/thoughtcrime/securesms/util/AutoDarkModeManager.java
```diff
@@ -0,0 +1,72 @@
...
+  public void onSensorChanged(SensorEvent event) {
+    parseAmbientLightValue(event);
+  }
+
+  @Override
+  public void onAccuracyChanged(Sensor sensor, int accuracy) {
+
+  }
+
+  private synchronized void parseAmbientLightValue(SensorEvent event) {
```
moxie0: Why is this method synchronized? If the idea is that it can be called off the UI thread, then it seems like all the values that it accesses would also need synchronized access from their writes on the UI thread.

src/org/thoughtcrime/securesms/util/AutoDarkModeManager.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+  }
+
+  private void stopListening() {
+    sensorManager.unregisterListener(this);
+  }
+
+  @Override
+  public void onSensorChanged(SensorEvent event) {
+    parseAmbientLightValue(event);
```
moxie0: Private methods might read instance state if necessary, but they should not have the side effect of modifying instance state unless they're initializer methods. The method you're calling is named "parseAmbientLightValue," but it does much more than parsing an argument -- it modifies instance state, updates listeners, even potentially destroys an activity.

fercarcedo:
I have refactored the code to be cleaner, and also now the sensor gets unregistered as soon as onPause is called. The behavior of unregistering the sensor after the first value is found is correct, I did that in order to preserve battery life and also to avoid annoying the user, since for example if the user is typing a message and then it becomes necessary to switch to the dark theme based on ambient light, the user will get annoyed due to the activity being restart while typing the message, but if you want another behavior, let me know.

fercarcedo:
I've also made a rebase to the pull request to update it to the new changes of Signal (and test it with the new 4.8.0 version)

FeuRenard:
This PR currently includes some commits from master. Seems like something went wrong during rebase.

Fixes #751.

fercarcedo:
I have removed the merge commit from the PR, which was causing some master commits to appear here. Sorry, that's my first time doing a PR

PR #6776, opened 2017-06-29
Prevent MitM attacks, updates http:// to https://

sergey-wowwow:
App used insecure HTTP protocol what allowed to an attacker to perform Man in the Middle attacks by intercepting user traffic and its probable modification.


moxie0:
i updated the play store listing

PR #6826, opened 2017-07-27
Remove canonical address database.

moxie0:
This is a large dangerous change that I could use help testing, particularly given the variety and inconsistency of phone numbers in the world. Replacing the Canonical Address Database should open the door for a lot of improvements if we can make the migration work well.


FeuRenard:
Could you please be a bit more specific about what and how to test?

nrizzio:
Forum thread: https://whispersystems.discoursehosting.net/t/new-pr-remove-canonical-address-database-call-for-testing/1189

moxie0:
The change touches just about everything. It removes the entire concept of the "recipient id," which is used in almost every activity and every database table, so this change has the potential to break functionality in almost any activity.

The worst case, though, is probably if something doesn't happen correctly with the migration. This change replaces every instance of a "recipient id," as well as every instance of an arbitrarily formatted number in every database table with an "address," which is either an e164 formatted number or a group identifier.

The process of generating e164 numbers for all existing recipient ids and arbitrarily formatted numbers has the slight potential to go awry given how inconsistent phone number formats are in all of the world's locales.

So to summarize it would be helpful to know:

a) Does this migration succeed without any problems on existing large Signal message databases (particularly for non-US users).

b) Did I miss something somewhere in some of the functionality for any of the activities it touches (almost all of them).

RiseT:
Just tried it. The current PR (using the [moxie/original_sin](https://github.com/WhisperSystems/Signal-Android/tree/moxie/original_sin) branch) mixes up all my conversations in conversation view. The converations are listed with an incorrect contact (or the contacts are listed with the conversation content of another contact, depending how you look at it).

Log:
https://gist.github.com/anonymous/53fa7319bc99e7fc90bcbece84628406

moxie0:
@RiseT It would be great if you could produce a sample data set that reproduces this. Something like a plaintext backup file that, once imported and taken through the migration, reproduces the behavior you're seeing.

RiseT:
Tried to reproduce this with a new database and 2 sample conversations/contacts. "Unfortunately", everything went well this time. I guess I have to wait a couple of days until my conversation list fills up a bit and then try again.

When the issue arose, there were some SMS conversations with contacts not listed in Android's address book (PayPal 2FA & a missed call notification from my mobile provider), if of importance.

FeuRenard:
I didn't observe any bugs yet with

- two single conversations
- one group
- self conversation (without my number being in contacts)

on Android 4.4.4 (CyanogenMod).

moxie0:
@RiseT Did you happen to save the incorrectly migrated database somewhere?

I'll hold the 4.9.0 release until one of us is able to reproduce what @RiseT saw.

RiseT:
@moxie0 

> Did you happen to save the incorrectly migrated database somewhere?

Unfortunately not.

> Also, was it literally every single conversation thread that was labeled with the incorrect contact, or just some of them?

All of them (about 5). I didn't look at the archived ones, though.

> Were the ones that were wrong "symmetirc," in that the conversation with Alice and Bob were swapped, or were they asymmetirc, in that Alice's conversation became Charlie, Charlie became Bob, and so on?

If I remember this correctly, asymmetric.

Question: What exactly should I focus on when trying to reproduce this?

- Is it about the specific contacts/senders/recipients (i. e. the phone numbers) I've had in the conversation list at that point in time? Should I try to re-create conversations with exactly those phone numbers?
- If yes, should I focus on Signal contacts, non-Signal contacts, or both?
- Are the conversations' contents of any importance (I guess not)?

moxie0:
> Question: What exactly should I focus on when trying to reproduce this?

Unfortunately I don't know. I have no idea how that would have happened, since it should be tottaly impossible (although obviously it isn't). If it was only four of five conversations, I'd try recreating exactly those.

RiseT:
@moxie0 
Managed to reproduce. Yay. It doesn't look exactly like the issue before, but it's definitely incorrect.

I've had 3 conversations with 3 different contacts: A, B, and C.
After installing Original Sin, conversation A and C both display A as the contact.

I suspect the culprit could be conversation B: That conversation is a 2FA SMS from online banking which somehow features a name (not from Android's address book) but not a phone number.

Small caveat: As I've had to do a re-install of Windows and thus Android Studio, because of the incompatibility of the produced builds I've had to re-install the unmodified Signal 4.8.1 and import the plaintext backup before installing Original Sin. But I guess this shouldn't matter.

So, do you have some more private channel in order to send you my plaintext backup?

moxie0:
@RiseT I was *just* able to reproduce this! 0ae60da68b86efd5916c27984e474f5e1b39c797 should fix. Thanks so much for doing all this testing and finding this issue, definitely let me know if you still see bad migrations with that commit.

RiseT:
Another migration test with https://github.com/WhisperSystems/Signal-Android/commit/0ae60da68b86efd5916c27984e474f5e1b39c797 and my plaintext backup reveals: Confirmed, issue fixed!

bjoli:
I have a large conversation history and I saw this made it into the beta channel.  My brother's conversation history became botched by the update, and I thought I could maybe be of any help if the same thing happens to me.

How do I best create a snapshot? The built-in  export feature only exports plain text stuff. Is that enough to to reproduce an eventual error?

/Linus

FeuRenard:
Some (currently 6) related complaints here #6838

moxie0:
@bjoli Could you provide any more information on in what way your brother's conversation history is "botched?"

moxie0:
@FeuRenard That's just the first moment those stack traces access the DB, so it's the first moment the migration runs.

moxie0:
In 4.9.0

PR #6913, opened 2017-09-02
Replace all instances of unsecure with insecure

nrizzio:
### Contributor checklist

 * Virtual device Nexus 5, Android Oreo

----------

### Description

The words insecure and unsecure are used interchangeably within Signal. Insecure is used more often than unsecure, and having only one form of the word increases clarity. This commit replaces all instances of unsecure with insecure.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -92,10 +92,10 @@
     <string name="ConversationItem_received_key_exchange_message_tap_to_process">Received key exchange message, tap to process.</string>
     <string name="ConversationItem_group_action_left">%1$s has left the group.</string>
     <string name="ConversationItem_click_for_details">Tap for details</string>
-    <string name="ConversationItem_click_to_approve_unencrypted">Tap for unsecured fallback</string>
+    <string name="ConversationItem_click_to_approve_unencrypted">Tap for insecure fallback</string>
     <string name="ConversationItem_click_to_approve_unencrypted_sms_dialog_title">Fallback to unencrypted SMS?</string>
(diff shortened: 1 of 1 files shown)


moxie0:
There was a whole discussion and intentional change around making these instances "unsecured" instead of "insecure." I forget all of the rationale, but I do think there is a subtle difference that makes sense in these instances. Thanks!

igor2x:
I know this pull request is closed, but can I add two cents.

Actually discussion on forum https://whispersystems.discoursehosting.net/t/why-is-double-wording-insecure-sms-and-unsecured-sms/1260 was a question from user shellSignal why there is a double wording when sending insecure/unsecured SMS. Based on this question alone @nrizzio created maybe too broad pull request for replacing ALL of words with "unsecured" to "insecure".

I do see the point @shellSignal asked on forum. It is little bit confusion. You select "Insecure SMS" and it appears "Unsecured SMS" in input box. User may ask himself "What??? That is not the option I have selected."

The oder point on forum was, major translations have this two strings translated with the same wording. I know we can argue there is a bug in all of translations, but translators probably try to fix the problem with simplification.

I am not arguing which of the words are correct "Insecure SMS" or "Unsecured SMS", because I am not a native English speaker and I most probably do not understand all of different meanings of this words, but what I am arguing is there is no point of having two different wording. It is little bit strange that user selects one option and it appears another in input text. Strange and confusing.

I think the wording of selection and input box should be the same and as @shellSignal written "this is a usability bug". Which one is better option, I don't know. Native English speakers should tell more.

Print screen of the problem:
image

Issue #1232, opened 2014-03-24
Search text in message/messages/conversation/conversations/chat/chats


Genis-Sage:
I would like the possibility to search , for words/phrases/numbers in chats, especially if they are old and therefore long.
For example if someone send me an address, but I can't remember the date,
it would be helpful if I can press options/search and then have a text field (and a search button).

inetpro:
+1

sigenc:
@moxie0 Thank you for opening this issue again.

Hello all,

i think this is one of the most wanted features for signal. Every one of my not so technical friends ask me why he can't search his messages. The Brown University has an interesting approach to handle this. Thanks to jaypatelani on the community forum, we know that there is some possibility to search in signal messages without compromising security and without doing major changes to the signal infrastructur. 

http://esl.cs.brown.edu/blog/signal/

> In this project, we set out to explore whether encrypted search could be used to add search functionality to the Signal messaging app securely and without compromising efficiency. The design we proposed incurs minimal changes to Signal's current architecture and has several advantages over the naive solutions which include using a plaintext in-memory index or using full-database encryption solutions like SQLCipher.

And my personal most anticipated feature that would come as a side effect of this:

> In the future, we plan on improving the project in two ways. First, we would like to store EDB and message encryption keys in either the Android Key Store or, depending on the device, in secure hardware. The benefit of this would be that the encrypted index would be secure even while the app is in open mode. Second, we would like to use encrypted search techniques to also hide some of the meta-data that Signal currently stores in plaintext like timestamps, sender and device ID.

justjoeyuk:
Is it worth moving towards https://realm.io for something like this?

wes-novack:
+1 
very surprised to see that we're still missing content search in Signal messenger on Android

moxie0:
@wes-novack Please do not bump issues: https://github.com/WhisperSystems/Signal-Android/blob/master/CONTRIBUTING.md#dont-bump-issues  

You can use the reaction emoji in GH to express support, thanks.

sandeepcmsm:
+1

PR #6952, opened 2017-09-10
add search for finding newest matching message in a conversation

gulp21:
### Contributor checklist

 * GT-I9300 (Galaxy SIII), LineageOS 14.1 (Android 7.1)
 * Nexus 4 Api 25 Emulator (Android 7.1.1)

----------

### Description

This PR addresses a part of #1232: Searching a text within a conversation. So the use case “You have sent me your address some time ago and I don’t want to scroll through the whole chat history.” is addressed, but the case “Someone has sent me that address, I don’t remember in which conversation.” is not.

Iterating through all messages is not a very fast solution, but still better than manual searching. That’s why only the loaded part of the conversation (500 messages by default) is searched, to get a fast feedback whether the information you’re looking for is in the conversation or not; if no match is found the user can decide to load the whole conversation and search again. If you don’t think that the user experience with this patch is good enough, I’m fine if you don’t want to merge this PR, but I considered this patch quite useful during the last month, so I wanted to share it.

### DIFF
--- a/res/menu/conversation_search.xml
+++ b/res/menu/conversation_search.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <item android:id="@+id/action_search_message"
+        android:title="@string/conversation_fragment__action_search_message"
+        app:actionViewClass="android.support.v7.widget.SearchView"
+        app:showAsAction="always"/>
+
+</menu>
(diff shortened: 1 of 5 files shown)


0x70b1a5:
+1. Been craving this change for a very long time, and as a frequent user it's the only thing stopping me from recommending Signal to more people.

strobelm:
First of all thanks for implementing this! Would be great to have this merged. 

Concerning performance: since the queries are independent one could perhaps use ``parallelStream``, ``filter`` and ``findFirst`` to speed things up.

moxie0:
Thanks but I think we just want to get the ciphertext out of the database so we can do this properly instead.

Issue #6964, opened 2017-09-14
Profiles: cannot share profile without adding to contacts after having replied


cascheberg:
I have:
- [x] searched open and closed issues for duplicates
- [x] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

----------------------------------------

### Bug description
In a new conversation with an unknown contact, users can't go back to choose "Don't add, but make my profile visible" after replying once.

#### Note
It's also not obvious that "add to contacts" will make one's profile visible, unless you know the details of the implementation. One extra line of text at the top could fix that.

### Steps to reproduce
- receive a message from someone not in your contacts
- reply to conversation

**Actual result:** `UnknownSenderView` will disappear. Now the only option left to share one's profile is to save that contact to phone contacts, which is not obvious.
**Expected result:** There should be an option in conversation settings to make one's user profile visible.

### Device info
**Signal version:** 4.10.9

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #7046, opened 2017-10-05
Allow refusing profile sharing

### LINKED ISSUE #6927: Profiles: no way to hide the "share profile with group" message
As the title says: if I don't want to share my profile with a group, there is no way to just hide the message that suggest doing so. There should be a way to change idea later and share the profile, but that bar is annoying.

An indicator to know if the profile is shared or not is also missing.

Not sure if this also happens in conversations with people not in the contacts list.

interfect:

- [X] I have read [how to contribute](https://github.com/WhisperSystems/Signal-Android/blob/master/CONTRIBUTING.md) to this project

### Contributor checklist

 * Oneplus 5, Android 7.1.1

----------

### Description
Fixes  #6927 by adding a "never ask again" button to the dialog
that comes up when you tap the banner.

I've tested this on my device. I see the new option in the dialog that appears when you tap on the banner, and I can select the never ask again button and make the banner disappear. Going back to the chat, the banner is still gone. So I'm fairly certain this solves the issue.

I only have the string for English, because I only speak English.

(I had trouble testing it, because my build isn't signed with the key used for the Play Store and I couldn't install it over the existing Signal to use its database. I couldn't get the bar on any group chats I made, and eventually I had to resort to asking all my Facebook friends to Signal group chat me. I've also clobbered my Signal keys about 3 times today. I wish there was some kind of testing guide.)

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -343,6 +343,7 @@
     <string name="GroupShareProfileView_share_your_profile_name_and_photo_with_this_group">Share your profile name and photo with this group?</string>
     <string name="GroupShareProfileView_do_you_want_to_make_your_profile_name_and_photo_visible_to_all_current_and_future_members_of_this_group">Do you want to make your profile name and photo visible to all current and future members of this group?</string>
     <string name="GroupShareProfileView_make_visible">Make visible</string>
+    <string name="GroupShareProfileView_never_ask_again">Never ask again</string>
 
     <!-- GroupMembersDialog -->
     <string name="GroupMembersDialog_me">Me</string>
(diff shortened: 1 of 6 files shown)


cascheberg:
@interfect You changed the database scheme, so you need to alter the table during the upgrade process of existing installations. There have been scheme changes before, so you should be able to figure out how to do that from the existing code.

Apart from that, I think that there is a better solution. I suggest to allow to directly hide the banner without showing the alert dialog and add a preference to conversation settings to allow to manually toggle (and override the default) profile sharing setting. Such an approach would also fix #6964, and it would allow to hide ones profile from users saved in contacts.

interfect:
I'll look around for the database upgrade code and try to duplicate it.

As for the other stuff, I agree that would probably be a more complete
solution, but I think it might be a bit beyond my current skill level as an
Android developer. I can give it a shot, but someone else would probably be
able to do it a lot faster than I could.

PR #7391, opened 2018-02-03
Call onPageSelected(0) during media preview init

### LINKED ISSUE #7385: Wrong MediaPreview title when entering via media item at position 0
I have:

### Bug description
[`initializeActionBar`](https://github.com/signalapp/Signal-Android/blob/master/src/org/thoughtcrime/securesms/MediaPreviewActivity.java#L127) is generally only called in [`onPageSelected(int position)`](https://github.com/signalapp/Signal-Android/blob/master/src/org/thoughtcrime/securesms/MediaPreviewActivity.java#L329). But this listener method won't be called if the first position of the ViewPager is 0.

### Steps to reproduce 1
- open a conversation
- go to "All media"
- tap the media item at the very top

### Steps to reproduce 2
- open a conversation
- scroll to the oldest media item (at the top)
- tap this media item

### Steps to reproduce 3
- open a conversation
- attach an image or video (don't send)
- tap this media item

**Actual result:** "Media Preview" (`@string/AndroidManifest__media_preview`) is displayed in the action bar.
**Expected result:** Sender name and date/"Draft" are displayed in the action bar. These information will be displayed after swiping to another media item and then returning to the initial one.

### Screenshots
directly after entering MediaPreview
screenshot_20180203-124610
after swiping sideways and back
screenshot_20180203-124617

### Device info

**Device:** Fairphone 2
**Android version:** 6.0.1
**Signal version:** 4.15.5

FeuRenard:
### Contributor checklist

 * AVD Nexus 5X, Android 7.1.1

----------

### Description
Fixes #7385

The changes proposed in this PR make sure `onPageSelected(0)`/`initializeActionBar()` is called when `MediaPreviewActivity` is entered at `ViewPager` position 0. To call the former method I need access to the `OnPageChangeListener` of the `ViewPager`. The `ViewPager` class doesn't provide a method to access its `OnPageChangeListener`(s). So I decided to keep this listener in a local variable for a short time and so I am able to call `onPageSelected(0)`. Note that it is sufficient to attach the listener before the adapter is set. This condition is still satified after this PR.

### Alternative considered approaches

We could also use a global variable to keep the listener. But I don't think this is necessary because it is only needed in the local scope you can see in the commit.

Alternatively extending `ViewPager` to be able to access the listeners at a later point is not easily possible because the listeners are kept in a private variable there. Also the diff would be much larger and not worth the effort imho.

### DIFF
--- a/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
+++ b/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
@@ -199,10 +199,8 @@ public class MediaPreviewActivity extends PassphraseRequiredActionBarActivity im
     if (conversationRecipient != null) {
       getSupportLoaderManager().restartLoader(0, null, this);
     } else {
-      ViewPager.OnPageChangeListener onPageChangeListener = new ViewPagerListener();
-      mediaPager.addOnPageChangeListener(onPageChangeListener);
       mediaPager.setAdapter(new SingleItemPagerAdapter(this, GlideApp.with(this), getWindow(), initialMediaUri, initialMediaType, initialMediaSize));
-      onPageChangeListener.onPageSelected(0);
+      initializeActionBar();
     }
   }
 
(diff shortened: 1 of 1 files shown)


FeuRenard:
Fixed by 42a8522

PR #5857, opened 2016-11-14
Hide copy button for messages without text

FeuRenard:
### Contributor checklist

- [x] I am following the [Code Style Guidelines](https://github.com/WhisperSystems/Signal-Android/wiki/Code-Style-Guidelines)
- [x] I have tested my contribution on these devices:
 * Sony Xperia U, Android 4.4.4 (CyanogenMod)
 * AVD Nexus 5X, Android 6.0
- [x] My contribution is fully baked and ready to be merged as is
- [x] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the `Fixes #1234` [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)
- [x] I have made the choice whether I want the [BitHub reward](https://github.com/WhisperSystems/Signal-Android/wiki/BitHub-Rewards) or not by omitting or adding the word `FREEBIE` in the commit message of my first commit

----------

### Description
This PR hides the copy button in a conversation's context menu as soon as a message without text (e.g. image only message) is marked.

Fixes #5802
// FREEBIE

### Screenshots
before
device-2016-11-14-203304
after
device-2016-11-14-202924

Issue #5802, opened 2016-10-23
Copy image-only message does nothing


FeuRenard:
I have:
- [x] searched open and closed issues for duplicates
- [x] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

---
### Bug description

If you copy an image (without text) in a conversation, nothing will actually be copied to clipboard. Except if you mark several images (without text), line breaks are copied to clipboard.
### Steps to reproduce
- Long-press an image only message in a conversation
- tap the copy icon
- long-press the input field
- if possible: tap "paste"

**Actual result:** No paste hint appears while long-pressing or another text is pasted that hung around in the clipboard.
**Expected result:** Copy button is hidden.

FeuRenard:
#7472 supersedes this issue.

PR #7410, opened 2018-02-14
Improve batch selection mode

art1fa:

### Contributor checklist

 * Samsung Galaxy S8, Android 7.0
 * Virtual Pixel 2, Android 8
 * Virtual Nexus 4, Android 4.4

----------

### Description
- First commit updates the batch selection mode to a more modern style and is now the same in Conversation and ConversationList, too.
- Second commit removes the 'copy' icon from the menu when there is any media message selected. Rationale of this change is, that copying media messages does not work (but forwarding does).
- Since there is enough space left on the toolbar: make the forward icon always visible. Hidden as the only option in the overfow menu, this action gets only little attention otherwise.

### Screenshots

screenrecord

#### Before - After:
anim

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -211,10 +211,9 @@ public class ConversationFragment extends Fragment
       {
         actionMessage = true;
         break;
-      }
-      else if (messageRecord.isMms()             &&
-              !messageRecord.isMmsNotification() &&
-              ((MediaMmsMessageRecord)messageRecord).containsMediaSlide())
+      } else if (messageRecord.isMms()              &&
+                 !messageRecord.isMmsNotification() &&
+                 ((MediaMmsMessageRecord)messageRecord).containsMediaSlide())
       {
         mediaMessage = true;
         break;
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -201,6 +201,7 @@ public class ConversationFragment extends Fragment
   private void setCorrectMenuVisibility(Menu menu) {
     Set<MessageRecord> messageRecords = getListAdapter().getSelectedItems();
     boolean            actionMessage  = false;
+    boolean            mediaMessage  = false;
 
     for (MessageRecord messageRecord : messageRecords) {
       if (messageRecord.isGroupAction() || messageRecord.isCallLog() ||
@@ -211,6 +212,13 @@ public class ConversationFragment extends Fragment
         actionMessage = true;
         break;
       }
+      else if (messageRecord.isMms()             &&
+              !messageRecord.isMmsNotification() &&
+              ((MediaMmsMessageRecord)messageRecord).containsMediaSlide())
+      {
+        mediaMessage = true;
+        break;
+      }
     }
 
     if (messageRecords.size() > 1) {
@@ -218,7 +226,7 @@ public class ConversationFragment extends Fragment
       menu.findItem(R.id.menu_context_details).setVisible(false);
       menu.findItem(R.id.menu_context_save_attachment).setVisible(false);
       menu.findItem(R.id.menu_context_resend).setVisible(false);
-      menu.findItem(R.id.menu_context_copy).setVisible(!actionMessage);
+      menu.findItem(R.id.menu_context_copy).setVisible(!actionMessage  && !mediaMessage);
     } else {
       MessageRecord messageRecord = messageRecords.iterator().next();
 
@@ -230,7 +238,7 @@ public class ConversationFragment extends Fragment
 
       menu.findItem(R.id.menu_context_forward).setVisible(!actionMessage);
       menu.findItem(R.id.menu_context_details).setVisible(!actionMessage);
-      menu.findItem(R.id.menu_context_copy).setVisible(!actionMessage);
+      menu.findItem(R.id.menu_context_copy).setVisible(!actionMessage && !mediaMessage);
     }
   }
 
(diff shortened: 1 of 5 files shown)


FeuRenard:
Thank you for your contribution! (And sorry for my nagging in advance...)

With your commit "Don't show 'copy' icon for media messages in selection mode" you address #5802. I attempted much the same with #5857. It is slightly different as it hides the copy button only when there is no text in one of the selected messages. I did this because the description (displayed on long-press) of the copy button says "Copy text". This is possible for messages including media and also text. The implementation in this PR would not allow copying text from a media message.

May I ask you to drop this commit in favor of #5857? It can be done by an interactive rebase and a forced push without breaking the rest of this PR.

art1fa:
Sure! Would you be so kind to send me an e-mail how to do it without breaking the PR?

FeuRenard:
In your local terminal/shell:
* `git checkout better_selection_mode`
* `git rebase -i 7acf1c5^` (This means "Start interactive rebase from \<commit> to HEAD (including \<commit>)". Further reading on this tool: git's [documentation on this](https://git-scm.com/docs/git-rebase#git-rebase---interactive) or Pro Git's section on [Rewriting History](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History))
* a text editor should open in the terminal. For me it's the program "Vi". There should be a list of the commits to edit at the top starting with a keyword which tells git what to do with the commit in the rebase process. "Pick" is the default and just keeps the commit as is. To remove/"drop" a commit, you can replace "pick" with "drop". To do this in "Vi" press "i" for entering the input mode. After editing exit the editor. I usually do this by first pressing the escape key and then entering the command ":x".
* you can check whether the commit was actually dropped with `git log`. Only proceed if the commit history matches your expectations.
* `git push -f origin better_selection_mode` to perform a forced push. You have to force this push with the flag "-f" because you are not just adding commits to the branch but also removing some of the existing ones.
* This PR should automatically be updated to only show the updated commit history.

You should only do this whole process I described on your own branches in case nobody branched them!

art1fa:
I think the action induced by the copy icon expects the user to copy the whole selected content, not only part of it (the text). Isn't it confusing when Signal does it differently than users learned?

I would say let's wait for more feedback until I eventually remove the commit.

And thanks for your instructions @FeuRenard ! But are you sure I don't have to reopen a new PR? The last time I did a force push to a forked repo, the PR didn't get synchronized as far as I could remember.

FeuRenard:
> But are you sure I don't have to reopen a new PR? The last time I did a force push to a forked repo, the PR didn't get synchronized as far as I could remember.

It usually worked for me in the past when the PR was open. You shouldn't do this while the PR is closed. That would break it.

Apart from this, [`git revert`](https://git-scm.com/docs/git-revert) is a much easier option and doesn't require a forced push.

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -201,11 +201,7 @@ private void initializeListAdapter() {
   private void setCorrectMenuVisibility(Menu menu) {
     Set<MessageRecord> messageRecords = getListAdapter().getSelectedItems();
     boolean            actionMessage  = false;
-
-    if (actionMode != null && messageRecords.size() == 0) {
-      actionMode.finish();
-      return;
-    }
```
moxie0: Why remove this?

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -216,14 +212,21 @@ private void setCorrectMenuVisibility(Menu menu) {
         actionMessage = true;
         break;
       }
+      else if (messageRecord.isMms()             &&
```
moxie0: Please move one line up " } else { "

src/org/thoughtcrime/securesms/ConversationFragment.java
art1fa: `ActionMode` will be finished in `onItemClick`, if the amout of selected records is `0`. `setCorrectMenuVisibility` is only called, if more than `0` records are selected.

src/org/thoughtcrime/securesms/ConversationFragment.java
art1fa: Done :+1:

src/org/thoughtcrime/securesms/ConversationFragment.java
moxie0: Is it necessary to move this? From the perspective of a reviewer, moving this changes the function promise from being idempotent to non-idempotent. I have to ask myself whether there are other callers to this method, if anyone else expects it to be idempotent, etc. Hard to answer those questions staring at a diff on a website, much easier to review if I don't have to ask them.

art1fa: The behavior in `onItemClick` is basically the same like in `ConversationListFragment`.
But I see your point and will bring those four lines back.

art1fa:
@moxie Updated the PR considering your code review requests. Comment if you want me to squash all commits into one :)

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -578,7 +585,13 @@ public void onItemClick(MessageRecord messageRecord) {
         ((ConversationAdapter) list.getAdapter()).toggleSelection(messageRecord);
         list.getAdapter().notifyDataSetChanged();
 
-        setCorrectMenuVisibility(actionMode.getMenu());
+        if (getListAdapter().getSelectedItems().size() == 0) {
+          actionMode.finish();
+        } else {
+          setCorrectMenuVisibility(actionMode.getMenu());
+          actionMode.setTitle(String.valueOf(getListAdapter().getSelectedItems().size()));
+        }
```
moxie0: Does this still need to be here?

art1fa: Mh we need to set the number of currently selected items in the title....and preferably only when those are greater than 0. It's basically the same like it's already there in `ConversationListFragment`, just including the call to `setCorrectMenuVisibility`

moxie0:
Thanks! Extra credit is animating the back arrow <-- into the X  =)

PR #7437, opened 2018-02-23
Upload Signal to F-Droid Android FOSS catalogue?

yvan-sraka:
https://f-droid.org/


sigenc:
you can close it. wan't happen. moxie said it at least 100 times

PR #7464, opened 2018-03-06
Shows contact number when viewing all group members. Fixes #7463

### LINKED ISSUE #7463: Group Chat: Impossible to find contact's number
I have:
- [ x] searched open and closed issues for duplicates
- [x ] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

### Bug description
Signal offers no way to see contact's number in group chats

### Steps to reproduce
Open any group chat in Signal and try to figure out a way to see the actual phone number of a contact.

**Actual result:** You can't
**Expected result:** You can. This is a really annoying problem in my use case because many of our employees have two cell phones: 1 work phone and 1 personal. I discovered this issue when an employee lost his work cell and started using his personal cell for everything. Every time someone started a work group chat it was impossible to tell whether they were using the right number for that employee.

### Device info

**Device:** Google Pixel XL
**Android version:** 7.1.2
**Signal version:** 4.15.5

ImDevinC:

### Contributor checklist

 * Android Emulator, Android 8.0 Oreo

----------

### Description

This issue was not tagged as "Needs Help", but it's something that also annoyed me so I figured I'd take a crack at implementing a simple solution (also my first submission to this repo, so figured I'd start small).  

I tested a few basic users, also tested to make sure that a user with more than one number used the same number that was added to the group. I also tested with a user that has a really long name, and it doesn't look the *best*, but I see it looks weird in a few other scenarios so it may be ok here.  
Thanks for taking a look!

**Sample Screenshots** *Note: *me* is not the number of my device, just a poorly chosen username*
screen shot 2018-03-05 at 9 43 33 pm
screen shot 2018-03-05 at 9 44 24 pm

### DIFF
--- a/src/org/thoughtcrime/securesms/GroupMembersDialog.java
+++ b/src/org/thoughtcrime/securesms/GroupMembersDialog.java
@@ -118,6 +118,11 @@ public class GroupMembersDialog extends AsyncTask<Void, Void, List<Recipient>> {
             name += " ~" + recipient.getProfileName();
           }
 
+          if (recipient.getAddress().isPhone() && !TextUtils.isEmpty(recipient.getAddress()
+                  .toPhoneString())) {
+            name += " " + recipient.getAddress().toPhoneString();
+          }
+
           recipientStrings.add(name);
         }
       }


moxie0:
I'm not sure I understand the problem that this is trying to solve. Why not just tap on the contact's name to see their number?

ImDevinC:
@moxie0 because when you do that, it shows the whole contact, not necessarily the number they're using in Signal. In the linked issue report, it was stated that the user had a colleague with two phone numbers but lost one phone. When trying to reply to a group chat, it was impossible to make sure they were using the new phone number that still belonged to the colleague.

moxie0:
I see, maybe we should have that tap jump to the recipient details screen in signal instead, then?

ImDevinC:
The recipient details screen doesn't seem to show the recipients number either, at least not that I can see. I only see name, shared media, chat settings and privacy settings (which has "View Safety Number" and "Block"). We could put this data on the recipient screen instead and then have the tap go there, but two clicks to get there seems like a lot when we could show it on the popup.

moxie0:
I'm willing to entertain either, but if it's on the dialog I think it has to be formatted differently. That's not what most people are looking at that dialog for, so I think it's way too prominent this way.

PR #7515, opened 2018-03-13
Be less conservative concerning the next reminder interval progression

FeuRenard:
### Contributor checklist

 * AVD Nexus 5X, Android 7.1.1

----------

### Description
I think you share the view that we should bother the user with the registration lock PIN reminder only as often as necessary. Then, consider a Signal user who doesn't use Signal very frequently. Say once a day. With the current implementation of the reminder the user has to enter his/her PIN three days in a row, although he/she remembered the PIN on the first try one day after activation of the feature.

I propose to consider the time since the last registration lock PIN reminder for setting the next reminder interval instead of the last interval set. This allows to skip some intervals if the user proves to remember the PIN for a time interval longer than the last interval set.

### DIFF
--- a/src/org/thoughtcrime/securesms/lock/RegistrationLockReminders.java
+++ b/src/org/thoughtcrime/securesms/lock/RegistrationLockReminders.java
@@ -6,15 +6,16 @@ import android.support.annotation.NonNull;
 
 import org.thoughtcrime.securesms.util.TextSecurePreferences;
 
-import java.util.HashMap;
-import java.util.Map;
+import java.util.Map.Entry;
+import java.util.NavigableMap;
+import java.util.TreeMap;
 import java.util.concurrent.TimeUnit;
 
 public class RegistrationLockReminders {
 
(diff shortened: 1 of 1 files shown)


FeuRenard:
41d06fc8c2e374ce1e6a0eec680d2a38765ed544 broke this, but maybe you like the idea?

moxie0:
Oh I think I get what you were going for. Is this not already the case though? The next reminder is scheduled for current time + interval, not last time scheduled + interval, right?

FeuRenard:
Sorry, then I was not really clear explaining this. current time + interval is the case and I wanted to change the selection of the interval. Currently you conservatively just take the next interval. But imagine the interval was 6 hours, but I entered the PIN after 1 day.

Current behaviour: set the next interval to 12 hours, the next interval after 6 hours (although I already showed I still remember the PIN after 1 day)
Desired behaviour: set the next interval to 3 days, the next interval after 1 day).

moxie0:
I see. Yeah I think I'd be fine with that if it's a simple change

PR #7516, opened 2018-03-13
Reduce image stutter by using subsampling more often.

### LINKED ISSUE #7208: Stuttering while zooming and scrolling large images
I have:

(Lazy copy pasting this from https://whispersystems.discoursehosting.net/t/stuttering-while-zooming-large-images/1591/1)

### Bug description

I'm experiencing stuttering when pinch-zooming large pictures in Signal.

Some examples when I zoom an image of size:

* 4032 x 3024 2.6 MB: stutters a lot
* 3840 x 2160 1.8 MB: stutters
* 2752 x 1548 0.9 MB: no stuttering, smooth zoom

I can't recall having this issue before. Maybe started experiencing this a few weeks ago? Currently running 4.13.1 and now 4.13.2
There's no stuttering in the Photos app with the same images.

It was speculated [here](https://whispersystems.discoursehosting.net/t/stuttering-while-zooming-large-images/1591/3) that the recent Glide update could have something to do with this

### Steps to reproduce
- Open a large image
- Try to zoom

**Actual result:** Zooming is slow and stutters
**Expected result:** Zooming should be butter smooth

### Device info

**Device:** Moto G4 Play
**Android version:** 6.0.1
**Signal version:** 4.13.1, 4.13.2

### Link to debug log
https://gist.github.com/952b9bdad590e3c1e7df9f5d93f20c57

I zoomed and scrolled a 4032 x 3024 (2.6 MB) image and then captured the log

greyson-signal:
Fixes #7208.

On some phones, zooming in on large images (say, a 4k image) would be slow and stutter a lot. This is due to ZoomingImageView using a regular PhotoView instead of a SubsamplingScaleImageView. ZoomingImageView decides whether or not to use a SubsamplingScaleImageView depending on whether the image has a dimension larger than BitmapUtil.getMaxTextureSize(). 

However, while that method will determine the maximum size of an image that is _possible_ to render, it doesn't necessarily mean it can be rendered _efficiently_. There is some size at which some phones will have to continually re-upload a bitmap to the GPU, as seen in this systrace taken from a Moto E.

moto-e-zoom-stutter

We could just use SubsamplingScaleImageView all the time (it has built-in logic to prevent tiling on smaller images), but given that it doesn't use Glide to load Bitmaps, we shouldn't be over-using it. As a result, I altered BitmapUtil.getMaxTextureSize() to have a cap of 2048, which from research and experimentation seems to be a good number (OpenGL operates in powers of 2 -- if 4096 is too big, 2048 is the next best bet).

Tested on the following devices. No scroll jank on images >= 2048x2048. No scroll jank on images < 2048x2048.

Example test image to induce jank:
signal-2018-03-13-104700

### DIFF
--- a/src/org/thoughtcrime/securesms/util/BitmapUtil.java
+++ b/src/org/thoughtcrime/securesms/util/BitmapUtil.java
@@ -277,7 +277,7 @@ public class BitmapUtil {
   }
 
   public static int getMaxTextureSize() {
-    final int IMAGE_MAX_BITMAP_DIMENSION = 2048;
+    final int MAX_ALLOWED_TEXTURE_SIZE = 2048;
 
     EGL10 egl = (EGL10) EGLContext.getEGL();
     EGLDisplay display = egl.eglGetDisplay(EGL10.EGL_DEFAULT_DISPLAY);
@@ -303,6 +303,6 @@ public class BitmapUtil {
 
     egl.eglTerminate(display);
 
-    return Math.max(maximumTextureSize, IMAGE_MAX_BITMAP_DIMENSION);
+    return Math.min(maximumTextureSize, MAX_ALLOWED_TEXTURE_SIZE);
   }
 }


2-4601:
Tested f544b7d7b416015042268086f29ff198cb1bff48, can confirm it works on my device. Great fix.

Issue #4242, opened 2015-10-12
Allow removing attachments


johannesjh:
Please add the possibility of deleting attachments from received and sent messages.

This functionality would benefit usage on devices with limited storage capacity. E.g., my personal situation is that the many pictures and videos I receive are filling up my precious storage space. I would like to delete the pictures but keep the message texts. (Also compare #4065 for a related request to allow storing messages on the external SD card).

jeremymasters:
I just long press the message with the attachment and then delete it.

merkste:
I think the request was to remove the attachment only while preserving the text.

johannesjh:
yes, exactly. to remove the attachment only while preserving the text.

langboost:
Other similar but slightly different use cases would call for the ability to not download the image by default. This is particularly nice when you have a group of people who feel the need to send an inspiring picture of a sunrise every morning with some sort of inspiring quote embedded into the image. Every day. Sometimes multiple times per day. :) I suppose I could always find new friends...

mcginty:
@langboost You can already do that. Settings -> Chats and Media -> Media auto-download.

jeremymasters:
I think he might be talking about a per-user basis.  Maybe that could work like the ringtones.  All users follow the default settings unless you change that particular user's settings.  Like "Never allow auto-downloads of anything from a specified set of users.  I guess that would be more like "blocked users" but "blocked auto-download users".

2-4601:
![could not resist](https://cloud.githubusercontent.com/assets/174176/10469533/a87cd4bc-720e-11e5-922a-aa8b12ae7252.jpg)

johannesjh:
To bring the conversation back to a serious level: This is what the functionality for removing attachments could look like in the UI.

![mockup](https://cloud.githubusercontent.com/assets/581188/10476262/5b94f114-724c-11e5-80c3-af8046d56833.png)

bhabba:
1. I think it would be very handy to also add a global option for auto-deletion of attachments after exceeding a certain size limit: Under _Settings_ -> _Chats and media_ -> _Message trimming_ there should be an option for <ins>_Delete old media_</ins> where you can set a maximum storage space limit in MiB. Ideally the current free storage space of the partition where Signal's attachments are stored should be shown besides as a helpful orientation.

  When exceeding the set limit the oldest media attachments should get deleted until the total media size has fallen below the limit again.

  Of course this option should be disabled by default.

[...]

Issue #2798, opened 2015-03-26
[Feature request] Add ability to delete image in image view


BT80:
Currently the only way to delete an image is by long-tapping the image and selecting the delete icon.

If you short-tap the image it is displayed in full screen mode. There is already a save icon in that view. My suggestion is to also add a delete icon in that view.

moxie0:
Thanks, but I don't think we're going to do this.

BT80:
That's sad! I thought it would be of benefit. My use case was as follows:
When selecting "All images" in the menu there is no posibility to delete an image from the past. The only way to delete it would be by scrolling up the chat history until you reach that specific image, long-tap it and then select the delete icon. Scrolling up on large chats can be very tedious.

2-4601:
I can imagine @BT80 use case becoming more and more common as the user base grows. Deleting one image out of hundreds is currently not very user friendly.
Please consider reopening this.

mcginty:
I think this could be useful, especially now that we have a message details page where that action could also exist. Reopening with the caveat that we will likely want to add the "caption" text underneath the image so they know they're deleting the whole message and not just the media.

Don42:
I think it would also be useful and in line with android ux guidelines, to be able to longpress select images in the "all images"-View. This would allow it to save or delete multiple images at once. This would make the view behave more like the native gallery.

2-4601:
Deleting multiple images should then have a dialog which tells explicitly that you are also deleting associated messages at the same time. As mcginty noted above deleting images in TS is not as trivial as it may first seem.

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #7522, opened 2018-03-14
Allow deletion of media through preview.

greyson-signal:
Added the ability to delete media via the media preview. A delete option is added to the action bar of the MediaPreviewActivity. When pressed:

* The attachment entry representing the media is deleted.
* If there are no other attachments for the linked MMS message, then the MMS entry is also deleted.
* The preview is closed.

[Video](https://github.com/signalapp/Signal-Android/files/1809293/delete-media-preview.zip)

**TODO**
- [ ] Verify files are physically removed from the disk.

**Test Cases**
* Deleted static image
* Deleted GIF
* Deleted video
* Ensured that messages are gone from the conversation view.
* Ensured that media is gone from the "all media" screen.
* Ensured that the delete option is not present for previews of unsent media.
* Simulated the DeleteMediaTask taking a long time, so the preview closes before the deletion is finished. The conversation is updated when the message is deleted.
* Ensured that if you open the preview from the "all media" screen and then delete it, the "all media" screen is updated to reflect that.

**Unable to Test**
* Deleting one attachment from a message with multiple attachments. While this is technically possible, there's no way to create this scenario in the app today.

**Test Devices**

**Follow-Up Tasks**
* Add a confirmation dialogue for deleting.
* Allow multi-select and delete from the "all media" screen.

### DIFF
--- a/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
+++ b/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
@@ -20,6 +20,7 @@ import android.Manifest;
 import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
 import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.database.Cursor;
 import android.net.Uri;
@@ -34,6 +35,7 @@ import android.support.v4.content.Loader;
 import android.support.v4.util.Pair;
 import android.support.v4.view.PagerAdapter;
 import android.support.v4.view.ViewPager;
+import android.support.v7.app.AlertDialog;
 import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.Menu;
@@ -46,9 +48,13 @@ import android.view.WindowManager;
 import android.widget.FrameLayout;
 import android.widget.Toast;
 
+import org.thoughtcrime.securesms.attachments.AttachmentId;
+import org.thoughtcrime.securesms.attachments.DatabaseAttachment;
 import org.thoughtcrime.securesms.components.MediaView;
 import org.thoughtcrime.securesms.components.viewpager.ExtendedOnPageChangedListener;
 import org.thoughtcrime.securesms.database.Address;
+import org.thoughtcrime.securesms.database.AttachmentDatabase;
+import org.thoughtcrime.securesms.database.DatabaseFactory;
 import org.thoughtcrime.securesms.database.MediaDatabase.MediaRecord;
 import org.thoughtcrime.securesms.database.loaders.PagingMediaLoader;
 import org.thoughtcrime.securesms.mms.GlideApp;
@@ -251,6 +257,48 @@ public class MediaPreviewActivity extends PassphraseRequiredActionBarActivity im
     }
   }
 
+  @SuppressLint("StaticFieldLeak")
+  private void deleteMedia() {
+    MediaItem mediaItem = getCurrentMediaItem();
+    if (mediaItem == null || mediaItem.attachment == null) {
+      return;
+    }
+
+    AlertDialog.Builder builder = new AlertDialog.Builder(this);
+    builder.setIconAttribute(R.attr.dialog_alert_icon);
+    builder.setTitle(R.string.MediaPreviewActivity_media_delete_confirmation_title);
+    builder.setMessage(R.string.MediaPreviewActivity_media_delete_confirmation_message);
+    builder.setCancelable(true);
+
+    builder.setPositiveButton(R.string.delete, (dialogInterface, which) -> {
+      new AsyncTask<Void, Void, Void>() {
+        @Override
+        protected Void doInBackground(Void... voids) {
+          if (mediaItem.attachment == null) {
+            return null;
+          }
+          Context      context         = MediaPreviewActivity.this.getApplicationContext();
+          AttachmentId attachmentId    = mediaItem.attachment.getAttachmentId();
+          long         mmsId           = mediaItem.attachment.getMmsId();
+          int          attachmentCount = DatabaseFactory.getAttachmentDatabase(context)
+                                                     .getAttachmentsForMessage(mmsId)
+                                                     .size();
+
+          if (attachmentCount <= 1) {
+            DatabaseFactory.getMmsDatabase(context).delete(mmsId);
+          } else {
+            DatabaseFactory.getAttachmentDatabase(context).deleteAttachment(attachmentId);
+          }
+          return null;
+        }
+      }.execute();
+
+      finish();
+    });
+    builder.setNegativeButton(android.R.string.cancel, null);
+    builder.show();
+  }
+
   @Override
   public boolean onPrepareOptionsMenu(Menu menu) {
     super.onPrepareOptionsMenu(menu);
@@ -258,7 +306,11 @@ public class MediaPreviewActivity extends PassphraseRequiredActionBarActivity im
     menu.clear();
     MenuInflater inflater = this.getMenuInflater();
     inflater.inflate(R.menu.media_preview, menu);
-    if (conversationRecipient == null) menu.findItem(R.id.media_preview__overview).setVisible(false);
+
+    if (!isMediaInDb()) {
+      menu.findItem(R.id.media_preview__overview).setVisible(false);
+      menu.findItem(R.id.delete).setVisible(false);
+    }
 
     return true;
   }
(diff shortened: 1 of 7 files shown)


src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
```diff
@@ -253,6 +245,88 @@ void deleteAttachmentsForMessage(long mmsId) {
     database.delete(TABLE_NAME, MMS_ID + " = ?", new String[] {mmsId + ""});
   }
 
+  /**
+   * Deletes the provided attachment. If this was the only attachment for its linked MMS message,
+   * then the MMS message will also be deleted.
+   */
+  public void deleteAttachmentAndPossiblyMmsMessage(AttachmentId id) {
```
moxie0: I don't love the idea of operations on the attachment database having side effects against the message database. I think the hierarchy is reversed, where one expects operations against the message database to modify the attachment database, but not really the inverse.

It seems like ideally there would be a single new method here that allows you to delete by AttachmentId, and then the calling code would determine whether it needs to call delete() on the MmsDatabase or if it needs to call delete() on the AttachmentDatabase.

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -251,14 +254,26 @@ private void saveToDisk() {
     }
   }
 
+  private void deleteMedia() {
```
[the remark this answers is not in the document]
greyson-signal: Yeah, I agree with this. I put it as a "follow-up task" in the PR description, but I can also do it in this PR.

src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
greyson-signal: Ok, that makes sense. I imagine the flow from the caller then would be something like:

```
// Get number of attachments for the message this media is attached to
// if (attachments == 1)
//   delete message
// else
//   delete attachment
```
My only concern here would be a race condition where an attachment is being added in another thread at the same time, causing us to delete the message when we shouldn't. Normally you'd get around these types of problems by having a dedicated single-thread executor for DB operations (or by using DB transactions like I am right now). Is that something we should look into?

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -251,14 +257,64 @@ private void saveToDisk() {
...
+            DatabaseFactory.getMmsDatabase(context).delete(mmsId);
+          } else {
+            DatabaseFactory.getAttachmentDatabase(context).deleteAttachment(attachmentId);
+          }
+          return null;
+        }
+
+        @Override
+        protected void onPostExecute(Void aVoid) {
+          finish();
```
moxie0: Did you want to wait for the asynchronous action to complete before finishing?

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -251,14 +257,64 @@ private void saveToDisk() {
...
+        protected Void doInBackground(Void... voids) {
+          if (mediaItem.attachment == null) {
+            return null;
+          }
+          Context context = MediaPreviewActivity.this.getApplicationContext();
+
+          long mmsId = mediaItem.attachment.getMmsId();
+          AttachmentId attachmentId = mediaItem.attachment.getAttachmentId();
+          int attachmentCount = DatabaseFactory.getAttachmentDatabase(context)
+              .getAttachmentsForMessage(mmsId).size();
```
moxie0: Please align on '=', and '.getAttachmentsForMessage' should be aligned with the '.' of 'getAttachmentDatabase' =)

src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
```diff
@@ -225,7 +225,7 @@ public void setTransferProgressFailed(AttachmentId attachmentId, long mmsId)
   }
 
   @SuppressWarnings("ResultOfMethodCallIgnored")
-  void deleteAttachmentsForMessage(long mmsId) {
+  public void deleteAttachmentsForMessage(long mmsId) {
```
moxie0: Is there a reason this has to be public now?

src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
```diff
@@ -234,23 +234,57 @@ void deleteAttachmentsForMessage(long mmsId) {
...
     }
 
     database.delete(TABLE_NAME, MMS_ID + " = ?", new String[] {mmsId + ""});
+    notifyAttachmentListeners();
+  }
+
+  public void deleteAttachment(@NonNull AttachmentId id) {
+    SQLiteDatabase database = databaseHelper.getWritableDatabase();
+
+    AttachmentFileInfo attachmentFileInfo = getAttachmentFileInfo(database, id);
```
moxie0: I think the target method is small enough that you can inline getAttachmentFileInfo, so you don't have to create a holder class.

src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
```diff
@@ -234,23 +234,57 @@ void deleteAttachmentsForMessage(long mmsId) {
...
+    }
+
+    database.delete(TABLE_NAME, PART_ID_WHERE, id.toStrings());
+    deleteAttachmentOnDisk(attachmentFileInfo.data, attachmentFileInfo.thumbnail);
+    notifyAttachmentListeners();
+  }
+
+
+  @Nullable
+  private AttachmentFileInfo getAttachmentFileInfo(
```
moxie0: first argument goes on the same line

src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
```diff
@@ -234,23 +234,57 @@ void deleteAttachmentsForMessage(long mmsId) {
...
+      @NonNull SQLiteDatabase database,
+      @NonNull AttachmentId id) {
```
moxie0: brace on new line

src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
```diff
@@ -234,23 +234,57 @@ void deleteAttachmentsForMessage(long mmsId) {
...
+    Cursor cursor = null;
+    AttachmentFileInfo info = null;
```
moxie0: align

src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
```diff
@@ -234,23 +234,57 @@ void deleteAttachmentsForMessage(long mmsId) {
...
+
+    try {
+      cursor = database.query(TABLE_NAME,
```
moxie0: I typically like to put the cursor assignment and construction outside the try block. But also, we should be able to do try-with-resources now.

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
greyson-signal: I probably spent way too much time going back and forth on this already haha. I can finish() immediately. I think my main concern was that you'd usually see the message for a split moment before it disappeared if you finish() first, whereas waiting for the operation to complete hides this. However, on the flip side, it means that the UI may feel slightly sluggish at that moment.

Maybe the real solution is to finish() immediately, then improve the conversation recyclerview transition animations in the future to be a little more motion-y :)

Either way, I can switch it to finish() immediately for now.

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
greyson-signal: Haha sorry, I'll get used to this code style in due time, just might take me a few PR's :)

src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
greyson-signal: Ah nope, good catch, artifact of an earlier scrapped change.

greyson-signal:
Squashing and merging into 4.18.0.

FeuRenard:
Fixes #2798
Fixes #4242

RiseT:
Question: All the strings contained here and in the [batch deleting commit](https://github.com/signalapp/Signal-Android/pull/7527) speak of "messages" that are being deleted. What is deleted here? Really the messages themselves or the attached media? If it is the latter, the strings should get changed accordingly.

Even when it's currently the whole message that get's deleted: Since you're planning to allow messages with multiple attachments/media in the (near?) future, "media" (or more generally "attachments") would be more future-proof in that regard.

greyson-signal:
@RiseT The messages themselves are deleted. If you sent text along with your media attachment, that gets deleted as well. So we just wanted to be clear about that. If/when we get to multiple attachments per message, we'll have to revisit this flow to make other changes, at which point we'll update the strings as appropriate :)

PR #7542, opened 2018-03-19
Strip EXIF metadata from JPEG attachments.

### LINKED ISSUE #5968: Image Meta Data Removal
This feature request is for an option to remove meta data from a picture being shared in a conversation to prevent location, device and other key data being shared.

greyson-signal:
This PR strips EXIF metadata from all JPEG attachments before sending. Previously, it was being stripped from most images by our re-scaling process. However, smaller images would slip through and keep their metadata, possibly exposing things like photo location. 

This PR removes all EXIF metadata by re-encoding the JPEG. This keeps all of the visual effects of the EXIF tags (like orientation) by updating the image data itself, while removing all EXIF tags.

Fixes #5968

**Test Cases**
* Sending a small JPEG (that doesn't get resized) with lots of metadata, including orientation
* Sending a small JPEG (that doesn't get resized) with lots of metadata, excluding orientation
* Sending a large JPEG (that does get resized)
* Sending a JPEG with no metadata
* Sending a GIF
* Sending a PNG

**Metadata Verification Process**
* Used the [Photo Exif Editor](https://play.google.com/store/apps/details?id=net.xnano.android.photoexifeditor&hl=en) app to view image metadata.
* Verified that if the image was rendering in the proper orientation (in both the thumbnail and the preview).

**Test Image**
cat-rotated
(Hilariously, the orientation field is not respected in my browser in the thumbnail, but it _is_ respected if you click the image and view it fullscreen).

**Things I Still Need to Test**
- [x] MMS messages
- [x] Group messages

**Test Devices**

### DIFF
--- a/src/org/thoughtcrime/securesms/jobs/SendJob.java
+++ b/src/org/thoughtcrime/securesms/jobs/SendJob.java
@@ -13,6 +13,7 @@ import org.thoughtcrime.securesms.mms.MediaConstraints;
 import org.thoughtcrime.securesms.mms.MediaStream;
 import org.thoughtcrime.securesms.mms.MmsException;
 import org.thoughtcrime.securesms.transport.UndeliverableMessageException;
+import org.thoughtcrime.securesms.util.MediaUtil;
 import org.thoughtcrime.securesms.util.Util;
 import org.whispersystems.jobqueue.JobParameters;
 
@@ -50,8 +51,8 @@ public abstract class SendJob extends MasterSecretJob {
     }
   }
 
-  protected List<Attachment> scaleAttachments(@NonNull MediaConstraints constraints,
-                                              @NonNull List<Attachment> attachments)
+  protected List<Attachment> scaleAndStripExifFromAttachments(@NonNull MediaConstraints constraints,
+                                                              @NonNull List<Attachment> attachments)
       throws UndeliverableMessageException
   {
     AttachmentDatabase attachmentDatabase = DatabaseFactory.getAttachmentDatabase(context);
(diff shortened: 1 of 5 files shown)


src/org/thoughtcrime/securesms/jobs/SendJob.java
```diff
@@ -74,4 +91,62 @@ protected void markAttachmentsUploaded(long messageId, @NonNull List<Attachment>
...
+
+        TiffOutputSet         outputSet          = getWhitelistedFields((JpegImageMetadata) metadata);
+        ExifRewriter          rewriter           = new ExifRewriter();
+        ByteArrayOutputStream noExifOutputStream = new ByteArrayOutputStream();
+
+        attachmentStream.close();
+        attachmentStream = attachmentDatabase.getAttachmentStream(attachmentId, 0);
+        rewriter.updateExifMetadataLossy(attachmentStream, noExifOutputStream, outputSet);
+
+        byte[]      noExifBytes       = noExifOutputStream.toByteArray();
```
greyson-signal: @moxie-signal This is a bit I feel a little iffy on, just because it involves a big memory allocation. The alternative is to get a raw OutputStream from the AttachmentDb, but we'd have to be careful that both the file and AttachmentDb row both get updated together, or not at all. I figured that since the image is always downscaled by this point it should be fine, but it's worth pointing out.

src/org/thoughtcrime/securesms/mms/MediaConstraints.java
```diff
@@ -81,4 +81,14 @@ public MediaStream getResizedMedia(@NonNull Context context, @NonNull Attachment
       throw new IOException(e);
     }
   }
+
+  public boolean canStripExif(@Nullable Attachment attachment) {
+    return attachment != null && MediaUtil.isJpeg(attachment);
+  }
```
moxie0: let's put this in the calling code

src/org/thoughtcrime/securesms/mms/MediaConstraints.java
```diff
@@ -81,4 +81,14 @@ public MediaStream getResizedMedia(@NonNull Context context, @NonNull Attachment
...
+
+  public MediaStream getExifStrippedMedia(@NonNull Context context, @NonNull Attachment attachment)
+      throws IOException
+  {
+    return getResizedMedia(context, attachment);
+  }
```
moxie0: let's have the calling code just do this directly, i'm just not sure about MediaConstraints knowing about exif

PR #7546, opened 2018-03-20
Fill in attachment sizes when sending media.

greyson-signal:
We need to populate the UriAttachment models with width and height so that when we build DatabaseAttachment models from them, the dimensions are saved to the database.

I did this by adding a width and height field to the UriAttachment constructor, and then making sure that got filled in during the sending flow. All images should have a width and height populated (including GIFs), but otherwise the fields are left empty (0, 0).

**Test Cases**
* Attached a JPEG
* Attached a PNG
* Attached a GIF
* Attached an audio file (read 0, 0)
* Attached a video (read 0, 0)
* Used logging to verify that the attachment models that were part of the conversation adapter had widths and heights set, setting us up to be able to use them for sizing the ThumbnailView appropriately.

**Test Devices**

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/Slide.java
+++ b/src/org/thoughtcrime/securesms/mms/Slide.java
@@ -22,6 +22,7 @@ import android.net.Uri;
 import android.support.annotation.DrawableRes;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.util.Pair;
 
 import org.thoughtcrime.securesms.attachments.Attachment;
 import org.thoughtcrime.securesms.attachments.UriAttachment;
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/mms/Slide.java
```diff
@@ -137,9 +138,19 @@ protected static Attachment constructAttachmentFromUri(@NonNull  Context context
                                                                    boolean  voiceNote)
   {
     try {
-      Optional<String> resolvedType    = Optional.fromNullable(MediaUtil.getMimeType(context, uri));
-      String           fastPreflightId = String.valueOf(SecureRandom.getInstance("SHA1PRNG").nextLong());
-      return new UriAttachment(uri, hasThumbnail ? uri : null, resolvedType.or(defaultMime), AttachmentDatabase.TRANSFER_PROGRESS_STARTED, size, fileName, fastPreflightId, voiceNote);
+      String                 resolvedType    = Optional.fromNullable(MediaUtil.getMimeType(context, uri)).or(defaultMime);
+      String                 fastPreflightId = String.valueOf(SecureRandom.getInstance("SHA1PRNG").nextLong());
+      Pair<Integer, Integer> dimens          = MediaUtil.getDimensions(context, resolvedType, uri);
```
moxie0: I'm not sure about doing these kinds of blocking operations inside here. I think this method should take width and height as parameters that are passed in from the calling code. In many cases that information should be more readily available to the calling code than having to load the entire URI.

greyson-signal: So right now, the code path looks something like:

(Activity gets uri for media in some fashion) ->
AttachmentManager#setMedia ->
AsyncTask#doInBackground ->
AsyncTask#getContentResolverSlideInfo() -> 
MediaType#createSlide() ->
ImageSlide#constructor ->
Slide.constructAttachmentForUri()

The next best place I could see us putting this is MediaType#createSlide(). We could read the dimensions there and pass them into the constructor for GifSlide and ImageSlide. Still, because at this point we only have a generic content Uri, we're going to have to read bytes from it to get the dimensions. Thankfully, the methods we use (except for GIFs, unfortunately) don't require reading in the whole stream, just the dimension-related parts.

Depending on how the uri is retrieved in the first part of the flow, we _may_ already have access to the dimensions without reading the bytes, but not always (for instance, picking from the system gallery does not provide us with dimensions). But it could be worth looking into though for the GIF-from-Giphy flow, for instance, which should have the width and height already without having to give it to glide to ingest.

greyson-signal:
Going to do this all in one big PR. See #7556

PR #7556, opened 2018-03-23
Render images in a conversation true-to-size.

greyson-signal:
Previously, we were always rendering images as squares. Instead of doing that, we now render them as close to true-to-size as possible (within reasonable min/max width/height boundaries).

[Demo Video](https://github.com/signalapp/Signal-Android/files/1839736/photo-sizing-demo.zip)

**Test Cases**
* 1000x50 image
* 50x1000 image
* 1x300 image
* 300x1 image
* Exif-rotated image
* GIFs
* JPEGs
* PNGs
* Tiny screens (320dp wide)
* Normal screens (360dp wide)
* Received images from clients that do not send width/height, verified we fallback properly to the "square" display method.

**Test Devices**


src/org/thoughtcrime/securesms/components/ThumbnailView.java
```diff
@@ -97,8 +212,17 @@ public void setBackgroundColorHint(int color) {
   }
 
   public void setImageResource(@NonNull GlideRequests glideRequests, @NonNull Slide slide,
-                               boolean showControls, boolean isPreview)
+                               boolean showControls, boolean isPreview) {
```
moxie0: brace on new line =)

src/org/thoughtcrime/securesms/components/ThumbnailView.java
```diff
@@ -97,8 +212,17 @@ public void setBackgroundColorHint(int color) {
...
+    setImageResource(glideRequests, slide, showControls, isPreview, 0, 0);
+  }
+
+  public void setImageResource(@NonNull GlideRequests glideRequests, @NonNull Slide slide,
+                               boolean showControls, boolean isPreview, int naturalWidth,
+                               int naturalHeight)
   {
+    dimens[WIDTH]  = naturalWidth;
+    dimens[HEIGHT] = naturalHeight;
```
moxie0: Out of curiosity, you don't have to request a redraw/remeasure here?

src/org/thoughtcrime/securesms/components/ThumbnailView.java
```diff
@@ -171,22 +295,28 @@ public void showProgressSpinner() {
...
+                        .diskCacheStrategy(DiskCacheStrategy.NONE), new FitCenter());
+  }
+
+  private GlideRequest applySizing(@NonNull GlideRequest request, @NonNull BitmapTransformation unavailableDimensSizing) {
+    Pair<Integer, Integer> targetDimens = getTargetDimensions(dimens, bounds);
+    if (targetDimens.first == 0 && targetDimens.second == 0) {
+      return request.transforms(unavailableDimensSizing, new RoundedCorners(radius));
+    }
+    return request.override(targetDimens.first, targetDimens.second)
+                  .transforms(new CenterCrop(), new RoundedCorners(radius));
```
moxie0: Is this necessary? For some reason I thought glide would do everything automatically based on the target view size?

src/org/thoughtcrime/securesms/mms/Slide.java
```diff
@@ -137,9 +138,19 @@ protected static Attachment constructAttachmentFromUri(@NonNull  Context context
                                                                    boolean  voiceNote)
   {
     try {
-      Optional<String> resolvedType    = Optional.fromNullable(MediaUtil.getMimeType(context, uri));
-      String           fastPreflightId = String.valueOf(SecureRandom.getInstance("SHA1PRNG").nextLong());
-      return new UriAttachment(uri, hasThumbnail ? uri : null, resolvedType.or(defaultMime), AttachmentDatabase.TRANSFER_PROGRESS_STARTED, size, fileName, fastPreflightId, voiceNote);
+      String                 resolvedType    = Optional.fromNullable(MediaUtil.getMimeType(context, uri)).or(defaultMime);
+      String                 fastPreflightId = String.valueOf(SecureRandom.getInstance("SHA1PRNG").nextLong());
+      Pair<Integer, Integer> dimens          = MediaUtil.getDimensions(context, resolvedType, uri);
```
moxie0: To revisit this agian, I still kind of want this to be passed in rather than having the model do all this i/o.

Off the top of my head, these URI constructions should either be coming from the media store, in which case perhaps we can query the content provider for that info like we do size, or some other source, in which case we can do the long thing of looking at the data itself?

src/org/thoughtcrime/securesms/components/ThumbnailView.java
greyson-signal: Oh you'd think :) There was an odd bug with Glide where sometimes it'd only render the middle sliver of an image. After trying a bunch of stuff (like forcing view invalidations at different points), this was the only thing that worked.

src/org/thoughtcrime/securesms/mms/Slide.java
greyson-signal: Talked offline, but I'm going to move the computation of the dimensions to be more directly linked to the existing background work already happening on the async task, and then also trickle down the dimensions for some of the attachment types from the pickers when they're available.

src/org/thoughtcrime/securesms/components/ThumbnailView.java
greyson-signal: I think in practice you don't notice because glide triggers typically triggers a near-immediate redraw, but I agree that I should add one to be safe.

greyson-signal:
@moxie-signal I've updated the PR to move the sizing to a more appropriate slot. I also trickled down the dimensions for GIF's from the Giphy picker. 

I could also look into trickling down the dimensions for media items from the "recent images" rail, but unfortunately the width/height columns in MediaStore are only for API 16+. Figured I could hold off on diving into that until we find performance to be a problem (getting dimensions from images should be relatively cheap). I haven't noticed any perf issues myself on my old S3 mini.

src/org/thoughtcrime/securesms/database/loaders/RecentPhotosLoader.java
```diff
@@ -8,6 +8,7 @@
 import android.provider.MediaStore;
 import android.support.v4.content.CursorLoader;
 
+import org.thoughtcrime.securesms.mms.MediaStream;
```
moxie0: Still need this?

src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -256,28 +260,35 @@ protected void onPostExecute(@Nullable final Slide slide) {
...
         try {
           cursor = context.getContentResolver().query(uri, null, null, null, null);
 
           if (cursor != null && cursor.moveToFirst()) {
-            String fileName = cursor.getString(cursor.getColumnIndexOrThrow(OpenableColumns.DISPLAY_NAME));
-            long   fileSize = cursor.getLong(cursor.getColumnIndexOrThrow(OpenableColumns.SIZE));
-            String mimeType = context.getContentResolver().getType(uri);
+            String                 fileName = cursor.getString(cursor.getColumnIndexOrThrow(OpenableColumns.DISPLAY_NAME));
+            long                   fileSize = cursor.getLong(cursor.getColumnIndexOrThrow(OpenableColumns.SIZE));
+            String                 mimeType = context.getContentResolver().getType(uri);
```
moxie0: what's going on here?

src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -256,28 +260,35 @@ protected void onPostExecute(@Nullable final Slide slide) {
...
+
+            if (width == 0 && height == 0) {
```
moxie0: Maybe this should be "||"? Seems like we should make another effort if either is true.

src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -302,8 +313,18 @@ protected void onPostExecute(@Nullable final Slide slide) {
           mediaSize = MediaUtil.getMediaSize(context, uri);
         }
 
+        if (mimeType == null) {
+          mimeType = MediaUtil.getMimeType(context, uri);
+        }
+
+        if (width == 0 && height == 0) {
```
moxie0: same, maybe "||"?

src/org/thoughtcrime/securesms/util/MediaUtil.java
```diff
@@ -100,6 +111,63 @@ public static long getMediaSize(Context context, Uri uri) throws IOException {
...
+      }
+    } else {
+      InputStream attachmentStream = null;
+      try {
+        if (MediaUtil.isJpegType(contentType)) {
+          attachmentStream = PartAuthority.getAttachmentStream(context, uri);
+          dimens = BitmapUtil.getExifDimensions(attachmentStream);
+        }
+        if (dimens == null) {
+          attachmentStream = PartAuthority.getAttachmentStream(context, uri);
```
moxie0: This is potentially reassigning attachmentStream, in which case the first created stream might not be closed

src/org/thoughtcrime/securesms/mms/AttachmentManager.java
greyson-signal: Ah shoot, I used to have a new variable here that shifted out the spacing, but I removed it and now just have to revert the spacing back :)

moxie0:
in 4.18.0

PR #7577, opened 2018-03-28
Show groups when searching for contacts.

### LINKED ISSUE #7202: [4.13.1] Unable to share to group from another app when not listed in recent chats.
I have:

### Bug description
When I try to share from an another app I cannot share to a group when it's not listed in the recent chat history. The recent chats only contains 5 items, which is a mixture of contacts and groups. When I scroll down to the contacts there are no groups. Also searching for the group did not find it.

### Steps to reproduce
- open an other app
- select share with signal
- try to share to a group, not listed in the recent chats.

**Actual result:** Unable to find the group.
**Expected result:** Should be able to select the group. I guess the best solution would be that the group is listed in the contact list. Which seems to be tracked here #6053

### Device info

**Device:** Samsung SM-G900F
**Android version:** 6.0.1
**Signal version:** 4.13.1

greyson-signal:
Currently, if you're searching for a contact to start a conversation with or send a share to (via the Android sharing system), groups do not appear. With this change, groups will now appear when searching, located under their own  heading.

I also took the opportunity to clean up ContactsCursorLoader to increase readability, especially now that groups are thrown into the mix, which ups the complexity of the data that is returned.

[Demo Videos](https://github.com/signalapp/Signal-Android/files/1857177/groups-in-search.zip)

Fixes #7202.

**Test Cases**
* Groups show up in search when choosing a share recipient
* Groups show up in search when choosing a new conversation recipient
* Groups do not show up when adding a member to a group during group creation
* Groups do not show up when adding a member to an existing group
* SMS contacts do not show up when adding a member to an existing group, as before
* Whether or not SMS enabled is respected on all of the above use cases
* When groups show up, headings will appear based on the following logic:
  * If there are both groups and contacts, contacts will appear first under a "Contacts" heading, and groups will appear below all contacts under a "Groups" heading.
  * If there are no groups, there will be no headings -- only contacts.
  * If there are groups but no normal contacts, there will be no headings -- only groups.

**Test Devices**
* [Moto X (2nd Generation), Android 7.1](https://www.gsmarena.com/motorola_moto_x_(2nd_gen)-6649.php)

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -90,6 +90,7 @@
     <!-- ContactsCursorLoader -->
     <string name="ContactsCursorLoader_recent_chats">Recent chats</string>
     <string name="ContactsCursorLoader_contacts">Contacts</string>
+    <string name="ContactsCursorLoader_groups">Groups</string>
 
     <!-- ContactsDatabase -->
     <string name="ContactsDatabase_message_s">Message %s</string>
--- a/src/org/thoughtcrime/securesms/ContactSelectionActivity.java
+++ b/src/org/thoughtcrime/securesms/ContactSelectionActivity.java
@@ -23,6 +23,7 @@ import android.support.v4.widget.SwipeRefreshLayout;
 import android.util.Log;
 
 import org.thoughtcrime.securesms.components.ContactFilterToolbar;
+import org.thoughtcrime.securesms.contacts.ContactsCursorLoader.DisplayMode;
 import org.thoughtcrime.securesms.util.DirectoryHelper;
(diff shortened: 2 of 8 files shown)


moxie0:
in 4.18.0

FeuRenard:
Is it intended that after tapping the pencil button there are no groups shown but they appear when I start searching?

greyson-signal:
@FeuRenard Yes, the intended behavior of this change was to show groups during search. I didn't change any non-search behavior. The ContactSelectionListFragment is used several places throughout the app. However, as mentioned in #7202, we'll be taking another look at how this fragment should work in a future release.

PR #7584, opened 2018-03-29
Setup domain fronting through Souq.

greyson-signal:
In preparation for Google shutting down domain fronting, we've moved to using Souq instead.

tcpdump of Signal, forced into domain fronting mode:
[Download](https://github.com/signalapp/Signal-Android/files/1860839/signal-egypt.zip)

**Test Devices**
* [Moto X (2nd Generation), Android 7.1](https://www.gsmarena.com/motorola_moto_x_(2nd_gen)-6649.php)


lgierth:
I'm curious, is there public information on google shutting down domain fronting soon, or is this just a precaution to avoid a single point of failur?

moxie0:
It has already been disabled. We don't use GH for discussion, though. Please check out the forums for that.

PR #7519, opened 2018-03-13
Only hide copy button for messages without text

FeuRenard:
### Contributor checklist

- [x] I am following the [Code Style Guidelines](https://github.com/signalapp/Signal-Android/wiki/Code-Style-Guidelines)
- [x] I have tested my contribution on these devices:
 * AVD Nexus 5X, Android 7.1.1
- [x] My contribution is fully baked and ready to be merged as is
- [x] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the `Fixes #1234` [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)

----------

### Description
Fixes #7472
Supersedes #5857

PR #7620, opened 2018-04-03
Allow copying of text from media messages.

### LINKED ISSUE #7472: Regression: No option in context menu to copy text of media message
I have:

### Bug description
It is no longer possible to copy the text of media messges. This behaviour is introduced by 4bdb2acd29f06ec0782f33e20b3c8290a002bef3.

### Steps to reproduce
- send a media message with text
- long-press this message

**Actual result:** No copy button available.
**Expected result:** A copy button should allow users to copy the text.

### Screenshots
bildschirmfoto von 2018-03-08 19-00-15

### Device info

**Device:** AVD Nexus 5X
**Android version:** 7.1.1
**Signal version:** 4.17.0

### Workaround
Use the forward button and select any conversation. Then copy the text from the text input box at the bottom of the conversation.

greyson-signal:
Previously, the copy option was disabled for all media messages, even if they have text. This change enables the copy option for media messages that have text. Only the text is copied, not the media itself.

Fixes #7472.

**Test Cases**
* Copy regular message with text
* Copy option appears and copies text on media message with text
* Copy option does not appear on media message without text
* Sent media message with whitespace as the text -- whitespace is trimmed, no copy option appears after sending
* Copy option appears when multiple messages are selected, including media messages
* Copy multiple messages at once, including media messages both with and without text
  * Message with no text first, in the middle, and last -- verified pasted text had no extra newlines

**Test Devices**
* [OnePlus One, Android 5.0.2](https://www.gsmarena.com/oneplus_one-6327.php)

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -201,6 +201,7 @@ public class ConversationFragment extends Fragment
   private void setCorrectMenuVisibility(Menu menu) {
     Set<MessageRecord> messageRecords = getListAdapter().getSelectedItems();
     boolean            actionMessage  = false;
+    boolean            hasText        = false;
 
     if (actionMode != null && messageRecords.size() == 0) {
       actionMode.finish();
@@ -214,6 +215,11 @@ public class ConversationFragment extends Fragment
           messageRecord.isIdentityVerified() || messageRecord.isIdentityDefault())
       {
         actionMessage = true;
+      }
+      if (messageRecord.getBody().length() > 0) {
+        hasText = true;
+      }
+      if (actionMessage && hasText) {
         break;
       }
     }
@@ -223,7 +229,6 @@ public class ConversationFragment extends Fragment
       menu.findItem(R.id.menu_context_details).setVisible(false);
       menu.findItem(R.id.menu_context_save_attachment).setVisible(false);
       menu.findItem(R.id.menu_context_resend).setVisible(false);
-      menu.findItem(R.id.menu_context_copy).setVisible(!actionMessage);
     } else {
       MessageRecord messageRecord = messageRecords.iterator().next();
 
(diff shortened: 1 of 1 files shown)


FeuRenard:
It's not possible to copy the text of multiple messages (including at least one with media) in one go after this change.

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -230,7 +223,7 @@ private void setCorrectMenuVisibility(Menu menu) {
       menu.findItem(R.id.menu_context_details).setVisible(false);
       menu.findItem(R.id.menu_context_save_attachment).setVisible(false);
       menu.findItem(R.id.menu_context_resend).setVisible(false);
-      menu.findItem(R.id.menu_context_copy).setVisible(!actionMessage  && !mediaMessage);
+      menu.findItem(R.id.menu_context_copy).setVisible(!actionMessage);
```
moxie0: Doesn't this mean that selecting two media messages with no text will expose the "copy" option?

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -242,7 +235,7 @@ private void setCorrectMenuVisibility(Menu menu) {
 
       menu.findItem(R.id.menu_context_forward).setVisible(!actionMessage);
       menu.findItem(R.id.menu_context_details).setVisible(!actionMessage);
-      menu.findItem(R.id.menu_context_copy).setVisible(!actionMessage && !mediaMessage);
+      menu.findItem(R.id.menu_context_copy).setVisible(!actionMessage && messageRecord.getBody().length() > 0);
```
moxie0: Seems strange that selecting a media message with no text that's after a message with text will not expose copy, but the inverse will?

src/org/thoughtcrime/securesms/ConversationFragment.java
FeuRenard: That's why I introduce a variable `emptyBody` to hide the copy button as soon as there is a message without text selected in PR #7519.

src/org/thoughtcrime/securesms/ConversationFragment.java
greyson-signal: So right now the logic is: if there's only one message selected, only show copy if there's text -- if >1 are selected, always show copy. I can change it so we always display copy based on whether there is text present.

greyson-signal:
Okee dokee, updated is so that we check to make sure that _some_ message has text in order to show 'copy'. Also, found there was a little bug with the text that was copied -- it'd put a newline for messages with no text. Fixed that too.

moxie0:
in 4.19.0

PR #7699, opened 2018-04-18
Add IR to censored country code list

mrphs:
Now that you've switched from Google meek, it makes sense to add +98 back to the list of country codes that should be using meek.

I tried to sign the cla but it gave me an error:
signal-cla


moxie0:
Thanks but the current implementation uses a domain front that's not in the alexa top 100/50 for +98. We'd need to identify something there in order to successfully domain front in that region.

PR #6165, opened 2017-02-04
Allow deleting attachments from media overview

haffenloher:
### Contributor checklist

- [x] I am following the [Code Style Guidelines](https://github.com/WhisperSystems/Signal-Android/wiki/Code-Style-Guidelines)
- [x] I have tested my contribution on these devices:
* Android 6 and 7.1 emulators
* Android 4.2.2 phone
- [x] My contribution is fully baked and ready to be merged as is
- [x] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the `Fixes #1234` [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)
- [x] I have made the choice whether I want the [BitHub reward](https://github.com/WhisperSystems/Signal-Android/wiki/BitHub-Rewards) or not by omitting or adding the word `FREEBIE` in the commit message of my first commit

----------

### Description

Batch-select and delete attachments.

The message itself is currently not deleted (even if it's empty); only the attachment is removed.

Fixes #4242
Fixes #6111

PR #7722, opened 2018-04-23
Add placeholder icons for missing quote thumbnails.

greyson-signal:
This PR puts in placeholder icons for quote thumbnails when they're not available. Also, for image attachments, if a thumbnail URI is not available, we'll first fallback to using the URI of the image itself before falling back to a placeholder icon.

Along the way, I did a small naming refactor for MediaUtil methods so that all of the names were consistent, and I also ended up refactoring how we set thumbnails/icons in QuoteView.

quote-thumbnail-placeholders

**Test Devices**
* [Galaxy S3 Mini, Android 4.2.2. API 17](https://www.gsmarena.com/samsung_i8200_galaxy_s_iii_mini_ve-6190.php)

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1924,11 +1924,11 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   public void onMediaSelected(@NonNull Uri uri, String contentType) {
     if (!TextUtils.isEmpty(contentType) && contentType.trim().equals("image/gif")) {
       setMedia(uri, MediaType.GIF);
-    } else if (MediaUtil.isImageType(contentType)) {
+    } else if (MediaUtil.isImage(contentType)) {
       setMedia(uri, MediaType.IMAGE);
-    } else if (MediaUtil.isVideoType(contentType)) {
+    } else if (MediaUtil.isVideo(contentType)) {
       setMedia(uri, MediaType.VIDEO);
-    } else if (MediaUtil.isAudioType(contentType)) {
+    } else if (MediaUtil.isAudio(contentType)) {
       setMedia(uri, MediaType.AUDIO);
     }
   }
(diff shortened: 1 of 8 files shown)


greyson-signal:
@PiCob Good question! Let's look at this scenario:

User A sends a message Y that quotes message X (which contains media)
User B receives message Y, but they had previously deleted message X

We actually attach a scaled-down thumbnail of message X's media inside of message Y for this very reason. So in this case, User B will display that thumbnail.

So really, this PR is for an extreme scenario where for some reason a thumbnail just isn't present at all -- meaning we didn't get a scaled-down thumbnail from the sender, and we don't have the message locally (or some other error happened).

I think you also may be talking about a scenario where the user deletes the media of a message, but not the message itself? I don't believe that the app itself presents that ability - deleting the media (via the media viewer, or "all media" screen) will also delete the message itself.

PiCob:
@greyson-signal 

Thanks for the fast response!

> I think you also may be talking about a scenario where the user deletes the media of a message, but not the message itself?

Correct (sorry if my description was unclear/confusing).

> I don't believe that the app itself presents that ability - deleting the media (via the media viewer, or "all media" screen) will also delete the message itself

Oh, really?! :open_mouth: Now that's not what (I guess) someone is expecting when you're going to **_delete media_**, since the media technically (and IMHO logically as well) is just a **_part_** of the message and should be handled appropriately (like anywhere else, e. g. when copying messages, creating backups etc.). The way it seems to be implemented right now does delete every piece of text (and that again is most often the most valuable content) contained in the message the media belongs to as well, if I get you right I'd even consider this behaviour as a bug, because I don't want to batch delete whole messages (including text) when cleaning up space by deleting pictures for example... :confused:

Is there a reason for not at least keeping the **text** of a message (if it's not a media-only message) and just insert a placeholder like in this commit (or even better: optionally keep/embed a thumbnail of the deleted media)? IIRC, the [PR #6165](https://github.com/signalapp/Signal-Android/pull/6165) did it that way, but was closed since you implemented the (batch) media deletion in a superior manner.

If you don't want to discuss here, we can move over to the Community forum or anywhere else - just let me know.

PiCob:
@greyson-signal Done: https://community.signalusers.org/t/media-deletion-actually-means-message-deletion/2680 (posting it here just to create a backlink). Thanks again!

PR #7780, opened 2018-05-10
Added ability to share contacts.

greyson-signal:
The "contact" option in the attachments tray now brings you through an optimized contact sharing flow, allowing you to select specific fields to share. The contact is then presented as a special message type, allowing you to interact with the card to add the contact to your system contacts, invite them to signal, initiate a signal message, etc.

[Demo](https://github.com/signalapp/Signal-Android/files/1990404/contact-sharing.zip)

**To-Do**
- [x] Update libsignal-service-java

**Follow-Up PR's**
* Name editing
* vCard import

**Test Devices**


src/org/thoughtcrime/securesms/database/helpers/SQLCipherOpenHelper.java
```diff
@@ -177,6 +178,10 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
         db.execSQL("ALTER TABLE part ADD COLUMN quote INTEGER DEFAULT 0");
       }
 
+      if (oldVersion < CONTACT_SHARING) {
+        db.execSQL("ALTER TABLE part ADD COLUMN linked_id INTEGER DEFAULT 0");
```
greyson-signal: Database explanation:

Contacts are stored as a pair of attachments. One is just a JSON blob of the serialized contact, and the other is the avatar photo. When building a message record, we want to be sure that the contact-related attachments (especially the avatar) don't show up as regular image attachments. To take care of this, I introduced the concept of an attachment "link".

The JSON blob is stored with a special mimetype (text/signal-contact). Contact avatars are stored like regular image attachments, but with a linked_id equal to the JSON blob's attachment _id. When the JSON blob attachment is found (which we can detect via it's mimetype), we see if there's any other contacts in the set that have a linked_id equal to the contact attachment's _id. If there is, those are removed and paired off as a contact.

Given that we can't read the JSON blob off the disk on the main thread, the contacts are given to the message record via a ContactRetriever. We use the ViewModel pattern inside of the ConversationItem to retrieve the contact from the ContactRetriever asynchronously.

greyson-signal:
[Cross-dependecy on libsignal-service-java](https://github.com/signalapp/libsignal-service-java/pull/55)

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -2072,11 +2113,26 @@ public void handleReplyMessage(MessageRecord messageRecord) {
       author = messageRecord.getIndividualRecipient();
     }
 
-    inputPanel.setQuote(GlideApp.with(this),
-                        messageRecord.getDateSent(),
-                        author,
-                        messageRecord.getBody(),
-                        messageRecord.isMms() ? ((MmsMessageRecord)messageRecord).getSlideDeck() : new SlideDeck());
+    if (messageRecord.isMms() && ((MmsMessageRecord) messageRecord).getSlideDeck().getSharedContactSlide() != null) {
```
moxie0: Why is this special cased? The contact should be a slide just like any other, and it's the quote view's responsibility to know how to render it, no?

src/org/thoughtcrime/securesms/ConversationAdapter.java
```diff
@@ -91,14 +91,14 @@
...
-  private final @NonNull  Calendar          calendar;
-  private final @NonNull  MessageDigest     digest;
+  private final @Nullable ItemClickListener  clickListener;
+  private final @NonNull  GlideRequests      glideRequests;
+  private final @NonNull  Locale             locale;
+  private final @NonNull  Recipient          recipient;
+  private final @NonNull  MmsSmsDatabase     db;
+  private final @NonNull  LayoutInflater     inflater;
+  private final @NonNull  Calendar           calendar;
+  private final @NonNull  MessageDigest      digest;
```
moxie0: Is this still necessary?

src/org/thoughtcrime/securesms/ConversationAdapter.java
```diff
@@ -166,14 +166,14 @@ public ConversationAdapter(@NonNull Context context,
...
-      this.calendar      = Calendar.getInstance();
-      this.digest        = MessageDigest.getInstance("SHA1");
+      this.glideRequests      = glideRequests;
+      this.locale             = locale;
+      this.clickListener      = clickListener;
+      this.recipient          = recipient;
+      this.inflater           = LayoutInflater.from(context);
+      this.db                 = DatabaseFactory.getMmsSmsDatabase(context);
+      this.calendar           = Calendar.getInstance();
+      this.digest             = MessageDigest.getInstance("SHA1");
```
moxie0: is this still necessary?

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -85,12 +96,13 @@
...
 
-  private static final long   PARTIAL_CONVERSATION_LIMIT = 500L;
+  private static final long PARTIAL_CONVERSATION_LIMIT = 500L;
+  private static final int  CODE_ADD_EDIT_CONTACT      = 77;
 
   private final ActionModeCallback actionModeCallback     = new ActionModeCallback();
   private final ItemClickListener  selectionClickListener = new ConversationFragmentItemClickListener();
 
-  private ConversationFragmentListener listener;
+  private ConversationFragmentListener    listener;
```
moxie0: necessary?

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -105,6 +117,7 @@
   private View                        composeDivider;
   private View                        scrollToBottomButton;
   private TextView                    scrollDateHeader;
+  private Contact pendingAddedContact;
```
moxie0: align please

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -652,6 +665,61 @@ protected void onPostExecute(Integer position) {
...
+    }
+
+    @Override
+    public void onMessageSharedContactClicked(@NonNull List<Recipient> choices) {
+      selectNumberAndOpenConversation(choices, null);
+    }
+
+    @Override
+    public void onInviteSharedContactClicked(@NonNull List<Recipient> choices) {
+      selectNumberAndOpenConversation(choices, getString(R.string.InviteActivity_lets_switch_to_signal, "https://sgnl.link/1KpeYmF"));
```
moxie0: Might as well just fire the sms compose intent here? That way if Signal isn't the default sms app, whatever is will open?

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -406,10 +415,60 @@ private void setBodyText(MessageRecord messageRecord) {
...
+
+        @Override
+        public void onMessageClicked(@NonNull List<Recipient> choices) {
+          if (eventListener != null && batchSelected.isEmpty()) {
+            eventListener.onMessageSharedContactClicked(choices);
+          } else {
+            passthroughClickListener.onClick(sharedContactView);
+          }
+        }
+      });
```
moxie0: it seems strange to me that this if block is so radically different from the others in the same method

src/org/thoughtcrime/securesms/MessageDetailsActivity.java
```diff
@@ -67,7 +67,9 @@
 /**
  * @author Jake McGinty
  */
-public class MessageDetailsActivity extends PassphraseRequiredActionBarActivity implements LoaderCallbacks<Cursor>, RecipientModifiedListener {
+public class MessageDetailsActivity extends PassphraseRequiredActionBarActivity implements LoaderCallbacks<Cursor>,
+                                                                                           RecipientModifiedListener
+{
```
moxie0: Did anything in this file change?

src/org/thoughtcrime/securesms/SignalExecutors.java
```diff
@@ -0,0 +1,9 @@
+package org.thoughtcrime.securesms;
+
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+
+public class SignalExecutors {
+
+  public static final Executor DATABASE = Executors.newSingleThreadExecutor();
```
moxie0: Given the infrequency of these messages, it does not feel to me like we should have a dedicated thread for them allocated at all times

src/org/thoughtcrime/securesms/components/SharedContactView.java
```diff
@@ -0,0 +1,207 @@
...
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+public class SharedContactView extends LinearLayout implements SharedContactInjector.Target,
+                                                               RecipientModifiedListener {
```
moxie0: brace on newline please

src/org/thoughtcrime/securesms/contactshare/ContactRepository.java
```diff
@@ -0,0 +1,281 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.concurrent.Executor;
+
+public class ContactRepository {
```
moxie0: We already have a ContactsDatabase, which is responsible for interfacing with the system contacts database. What is the difference between ContactsDatabase and ContactRepository, and is there a reason for both to exist?

src/org/thoughtcrime/securesms/contactshare/model/ContactReader.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class ContactReader {
+
+  private final InputStream inputStream;
+  private final Contact     contact;
+
+  public ContactReader(@NonNull InputStream inputStream) throws IOException {
+    this.inputStream = inputStream;
+
+    DataInputStream dataStream    = new DataInputStream(inputStream);
```
moxie0: In general, try to never use DataInputStream, it will somehow eventually lead to doom.

src/org/thoughtcrime/securesms/contactshare/model/ContactReader.java
```diff
@@ -0,0 +1,64 @@
...
+      throw new IOException("Invalid contact length: " + contactLength);
+    }
+
+    byte[] contactBytes = new byte[contactLength];
+
+    int totalRead = 0;
+    int read      = 0;
+    while ((read = inputStream.read(contactBytes, totalRead, contactBytes.length - totalRead)) > 0 && totalRead < contactBytes.length) {
+      totalRead += read;
+    }
```
moxie0: readfully exists already in a util class

src/org/thoughtcrime/securesms/contactshare/model/ContactReader.java
```diff
@@ -0,0 +1,64 @@
...
+
+    if (totalRead < contactBytes.length) {
+      throw new IOException("Failed to read all of the contact bytes.");
+    }
+
+    try {
+      contact = (Contact) new ObjectInputStream(new ByteArrayInputStream(contactBytes)).readObject();
```
moxie0: java serialization should always be an absolute last resort, it will always lead to doom

src/org/thoughtcrime/securesms/contactshare/model/ContactStream.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.SequenceInputStream;
+import java.util.Arrays;
+import java.util.Collections;
+
+public final class ContactStream extends InputStream {
```
moxie0: Is this necessary? Seems strange that we have both a ContactStream, a ContactReader, and a normal InputStream

src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
```diff
@@ -360,6 +372,30 @@ void insertAttachmentsForMessage(long mmsId, @NonNull List<Attachment> attachmen
     for (Attachment attachment : quoteAttachment) {
       insertAttachment(mmsId, attachment, true);
     }
+
+    for (SharedContact sharedContact : sharedContacts) {
+      insertContactAttachment(mmsId, sharedContact);
+    }
```
moxie0: why does this need to be special cased? seems like we're trying to go for a generic attachment thing here

src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
```diff
@@ -338,16 +347,19 @@ public void insertAttachmentsForPlaceholder(long mmsId, @NonNull AttachmentId at
 
     if (database.update(TABLE_NAME, values, PART_ID_WHERE, attachmentId.toStrings()) == 0) {
       //noinspection ResultOfMethodCallIgnored
-      dataInfo.file.delete();
+      newDataInfo.file.delete();
     } else {
+      if (existingDataInfo != null) {
+        existingDataInfo.file.delete();
+      }
```
moxie0: I think all this stuff should get pushed up to the caller

src/org/thoughtcrime/securesms/glide/KeyedInputStream.java
```diff
@@ -0,0 +1,24 @@
...
+
+import android.support.annotation.NonNull;
+
+import com.bumptech.glide.load.Key;
+
+import java.io.FilterInputStream;
+import java.io.InputStream;
+import java.security.MessageDigest;
+
+public class KeyedInputStream extends FilterInputStream implements Key {
```
moxie0: Do we not already have the ability to load via inputstream? Can the key no longer be specified in the glide call with the new api?

src/org/thoughtcrime/securesms/contactshare/ContactRepository.java
greyson-signal: In the new architecture pattern Google is pushing, a repository is the abstraction layer over top of your data. The idea is that you get all of your data from a repository, but you don't care how the repository gets the data. Here, the repository fetches data from the database, but the idea is that you could change how the repository gets data in the future (if, say, we changed to something other than SQLite). You could also imagine other scenarios where a repository gives you back local data, then initiates a fetch for remote data, etc.

It also handles any threading concerns and can expose friendlier interfaces than a raw DB method would (i.e. converts cursors into models and stuff).

More details here: https://developer.android.com/jetpack/docs/guide

src/org/thoughtcrime/securesms/glide/KeyedInputStream.java
greyson-signal: InputStreams are not valid Glide models by default, no. If there is a way to specify a key on an ad-hoc basis, I don't see it. Creating a new Loader/Fetcher was what my research pointed me to.

src/org/thoughtcrime/securesms/ConversationItem.java
greyson-signal: Talked offline, but basically it's more complicated because the view itself is more complicated. It has various events it can emit besides just a "click", which is the only event the other views report. I can move this out to a separate method though or something so it looks cleaner.

src/org/thoughtcrime/securesms/ConversationActivity.java
greyson-signal: Talked offline, but there's issues rendering the SharedContactSlide inside of the QuoteView, because the presentation data has to be fetched asynchronously, causing funny redraws, as well as making the code in QuoteView less clear. Also, the outbound Quote proto-filling code is made more complex due to having to special-case the contact attachment. 

Options are: (1) Keep the code as-is, which will just set the text of the quote equal to the contact name. (2) Pass the contact in as a slide, but take additional time to do a light refactor of the QuoteView and clean up other UI issues (we could set the QuoteView to have a fixed width when rendering a contact to avoid glitchiness, maybe?)

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -652,6 +664,59 @@ protected void onPostExecute(Integer position) {
...
+      if (getContext() != null && getActivity() != null) {
+        Bundle bundle = ActivityOptionsCompat.makeSceneTransitionAnimation(getActivity(), avatarTransitionView, "avatar").toBundle();
+        ActivityCompat.startActivity(getActivity(), SharedContactDetailsActivity.getIntent(getContext(), sharedContactSlide), bundle);
+      }
+    }
+
+    @Override
+    public void onAddToContactsClicked(@NonNull SharedContactSlide sharedContactSlide, @NonNull Contact sharedContact) {
+      if (getContext() != null && sharedContactSlide.getUri() != null) {
+        pendingAddedContact = sharedContact;
```
moxie0: i really don't like the idea of stashing something as an instance variable

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -218,6 +222,7 @@ public void bind(@NonNull MessageRecord      messageRecord,
     setSimInfo(messageRecord);
     setExpiration(messageRecord);
     setQuote(messageRecord);
+//    setSharedContacts(messageRecord);
```
moxie0: why is this still here?

src/org/thoughtcrime/securesms/SignalExecutors.java
```diff
@@ -0,0 +1,17 @@
+package org.thoughtcrime.securesms;
+
+import java.util.concurrent.Executor;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+public class SignalExecutors {
+
+  public static final Executor DATABASE = newBoundedCachedThreadPool(1);
```
moxie0: I don't think we should be using an executor for this feature

src/org/thoughtcrime/securesms/contacts/avatars/ProfileContactPhoto.java
```diff
@@ -27,6 +29,16 @@ public InputStream openInputStream(Context context) throws IOException {
     return AvatarHelper.getInputStreamFor(context, address);
   }
 
+  @Override
+  public @Nullable Uri getUri(@NonNull Context context) {
+    return Uri.fromFile(AvatarHelper.getAvatarFile(context, address));
```
moxie0: does this work? there doesn't need to be a file provider or soemthing?

src/org/thoughtcrime/securesms/contactshare/ContactReader.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class ContactReader {
+
+  private final InputStream inputStream;
+  private final Contact     contact;
+
+  public ContactReader(@NonNull InputStream inputStream) throws IOException {
+    this.inputStream = inputStream;
+
+    DataInputStream dataStream    = new DataInputStream(inputStream);
```
moxie0: please do not ever use datainputstream, it will lead to doom

src/org/thoughtcrime/securesms/contactshare/ContactReader.java
```diff
@@ -0,0 +1,62 @@
...
+      throw new IOException("Invalid contact length: " + contactLength);
+    }
+
+    byte[] contactBytes = new byte[contactLength];
+
+    int totalRead = 0;
+    int read      = 0;
+    while ((read = inputStream.read(contactBytes, totalRead, contactBytes.length - totalRead)) > 0 && totalRead < contactBytes.length) {
+      totalRead += read;
+    }
```
moxie0: readFully already exists in util

src/org/thoughtcrime/securesms/contactshare/SharedContactInjector.java
```diff
@@ -0,0 +1,99 @@
...
+      @Override
+      protected ResolvedContact doInBackground(Void... voids) {
+        try {
+          ContactReader   reader     = new ContactReader(PartAuthority.getAttachmentStream(context, contactSlideUri));
+          List<Recipient> recipients = ContactUtil.getRecipients(context, reader.getContact());
+
+          for (Recipient recipient : recipients) {
+            recipient.resolve();
+            // TODO(greyson): Don't do this -- maybe do it indirectly through the DB?
+            Recipient.applyCached(recipient.getAddress(), cached -> {
```
moxie0: please fix this

src/org/thoughtcrime/securesms/database/AttachmentDatabase.java
```diff
@@ -315,18 +323,19 @@ private void deleteAttachmentOnDisk(@Nullable String data, @Nullable String thum
   public void insertAttachmentsForPlaceholder(long mmsId, @NonNull AttachmentId attachmentId, @NonNull InputStream inputStream)
```
moxie0: My understanding was that no changes to attachmentdb should be required?

src/org/thoughtcrime/securesms/database/MmsDatabase.java
```diff
@@ -666,7 +669,7 @@ public long copyMessageInbox(long messageId) throws MmsException {
     contentValues.put(CONTENT_LOCATION, contentLocation);
     contentValues.put(STATUS, Status.DOWNLOAD_INITIALIZED);
     contentValues.put(DATE_RECEIVED, generatePduCompatTimestamp());
-    contentValues.put(PART_COUNT, retrieved.getAttachments().size());
+    contentValues.put(PART_COUNT, retrieved.getAttachments().size() + retrieved.getSharedContacts().size());
```
moxie0: can contacts not be an attachment like any other?

src/org/thoughtcrime/securesms/database/RecipientDatabase.java
```diff
@@ -448,6 +448,7 @@ public void finish() {
               recipient.setSystemContactPhoto(Util.uri(entry.getValue().photoUri));
               recipient.setCustomLabel(entry.getValue().phoneLabel);
               recipient.setContactUri(Util.uri(entry.getValue().contactUri));
+              recipient.setIsSystemContact(true);
```
moxie0: how does this ever get set to false?

src/org/thoughtcrime/securesms/jobs/PushDecryptJob.java
```diff
@@ -522,17 +525,18 @@ private void handleMediaMessage(@NonNull SignalServiceEnvelope envelope,
...
+    Optional<QuoteModel> quote          = getValidatedQuote(message.getQuote());
+    IncomingMediaMessage mediaMessage   = new IncomingMediaMessage(Address.fromExternal(context, envelope.getSource()),
+                                                                   message.getTimestamp(), -1,
+                                                                   message.getExpiresInSeconds() * 1000L, false,
+                                                                   Optional.fromNullable(envelope.getRelay()),
+                                                                   message.getBody(),
+                                                                   message.getGroupInfo(),
+                                                                   message.getAttachments(),
+                                                                   quote,
+                                                                   message.getSharedContacts().or(Collections.emptyList()));
```
moxie0: This is generally the boundary between service and app. I don't think IncomingMediaMessage should have any SignalService* instance variables

src/org/thoughtcrime/securesms/jobs/PushSendJob.java
```diff
@@ -158,6 +171,41 @@ protected void notifyMediaMessageDeliveryFailed(Context context, long messageId)
     return Optional.of(new SignalServiceDataMessage.Quote(quoteId, new SignalServiceAddress(quoteAuthor.serialize()), quoteBody, quoteAttachments));
   }
 
+  List<SharedContact> getSharedContactsFor(List<Attachment> attachments) {
+    List<SharedContact> contacts = new LinkedList<>();
+
+    for (Attachment attachment : attachments) {
+      if (!MediaUtil.SHARED_CONTACT.equals(attachment.getContentType()) || attachment.getDataUri() == null) {
+        continue;
```
moxie0: please don't use continue

src/org/thoughtcrime/securesms/contacts/avatars/ProfileContactPhoto.java
greyson-signal: Tested, works fine. Androdi understands `file://` Uri's by default.

src/org/thoughtcrime/securesms/ConversationFragment.java
greyson-signal: If we opt to refresh the entire recipient list, then we won't need to.

src/org/thoughtcrime/securesms/SignalExecutors.java
greyson-signal: What would you use instead? ContactRepository does heavy work that needs to take place off of the main thread. AsyncTasks aren't necessary because LiveData will publish events to the main thread on your behalf.

moxie0: We often use asynctask without making use of onpostexecute to take advantage of an existing worker thread pool. We should not be allocating an executor for such a trivial feature

build.gradle
```diff
@@ -75,7 +76,7 @@ dependencies {
     compile('org.whispersystems:libpastelog:1.1.2') {
         exclude group: 'com.squareup.okhttp3', module: 'okhttp'
     }
-    compile 'org.whispersystems:signal-service-android:2.7.5'
+    compile 'org.whispersystems:signal-service-android:2.7.6-RC2'
```
greyson-signal: @moxie0  Can you look at https://github.com/signalapp/libsignal-service-java/pull/55 and then cut a release?

src/org/thoughtcrime/securesms/contactshare/ContactUtil.java
```diff
@@ -0,0 +1,216 @@
...
+      return phoneNumber;
+    }
+  }
+
+  public static @NonNull String getNormalizedPhoneNumber(@NonNull Context context, @NonNull String number) {
+    Address address = Address.fromExternal(context, number);
+    return address.serialize();
+  }
+
+  public static void selectRecipient(@NonNull Context context, @NonNull List<Recipient> choices, @NonNull Locale locale, @NonNull RecipientSelectedCallback callback) {
```
moxie0: Annotate mainthread? also can we change the method signature to indicate that it will be displaying a dialog?

src/org/thoughtcrime/securesms/database/MmsDatabase.java
```diff
@@ -902,6 +968,46 @@ public void deleteThread(long threadId) {
...
+
+        Avatar  updatedAvatar  = new Avatar(attachmentId,
+                                            contact.getAvatarAttachment(),
+                                            contact.getAvatar() != null && contact.getAvatar().isProfile());
+        Contact updatedContact = new Contact(contact.getName(),
+                                             contact.getOrganization(),
+                                             contact.getPhoneNumbers(),
+                                             contact.getEmails(),
+                                             contact.getPostalAddresses(),
+                                             updatedAvatar);
```
moxie0: seems like a copy constructor might help

src/org/thoughtcrime/securesms/database/MmsDatabase.java
```diff
@@ -902,6 +968,46 @@ public void deleteThread(long threadId) {
     deleteThreads(singleThreadSet);
   }
 
+  private void updateMessageSharedContacts(long mmsId, @NonNull Map<Attachment, AttachmentId> insertedAttachmentIds, @NonNull List<Contact> contacts) {
```
moxie0: let's have this construct and return the json, then the caller do the db operation. less side effects in a private method that way

src/org/thoughtcrime/securesms/database/MmsDatabase.java
```diff
@@ -595,6 +608,50 @@ public OutgoingMediaMessage getOutgoingMessage(long messageId)
...
+          DatabaseAttachment attachment = attachmentIdMap.get(contact.getAvatar().getAttachmentId());
+          contacts.add(new Contact(contact.getName(),
+                                   contact.getOrganization(),
+                                   contact.getPhoneNumbers(),
+                                   contact.getEmails(),
+                                   contact.getPostalAddresses(),
+                                   new Avatar(contact.getAvatar().getAttachmentId(), attachment, contact.getAvatar().isProfile())));
+
+          if (attachment != null) {
+            attachments.remove(attachment);
```
moxie0: probably better for a private method not to modify an argument it's given. particularly given that there are multiple callers with different expectations, it seems

1) this is modifying "associatedAttachments," which isn't actually referenced after this is called. so if the intention is to retroactively remove from "attachments," I don't think that's happening here.
2) if the intention is to remove from attachments, it looks like it should be possible to reorder its construction and filter appropriately after this returns.

src/org/thoughtcrime/securesms/database/RecipientDatabase.java
moxie0: if someone is removed from system contacts, i don't see how it would get set to false until there's an app restart. basically the inverse of the current problem. the sync adapter is run on removal as well as addition

src/org/thoughtcrime/securesms/util/CommunicationActions.java
```diff
@@ -0,0 +1,68 @@
...
+          activity.startActivity(activityIntent);
+        })
+        .execute();
+  }
+
+  public static void startConversation(@NonNull  Context context,
+                                       @NonNull  Address address,
+                                                 long    threadId,
+                                       @Nullable String  text)
+  {
```
moxie0: Does anyone who calls this actually have the threadId to start with? If not, let's just put the async task here?

src/org/thoughtcrime/securesms/util/MediaUtil.java
```diff
@@ -46,6 +44,7 @@
   public static final String AUDIO_AAC         = "audio/aac";
   public static final String AUDIO_UNSPECIFIED = "audio/*";
   public static final String VIDEO_UNSPECIFIED = "video/*";
+  public static final String SHARED_CONTACT    = "text/signal-contact";
```
moxie0: What is this mime type for? Seems potentially dangerous to use "text/" as a prefix, since code might just check for prefix?

greyson-signal: Ah, I missed a spot :) We no longer need this.

res/layout/conversation_item_sent.xml
```diff
@@ -47,6 +47,12 @@
                     app:message_type="outgoing"
                     tools:visibility="visible"/>
 
+            <org.thoughtcrime.securesms.components.SharedContactView
+                    android:id="@+id/shared_contact_view"
+                    android:layout_width="@dimen/media_bubble_max_width"
+                    android:layout_height="wrap_content"
+                    android:visibility="gone"/>
```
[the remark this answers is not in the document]
greyson-signal: That's fair, I was thinking that GONE was enough, but from what I'm reading, I think GONE views are still inflated (but not drawn or measured), while ViewStubs aren't inflated at all. Switching.

moxie0:
lgtm

vgerasymenko:
Any plans to add such ability for sending and receiving contacts in vCard format in MMS? Such support was requested many times, for example: https://github.com/signalapp/Signal-Android/issues/535.

greyson-signal:
@vgerasymenko I believe that if you start in your contacts app and share a contact _into_ Signal that it will attach as a vCard when using MMS. Using the contact share button _inside of_ signal when using MMS will simply share the raw phone number.

vgerasymenko:
@greyson-signal Yes, and this is not so usable, as new contact sharing flow. Moreover, received MMS with vCards displayed as text in vCard format, which is even less usable. Any plans to make this cases better?

PR #7812, opened 2018-05-17
Convert vCard attachments into contact shares.

greyson-signal:
When you share a vCard from an external app (like the Contacts app) into Signal, we'll now convert it to a pretty Shared Contact message and allow you to choose which fields of the contact you wish to send.

To do this, we imported a new vCard parsing library. Then, the ContactShareEditActivity/ViewModel was adjusted to take in raw Uri's instead of system contact ID's. The ContactRepository will then convert the Uri to Contact by recognizing the authority and parsing as appropriate (either by reading the system contacts or the vCard).

Care was taken to ensure that vCard shares to non-signal users still use the legacy attachment flow.
[Demo Video](https://github.com/signalapp/Signal-Android/files/2014399/vcard-sharing.zip)
Demo

**Test Devices**

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -426,11 +426,13 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
       setMedia(data.getData(), MediaType.AUDIO);
       break;
     case PICK_CONTACT:
-      if (isSecureText && !isSmsForced()) {
-        openContactShareEditor(data.getData());
-      } else {
-        addAttachmentContactInfo(data.getData());
-      }
+      // TODO(greyson): Re-enable shared contact sending after receiving has been enabled for a few releases
+      addAttachmentContactInfo(data.getData());
+//      if (isSecureText && !isSmsForced()) {
+//        openContactShareEditor(data.getData());
+//      } else {
+//        addAttachmentContactInfo(data.getData());
+//      }
       break;
     case GET_CONTACT_DETAILS:
       sendSharedContact(data.getParcelableArrayListExtra(ContactShareEditActivity.KEY_CONTACTS));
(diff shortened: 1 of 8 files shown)


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1393,12 +1393,16 @@ private void setMedia(@Nullable Uri uri, @NonNull MediaType mediaType) {
 
   private void setMedia(@Nullable Uri uri, @NonNull MediaType mediaType, int width, int height) {
     if (uri == null) return;
-    attachmentManager.setMedia(glideRequests, uri, mediaType, getCurrentMediaConstraints(), width, height);
+
+    if (MediaType.VCARD.equals(mediaType) && isSecureText) {
+      openContactShareEditor(uri);
+    } else {
+      attachmentManager.setMedia(glideRequests, uri, mediaType, getCurrentMediaConstraints(), width, height);
+    }
```
moxie0: Is there a reason to go through ConversationActivity at all if we really want a different activity?

src/org/thoughtcrime/securesms/contactshare/ContactRepository.java
```diff
@@ -45,11 +52,18 @@
...
-        Contact contact = getContact(id);
+      List<Contact> contacts = new ArrayList<>(contactUris.size());
+      for (Uri contactUri : contactUris) {
+        Contact contact;
+
+        if (ContactsContract.AUTHORITY.equals(contactUri.getAuthority())) {
+          contact = getContactFromSystemContacts(ContactUtil.getContactIdFromUri(contactUri));
+        } else {
+          contact = getContactFromVcard(contactUri);
+        }
```
moxie0: This is kind of strange. Typically the idea behind using a Uri is to get a consistent abstract interface to loading data, but here that's not what is happening. Not only is the data different, the interface for loading it is too.

greyson-signal: I guess I was just carrying this "inconsistent underlying data" over from Android. Android gives us both system contacts and vCards as Uri's. At some point, we need to translate the Uri into a Contact model. I thought the most natural place to do this was the ContactRepository.

I suppose an alternative would be to have two interfaces on ContactRepository: one to load from a contact ID (we previously parsed the ID out of the Uri on ConversationActivity), and one to load from a Uri. The Uri path would assume that it is a vCard.

What do you think?

PR #7824, opened 2018-05-22
Don't show a contact search result for someone you have a conversation with.

greyson-signal:
Right now, if you search for a contact "Susan" that you also have a conversation with, then she will show up as both a conversation result and a contact result. This change prevents showing these duplicate results by excluding them from the contact results.

I also found a bug in SearchViewModel with our debouncer. Because the debouncer delay could fire off a query after the SearchFragment was destroyed, it could leave around an unclosed cursor. This is prevented by simply clearing the debouncer when the ViewModel is cleared.


london52:
Thank you

PR #7892, opened 2018-06-11
Show a banner in the event of a service outage.

greyson-signal:
We will now determine if there has been a service outage and render a banner at the top of the conversation list if we detect that there has been one.

To avoid hammering the status endpoint with requests, we only check for outages after a push message has failed to send. If an outage has been detected, we throttle subsequent outage requests for the same reason.

Also, given that making a network request is a potentially long operation, it was no longer suitable to run the reminder `isEligible()` checks in an anonymous AsyncTask, as that could result in leaking the ConversationListActivity. As a result, I took the opportunity to start a ViewModel class for the ConversationList and moved the reminder code in there.

**To-do**
- [ ] Get final copy for the banner.

**Test Devices**


src/org/thoughtcrime/securesms/jobs/PushSendJob.java
```diff
@@ -68,7 +68,13 @@ protected final void onSend(MasterSecret masterSecret) throws Exception {
       throw new TextSecureExpiredException("Too many signed prekey rotation failures");
     }
 
-    onPushSend();
+    try {
+      onPushSend();
+      TextSecurePreferences.setNeedsOutageCheck(context, false);
+    } catch (Exception e) {
```
moxie0: In general, please try to avoid catching and rethrowing unwrapped exceptions. You can set this value in onCancelled() rather than catching anything here.

src/org/thoughtcrime/securesms/components/reminder/ServiceOutageReminder.java
```diff
@@ -0,0 +1,76 @@
...
+
+    long timeSinceLastCheck = System.currentTimeMillis() - TextSecurePreferences.getLastOutageCheckTime(context);
+    if (timeSinceLastCheck < OUTAGE_CHECK_INTERVAL) {
+      return true;
+    }
+
+    TextSecurePreferences.setLastOutageCheckTime(context, System.currentTimeMillis());
+
+    try {
+      InetAddress address = InetAddress.getByName(BuildConfig.SIGNAL_SERVICE_STATUS_URL);
```
moxie0: Doesn't feel to me like we should be doing network I/O in this class. Would it not be more appropriate to leave the reminder hierarchy as-is, and then simply kick off a job to do the lookup and post an event in onResume() for ConversationListActivity and ConversationActivity?

greyson-signal: Yeah that sounds better to me.

src/org/thoughtcrime/securesms/jobs/ServiceOutageDetectionJob.java
```diff
@@ -0,0 +1,73 @@
...
+      if (IP_SUCCESS.equals(address.getHostAddress())) {
+        Log.w(TAG, "Service is available.");
+        TextSecurePreferences.setServiceOutage(context, false);
+      } else if (IP_FAILURE.equals(address.getHostAddress())) {
+        Log.w(TAG, "Service is down.");
+        TextSecurePreferences.setServiceOutage(context, true);
+      } else {
+        Log.w(TAG, "Service status check returned an unrecognized IP address. Assuming outage.");
+        TextSecurePreferences.setServiceOutage(context, true);
+      }
```
moxie0: Don't we want to post an event here if this is being toggled, to cause the activity to re-render with or without the reminder header?

greyson-signal: So we actually post the event inside of `TextSecurePreferences#setServiceOutage`, similar to how `TextSecurePreferences#setNeedsSqlCipherMigration` does it.

greyson-signal: Moved the event posting to here.

moxie0:
lgtm

PR #7894, opened 2018-06-12
Upgrade SQLCipher to respect a max window size.

greyson-signal:
Previously, SQLCipher's memory usage would grow indefinitely, up until it hit the end of the cursor. This leaves us open to OOM errors on large datasets. We've now switched to a release where the
memory used by the cursor can be bounded.

Performance seems largely unaffected. Basic query times appear to be the same on my 2014 Moto E. The main difference is that because there's a memory ceiling on a cursor, if it hits that memory ceiling, it has to requery and refill the cursor window. 

On a conversation with 18,500 messages (again, on my 2014 Moto E), this takes just under 1 second. Because the cursor is refilled on the main thread, that means the UI has a very noticeable "hiccup". However, I only experienced this hiccup when I set the memory ceiling very low (128 KB). At it's current value of 4MB, this hiccup should only be hit every 5000 messages or so (all depending on the message size, of course). On my 2016 Pixel, this hiccup is only ~250ms.

That said, while I think the hiccup is an acceptable compromise to avoid OOM errors on large conversations, it'd still be a good idea in the future to make some architectural changes to avoid the main-thread-requery pattern that is common with cursors.

**Test Devices**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -125,7 +125,7 @@ dependencies {
     }
     compile 'com.codewaves.stickyheadergrid:stickyheadergrid:0.9.4'
     compile 'com.github.dmytrodanylyk.circular-progress-button:library:1.1.3-S2'
-    compile 'org.signal:android-database-sqlcipher:3.5.9-S1'
+    compile 'org.signal:android-database-sqlcipher:3.5.9-S2'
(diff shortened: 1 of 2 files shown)


moxie0:
lgtm

PR #7916, opened 2018-06-20
Update scheduling of ServiceOutageDetectionJob.

greyson-signal:
Previously, we were running this job in PushSendJob#onCanceled(). However, with the new retry logic, this won't happen for 24 hours.

Instead, we now schedule the job in PushSendJob#onRetry().

It's worth noting that ServiceOutageDetectionJob throttles itself to be once every 60s and is non-persistent.

**Test Devices**


moxie0:
lgtm

Issue #7808, opened 2018-05-17
New group members have a diverged view on the expiration timer


Trolldemorted:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
"When a new user is added to a group for which a disappearing messages timer has been set, the first message the new users sends also resets the timer ("NewUser set disappearing messages time to Off")."

Legacy issue: https://github.com/signalapp/Signal-Android/issues/6421

In https://github.com/signalapp/libsignal-service-java/commit/c8d7c3c00445a81b81e0a7305151cda4534ba299 the multidevice group and contact updates received an expiration timer field, and it is filled correctly.

**MultiDeviceGroupUpdateJob.java**
```java
Recipient         recipient       = Recipient.from(context, Address.fromSerialized(GroupUtil.getEncodedId(record.getId(), record.isMms())), false);
Optional<Integer> expirationTimer = recipient.getExpireMessages() > 0 ? Optional.of(recipient.getExpireMessages()) : Optional.absent();

out.write(new DeviceGroup(record.getId(), Optional.fromNullable(record.getTitle()),
                          members, getAvatar(record.getAvatar()),
                          record.isActive(), expirationTimer));
```

However, mundane group updates have no such field, and the expiry timer is not applied to the group message itself.

**PushGroupUpdateJob.java**
```java
SignalServiceGroup groupContext = SignalServiceGroup.newBuilder(Type.UPDATE)
                                                        .withAvatar(avatar)
                                                        .withId(groupId)
                                                        .withMembers(members)
                                                        .withName(record.get().getTitle())
                                                        .build();

SignalServiceDataMessage message = SignalServiceDataMessage.newBuilder()
                                                           .asGroupMessage(groupContext)
                                                           .withTimestamp(System.currentTimeMillis())
                                                           .build();
```

You could either add a field to the group update, or set the timer for the update message itself.
cc @scottnonnenberg-signal @michaelkirk-signal @charlesmchen-signal I assume your clients are affected as well.

### Steps to reproduce
- set timer
- add new member to group
- write a message from new member's device

**Actual result:** The new member's devices reset the disappearing messages timer
**Expected result:** The group update should contain the new timer, and the new devices should honor it

PR #7921, opened 2018-06-21
Send/receive the expiring message timer for group updates.

greyson-signal:
We will now check for updates to the expiring message time when receiving a group update event, as well as send the expiring message time when sending a group update event.

The primary use case of doing this is so when a new person is added to a group, they will also be told the expiring message timer. Previously, someone who was just added to the group wouldn't get the expiring message time until an actual message was sent.

**Test Devices**

### DIFF
--- a/src/org/thoughtcrime/securesms/jobs/PushDecryptJob.java
+++ b/src/org/thoughtcrime/securesms/jobs/PushDecryptJob.java
@@ -380,8 +380,13 @@ public class PushDecryptJob extends ContextJob {
   private void handleGroupMessage(@NonNull SignalServiceEnvelope envelope,
                                   @NonNull SignalServiceDataMessage message,
                                   @NonNull Optional<Long> smsMessageId)
+      throws MmsException
   {
     GroupMessageProcessor.process(context, envelope, message, false);
+
+    if (message.getExpiresInSeconds() != 0 && message.getExpiresInSeconds() != getMessageDestination(envelope, message).getExpireMessages()) {
+      handleExpirationUpdate(envelope, message, Optional.absent());
+    }
 
     if (smsMessageId.isPresent()) {
       DatabaseFactory.getSmsDatabase(context).deleteMessage(smsMessageId.get());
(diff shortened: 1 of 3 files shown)


Trolldemorted:
I assume this will fix https://github.com/signalapp/Signal-Android/issues/7808 when Signal-iOS does the same?

greyson-signal:
@Trolldemorted Ah yes, good catch :) iOS is doing the same.

src/org/thoughtcrime/securesms/jobs/PushDecryptJob.java
```diff
@@ -380,9 +380,14 @@ private long handleSynchronizeSentEndSessionMessage(@NonNull SentTranscriptMessa
   private void handleGroupMessage(@NonNull SignalServiceEnvelope envelope,
                                   @NonNull SignalServiceDataMessage message,
                                   @NonNull Optional<Long> smsMessageId)
+      throws MmsException
   {
     GroupMessageProcessor.process(context, envelope, message, false);
 
+    if (message.getExpiresInSeconds() != getMessageDestination(envelope, message).getExpireMessages()) {
```
moxie0: and not 0

Trolldemorted:
@greyson-signal this PR only addresses the group updates which are automatically sent in response to explicit info requests. Do you want to add the expiration timer [to normal group updates](https://github.com/signalapp/Signal-Android/blob/master/src/org/thoughtcrime/securesms/jobs/PushGroupSendJob.java#L171) as well?

greyson-signal:
@Trolldemorted You are literally linking to a line I changed in this PR :)

Trolldemorted:
@greyson-signal woops, I confused the classes and got it the other way round. You got it right for mundane group updates, but not yet for requested ones I think: https://github.com/signalapp/Signal-Android/blob/4c2269175b6477c42f8aa6ec5cd4a376cc27a0f9/src/org/thoughtcrime/securesms/jobs/PushGroupUpdateJob.java#L91

PR #7926, opened 2018-06-22
Fix backup/import issue with expiring messages.

greyson-signal:
There was an issue where we were backing up group receipts and attachments that were for expiring messages (which are already excluded from the backup).

This PR excludes these items from the backup, and for backups made before this change, this PR also deletes these invalid entries at the end of the restore process.

**Test Devices**


greyson-signal:
Merged into master. 71a34dac5ff05bd22d77946afec1f4b9ee8f0a90

PR #7965, opened 2018-07-04
Filter out inactive groups during contact selection

mruddy:

### Contributor checklist

 * Nexus 5X, Android 8.1.0

----------

### Description

When "Leave group" is selected, the group is marked as not active in the local group database. However, after leaving a group, the inactive group still appears in the filtered contact list search results that appear when attempting to compose a new message. This is confusing to a number of fellow users that I have interacted with via groups. Once a group is left, a user must have another user of the group re-add them before the user that left the group can again send a message to the group. Thus the most intuitive action is to filter the results from possible contacts for new message compositions. That is what this PR does.

This effectively allows "deleting" a group (from the user's high-level perspective) by first leaving it, and then deleting all threads associated with the left group. The user will not see inactive groups while attempting to create new messages, unless re-invited (due to some "recents" processing, the inactive group still hangs around for sharing/invite actions, but these are out of scope for what I am interested in).

Technical note: The reason why I made the change like this is because when I originally made the change like the following diff, the conversation search (aka thread search filter) stopped working as I expected it too (basically that filter would not match existing threads/conversations). The reason is because `ContactAccessor.getNumbersForThreadSearchFilter` depends on `GroupDatabase.getGroupsFilteredByTitle` too.

```
diff --git a/src/org/thoughtcrime/securesms/database/GroupDatabase.java b/src/org/thoughtcrime/securesms/database/GroupDatabase.java
index 490d3ee0..ad7fcca8 100644
--- a/src/org/thoughtcrime/securesms/database/GroupDatabase.java
+++ b/src/org/thoughtcrime/securesms/database/GroupDatabase.java
@@ -105,8 +105,8 @@ public class GroupDatabase extends Database {
 
   public Reader getGroupsFilteredByTitle(String constraint) {
     @SuppressLint("Recycle")
-    Cursor cursor = databaseHelper.getReadableDatabase().query(TABLE_NAME, null, TITLE + " LIKE ?",
-                                                                                        new String[]{"%" + constraint + "%"},
+    Cursor cursor = databaseHelper.getReadableDatabase().query(TABLE_NAME, null, TITLE + " LIKE ? AND " + ACTIVE + " = ?",
+                                                                                        new String[]{"%" + constraint + "%", "1"},
                                                                                         null, null, null);
```

### DIFF
--- a/src/org/thoughtcrime/securesms/contacts/ContactsCursorLoader.java
+++ b/src/org/thoughtcrime/securesms/contacts/ContactsCursorLoader.java
@@ -204,11 +204,13 @@ public class ContactsCursorLoader extends CursorLoader {
     try (GroupDatabase.Reader reader = DatabaseFactory.getGroupDatabase(getContext()).getGroupsFilteredByTitle(filter)) {
       GroupDatabase.GroupRecord groupRecord;
       while ((groupRecord = reader.getNext()) != null) {
-        groupContacts.addRow(new Object[] { groupRecord.getTitle(),
-                                            groupRecord.getEncodedId(),
-                                            ContactsContract.CommonDataKinds.Phone.TYPE_CUSTOM,
-                                            "",
-                                            ContactsDatabase.NORMAL_TYPE });
+        if (groupRecord.isActive()) {
+          groupContacts.addRow(new Object[] { groupRecord.getTitle(),
+                                              groupRecord.getEncodedId(),
+                                              ContactsContract.CommonDataKinds.Phone.TYPE_CUSTOM,
+                                              "",
+                                              ContactsDatabase.NORMAL_TYPE });
+        }
       }
     }
     return groupContacts;


mruddy:
As example motivation for this change, consider the case where a user, Alice, has an active group named "Work" for interacting with co-workers at her job on job related tasks. Now, Alice quits her job and leaves the group "Work". Alice starts another job and is added to a group named "Work" at that job with a different set of signal users. Without this change, both the new active "Work" group and the old inactive one would display in the contact selection screen for Alice. With this change, Alice would only see her active "Work" group. The old "Work" group would only display again if it became active by someone at the old job re-adding her to the group. This is just one of the side-effects of using logical group deletion + no unique constraint on group name instead of requiring group names to be unique (but this is a separate subjective matter outof scope for this PR).
In summary, by only displaying the active groups for contact selection, we are more closely aligning with user expectations in an environment of possibly duplicating/overlapping group names.

mruddy:
do i need to do something to get this merged?

mruddy:
closing because this has been ignored long enough that i no longer feel like contributing it

PR #7973, opened 2018-07-07
Throttle background contact syncs to once every 6 hours.

greyson-signal:
Unfortunately, there's apps out there that trigger contact changes very frequently. Because we listen to the system for contact changes to tell us when to sync, that could result in us sending an abundance of contact syncs to linked desktop instances.

This throttles these background syncs to happen no more than once every 6 hours. If a sync was throttled in the background, a dirty flag is set, and the sync request will be made the next time the app opens.

**Test Devices**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,8 +66,9 @@ dependencies {
     compile 'com.android.support:preference-v14:27.0.2'
     compile 'com.android.support:gridlayout-v7:27.0.2'
     compile 'com.android.support:multidex:1.0.2'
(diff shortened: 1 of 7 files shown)


tuxayo:
Will adding a new contact and then opening Signal to send a message work seamlessly?

I think I **already** had to wait before seeing a contact in Signal and thought that it was a bug.

greyson-signal:
@tuxayo I've updated the PR to execute a contact sync when the app opens if any were skipped while the app was backgrounded.

PR #7981, opened 2018-07-12
Visual tweaks

greyson-signal:
The first large batch of changes to improve Signal's look.

The most interesting java change is likely `ConversationItem`, which renders the message items. There's a lot of work there around new rendering rules that take into consideration neighboring messages to collapse footers, round corners, show/hide the author, etc.

It's also worth noting that because the position of the timestamp information changes based on message type (solo-images and shared contacts render the timestamp in a different location), the footer has been broken out to `ConversationItemFooter` so it can be re-used in different areas.

Also, because the way we show images has changed a bit, I've wrapped `ThumbnailView` into a class `ConversationItemThumbnail` so we can render the `ConversationItemFooter` and timestamp background shade without making `ThumbnailView` overly-specific.

**Test Devices**
* [Galaxy S3 Mini, Android 4.2.2. API 17](https://www.gsmarena.com/samsung_i8200_galaxy_s_iii_mini_ve-6190.php)


duchampdev:
Really really nice! Thanks for the great redesign!

PR #8018, opened 2018-07-21
UX improvements: Unlocking Signal with fingerprint

art1fa:
### Contributor checklist

 * Samsung S8, Android 8.0
 * Emulated Pixel 2, Android 8.0    

----------

### Description
- Add Fingerprint recognition status text
- Action bar: Fix signal icon size + color of settings icon
- Change fingerprint success / failed colors [according to Material Design guidelines](https://material.io/design/platform-guidance/android-fingerprint.html#authentication); plus, change icon accordingly 
- Do not reset fingerprint icon to blue background on success
- Fix behavior, where fingerprint failed animation starts, when e.g. switching to recent apps (Authentification error)
- Change activity background to primary signal color (second commit; light theme only)

### Before:
Light theme
fingerprint resized

Dark theme
screenshot_1532179442 resized

Fingerprint recognition failed
screenrecord

Fingerprint recognition success
screenrecord

### After:
Light theme
screenshot_1532177854 resized
Dark theme
screenshot_1532179770 resized
Fingerprint recognition failed, then success
screenrecord


Dyras:
Just a quick question: Sometimes when I open the app I get the "Fingerprint failed" shaking animation without having touched the fingerprint scanner. Will this fix that?

art1fa:
Yes this will be fixed.

src/org/thoughtcrime/securesms/PassphrasePromptActivity.java
```diff
@@ -340,36 +345,35 @@ protected void cleanup() {
     @Override
     public void onAuthenticationError(int errMsgId, CharSequence errString) {
       Log.w(TAG, "Authentication error: " + errMsgId + " " + errString);
-      onAuthenticationFailed();
```
greyson-signal: Hmm, why did you choose to remove this?

art1fa: If there's an authentication error, there is no need to call the `onAuthenticationFailed` method...
because the latter is 
> called when a fingerprint is valid but not recognized. 

[(see docs)](https://developer.android.com/reference/android/hardware/fingerprint/FingerprintManager.AuthenticationCallback.html#onAuthenticationError(int,%20java.lang.CharSequence))

This is also the cause of the issue @Dyras pointed out...if you e.g. open Recent Apps, the authentication is aborted (AuthenticationError), but there is no failed recognition, and so there should be no animation / color change of the fingerprint icon.

greyson-signal:
Merged into 4.25 in d34f7fa3. Thank you!

greyson-signal:
This caused #8106, so I ended up reverting it. I do like the new icons and stuff though, so I'd like to get some of these improvements back in, just with more testing around the different types of locking mechanisms.

Issues I saw:
- We don't hide the prompt and status text when not using a fingerprint lock
- The new fixed height of the view makes it look silly when hiding the various prompts

art1fa:
Weird, I also tested on Android 6 (Emulated Nexus 5X) and I got the system UI for typing the password, which is also the same on my Samsung S8.

screen_android resized

What did I miss? In what cases is the built-in password input shown?
For Android earlier than Lollipop, no screen lock is applied (though activated), the logs say "Not supported".

greyson-signal:
@art1fa On multiple devices I've tested (Nexus 5X and Moto E), after I make it past that system prompt, we still show the screen lock screen for a short period (a little under a second, I'd say), and during that time you can see it references fingerprints. Also, because the layout was set to a fixed height, if you just hide the fingerprint text, it doesn't look good. So basically another pass just needs to be done to account for these different scenarios.

One option would be to just get in the new icons first. Then maybe we can add the descriptive text just for fingerprints without fixing the layout height?

art1fa:
With 5a4e7e5 the layout issues are resolved.
What I changed: 
- Put all fingerprint layout items in a LinearLayout container, which will be hidden if not no fingerprint available
- Small refactorings: `fingerPrintPrompt` -> `fingerPrintIcon` (because it's just the ImageView), and `fingerprintAuthContainer` is the View encapsulating all fingerprint related layout elements
- Polished "Tap to unlock" button

screenshot_20180815-181358_signal resized

Meteor0id:
Why not add an additional screen:
- When the user touches the screen at the location of the fingerprint logo, show a notification explaining that he/she needs to touch the fingerprint scanner instead of the touchscreen.

To you, me , and anyone who has already used this feature; this is obvious. And with this improved UI it clearly states to touch the touchsensor. Yet there are always people who pick up on such things a bit slower than most of us, and there is no reason not to help them along. And sometimes the "touchsensor" is not immidiately understood due to language differences.

"(!) Please touch the fingerprint scanner, instead of the screen."

Some users may get this warning by accidentally brushing the screen, to them it should not get in the way of quicly unlocking the device, so they should not have to wait for the message to disapear again.
There is no need for an OK button, this message disapreas on its own after a few seconds, and reapears any time the user attempts to unlock by thouching the screen. Don't leave it on the screen, but only show it when the user actually makes this mistake, that way he/she realises this notification is a direct response to their action.

#GrandmaProofDesign

PR #8031, opened 2018-07-24
Increase call control button sizes

art1fa:
### Contributor checklist

 * Samsung S8, Android 8.0
 * Emulated Pixel 2, Android 8.0

----------

### Description
- Increase the call control button paddings. Now the buttons match the recommended size (48dp), look more familiar and are easier to tap.

#### Before:
call_controls resized

#### After:
call_controls_new resized

### DIFF
--- a/res/drawable/webrtc_bluetooth_button.xml
+++ b/res/drawable/webrtc_bluetooth_button.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <layer-list xmlns:android="http://schemas.android.com/apk/res/android">
     <item android:drawable="@drawable/webrtc_control_background"/>
-    <item android:top="5dp"
-          android:left="5dp"
-          android:right="5dp"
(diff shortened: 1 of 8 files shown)


greyson-signal:
I think this looks nice during audio calls, but during video calls, that row of icons moves up to the top of the screen. And then the large icons just take away space from the video call :( 

One solution would be to animate the size of the icons in between the two different states (audio and full-screen video). That way we could get larger icons during audio calls without losing any space during video calls compared to today.

ParkerK:
Would it be possible to add a way to hide the buttons all together during a video call if a user wants? Maybe hide them after a few seconds or on click, and then re-show them when the screen is tapped (similar to other video chat's implementations)

greyson-signal:
@ParkerK Yeah I'd like to take another look at this screen and make some larger changes. However, in terms of this PR, we should keep the changes small. I like having the bigger buttons in the voice-call screen, we just need to downsize them in a video call.

PR #8039, opened 2018-07-26
Added new logger.

greyson-signal:
We have a problem where our debug logs don't go back very far. Right now, we rely on locally polling logcat. If we want to save more logs, we'll have to save them ourselves.

To do so, I've created a new logging layer called that mimics the interface of `Log`. You can add different `Loggers` to the new system. To start, I've added two Loggers: `AndroidLogger` and `PersistentLogger`. AndroidLogger just forwards everything to the standard Log command, while PersistentLogger persists logs on disk.

PersistentLogger works as follows:
- All work is done on a dedicated single-thread executor to prevent slowing down other parts of the app.
- Logs are written to a file until it hits a certain line length, at which point that file will be closed and a new one is started.
- When a new log is started, we'll delete any log file that is over a specified limit (e.g. right now we keep no more than 5 log files).
- A new global UncaughtExceptionHandler has been added to log crashes.
- Log files are stored in the cache directory in order to allow the system to clean them up as needed.
- During log submission, we'll include both the logcat logs and these new logs until we can verify their stability, at which point we may be able to remove logcat logs.

This PR just introduces the logger and not any usages (to prevent diff noise).

**Test Devices**
- [Moto E (2nd Gen), Android 5.1, API 22](https://www.gsmarena.com/motorola_moto_e_(2nd_gen)-6986.php)
- [Nexus 5X, Android 8.1, API 27](https://www.gsmarena.com/lg_nexus_5x-7556.php)

### DIFF
--- a/src/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/src/org/thoughtcrime/securesms/ApplicationContext.java
@@ -43,6 +43,10 @@ import org.thoughtcrime.securesms.jobs.MultiDeviceContactUpdateJob;
 import org.thoughtcrime.securesms.jobs.requirements.MasterSecretRequirementProvider;
 import org.thoughtcrime.securesms.jobs.requirements.ServiceRequirementProvider;
 import org.thoughtcrime.securesms.jobs.requirements.SqlCipherMigrationRequirementProvider;
+import org.thoughtcrime.securesms.logging.AndroidLogger;
+import org.thoughtcrime.securesms.logging.CustomSignalProtocolLogger;
+import org.thoughtcrime.securesms.logging.PersistentLogger;
+import org.thoughtcrime.securesms.logging.UncaughtExceptionLogger;
 import org.thoughtcrime.securesms.push.SignalServiceNetworkAccess;
 import org.thoughtcrime.securesms.service.DirectoryRefreshListener;
 import org.thoughtcrime.securesms.service.ExpiringMessageManager;
@@ -55,7 +59,6 @@ import org.webrtc.PeerConnectionFactory.InitializationOptions;
 import org.webrtc.voiceengine.WebRtcAudioManager;
 import org.webrtc.voiceengine.WebRtcAudioUtils;
 import org.whispersystems.libsignal.logging.SignalProtocolLoggerProvider;
-import org.whispersystems.libsignal.util.AndroidSignalProtocolLogger;
 
 import java.util.HashSet;
 import java.util.Set;
@@ -73,11 +76,12 @@ import dagger.ObjectGraph;
  */
 public class ApplicationContext extends MultiDexApplication implements DependencyInjector, DefaultLifecycleObserver {
 
-  private static final String TAG = ApplicationContext.class.getName();
+  private static final String TAG = ApplicationContext.class.getSimpleName();
 
   private ExpiringMessageManager expiringMessageManager;
   private JobManager             jobManager;
   private ObjectGraph            objectGraph;
+  private PersistentLogger       persistentLogger;
 
   private volatile boolean isAppVisible;
 
@@ -90,6 +94,7 @@ public class ApplicationContext extends MultiDexApplication implements Dependenc
     super.onCreate();
     initializeRandomNumberFix();
     initializeLogging();
+    initializeCrashHandling();
     initializeDependencyInjection();
     initializeJobManager();
     initializeExpiringMessageManager();
@@ -105,7 +110,6 @@ public class ApplicationContext extends MultiDexApplication implements Dependenc
   public void onStart(@NonNull LifecycleOwner owner) {
     isAppVisible = true;
     Log.i(TAG, "App is now visible.");
-
     executePendingContactSync();
   }
 
@@ -134,12 +138,24 @@ public class ApplicationContext extends MultiDexApplication implements Dependenc
     return isAppVisible;
   }
 
+  public PersistentLogger getPersistentLogger() {
+    return persistentLogger;
+  }
+
   private void initializeRandomNumberFix() {
     PRNGFixes.apply();
   }
 
   private void initializeLogging() {
-    SignalProtocolLoggerProvider.setProvider(new AndroidSignalProtocolLogger());
+    persistentLogger = new PersistentLogger(this);
+    org.thoughtcrime.securesms.logging.Log.initialize(new AndroidLogger(), persistentLogger);
+
+    SignalProtocolLoggerProvider.setProvider(new CustomSignalProtocolLogger());
+  }
+
+  private void initializeCrashHandling() {
+    final Thread.UncaughtExceptionHandler originalHandler = Thread.getDefaultUncaughtExceptionHandler();
+    Thread.setDefaultUncaughtExceptionHandler(new UncaughtExceptionLogger(originalHandler, persistentLogger));
   }
 
   private void initializeJobManager() {
--- a/src/org/thoughtcrime/securesms/logging/LogFile.java
+++ b/src/org/thoughtcrime/securesms/logging/LogFile.java
@@ -1,0 +1,139 @@
+package org.thoughtcrime.securesms.logging;
+
+import android.support.annotation.NonNull;
+
+import org.thoughtcrime.securesms.util.Conversions;
+import org.thoughtcrime.securesms.util.Util;
+
+import java.io.BufferedInputStream;
+import java.io.BufferedOutputStream;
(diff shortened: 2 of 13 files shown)


src/org/thoughtcrime/securesms/logging/Bugle.java
```diff
@@ -0,0 +1,132 @@
...
+        loggers[i].i(tag, message, t);
+      }
+    } else {
+      Log.i(tag, message, t);
+    }
+  }
+
+  public static void w(String tag, String message, Throwable t) {
+    if (loggers != null) {
+      for (int i = 0; i < loggers.length; i++) {
```
moxie0: for each?

src/org/thoughtcrime/securesms/logging/PersistentLogger.java
```diff
@@ -0,0 +1,301 @@
...
+      try {
+        writer.flush();
+      } catch (IOException e) {
+        Log.w(TAG, "Failed to flush file.", e);
+      }
+    });
+  }
+
+  @WorkerThread
+  private void initializeWriter() {
```
moxie0: From what I can tell, this means that every time the app starts, a new file is created, all the data from a previous file is copied to the new one, and they're both left there? Why not just use the existing file?

src/org/thoughtcrime/securesms/logging/PersistentLogger.java
```diff
@@ -0,0 +1,301 @@
...
+      }
+    });
+  }
+
+  private Writer openNewWriter() throws IOException {
+    shiftLogFiles();
+    trimLogFilesOverMax();
+
+    File file = getUniqueFile();
+    TextSecurePreferences.setLogFilePath(context, 0, file.getPath());
```
moxie0: This is a private method with 4 pretty substantial side effects

src/org/thoughtcrime/securesms/logging/PersistentLogger.java
```diff
@@ -0,0 +1,301 @@
...
+  private final Context          context;
+  private final Executor         executor;
+  private final AttachmentSecret attachmentSecret;
+
+  private Writer writer;
+  private int    lineCount;
+
+  public PersistentLogger(Context context) {
+    this.context          = context.getApplicationContext();
+    this.attachmentSecret = AttachmentSecretProvider.getInstance(context).getOrCreateAttachmentSecret();
```
moxie0: I don't think we should reuse this secret here

src/org/thoughtcrime/securesms/logging/PersistentLogger.java
```diff
@@ -0,0 +1,301 @@
...
+        }
+      }
+
+      future.set(new SequenceInputStream(Collections.enumeration(streams)));
+    });
+
+    return future;
+  }
+
+  public void forceFlush() {
```
moxie0: It looks like the place you're using this wants it to be a blocking operation?

src/org/thoughtcrime/securesms/logging/PersistentLogger.java
```diff
@@ -0,0 +1,301 @@
...
+        if (lineCount >= MAX_LOG_LINES) {
+          Util.close(writer);
+          writer    = openNewWriter();
+          lineCount = 0;
+        }
+
+        String entry = buildLogEntry(level, tag, message, t);
+        writer.write(entry);
+        lineCount += getLineCount(entry);
+
```
moxie0: If you're using a buffered output stream, you should probably always flush here, otherwise you're opening yourself up to data loss, right?

src/org/thoughtcrime/securesms/logging/Bugle.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+
+  public static abstract class Logger {
+    public abstract void v(String tag, String message, Throwable t);
+    public abstract void d(String tag, String message, Throwable t);
+    public abstract void i(String tag, String message, Throwable t);
+    public abstract void w(String tag, String message, Throwable t);
+    public abstract void e(String tag, String message, Throwable t);
+    public abstract void wtf(String tag, String message, Throwable t);
```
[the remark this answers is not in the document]
greyson-signal: I use levels besides w(). I try to use them for what they are. Is there a reason you only use w()?

src/org/thoughtcrime/securesms/logging/PersistentLogger.java
greyson-signal: I've gone back and forth on it. I suppose I can flush here -- we still get the benefit of a buffer when writing long log entries. We just wouldn't get the benefit of buffering multiple lines. It does simplify other areas of the codebase though, so maybe we can flush after every line and see if there's any negative side effects.

src/org/thoughtcrime/securesms/logging/PersistentLogger.java
greyson-signal: This line is removed. Don't know why Github is still showing it.

src/org/thoughtcrime/securesms/logging/PersistentLogger.java
greyson-signal: I have to read the old file to get the line count. It's a bug that the previous file wasn't being deleted. I'll take another look at this flow.

src/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -134,12 +140,25 @@ public boolean isAppVisible() {
...
+  public PersistentLogger getPersistentLogger() {
+    return persistentLogger;
+  }
+
   private void initializeRandomNumberFix() {
     PRNGFixes.apply();
   }
 
   private void initializeLogging() {
+    // TODO(greyson): Remember to update this to use the new logger
```
moxie0: why wait?

src/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -220,7 +239,7 @@ private void initializeWebRtc() {
                                                             .setEnableVideoHwAcceleration(true)
                                                             .createInitializationOptions());
     } catch (UnsatisfiedLinkError e) {
-      Log.w(TAG, e);
+      android.util.Log.w(TAG, e);
```
moxie0: why util?

src/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -233,7 +252,7 @@ protected Void doInBackground(Void... params) {
           try {
             ProviderInstaller.installIfNeeded(ApplicationContext.this);
           } catch (Throwable t) {
-            Log.w(TAG, t);
+            android.util.Log.w(TAG, t);
```
moxie0: why util?

src/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -104,15 +110,15 @@ public void onCreate() {
   @Override
   public void onStart(@NonNull LifecycleOwner owner) {
     isAppVisible = true;
-    Log.i(TAG, "App is now visible.");
+    android.util.Log.i(TAG, "App is now visible.");
```
moxie0: why util?

src/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -104,15 +110,15 @@ public void onCreate() {
...
 
     executePendingContactSync();
   }
 
   @Override
   public void onStop(@NonNull LifecycleOwner owner) {
     isAppVisible = false;
-    Log.i(TAG, "App is no longer visible.");
+    android.util.Log.i(TAG, "App is no longer visible.");
```
moxie0: why util?

src/org/thoughtcrime/securesms/logging/LogFile.java
```diff
@@ -0,0 +1,146 @@
...
+    void writeEntry(@NonNull String entry) throws IOException {
+      byte[] iv = new byte[16];
+      new SecureRandom().nextBytes(iv);
+
+      byte[] plaintext = entry.getBytes();
+      try {
+        cipher.init(Cipher.ENCRYPT_MODE, new SecretKeySpec(secret, "AES"), new IvParameterSpec(iv));
+
+        byte[] ciphertext1 = cipher.update(plaintext);
+        byte[] ciphertext2 = cipher.doFinal();
```
moxie0: byte[] ciphertext = cipher.doFinal(plaintext);

PR #8041, opened 2018-07-26
Updated License Date

shaunleejx:
Updated the year of the license to the current year.


nrizzio:
I'm pretty sure the copyright year in that context refers to the year the license was written, not the year that Signal's code was written.

bitmold:
@nrizzio  is right, it says in the license document that people using the license are not permitted to change it.

>  Everyone is permitted to copy and distribute verbatim copies
>  of this license document, but changing it is not allowed.

Issue #8058, opened 2018-07-31
UI: Rich Text Can be Pasted into a Signal Message 


bitmold:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Pasting rich text, such as HTML markup from a website into a Signal message will carry over the text's original formatting. In some cases, it looks silly, and in others, such as with enormous text, it makes the text field unusable (or at least undesirable)

Since Signal can't send texts of different sizes, colors, boldness, etc. we should sanitize text that gets pasted into the app. 

I would like to note that actually sending the message will strip the formatting and not cause any actual problems. This mostly just looks confusing to users and could potentially be very confusing depending on the nature of the formatted text pasted into Signal. 

### Steps to reproduce
- Navigate to a site with rich text, such as this one https://www.marxists.org/reference/archive/debord/society.htm
- Copy and paste some text. I chose: 

> 1.
> In societies where modern conditions of production prevail, all of life presents itself as an immense accumulation of spectacles

 since the "1." in that quote is formatted with the `<h5>` HTML element. 
- Open Signal and paste the text into a message.

**Actual result:** The external markup has found its way into Signal! 
**Expected result:** The markup should be stripped from the selected text as if the user were to have just typed the pasted quote by hand.

### Screenshots
Here's a screenshot of Signal after following the steps outlined above
![image](https://user-images.githubusercontent.com/22125581/43484873-9d5cc2dc-94dd-11e8-841c-0de55eb83fc1.png)
As you can see, the formatting, size, and alignment of the "1."'s `<h5>` element is preserved 

### Device info
**Device:** Alcatel_5044R
**Android version:** 7.0
**Signal version:** 4.24.5

bitmold:
Recreating a PR for this since my last one from 2018 didn't follow the style guidelines  used by this project

bitmold:
This is still a bug it's existed for years at one point I submitted a PR for it

bitmold:
# Literally still happens

alex-signal:
Hi! I see you opened a PR a while back that had a potential fix for this issue, but closed it. Was that due to inactivity or due to realizing something didn't quite work right?

alex-signal:
[...]

PR #8059, opened 2018-07-31
UI Fix: Prevents Rich Text From Being Pasted Into a Message At Composition Time 

bitmold:
### Contributor checklist

 * Alcatel_5044R, Android 7.0
- [oops i didn't know at the time of writing the commit message and I'm honestly not sure how to change this i will be better about it next time] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the `Fixes #1234` [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)

This fixes issue #8058 !

----------

### Description
This PR fixes the bug described in issue #8058 with regards to external rich text formatting being able to be pasted into a Signal message that the user is composing. 

The patch works by intercepting paste events via `onTextContextMenuItem(int)` in `EmojiEditText` and taking one of the following steps depending on the version of android: 
- If Android 6.0 or higher, simply set the paste event (`android.R.id.paste`) to be the newly added `android.R.id.pasteAsPlainText`
- If < 6.0, the clipped text is sanitized of all formatting and then pasted into the clipboard. The paste proceeds, and after the text is inserted into the `EmojiEditText` the original unsanitized `ClipData` is returned to the clipboard. I think that this behavior is preferable because a user might want that formatting to be pasted somewhere else and, more importantly, it would be nasty of us to silently modify the clipboard data when a user pastes. 
  * *To test this patch, I would hardcode both approaches to be used one at a time (since I do not actually have a device that is running anything below Marshmallow).*

### Screenshots 
Here's an image of a broken paste (taken from issue #8058) 
image

Here's an image of pasting the same selection after applying the code in this patch. If I were to have sent the unformatted message in the last screenshot, it would be arranged like how the text is pasted here.

image

### DIFF
--- a/src/org/thoughtcrime/securesms/components/emoji/EmojiEditText.java
+++ b/src/org/thoughtcrime/securesms/components/emoji/EmojiEditText.java
@@ -43,33 +43,28 @@ public class EmojiEditText extends AppCompatEditText {
     setSelection(start + emoji.length());
   }
 
-  /*
-  Paste events are watched here so that rich text is never inserted into an EmojiEditText
-   */
   @Override
   public boolean onTextContextMenuItem(int id) {
-    // we only care about the paste option
-    if (id != android.R.id.paste) return super.onTextContextMenuItem(id);
-    // the paste option was selected
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
-      // the system handles plain text pasting perfectly fine on M and above
-      return super.onTextContextMenuItem(android.R.id.pasteAsPlainText);
-    } else { // manual fallback for pre-M versions of Android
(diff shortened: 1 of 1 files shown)


duchampdev:
That's a really nice bug fix!
But I'd suggest to remove (some of) the comments in the source code in order to comply with the Code Style Guidelines.

bitmold:
@duchampdev that is a good suggestion, I think I can refactor some of the method lvl variable names to make the comments unnecessary, I'll get on this tonight, thank you very much for your feedback :+)

duchampdev:
@bitmold I applied your PR and tested it on an Android 5 virtual device - seems to be working fine :)

PR #8063, opened 2018-08-01
Add the ability to create conversation shortcuts.

greyson-signal:
Added the ability to create a shortcut to a conversation. A new option "Add to home screen" will appear in the three-dot menu of a conversation. Clicking it will prompt your launcher to add a shortcut to that conversation to your home screen.

Notable Changes:
- Due to restrictions on how addresses and threadId's need to be passed, a new "entrypoint" activity was created, `ShortcuteLauncherActivity`, which will do threadId lookups and otherwise format to be passed to the ConversationActivity. The backstack is still built as expected.

Special thanks to @leonardehrenfried for getting this started!

**Test Devices**
* [Moto E (2nd Gen), Android 5.1, API 22](https://www.gsmarena.com/motorola_moto_e_(2nd_gen)-6986.php) -- Google Now Launcher
* [Nexus 5X, Android 8.1, API 27](https://www.gsmarena.com/lg_nexus_5x-7556.php) -- Google Now Launcher, Action Launcher, Evie Launcher
* [Google Pixel 2, Android P Beta](https://www.gsmarena.com/google_pixel_2-8733.php) -- Pixel Launcher, Evie Launcher


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1683,6 +1742,22 @@ protected Void doInBackground(Long... params) {
     }.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR, threadId);
   }
 
+  private void fetchAndNotifyThreadIdIfAvailable(final Recipient recipientToCheck) {
+    SafeAsyncTask.execute(this, () -> {
+      return DatabaseFactory.getThreadDatabase(ConversationActivity.this).getThreadIdFor(recipientToCheck);
+    }, (threadId) -> {
+      if (threadId != ConversationActivity.this.threadId) {
```
moxie0: This activity is reused, so if this query is slow it'll load the wrong conversation into the fragment under some circumstances

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -774,6 +786,50 @@ private void handleViewMedia() {
...
+        }
+      }
+
+      if (icon == null) {
+          icon = IconCompat.createWithResource(this, recipient.isGroupRecipient() ? R.mipmap.ic_group_shortcut
+                                                                                  : R.mipmap.ic_person_shortcut);
+      }
+
+      return icon;
+    }, (icon) -> {
```
moxie0: Why do you want this to not run if the activity lifecycle has changed? If I select "add shortcut" and immediately go back to the conversation list before the background process completes, I would expect the shortcut to still get added.

greyson-signal: That's fair, in this case we can just use a regular AsyncTask.

isaackramer:
Was this ever implemented? It would be very useful, in my opinion.

leonardehrenfried:
Yes, I've been using it for over 2 years now.

PR #8067, opened 2018-08-02
Updated Polish translation

ppwozniak:
Updated Polish translations


moxie0:
Thanks @ppwozniak, but translations happen through Transifex. These strings will just be overwritten with what's pulled from there.

PR #8091, opened 2018-08-09
Switch to WorkManager.

greyson-signal:
In preparation for the pending `targetSdk 26` requirement, we're changing `JobScheduler` to use the new `WorkManager` to execute tasks. The existing JobScheduler interface has been kept as close as possible to the original to reduce the amount of code changes. here's a general outline of the code changes.

- Changed `Job` to subclass the new `Worker` class.
- Job now essentially does what `JobConsumer` used to do -- i.e. it runs an `onRun()` method and returns the proper value (`SUCCESS`, `FAILURE`, `RETRY`) based on whether or not certain exceptions are thrown. This let's us keep all subclasses of `Job` identical, save the next step:
- Because WorkManager requires all data be persisted via a `Data` object, two methods have been added to Job: `serialize()` and `initialize()`. These methods are responsible for saving and restoring instance state. So really, the largest changes in this diff come from having to do that for every job.
- WorkManager also requires that every `Worker` subclass have a no-arg constructor, so one has been added to every job.
- We migrate all outstanding persisted Jobs to run on the new JobManager after app upgrade. Our reliance of java serialization for persistence means that we unfortunately have to carry around all of our existing Requirement classes for the foreseeable future, even though we don't use them anymore (outside of migration).
- To prevent the use of any new Requirements (which could only actually work if they're manually persisted in the data bundle, which I've done for the existing ones), I've removed them from JobParameters, instead using named setters for the requirements we already have.
- Had to update our compileSdk to 28 to get WorkManager to work.
  - Also had to update all of our support library versions

**Test Devices**
* Android 9 Emulator under various network conditions


src/org/thoughtcrime/securesms/components/camera/QuickAttachmentDrawer.java
```diff
@@ -254,7 +254,7 @@ protected void onSizeChanged(int w, int h, int oldw, int oldh) {
   @Override
   protected boolean drawChild(@NonNull Canvas canvas, @NonNull View child, long drawingTime) {
     boolean result;
-    final int save = canvas.save(Canvas.CLIP_SAVE_FLAG);
```
greyson-signal: It appears that both this method and constant have been removed in SDK 28. I tested the attachment drawer on a couple phones after this change, and i haven't noticed any oddities.

src/org/thoughtcrime/securesms/jobmanager/JobParameters.java
```diff
@@ -30,46 +34,79 @@
...
   private final List<Requirement> requirements;
-  private final boolean           isPersistent;
+  private final boolean           requiresNetwork;
+  private final boolean           requiresMasterSecret;
+  private final boolean           requiresSqlCipher;
   private final int               retryCount;
   private final long              retryUntil;
   private final String            groupId;
-  private final boolean           wakeLock;
-  private final long              wakeLockTimeout;
```
greyson-signal: I remove some fields and stuff, but because we specified a serialVersionUID, this shouldn't affect the migration.

src/org/thoughtcrime/securesms/jobmanager/persistence/PersistentStorage.java
```diff
@@ -43,38 +41,18 @@
...
-
-    long id = databaseHelper.getWritableDatabase().insert(TABLE_NAME, null, contentValues);
-    job.setPersistentId(id);
   }
 
   public List<Job> getAllUnencrypted() {
     return getJobs(null, ENCRYPTED + " = 0");
   }
 
-  public List<Job> getAllEncrypted(EncryptionKeys keys) {
```
greyson-signal: I removed this because we never call `JobManager.setEncryptionKeys()` today, meaning we would never read these anyway.

src/org/thoughtcrime/securesms/jobmanager/requirements/NetworkBackoffRequirement.java
```diff
@@ -31,15 +31,6 @@ public boolean isPresent(@NonNull Job job) {
...
   public void onRetry(@NonNull Job job) {
-    Log.i(TAG, "onRetry()");
-
-    if (!(new NetworkRequirement(context).isPresent())) {
-      Log.i(TAG, "No network. Resetting run stats.");
-      job.resetRunStats();
-      return;
-    }
-
-    BackoffReceiver.setUniqueAlarm(context, NetworkBackoffRequirement.calculateNextRunTime(job));
```
greyson-signal: Need the class for migration purposes, but because it'll never run, I removed the bodies because they reference `Job` methods that no longer exist.

src/org/thoughtcrime/securesms/jobs/SmsSendJob.java
```diff
@@ -192,16 +211,16 @@ private SmsManager getSmsManagerFor(int subscriptionId) {
...
+                                                 .withMasterSecretRequirement()
                                                  .withRetryCount(15)
                                                  .withGroupId(name);
 
-    if (TextSecurePreferences.isWifiSmsEnabled(context)) {
-      builder.withRequirement(new NetworkOrServiceRequirement(context));
-    } else {
-      builder.withRequirement(new ServiceRequirement(context));
-    }
+    // TODO(greyson): Handle this appropriately
```
greyson-signal: We need to figure out the best approach for scheduling SMS/MMS messages given WorkManager doesn't let you schedule based on telephony availability.

src/org/thoughtcrime/securesms/jobmanager/JobParameters.java
greyson-signal: Also worth noting that WorkManager acquires all the WakeLock stuff for you, so no need to do it ourselves anymore.

src/org/thoughtcrime/securesms/jobmanager/JobManager.java
```diff
@@ -1,252 +1,104 @@
...
-     * after deserializing a persistent job.
-     *
-     * @param dependencyInjector The injector to use.
-     * @return The builder.
-     */
-    public Builder withDependencyInjector(DependencyInjector dependencyInjector) {
-      this.dependencyInjector = dependencyInjector;
-      return this;
+  private boolean getRequiresMasterSecret(Job job) {
+    if (job.getJobParameters() != null) {
```
moxie0: maybe assert not null on theses?

src/org/thoughtcrime/securesms/jobs/AttachmentDownloadJob.java
```diff
@@ -62,6 +70,23 @@ public AttachmentDownloadJob(Context context, long messageId, AttachmentId attac
     this.manual       = manual;
   }
 
+  @Override
+  protected void initialize(Data data) {
+    messageId    = data.getLong(KEY_MESSAGE_ID, -1);
+    partRowId    = data.getLong(KEY_PART_ROW_ID, -1);
+    partUniqueId = data.getLong(KEY_PAR_UNIQUE_ID, -1);
+    manual       = data.getBoolean(KEY_MANUAL, false);
```
moxie0: Can we assert that these exist rather than accepting defaults?

src/org/thoughtcrime/securesms/jobs/MmsDownloadJob.java
```diff
@@ -45,32 +43,54 @@
...
     this.messageId = messageId;
     this.threadId  = threadId;
     this.automatic = automatic;
   }
 
+  @Override
+  protected void initialize(Data data) {
+    messageId = data.getLong(KEY_MESSAGE_ID, -1);
+    threadId  = data.getLong(KEY_THREAD_ID, -1);
+    automatic = data.getBoolean(KEY_AUTOMATIC, false);
```
moxie0: Can we assert these exist rather than accepting defaults?

src/org/thoughtcrime/securesms/jobs/MmsSendJob.java
```diff
@@ -48,29 +46,43 @@
...
                                 .create());
 
     this.messageId = messageId;
   }
 
   @Override
-  public void onAdded() {
-    Log.i(TAG, "onAdded() messageId: " + messageId);
+  protected void initialize(Data data) {
+    messageId = data.getLong(KEY_MESSAGE_ID, -1);
```
moxie0: Can we assert this exists?

src/org/thoughtcrime/securesms/jobs/AttachmentDownloadJob.java
greyson-signal: Unfortunately there's no `Data.hasKey()` equivalent, but I did my best via the `SafeData` class.

src/org/thoughtcrime/securesms/jobs/MmsDownloadJob.java
greyson-signal: Unfortunately there's no `Data.hasKey()` equivalent, but I did my best via the `SafeData` class.

src/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -89,27 +89,45 @@
...
+      isInitialized = true;
+      notifyAll();
+    }
+  }
+
+  public void ensureInitialized() {
+    synchronized (this) {
+      while (!isInitialized) {
+        try {
+          wait();
```
moxie0: Util.wait(this) will take care of the try/catch for you

src/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -89,27 +89,45 @@
   private PersistentLogger       persistentLogger;
 
   private volatile boolean isAppVisible;
+  private volatile boolean isInitialized;
```
moxie0: fwiw this doesn't have to be volatile, since you're accessing it from within synchronized blocks only

PR #8095, opened 2018-08-10
Show attachment button while composing text.

greyson-signal:
Previously, we'd only show the attachment button when the user had yet to enter any text. To add an attachment after text was entered, you'd have to go to the three-dot menu. Now we just show a little attach button in the text area.

I also took the opportunity to clean up other button paddings and stuff in the compose area so things look better and react to text sizes more predictably.

[Video Demo](https://github.com/signalapp/Signal-Android/files/2278703/attachment-button.zip)

**Test Devices**

### DIFF
--- a/res/layout/conversation_input_panel.xml
+++ b/res/layout/conversation_input_panel.xml
@@ -54,30 +54,32 @@
                 <LinearLayout
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content"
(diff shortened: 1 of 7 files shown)


Meteor0id:
Discussing enhancement of this PR:
https://community.signalusers.org/t/swap-attachment-and-audio-recording-icons-in-compose-bar/4793

Issue #6051, opened 2017-01-14
Possibility to rotate and crop images (in scribble mode)


mawed:
I have:
- [x] searched open and closed issues for duplicates
- [X] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

----------------------------------------

### Feature request description
When shooting or selecting an image, it get's put into the scribble-mode where it is possible to edit the picture with doodles, notes etc.
I think it would be great if there was a possibility to rotate and/or crop the image before sending.
(This could be done also with more image-editing functions as filters, contrast, brightness....)

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

PR #8185, opened 2018-09-07
Added image crop feature

aleb2000:
### Contributor checklist

 * Device Huawei P8 Lite (2017), Android 8.0.0

----------

### Description
Added the crop image feature.
There is button in the RemovableEditableMediaView component on the bottom left corner.
I have done 2 commits, the first one is the code for the feature itself, the second one was to remove some useless imports and can be ignored.
To crop the image it doesn't rely on any library because I did not want to make this commit too big; there already is a cropping library that is used to crop the profile image when the user chooses one, however such library is deprecated and no longer maintained. As such, the way I used to crop the image will only work if there is an app installed that provide such functionality (even the default gallery has this feature so this shouldn't be a problem, but I've noticed that on my huawei device it can cause problems), for this reason I suggest to import a new library that is still up to date and maintained and provide the cropping functionality (such as https://github.com/Yalantis/uCrop or https://github.com/ArthurHub/Android-Image-Cropper) and to use it for both cropping the images on messages as well as the profile picture. On another note, both libraries I linked also provide a rotate functionality which will make adding the image rotation a lot easier.

This feature was mentioned in an old and now closed issue #6051 as well as on this forum thread https://whispersystems.discoursehosting.net/t/ability-to-edit-images-before-sharing-crop-rotate-flip-rebalance-draw-on/2446


aleb2000:
This PR should be reworked to be compatible with the new camera workflow introduced in 4.27 .

PR #8260, opened 2018-10-08
Ensure download of GCM messages.

greyson-signal:
After switching to WorkManager, we ran into situations where jobs to download messages after receiving a GCM push were being delayed. To combat this, this PR makes efforts to download the messages ASAP, starting a foreground service if necessary.

**Test Devices**
* [Google Pixel 2, Android 9.0, API 28](https://www.gsmarena.com/google_pixel_2-8733.php)

### DIFF
--- a/src/org/thoughtcrime/securesms/gcm/GcmBroadcastReceiver.java
+++ b/src/org/thoughtcrime/securesms/gcm/GcmBroadcastReceiver.java
@@ -2,8 +2,13 @@ package org.thoughtcrime.securesms.gcm;
 
 import android.content.Context;
 import android.content.Intent;
+import android.os.PowerManager;
 import android.support.v4.content.WakefulBroadcastReceiver;
 import android.text.TextUtils;
+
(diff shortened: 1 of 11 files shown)


src/org/thoughtcrime/securesms/gcm/GcmBroadcastReceiver.java
```diff
@@ -44,8 +62,61 @@ private void handleReceivedMessage(Context context, String data) {
...
+              callback.finish();
+            }
+            taskCompleted.set(true);
+          }
+          Log.i(TAG, "Processing complete.");
+        }
+      }
+    }.start();
+
+    new Thread("GcmForegroundServiceTimer") {
```
moxie0: Do we need to start this thread if we started the foreground service already in the doze case?

moxie0:
One other suggestion is to use a single lock for the incoming message pipeline. Right now it looks like there are three different locks, and while I don't think that's actually a problem right now, it does potentially increase the risk of deadlock.

PR #8281, opened 2018-10-17
Updated emoji set.

greyson-signal:
Includes display support for more genders, and more notably, skin tones.These are not currently selectable in the UI, but they will be rendered properly when other clients send them (or when they're input via the system keyboard).

signal-attachment-2018-10-17-112831


kralokgupta:
Sir i want earn money from here...i'm php developer...can you help me?

PR #8332, opened 2018-11-01
Add support for typing indicators.

greyson-signal:
Adds the ability to see when other people in the conversation are typing. Opt-in for existing users.

**To-Do**

- [x] Intro screen
- [x] Better enter/exit animation

**Test Devices**
* [Google Pixel, Android 9.0, API 28](https://www.gsmarena.com/google_pixel-8346.php)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -149,6 +149,7 @@ dependencies {
     }
 }
 
+// TODO: Update dependencies after getting final libsignal
 dependencyVerification {
     verify = [
         'com.android.support:design:7874ad1904eedc74aa41cffffb7f759d8990056f3bbbc9264911651c67c42f5f',
--- a/build.gradle
+++ b/build.gradle
@@ -76,8 +76,8 @@ dependencies {
     compile 'com.google.android.gms:play-services-maps:9.6.1'
     compile 'com.google.android.gms:play-services-places:9.6.1'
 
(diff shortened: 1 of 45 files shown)


Diapolo:
Can this be disabled :)?

greyson-signal:
@deutrino No, it won't undo your setting if you re-register within an existing installation. It's also worth noting that this is a setting that can be changed at any time, just like read receipts.

glU79z:
@greyson-signal Considering Signal's privacy first approach the default for new users should also be opt-in and not opt-out. There could be a splash screen after the registration aksing the user whether he/she wants typing indicators and reading receipts which both default to "turn off". If the user wants these features because he knows them from other messengers, he/she will be able to use them directly. If the user is more privacy focused, he/she will also be satisfied by the default configuration.

greyson-signal:
@glU79z I consider this very analogous to read receipts, which we also default to 'on', and I think that makes sense for most users. That said, let's all please keep discussion about this on [the forum](https://community.signalusers.org/c/feature-requests/android-feature-requests). Thanks!

hugopilot:
Wait? This is cancelled?

greyson-signal:
@hugopilot Oh no, just merged into [4.31](https://github.com/signalapp/Signal-Android/commits/4.31) :) I must have forgot to push the rebase before merging, so it didn't detect that it got merged. But it's all good!

PR #8407, opened 2018-11-29
Multiple Image send via Preview ("+" Symbol in Chats) + Remit multiple content in Chats

tomgithub2018:
Please add the Function to send multiple images via the preview in opened chats. "+" Symbol in a chat shows the images stored on the device, but you can only send one image.

And it would be great t remit multiple content from one chat to another chat.
Actually i can only remit one file/text/content to another chat and not choose more than one content.

Regards and Thanks! Signal is great.


greyson-signal:
Hi, I don't know if you meant to open a pull request, but this just has my own commits in it :)
As far as the request goes, please keep feature requests on [the forum](https://community.signalusers.org/c/feature-requests/android-feature-requests). Thanks!

PR #8409, opened 2018-11-30
Browser access to Signal

AlecClaire:
Hey guys

Suggestion for further development of Signal:

- can you make it (every) browser accessible? I know there is an extension for Chrome but it's so heavy. Also the Win app is heavy and often freezes and I have i3 CPU and 4GB ram. It would be much more convenient just to pull up a new tab in the browser with Signal.

Keep up the good work


rainerzufall:
As mentioned in the bug report template, Signal devs don't use Github for feature requests or discussion topics. Please move this to the [forum ](https://community.signalusers.org/c/feature-requests/android-feature-requests) and close this thread. Thanks!

PR #8410, opened 2018-11-30
Emoji sourtcuts fix and other emoji suggestions

AlecClaire:
Hey guys

Suggestion for further development of Signal:

- when i type ":D" (or anything similar an well known) it doesn't convert to the particular emoji
- the second thing is emojis should pop up as a suggestions when typing a particular words; typing ":joy:" isn't particularity intuitive and it;'s burdensome wasting precious seconds
- some GIF library would be nice as it it's in devil's Facebook Messenger

Keep up the good work


rainerzufall:
Hej @AlecClaire. While I agree with some of the feature requests themselves, as mentioned in the bug report template, Signal devs don't use Github for feature requests or discussion topics. Please move this to the [forum ](https://community.signalusers.org/c/feature-requests/android-feature-requests) and close this thread. Thanks!

ps. giphy support is there already for a long time.

PR #8417, opened 2018-12-01
Allow Signal installation in external storage

tezeb:
### Contributor checklist
 * Samsung Galaxy A3, Android 8.0
 * Innos D10F, Android 7.1.2

----------

### Description
Since Android API 8, it is possible to store apk file in external storage. This allows users of phones with low amount of internal memory to install more aps. To keep backward compatibility and allow the use of external storage only when android device supports it and user did configured it, the safe "auto" value is selected.
As Signal does not qualify in any of the "Should Not" categories listed here:
https://developer.android.com/guide/topics/data/install-location#ShouldNot
it should be perfectly safe to use.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           xmlns:tools="http://schemas.android.com/tools"
+          android:installLocation="auto"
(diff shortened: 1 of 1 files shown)


greyson-signal:
To the contrary, Signal uses several  "Should Not" features:

* Services
* Alarm Services
* Account Managers
* Sync Adapters
* Broadcast Receivers listening for "boot completed"

Thank you, but I think for now we're better off keeping things as they are.

PR #8456, opened 2018-12-15
Fix case of 'Signal call' text in call view

jancborchardt:

After signing the CLA and connecting it to my Github account, the page came up blank and the error console said this, so not sure if it was sent in:
```
The character encoding of the HTML document was not declared.
The document will render with garbled text in some browser configurations if the document contains characters from outside the US-ASCII range.
The character encoding of the page must be declared in the document or in the transfer protocol.
```

### Contributor checklist
- Just a small 1 character case change so I didn’t do testing as I don’t have a full-blown dev environment setup, sorry.

----------

### Description
The expression is written "Signal call" with all-lowercase "call" everywhere. Except in the call view below the name of the person you are calling, where it is "Signal Call" with uppercase "C". This fixes that.

Here is a screenshot of how it looks on master, with "Signal **C**all" casing:
signal call name ellipsis

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -104,7 +104,7 @@
 
     <!-- ContactsDatabase -->
     <string name="ContactsDatabase_message_s">Message %s</string>
-    <string name="ContactsDatabase_signal_call_s">Signal Call %s</string>
+    <string name="ContactsDatabase_signal_call_s">Signal call %s</string>
 
     <!-- ContactNameEditActivity -->
     <string name="ContactNameEditActivity_given_name">Given name</string>


calebTree:
It appears that is because "Signal Call" is the string value of the string identifier "_ContactsDatabase_signal_call_s_". The only string which is "Signal call" is in the value of the "_conversation_callable_secure__menu_call_" string. Yet that is never seen by the user because it is in an attribute of the locked phone icon that is correctly set as always an 'action'. In either case the string value isn't going to interfere with anything.

calebTree:
Hi, if your intention was to to change the string when making that type of call to read "Signal call" instead of "Signal Call" then yes the change you made in the strings.xml file in the resources folder does just that accurately. The way I understood your intention was to fix broken functionality because of the mismatched letter case from the string identifier and the string value of that feature in which case changing the value of the string does not affect functionality only aesthetic preference. The android training course https://developer.android.com/training/basics/firstapp/ teaches that relation well.

jancborchardt:
Added a screenshot to the original post to clarify: https://github.com/signalapp/Signal-Android/pull/8456#issue-238935110

jancborchardt:
Can this me merged or should I close the pull request? :)

calebTree:
Hi, thanks I'm not sure if you're asking me but I don't have a pull request on this concern.

jancborchardt:
Superseded by https://github.com/signalapp/Signal-Android/pull/9536 which fixes the conflicts with master as well.

PR #8465, opened 2018-12-18
Fix restore from an old backup version

### LINKED ISSUE #8011: FullBackupImporter.trimEntriesForExpiredMessages() crashes.
I was on CopperheadOS Noise (copy of Signal) due to the forced upgrade Noise stopped working and I had no option but to move to Signal. It was not possible to install Signal while Noise was still installed, needless to say Noise created a broken backup so I'm stuck with all my (important) messages contained in a broken backup.

While restoring the messages Signal crashes:

[debug.log](https://github.com/signalapp/Signal-Android/files/2212342/debug.log)

I understand the SQL statements are broken (https://github.com/signalapp/Signal-Android/commit/7f7aab044cc6ca2989c3f949c7dbeff13ec39dd6), so guess I need to find a way to fix them.
As far as I understand it I have now two options:
* [proper fix] Fix the parser so that it does not choke on broken backups, (even a partial restore where broken messages are discarded is way better than no messages at all).
* [workaround] Decrypt my backup and fix the SQL manually, I looked into this but it seems Signal uses it's own encryption format for the file so then I would have to build a tool that can decrypt and encrypt the file so at this point it seems easier to try to use the application in order to work on the file. 

My backup is 900mb, and needless to say I cannot post it.
It always fails on the same message and it looks like most messages work.

The log indicates that backup.FullBackupImporter.trimEntriesForExpiredMessages causes the problem, so I'll start looking at this function to try to figure out how it can be fixed or at least improved to work around the problem.

Any help appreciated for how to fix it, I don't think I'm the only one with this problem as Noise expired today.

elkhadiy:
### Contributor checklist

 * Device OnePlus2, OxygenOS 3.5.6, Android 6.0.1
 * Virtual device Nexus 5X API 28 x86, Android 9.0 (Google Play)

----------

### Description

Fixes #8011 

I fixed this since I had trouble restoring a save made in `v4.29.7` into the latest build.

I found that this was the offending hunk in `trimEntriesForExpiredMessages()` (using debug logs):

```Java
try (Cursor cursor = db.query(ThreadDatabase.TABLE_NAME, new String[]{ThreadDatabase.ID}, ThreadDatabase.EXPIRES_IN + " > 0", null, null, null, null)) {
    while (cursor != null && cursor.moveToNext()) {
      DatabaseFactory.getThreadDatabase(context).update(cursor.getLong(0), false);
    }
}
```

What I did was basically move the `trimEntriesForExpiredMessages()` function to `RegistrationActivity.java` so it could be called in the restore handler `handleRestore()` right after `DatabaseFactory.upgradeRestored(context, database);` where we are sure the `database` would be on the latest schema.

A better place for this function might be the `SQLCipherOpenHelper.onUpgrade()` or the `DatabaseFactory.upgradeRestored()` methods?

### DIFF
--- a/src/org/thoughtcrime/securesms/RegistrationActivity.java
+++ b/src/org/thoughtcrime/securesms/RegistrationActivity.java
@@ -8,6 +8,7 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.database.Cursor;
 import android.graphics.Color;
 import android.net.Uri;
 import android.os.AsyncTask;
@@ -52,6 +53,7 @@ import org.greenrobot.eventbus.EventBus;
 import org.greenrobot.eventbus.Subscribe;
 import org.greenrobot.eventbus.ThreadMode;
 import org.thoughtcrime.securesms.animation.AnimationCompleteListener;
+import org.thoughtcrime.securesms.attachments.AttachmentId;
 import org.thoughtcrime.securesms.backup.FullBackupBase;
 import org.thoughtcrime.securesms.backup.FullBackupImporter;
 import org.thoughtcrime.securesms.components.registration.CallMeCountDownView;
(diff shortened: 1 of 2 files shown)


greyson-signal:
Fixed in 923016f12c32536f7edf626c2bc08abe45fa7087. Thanks!

PR #8504, opened 2019-01-04
Updates sdk.dir for macOS

JasonMFry:
It's a readme update.


JasonMFry:
Something doesn't like relative paths.

Issue #8395, opened 2018-11-24
Clicking on a quoted message


ghost:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
On Signal4Android, tapping on the quoted message (during drafting) to scroll back to the original message does not work.  Tapping on a quoted-message that is already in the conversation-history DOES work, the problem only occurs when one is composing a quoted-reply

### Steps to reproduce
- Receive a message
- Click on it to respond to it
- Click on the quoted message above text-area

**Actual result:** screen does not scroll back to quoted message
**Expected result:** screen should scroll back to quoted message, as it happens on Signal4Desktop (Linux/Windows)

### Device info

**Device:** Samsung S9+
**Android version:** 8+
**Signal version:** 4.30.8

PR #8582, opened 2019-02-01
Add functionality to scroll to a quoted message by clicking on the quote view

julKali:

### Contributor checklist

 * Huawei P10 Pro, Android 8.0.0
 * Nokia 8, Android 9.0.0

----------

### Description
The quote view now keeps track of the original message's id (if the message is not missing) to allow scrolling to the original message in the conversation when the quote view is tapped.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -689,6 +689,10 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     finish();
   }
 
+  public void handleQuoteViewClick(long originalMessageId) {
+    fragment.smoothScrollToMessageId(originalMessageId);
+  }
+
   private void handleSelectMessageExpiration() {
     if (isPushGroupConversation() && !isActiveGroup()) {
       return;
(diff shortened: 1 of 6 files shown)


greyson-signal:
1. I don't know if this is strictly necessary. If the user just quoted a message, they're likely already looking at it. In addition, it could just cause jumpiness if, for example, you go to tap the X, miss, and then scroll up to the message.
2. If we did want to do this, I imagine we'd want to leverage existing methods like `ConversationFragment#jumpToMessage` or any of the code in `ConversationFragment#onQuoteClicked`.

mcguire-james:
@greyson-signal I think arguing a simple but useful feature like this a necessary would not be doing it justice. Something doesn't have to be necessary for it to add value. This feature request is a prime example of that. Sure it's necessary at all. But it's very useful. Myself and many other people I know use this feature (in WhatsApp) in conjunction with swipe to reply/"quote" a previous message when sending a new message. That quoted message could three or four scrolls back up in the chat history. To be able to simply tap on the "quoted" message and have the view scroll up and highlight the exact message being quoted is very useful for providing context or reminding of what has been said within the chat. This is even more apparent in large group chat where there can be multiple conversations and/topics being discussed at the same time.

I for one would love to have this feature in Signal.

greyson-signal:
@mcguire-james  To be clear, tapping a quote in the conversation will *already* jump to that message. This PR isn't doing that. This PR is adding the same functionality to a quote in the message compose box (i.e. a quote that hasn't been sent yet). WhatsApp doesn't do this.

The points I made in an earlier comment still stand.

julKali:
Okay so I finally revisited this topic and I came to the conclusion that you are right (especially the point that the user might misclick). I will close this PR, also because the code, if we ever want to implement this, should be rewritten anyways. However, I would suggest also closing #8395 as that is the original issue for this PR.

Issue #7635, opened 2018-04-04
Hardware echo cancellation does not work on many devices


equaeghe:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------
This is a revival of issue #6241. The posts there contain valuable information that cannot concisely be copied here. To deal with this issue, a whitelist was in place initially, which worked in the sense that only known ‘good’ devices would use built-in echo cancellation. Afterwards a switch was made to a blacklist, which was however not updated, leaving many people with non-usable call functionality.

*I suggest others experiencing this issue to wait with commenting until the developers have indicated what info they need. The material in issue #6241 should give them a good starting point.*

### Bug description
Echo cancellation does not work on multiple devices and for various Android versions.
For example, my Samsung GT-I9192 Galaxy S4 Mini Duos (serranodsxx) on Android 7.1.2 (LineageOS) is affected.

### Steps to reproduce
- call someone with Signal, using with affected device/OS (other party's device/OS seems irrelevant)
- have conversation

**Actual result:** Other party hears disturbing echo
**Expected result:** Other party does not hear disturbing echo

Depending on the device, the results vary when using headphones or not. On my device, the echo becomes *worse* when using headphones, so in a situation where the echo cannot be caused by sound being captured by my microphone. (My guess is that the echo cancellation functionality actually *adds* the echo.) For others, the echo disappears in this situation.

Suggested potential workarounds until the root cause can be addressed are to update the blacklist, to include an ‘advanced’ option that switches hardware echo cancellation on or off, or to disable echo cancellation when using headphones. A workaround is needed to avoid chasing users to Skype or so. (I managed to convince my family to switch to Signal and then the whitelist-to-blacklist move forced us back to Skype.)

### Device info

**Device:** Samsung GT-I9192 Galaxy S4 Mini Duos (serranodsxx)
**Android version:** 7.1.2
**Signal version:** 4.17.5

### Link to debug log

Not applicable.

moxie0:
I'm happy to add devices to the blacklist, but only if we can confirm that they are a problem with stock Android.

moxie0:
@sluger We'll need the ro.product.model string for that.

sluger:
[...]

PR #8630, opened 2019-02-22
Adds Fairphone 2 to HARDWARE_AEC_BLACKLIST

irl:
The Fairphone 2 has a lot of very loud echo when making Signal calls. No option is available in the app to test disabling hardware echo cancellation, so this change adds the model to a hardcoded list and then this can be tested once an update has shipped.

Related discussion can be found in #7635 for other devices.


arnt:
FWIW I wrote the same line of code, in #8733. I think the best course of action would be for me to close #8733 and you to do sign the covenant so that this PR can be merged.

irl:
So that I can avoid signing the CLA, please instead merge #8733.

PR #7727, opened 2018-04-24
Add 'quick reply'.

greyson-signal:
If you have a message selected, simply typing will begin composing a reply. You don't have to click the reply button up top.

[Demo Video](https://github.com/signalapp/Signal-Android/files/1940575/quick-reply.zip)

**Test Devices**
* [Moto E (2nd Gen), Android 5.1, API 22](https://www.gsmarena.com/motorola_moto_e_(2nd_gen)-6986.php)

PR #8849, opened 2019-05-31
Swipe to reply

chankruse:

### Contributor checklist

 * Samsung Galaxy S9, 8.0.0
 * Pixel 3, 10

----------

### Description
Enable swipe-to-reply on conversation screen. Updated to address merge conflicts.


chankruse:
Sorry for the multiple PRs. I'm not used to using GIT with multiple users, and messed up the previous PR when merging your master to resolve merge conflicts.

Meteor0id:
I would prefer to see https://github.com/signalapp/Signal-Android/pull/7727 merged (reply by starting to type). If that one is merged than swipe to reply seems close to obsolete to me.

chankruse:
@Meteor0id These are fairly different designs. #7727 still needs a long press, which is a much more laborious user input than a swipe.

chankruse:
@Meteor0id actally, both systems could be implemented concurrently. No need to choose 1 over the other. Whatsapp has both.

SurajVerma:
:+1: I would love to see this merged, as long press to reply is quite inconvenient and involves 2 steps. Hope @moxie0 merge this PR soon.

Meteor0id:
I was actually looking for select and start typing, instead of long press and swiping
But certainly swiping and select-type could could both be implemented.

injainja:
+1

mcguire-james:
@Meteor0id I concur with @chankruse and @SurajVerma a swipe on a message to have it "linked" to your current message that you're typing out is far more intuitive than a long press. They actually have different use cases. A long press on a message would be more to begin a multi-select of messages rather than a quick reply to a single message.

Meteor0id:
@mcguire-james I agree that long pressing is more suited for multi-select than for quick reply. Long press would add an unwanted delay to the reply, which is a direct contradiction to the "quick" of quick reply.

Swiping as a way of selection messages might feel ok to some users, but I think the best solution would be to add both swiping and short tab and starting to type.

So I would be in favor of implementing both this pr snd #7727 but with the annotation that 7727 should be altered to eork with a short tab, not a long one.

umerf52:
Any updates on this?

greyson-signal:
Just played with it. Swipe-to-reply is something that we want, I just have to coordinate with our designers. I think visually this is pretty close, but we're going to want to change some things. Also, some things with touch interactions need to be tightened up. For instance, you can't currently drag the scrubber on a voice note because the slide action is stolen by the ItemTouchHelper.

That said, this is a great base, and a lot of effort was clearly put into this, so thank you very much @chankruse for the contribution! I just have to find time to go through and clean it up. Thanks!

SurajVerma:
@greyson-signal  :smile:

Meteor0id:
> @Meteor0id actally, both systems could be implemented concurrently. No need to choose 1 over the other. Whatsapp has both.

For some reason Whatsapp has removed tap to reply. Their mistake though, as with Android 10 (Android Q) swiping from the edge of the screen inward will become the standard gesture for going backwards one view. [https://developer.android.com/preview/features/gesturalnav#conflicting-gestures](https://developer.android.com/preview/features/gesturalnav#conflicting-gestures)
Therefore swipe to reply would conflict with native Android behavior, while tap and start typing does not.

chankruse:
@Meteor0id I don't think you're understanding how swipe-to-reply works. You don't swipe from the edge of the screen (which, as you correctly mentioned, would go back in Android Q). You can simply swipe on the message itself.

Meteor0id:
@chankruse i understand perfectly well thank you.
The issue is that the user will need to get used to swiping, but at the same time needs to stay away from the border when doing so. This is error prone, the user can't see where the middle of the screen ends and where the border begins. Especially on small messages with limited surface area, swiping to reply could accodentally cause a user to back out of the conversation.

Signal could disable the Android Q back gesture, but why mess with default navigation gestures, some users might prefer to use that.

Pretty sure this gesture was added to Android Q to prevent apps from all coming up with different swiping behavior.

SurajVerma:
I agree with @chankruse here, the same function works all fine in iOS devices and various Android devices which are already offering gesture based navigation. I have used the same on iOS (iPhone X) and Android (Xiaomi and OnePlus devices) both and never ever triggered back action while using swipe to reply feature in WhatsApp and Telegram. Swiping will also feel more intuitive to new Signal users coming from WhatsApp and/or Telegram as both are widely used and a lot of Signal users either used these two messaging apps or are still using.

chankruse:
@Meteor0id I disagree that this is an error prone approach. Besides having tested it myself thoroughly, it behaves the exact same way as the swipe-to-action approach any other app takes (eg: Gmail for deleting/archiving) and I certainly doubt Google and any other company that uses swipe-to-action will disable this feature in Q. As @SurajVerma has pointed out, iOS also has the swipe edge to go back and doesn't interfere in the WhatsApp app. Hopefully one of the UX experts on your team can review and see that this is a much better approach that tap to reply, which has its own usability issues. WhatsApp removed this design because it's a terrible cumbersome approach for replying to messages.

u32i64:
Just to clarify, Signal iOS already has Swipe to Reply in the stable App Store version for some weeks now. The direction of the swipe there is only

```text
left >>>>>>>>>>>>>>>>>> right
```

(not sure how it works with RTL locales though).

DuckyCrayfish:
This PR can be closed as of this commit: 007ea43dc82be4746dcd5cdc4521bd2bbc72ab57.

alex-signal:
Thank you!

Unfortunately we will not merge this as this issue has since been resolved:

Either by:
- another pull request,
- an external factor,
- changes to the minimum API level, etc.

Issue #8682, opened 2019-03-12
Media Gallery UI (Landscape Mode)


Xashyar:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description

**Actual result:** The Header Toolbar Obfuscates the Full Screen View.
**Expected result:** Header Toolbar Should Fade Away with the Navigation Menu.
### Screenshots
![signal-attachment-2019-03-13-001821](https://user-images.githubusercontent.com/34171942/54235428-770e2400-4526-11e9-825e-46d9103cb5bf.png)

jatinkumarg:
Hey, can I work on this?

Xashyar:
> 
> 
> Hey, can I work on this?
👍

jatinkumarg:
@Xashyar Just to be sure, do you want me to completely remove the Header Toolbar?

Xashyar:
@jatinkumarg 
No, just that it should fade away by tapping on the underneath media (photo, video, etc.).

and I also suggest contacting @greyson-signal since the [PRs](https://github.com/signalapp/Signal-Android/pulls) rarely get merged.

Xashyar:
@danmoop
Hi Danill, sorry for pinging you along, I thought this issue might be similar to #8915, would you mind taking a look as well? 

Thanks in Advance.

danmoop:
@Xashyar Hello. Sure I checked the issue #8915. I think there is a solution to this problem, it will look like this:
``` java
@Override
  public void onConfigurationChanged(Configuration newConfig) {
    if(newConfig.orientation == Configuration.ORIENTATION_LANDSCAPE) {
      toolbar.hide(); // when landscape mode is on -> hide toolbar
    } else {
      toolbar.show(); // when we are back to portrait -> show toolbar
    }
  }
```
And the file that handles menu buttons is [this one](https://github.com/signalapp/Signal-Android/blob/master/src/org/thoughtcrime/securesms/MediaPreviewActivity.java)
The problem is that I can't access toolbar. I can access buttons (save, forward, etc.) but **this box** I cannot. It wants to hide me :(

Xashyar:
The issue no longer exists.

PR #8915, opened 2019-07-08
Fixes #8912 & #8682

### LINKED ISSUE #8912: Image editor: Keyboard doesn't hide when you select highlighter, pen, or crop

### Bug description
In the image editor, having the keyboard active does not hide it when you use tools that do not use the keyboard.

### Steps to reproduce
1. Select some images to send
2. Make the keyboard show by tapping on composeText or "Add a caption..."
3. Press any of the highlighter, marker, or crop tools

**Actual result:** Keyboard does not hide for tools that don't use the keyboard
**Expected result:** Keyboard hides

### Device info

**Device:** Samsung Galaxy S7 Edge and Samsung Galaxy S10
**Android version:** both 8.0.0
**Signal version:** both 4.42.3

### Debug log
Seems easily reproducible

danmoop:

### Contributor checklist

 * Physical Device - Xiaomi Redmi 4X

----------

### Description

This PR fixes the problem with a keyboard when you try to edit an image (draw, crop etc.)
Also works when multiple images are selected.


src/org/thoughtcrime/securesms/scribbles/ImageEditorFragment.java
```diff
@@ -240,6 +242,11 @@ public void onModeStarted(@NonNull ImageEditorHud.Mode mode) {
 
     controller.onTouchEventsNeeded(mode != ImageEditorHud.Mode.NONE);
 
+    InputMethodManager keyboard = (InputMethodManager) getActivity().getSystemService(Activity.INPUT_METHOD_SERVICE);
+
+    if (keyboard.isActive())
+      keyboard.hideSoftInputFromWindow(getActivity().getCurrentFocus().getWindowToken(), 0);
```
alan-signal: I tried your changes and this crashed when I pressed the tick to exit the pen mode.

Instead of doing this here, why don't you look at when the message entry box gets hidden. That is the logical place to hide the keyboard too.

danmoop: Ok, I will try to figure this out, thanks

danmoop: What device are you using? I tested this one more time on my Xiaomi Redmi 4X and it seems to be working fine. I can't understand what is the problem.
When I exit pen / highlight / crop mode by pressing that tick button it doesn't crash. 
By the way you told me about that message entry box. Well, I can't find a line when it hides. Of course it is logical to hide keyboard together with it, in my solution it hides when you choose the mode, so it is almost similar.
After doing some attempts trying to figure out the `ImageEditorFragment` structure I am out of possible solutions.

alan-signal: I tried on a oneplus 5T, and I think the currentFocus was `null`.

> Of course it is logical to hide keyboard together with it, in my solution it hides when you choose the mode, so it is almost similar.

Yes but it's more maintainable and predictable if the two (hide message box and hide keyboard) are very close together.

At some point it "requests full screen": https://github.com/signalapp/Signal-Android/blob/b5aa46bb672296a40eb6f5aee555dfa0c402c774/src/org/thoughtcrime/securesms/scribbles/ImageEditorHud.java#L193

Follow that down and you will see how it makes the message box disappear.

danmoop: Ok, I made a new commit, please have a look!

src/org/thoughtcrime/securesms/scribbles/ImageEditorFragment.java
```diff
@@ -308,7 +311,15 @@ public boolean isCropAspectLocked() {
...
   @Override
-  public void onRequestFullScreen(boolean fullScreen) {
+  public void onRequestFullScreen(boolean fullScreen, ImageEditorHud.Mode mode) {
+
+    View view = getActivity().getCurrentFocus();
+
+    if (view != null && mode != ImageEditorHud.Mode.TEXT) {
+      InputMethodManager imm = (InputMethodManager) getContext().getSystemService(Context.INPUT_METHOD_SERVICE);
+      imm.hideSoftInputFromWindow(view.getWindowToken(), 0);
+    }
```
alan-signal: This is not where the message box is hidden. I think this should be along side that.

danmoop: [That is where it hides](https://github.com/signalapp/Signal-Android/blob/master/src/org/thoughtcrime/securesms/mediasend/MediaSendFragment.java#L514). But the problem is how can I access mode that has been chosen ( we can't hide keyboard if "TEXT" mode has been chosen.

alan-signal: I would think that `requestFullScreen` could take two arguments. Just like you've added a mode, but instead, just a boolean:

```
requestFullScreen(boolean fullscreen, boolean hideKeyboard)
```

alan-signal: Then also note that from within `MediaSendFragment` there is available to you a much cleaner method to hide the keyboard. **Maybe** you test if the message box has the keyboard focus and not need the extra parameter after all.

danmoop: Yeah I got the same idea, let me try that!

danmoop: Check out the latest commit, seems to be working right!

danmoop: By the way we need to know `MODE` anyway because if it is `TEXT` then keyboard hides and we can't add text to the image. But there is `if` so works fine.

src/org/thoughtcrime/securesms/mediasend/MediaSendFragment.java
```diff
@@ -511,7 +512,11 @@ protected void onPostExecute(List<Media> media) {
     return future;
   }
 
-  public void onRequestFullScreen(boolean fullScreen) {
+  public void onRequestFullScreen(boolean fullScreen, ImageEditorHud.Mode mode) {
+    if (mode != ImageEditorHud.Mode.TEXT && hud.isKeyboardOpen()) {
+      hud.hideSoftkey(composeText, null);
```
alan-signal: You found it 👍 

So there's no way to replace `mode != ImageEditorHud.Mode.TEXT` with `composeText.hasFocus()` or some other method that detects which textbox has input?

danmoop: Your solutions works but the only thing is that keyboard doesn't hide after we add some text on the image and press tick button. And if we press on buttons there is no text added anywhere.
So I think passing mode isn't too bad, everything hides when it should.

danmoop:
Hello! Is there anybody active here? I wonder if this PR gets merged or reviewed.

alan-signal:
One PR per issue please, separate #8682 solution

src/org/thoughtcrime/securesms/mediasend/MediaSendFragment.java
```diff
@@ -137,6 +157,154 @@ private void initViewModel() {
...
+    }.execute();
+  }
+
+  private static ListenableFuture<Bitmap> render(@NonNull Context context, @NonNull EditorModel model) {
+    SettableFuture<Bitmap> future = new SettableFuture<>();
+
+    AsyncTask.THREAD_POOL_EXECUTOR.execute(() -> future.set(model.render(context)));
+
+    return future;
+  }
```
alan-signal: I don't understand where all of these changes have come from. I suspect you have made an error during rebase.

src/org/thoughtcrime/securesms/mediasend/MediaSendActivity.java
```diff
@@ -62,6 +62,8 @@
 import org.thoughtcrime.securesms.providers.BlobProvider;
 import org.thoughtcrime.securesms.recipients.Recipient;
 import org.thoughtcrime.securesms.scribbles.ImageEditorFragment;
+import org.thoughtcrime.securesms.scribbles.ImageEditorHud;
```
alan-signal: I don't like that this activity now knows about hud modes.

I reviewed before and was after something cleaner.

src/org/thoughtcrime/securesms/mediasend/MediaSendFragment.java
```diff
@@ -13,6 +13,26 @@
...
+import org.thoughtcrime.securesms.components.emoji.MediaKeyboard;
+import org.thoughtcrime.securesms.contactshare.SimpleTextWatcher;
+import org.thoughtcrime.securesms.imageeditor.model.EditorModel;
+import org.thoughtcrime.securesms.logging.Log;
+import org.thoughtcrime.securesms.mediapreview.MediaRailAdapter;
+import org.thoughtcrime.securesms.mms.GlideApp;
+import org.thoughtcrime.securesms.providers.BlobProvider;
+import org.thoughtcrime.securesms.recipients.Recipient;
+import org.thoughtcrime.securesms.scribbles.ImageEditorFragment;
+import org.thoughtcrime.securesms.scribbles.ImageEditorHud;
```
alan-signal: I don't like that this fragment now knows about hud modes.

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -390,6 +391,16 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
 
+  @Override
+  public void onConfigurationChanged(Configuration newConfig) {
+    super.onConfigurationChanged(newConfig);
+    if(newConfig.orientation == Configuration.ORIENTATION_LANDSCAPE) {
+      getSupportActionBar().hide();
+    } else {
+      getSupportActionBar().show();
+    }
+  }
```
alan-signal: This should not be in same PR, but I've tested it, what about when open the activity and are already in landscape?

src/org/thoughtcrime/securesms/mediasend/MediaSendFragment.java
alan-signal: I see, looks like you didn't rebase, but merged, either way, these should not be here.

image

PR #8962, opened 2019-07-29
Warn when setting a system screenlock on API < 21. Fixes #8603

### LINKED ISSUE #8603: Screen Lock Not Working on Android 4
- [ x I am submitting a bug report for existing functionality that does not work as intended

### Bug description
After Signal's inactivity timeout, Signal does not invoke Android's screen lock.
Also when locking Signal in the message center pull down, it remains unlocked.
This is the same issue reported at https://github.com/signalapp/Signal-Android/issues/7962 for Android 4.3, and it also occurs on my Samsung GT-I9105P running Android 4.1.2

### Steps to reproduce
- Set Android screen lock to PIN + to automatically lock immediately after sleep (2 min)
- Set Signal to screen lock + inactivity timeout to 1 minute
- Open Signal after more than 1 minute
- Or open Signal after locking it in the message center

**Actual result:** Signal opens without invoking the PIN lock screen
**Expected result:** PIN lock screen prevents direct access to Signal

### Device info

**Device:** Sony D6503
**Android version:** 4.4.2
**Signal version:** 4.33.5

### Link to debug log

https://debuglogs.org/a21104cfefffd3f8a73b79ad157334e0fac3f2c71051b217d6057427d44e403d

johanw666:
### Contributor checklist

 * Sony Xperia Z1 Compact, Android 4.4.4
 * Sony Xperia M4 Aqua, Android 5.0
 * Sony Xperia Z3 Compact, Android 6.0.1

----------

### Description
Locking Signal via a the Android screenlock can not be called on Android 4.4, but the user interface does not warn the user when he sets a screenlock on Android 4.4. It just silently does not work. This commit shows a user on Android 4.4 a popup screen that setting a screenlock is not possible and sets the toggle to disabled so the UI and the program agree with each orther.

On API >= 21 it changes nothing.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -1336,6 +1336,8 @@
     <string name="preferences_advanced__always_relay_calls">Always relay calls</string>
     <string name="preferences_app_protection__app_access">App access</string>
     <string name="preferences_app_protection__communication">Communication</string>
+    <string name="preferences_app_protection__screenlock_requires_lollipop">Using the system screenlock requires Android 5.0 (Lollipop) or higher.</string>
+    <string name="preferences_app_protection__android_version_too_low">Android version too low</string>
     <string name="preferences_chats__chats">Chats</string>
     <string name="preferences_notifications__messages">Messages</string>
     <string name="preferences_notifications__events">Events</string>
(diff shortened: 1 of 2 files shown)


ngsilverman:
Rather than waiting for the user to attempt to turn on screen lock before notifying them that the option is not available on their Android version, what would you think of disabling the toggle row, and perhaps amending the description to explain why it's disabled?

Another option could be to hide that row altogether, though that could create confusion for users looking at the app on two different phones and seeing it only on one, or talking to a friend about screen lock when they're using an older version of Android, etc.

johanw666:
> Rather than waiting for the user to attempt to turn on screen lock before notifying them that the option is not available on their Android version, what would you think of disabling the toggle row, and perhaps amending the description to explain why it's disabled?

I have no problems with that option, it is a UI choice. Disabeling the field would require less code (no event handler override) so it looks cleaner. Ultimately it is the choice of OWS if and how they implement it. I'm willing to recode the solution on request.

PR #9022, opened 2019-09-08
Add checksum-dependency-plugin to verify checksums and PGP signatures for dependency artifacts

vlsi:
See https://github.com/vlsi/vlsi-release-plugins#checksum-dependency-plugin

### Contributor checklist

----------

### Description

`checksum-dependency-plugin` is a superset of `gradle-witness`, and it enables to increase the level of security.

Key features:
* Gradle plugins can be verified (grade-witness doesn't track plugins)
* All Gradle configurations are supported (e.g. `java-library` plugin is supported). `checksum-dependency-plugin` intercepts detached configurations as well (e.g. the ones that are created on demand)
* PGP can be used for verification. PGP can be used with or without checksum. PGP enables to detect and prevent issues like https://blog.autsoft.hu/a-confusing-dependency/

Even though CONTRIBUTING.md says `PGP is our guide for what not to do`, I still think PGP for dependency verification is an improvement over existing gradle-witness, and it does not really introduce PGP to the app itself.

Note: I have not removed `gradle-witness` yet, however this PR makes `gradle-witness` obsolete.

### DIFF
--- a/buildSrc/checksum.xml
+++ b/buildSrc/checksum.xml
@@ -1,0 +1,11 @@
+<?xml version='1.0' encoding='utf-8'?>
+<dependency-verification version='1'>
+  <trust-requirement pgp='GROUP' checksum='NONE' />
+  <ignored-keys />
+  <trusted-keys />
+  <dependencies>
+    <dependency group='com.android.tools.build' module='apksig' version='3.3.2'>
+      <sha512>4F33A3FB4BC8268A19D8D1912B7F02F31AE5ECB65BC4BB62F5FB84235EC072D5D4265D35934DDA3BD48239F65939076B279F71565A095E90729ED8BEC36F83E5</sha512>
+    </dependency>
+  </dependencies>
+</dependency-verification>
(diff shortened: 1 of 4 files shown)


vlsi:
Note: default configuration is satisfied if a dependency is signed with a relevant group PGP key.
Certain dependencies miss PGP (e.g. `aar` files for some reason), then checksum-dependency resorts to SHA-512.

vlsi:
Hi, I see the PR has not yet been reviewed.

You might be interested that [Gradle 6.2 introduces in-core dependency verification](https://github.com/gradle/gradle/issues/10443#issuecomment-568741472)

It means you might consider upgrading Gradle and using in-core verification.

The documentation for the new feature can be reviewed here: https://github.com/gradle/gradle/pull/11755

From what I know Gradle would cover more cases when compared with `checksum-dependency-plugin`. For instance, it will be able to verify `pom.xml` which are implicitly fetched by Gradle when resolving transitive dependencies and probably other cases.

Some bits can be previewed in the current release candidates, release nightly builds and master nightly builds (see https://gradle.org/releases/ )

It would be nice if you could preview the feature and provide your feedback.

alan-signal:
Interesting thank you, I'm keeping an eye on it.

PR #9152, opened 2019-11-02
Silence ringer with volume buttons

### LINKED ISSUE #8338: Volume keys don't mute ringing during call arrival

### Bug description
Given a certain standard set by "regular" calling and calling through other messaging apps, I expect that hitting the "volume down" button while receiving a call would mute the ringing (sound). However, this is not what happens with Signal.

### Steps to reproduce
- Ensure your phone is not on mute (ringtone volume > 0).
- have someone call you on Signal.
- Wait for the ringing.
- Hit the volume down button.

**Expected result:** 
Ringing sound muted while the incoming call keeps on incoming (not dropped).

**Actual result:**
 The volume keys adjust the "call volume".
Alternatively, if one has the top-menu (the one holding the wifi/ location/ sound/ torch and so on toggle-buttons) pulled down, hitting the volume keys during an incoming call would adjust the "ringtone" volume, which would adjust the volume of the ringing.
Just to stress the point, however, in the latter case, even if the top-menu being down was not necessary, that (adjusting the ringing volume) is *not* the desired behaviour. Assuming that the ringing is there to bring to one's attention the fact that a call is being received and not to play music, there is no reason to adjust the volume of the ring in "real time". Supposedly the ring-muting is there in particular for those situation when one's phone starts ringing and it is a really, really bad time for it to ring. You want to kill the sound as quickly as possible. 

### Device info

**Device:** Samsung J5
**Android version:** 6.0.1
**Signal version:** 4.29.7

di72nn:

### Contributor checklist

 * Mi Pad 4 Plus, Android 9.0 (LineageOS 16.0).

----------

### Description

Silences incoming call ringer with a volume button press (as in the default Android phone app and many other messenger apps).

That was already implemented for "screen off" event (which is kind of a different use case), I only added handling of volume keys.

~I haven't tested it since I don't have an extra Signal account and I don't want to mess with my main account. The change is small, and you'll have to test it yourself anyway. But if you **really** want me to test it first, I guess I can think of something.~ Tested - works as expected.

Fixes #8338.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/WebRtcCallActivity.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/WebRtcCallActivity.java
@@ -116,8 +116,12 @@ public class WebRtcCallActivity extends Activity {
   @Override
   public boolean onKeyDown(int keyCode, KeyEvent event) {
     if (keyCode == KeyEvent.KEYCODE_VOLUME_UP || keyCode == KeyEvent.KEYCODE_VOLUME_DOWN) {
-      handleSilenceRinger();
-      return true;
+      WebRtcViewModel model = EventBus.getDefault().getStickyEvent(WebRtcViewModel.class);
+
+      if (model != null && model.getState() == WebRtcViewModel.State.CALL_INCOMING) {
+        handleSilenceRinger();
+        return true;
+      }
     }
     return super.onKeyDown(keyCode, event);
   }
--- a/app/src/main/java/org/thoughtcrime/securesms/WebRtcCallActivity.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/WebRtcCallActivity.java
@@ -29,6 +29,7 @@ import androidx.annotation.NonNull;
 import androidx.appcompat.app.AlertDialog;
 
 import org.thoughtcrime.securesms.logging.Log;
+import android.view.KeyEvent;
 import android.view.View;
 import android.view.Window;
 import android.view.WindowManager;
@@ -113,6 +114,15 @@ public class WebRtcCallActivity extends Activity {
   }
 
   @Override
+  public boolean onKeyDown(int keyCode, KeyEvent event) {
+    if (keyCode == KeyEvent.KEYCODE_VOLUME_UP || keyCode == KeyEvent.KEYCODE_VOLUME_DOWN) {
+      handleSilenceRinger();
+      return true;
+    }
+    return super.onKeyDown(keyCode, event);
+  }
+
+  @Override
   public void onConfigurationChanged(Configuration newConfiguration) {
     super.onConfigurationChanged(newConfiguration);
   }
@@ -182,6 +192,12 @@ public class WebRtcCallActivity extends Activity {
   private void handleFlipCamera() {
     Intent intent = new Intent(this, WebRtcCallService.class);
     intent.setAction(WebRtcCallService.ACTION_FLIP_CAMERA);
+    startService(intent);
+  }
+
+  private void handleSilenceRinger() {
+    Intent intent = new Intent(this, WebRtcCallService.class);
+    intent.setAction(WebRtcCallService.ACTION_SILENCE_RINGER);
     startService(intent);
   }
 
--- a/app/src/main/java/org/thoughtcrime/securesms/service/WebRtcCallService.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/service/WebRtcCallService.java
@@ -118,6 +118,7 @@ public class WebRtcCallService extends Service implements CallManager.Observer,
   public static final String ACTION_BLUETOOTH_CHANGE            = "BLUETOOTH_CHANGE";
   public static final String ACTION_WIRED_HEADSET_CHANGE        = "WIRED_HEADSET_CHANGE";
   public static final String ACTION_SCREEN_OFF                  = "SCREEN_OFF";
+  public static final String ACTION_SILENCE_RINGER              = "SILENCE_RINGER";
   public static final String ACTION_IS_IN_CALL_QUERY            = "IS_IN_CALL";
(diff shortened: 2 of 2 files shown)


Xashyar:
@di72nn 
Can `android.view.KeyEvent;` event be used for answering calls through Handsfree/Headphone buttons?

Xashyar:
@di72nn 
Can `android.view.KeyEvent;` be used for answering calls through Handsfree/Headphone buttons?

di72nn:
@Xashyar 
I'm not sure - it depends on the key code it generates. If it's a "media button" code, see [this guide](https://developer.android.com/guide/topics/media-apps/mediabuttons#foreground-activity), otherwise it indeed should be a simple `KeyEvent`.

di72nn:
Rebased and tested. Works as expected.

di72nn:
Rebased again.

nutts0:
Could this please be merged. It is an essential feature

app/src/main/java/org/thoughtcrime/securesms/WebRtcCallActivity.java
```diff
@@ -112,6 +113,15 @@ public void onPause() {
     EventBus.getDefault().unregister(this);
   }
 
+  @Override
+  public boolean onKeyDown(int keyCode, KeyEvent event) {
+    if (keyCode == KeyEvent.KEYCODE_VOLUME_UP || keyCode == KeyEvent.KEYCODE_VOLUME_DOWN) {
+      handleSilenceRinger();
+      return true;
```
alex-signal, line 120: This looks like it will silence the ringer whenever vol up or vol down is pressed. There are a couple of problems here:

1. We only want this to be the case when the user sees this activity from their lock screen. 
2. We also only want this to be the case when the call's state matches those within the SILENCE RINGER intent handling.

2. can be solved by using `getStickyEvent` and checking against the event state, as is done in a few other places.

1. might be a trickier solve.

This (only silencing when in 'lock screen' mode) is in line with what other calling apps do. For example, if you enter this screen from the notification bar while your phone is unlocked, the volume buttons would only modify volume (as it does in Signal today).

alex-signal (changes requested):
Thank you!

Thanks for the PR! And sorry it took a while to get to. I've left some feedback regarding the functionality in another comment.

app/src/main/java/org/thoughtcrime/securesms/WebRtcCallActivity.java
gtsiolis: > We only want this to be the case when the user sees this activity from their lock screen.

@alex-signal are you sure about this? I think most, if not all, calling apps allow you to silence the call, using the volume buttons, when the phone is unlocked and you are seeing the notification or the call activity. 📞 

PS. Thanks for adding this @di72nn!

di72nn: >  2. We also only want this to be the case when the call's state matches those within the SILENCE RINGER intent handling.
> 
> 3. can be solved by using `getStickyEvent` and checking against the event state, as is done in a few other places.

Added a condition.



> 1. We only want this to be the case when the user sees this activity from their lock screen.
> 
> This (only silencing when in 'lock screen' mode) is in line with what other calling apps do. For example, if you enter this screen from the notification bar while your phone is unlocked, the volume buttons would only modify volume (as it does in Signal today).

Which apps do that? The default Android dialer silences ringer with volume buttons regardless of the way the call screen becomes visible.
How does the way the screen appears matter anyway?

alex-signal: This is secondary information, but according to one of our designers, Telegram, WhatsApp, and Skype all behave the way we currently do when this screen appears while a user has their phone unlocked. 

The silencing only works if the screen is locked (though Skype is the odd one out, and behaves as Signal does now.)

If I'm incorrect here, please let me know and I'll be more than happy to investigate how these apps operate myself, but design doesn't want to sway from what other major messaging clients are doing here.

Xashyar: > The silencing only works if the screen is locked (though Skype is the odd one out, and behaves as Signal does now.)

I think the _immediate_ silencing should work in locked state as well. It's hard to imagine a scenario against this, i.e. the user needing to _incrementally_ lower the volume in the very moment the phone rings.

alex-signal: > I think the immediate silencing should work in locked state as well.

This is what I'm saying, but only in this state. We don't currently do this in Signal which is a mistake. Outside of a locked state, the volume controls should behave as they do now.

Xashyar:
Still reproducible on the latest Signal-Android.

Xashyar:
This is unfortunately still reproducible on pre-Android 8.0 devices.

Xashyar:
This issue is still reproducible on pre-API-26 devices.

cody-signal:
If this can be rebased and fixed I'll merge it.

cody-signal:
Please feel free to re-open when fixed. Otherwise, closing.

rajsinghUSA:
Still can't silence with volume keys

di72nn:
Rebased and addressed the "silence only in lockscreen mode" comment (at least I think I did - I haven't been able to test it since Signal just doesn't let me request that damn SMS code to log in :face_exhaling: ).

I updated the same branch. Not sure if this PR can be reopened - I certainly can't do it.

wertzuz:
@cody-signal

tylla:
@cody-signal Could you eventually take a look at this if it could be merged?

tylla:
@cody-signal, @alex-signal  As I see, this was closed without merging the commits. Could any of you take a look and merge them?

tylla:
@cody-signal @alex-signal I don't want to be annoying (yeah, I know I am, really sorry), but pretty-please can you take a look at this PR and see if it can be merged? We are so close to finishing this feature.
Thank you in advance!

PR #8798, opened 2019-05-06
Changed security scheme in 3 classes to diminish the vulnerabilities …

rgonzalezp:
…in the app

### First time contributor checklist

- [X] I have read [how to contribute](https://github.com/signalapp/Signal-Android/blob/master/CONTRIBUTING.md) to this project
- [X] I have signed the [Contributor License Agreement](https://whispersystems.org/cla/)

### Contributor checklist

- [x] I am following the [Code Style Guidelines](https://github.com/signalapp/Signal-Android/wiki/Code-Style-Guidelines)
- [x] I have tested my contribution on these devices:
 * Device One Plus 6, Android 9, OxygenOS version 9.0.5
- [x] My contribution is fully baked and ready to be merged as is
- [x] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the `Fixes #1234` [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)

----------

### Description

My Pullrequest proposes to fix a vulnerability presented when we use the security schema AES/CBC/PKCS5Padding in the classes MasterCipher.java, LogFile.java and ClassicDecryptingPartInputStream.java. I Changed the schema from AES/CBC/PKCS5Padding that can be vulnerable to oracle padding attacks to AES/GCM/NoPadding, which is the preferred method. These vulnerabilities where obtained running an analysis with SonarQube

PR #9215, opened 2019-11-22
Use Galois/Counter Mode 

jsprieto10:
----------

### Description

After analyzing the code with sonarqube, sonar detects that the Advanced Encryption Standard (AES) encryption algorithm can be used with various modes. Some combinations are not secured:

Electronic Codebook (ECB) mode: Under a given key, any given plaintext block always gets encrypted to the same ciphertext block. Thus, it does not hide data patterns well. In some senses, it doesn't provide serious message confidentiality, and it is not recommended for use in cryptographic protocols at all.
Cipher Block Chaining (CBC) with PKCS#5 padding (or PKCS#7) is susceptible to padding oracle attacks.
In both cases, Galois/Counter Mode (GCM) with no padding should be preferred.

### DIFF
--- a/src/org/thoughtcrime/securesms/crypto/ClassicDecryptingPartInputStream.java
+++ b/src/org/thoughtcrime/securesms/crypto/ClassicDecryptingPartInputStream.java
@@ -59,7 +59,7 @@ public class ClassicDecryptingPartInputStream {
       byte[]          ivBytes    = new byte[IV_LENGTH];
       readFully(fileStream, ivBytes);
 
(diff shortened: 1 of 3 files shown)


jsprieto10:
Sorry, how do i do that, i am knew doing pull requests

greyson-signal:
Of the things you've changed, only the LogFile is still in use (the rest are there for legacy purposes). And changing the LogFile would require some sort of migration, or you'd just break all of the existing logs. I can ask someone about switching to a no-padding scheme.

alan-signal:
This PR is identical to #8798

Issue #9220, opened 2019-11-24
ConversationActivity memory leaks


diegoramosb:
- [X] I have searched open and closed issues for duplicates
- [X] I am submitting a bug report for existing functionality that does not work as intended
- [X] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [X] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
The ConversationActivity consumes more and more memory as I open and close conversations.

### Steps to reproduce
- Open Signal
- Open a conversation
- Go back using the back button
- Repeat the two previous steps

**Actual result:** As you can see in the profiling below, the app starts consuming more and more memory. The GC events don't free enough memory.
**Expected result:** The GC events should free more memory as the ConversationActivity is closed.

### Screenshots
![unnamed](https://user-images.githubusercontent.com/18556257/69491411-f9473a00-0e62-11ea-889d-04172366b0a0.png)

### Device info
**Device:** Motorola Moto G5S
**Android version:** 8.1.0
**Signal version:** 4.50.5

### Link to debug log
https://debuglogs.org/be93fd10e778cc3cb2e16ac0d3a172b3d674b6beb36971e3169a95b027b2647e

Note: the profiling was done in a OnePlus 7 Pro with Android 10 because it was done before I read the guidelines. However, the behaviour has been tested several times.

PR #9221, opened 2019-11-24
Memory leaks

diegoramosb:

### Contributor checklist
 * OnePlus 7 Pro Android 10 

----------

### Description
As Described in #9220, there are memory leaks in the ConversationActivity. This PR aims to solve the ones that happen when opening and closing the activity. The AsyncTasks created in the initializeSecurity() and saveDraft() methods have implicit references to the activity. By moving the code of these AsyncTasks to separate classes and making weak references to the activity, these references are not created and the memory can be liberated. As seen in the memory profiling below, when the app is put in background, memory can be freed up.
asdf

### DIFF
--- a/src/org/thoughtcrime/securesms/util/task/InitializeSecurityAsyncTask.java
+++ b/src/org/thoughtcrime/securesms/util/task/InitializeSecurityAsyncTask.java
@@ -1,0 +1,86 @@
+package org.thoughtcrime.securesms.util.task;
+
+import android.content.Context;
+import android.os.AsyncTask;
+
+import org.thoughtcrime.securesms.contacts.sync.DirectoryHelper;
+import org.thoughtcrime.securesms.conversation.ConversationActivity;
+import org.thoughtcrime.securesms.database.RecipientDatabase;
+import org.thoughtcrime.securesms.logging.Log;
+import org.thoughtcrime.securesms.recipients.Recipient;
+import org.thoughtcrime.securesms.util.TextSecurePreferences;
+import org.thoughtcrime.securesms.util.Util;
+import org.thoughtcrime.securesms.util.concurrent.SettableFuture;
+
+import java.io.IOException;
+import java.lang.ref.WeakReference;
+
+public class InitializeSecurityAsyncTask extends AsyncTask<Recipient, Void, boolean[]> {
+
+
+    private long threadId;
+
+    private boolean currentSecureText;
+
+    private boolean currentIsDefaultSms;
+
+    private SettableFuture<Boolean> future;
+
+    private WeakReference<Context> contextWeakReference;
+
+    private WeakReference<ConversationActivity> activityWeakReference;
+
+    private String tag;
+
+    public InitializeSecurityAsyncTask(long threadId, boolean currentSecureText, boolean currentIsDefaultSms, SettableFuture<Boolean> future, WeakReference<Context> contextWeakReference, WeakReference<ConversationActivity> activityWeakReference, String tag) {
+        this.threadId = threadId;
+        this.currentSecureText = currentSecureText;
+        this.currentIsDefaultSms = currentIsDefaultSms;
+        this.future = future;
+        this.contextWeakReference = contextWeakReference;
+        this.activityWeakReference = activityWeakReference;
+        this.tag = tag;
+    }
+
+    @Override
+    protected boolean[] doInBackground(Recipient... params) {
+        Context context = contextWeakReference.get();
+        Recipient recipient = params[0].resolve();
+        Log.i(tag, "Resolving registered state...");
+        RecipientDatabase.RegisteredState registeredState;
+
+        if (recipient.isPushGroup()) {
+            Log.i(tag, "Push group recipient...");
+            registeredState = RecipientDatabase.RegisteredState.REGISTERED;
+        } else {
+            Log.i(tag, "Checking through resolved recipient");
+            registeredState = recipient.resolve().getRegistered();
+        }
+
+        Log.i(tag, "Resolved registered state: " + registeredState);
+        boolean signalEnabled = TextSecurePreferences.isPushRegistered(context);
+
+        if (registeredState == RecipientDatabase.RegisteredState.UNKNOWN) {
+            try {
+                Log.i(tag, "Refreshing directory for user: " + recipient.getId().serialize());
+                registeredState = DirectoryHelper.refreshDirectoryFor(context, recipient, false);
+            } catch (IOException e) {
+                Log.w(tag, e);
+            }
+        }
+
+        Log.i(tag, "Returning registered state...");
+        return new boolean[]{registeredState == RecipientDatabase.RegisteredState.REGISTERED && signalEnabled,
+                Util.isDefaultSmsProvider(context)};
+    }
+
+    @Override
+    protected void onPostExecute(boolean[] result) {
+        if (result[0] != currentSecureText || result[1] != currentIsDefaultSms) {
+            Log.i(tag, "onPostExecute() handleSecurityChange: " + result[0] + " , " + result[1]);
+            activityWeakReference.get().handleSecurityChange(result[0], result[1]);
+        }
+        future.set(true);
+        activityWeakReference.get().onSecurityUpdated();
+    }
+}
--- a/src/org/thoughtcrime/securesms/util/task/InitializeSecurityAsyncTask.java
+++ b/src/org/thoughtcrime/securesms/util/task/InitializeSecurityAsyncTask.java
@@ -15,72 +15,75 @@ import org.thoughtcrime.securesms.util.concurrent.SettableFuture;
 import java.io.IOException;
 import java.lang.ref.WeakReference;
 
+/**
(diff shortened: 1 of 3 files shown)


EukalyptusX:
Screenshot_20220128-193455

EukalyptusX:
Thanks stale bot. But it's still not fixed. I won't let this get closed :-)

This is just ridiculous.

src/org/thoughtcrime/securesms/util/task/InitializeSecurityAsyncTask.java
```diff
@@ -0,0 +1,89 @@
...
+
+  private String tag;
+
+  public InitializeSecurityAsyncTask(long threadId, boolean currentSecureText, boolean currentIsDefaultSms, SettableFuture<Boolean> future, WeakReference<Context> contextWeakReference, WeakReference<ConversationActivity> activityWeakReference, String tag) {
+    this.threadId = threadId;
+    this.currentSecureText = currentSecureText;
+    this.currentIsDefaultSms = currentIsDefaultSms;
+    this.future = future;
+    this.contextWeakReference = contextWeakReference;
+    this.activityWeakReference = activityWeakReference;
```
alex-signal, line 44: Activity and Context are going to be the same object here so no need to capture both, we can simply capture the activity.

src/org/thoughtcrime/securesms/util/task/InitializeSecurityAsyncTask.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.io.IOException;
+import java.lang.ref.WeakReference;
+
+/**
+* AsyncTask class to initialize the security of a conversation in background
+**/
+public class InitializeSecurityAsyncTask extends AsyncTask<Recipient, Void, boolean[]> {
+
+
```
alex-signal, line 23: extra space

src/org/thoughtcrime/securesms/util/task/InitializeSecurityAsyncTask.java
```diff
@@ -0,0 +1,89 @@
...
+  private long threadId;
+
```
alex-signal, line 25: These spaces are unnecessary, all of these should be final, and they should be aligned as per our style:

```
private final boolean                currentIsDefaultSms
private final SettableFuture<Boolean future;
```

src/org/thoughtcrime/securesms/util/task/InitializeSecurityAsyncTask.java
```diff
@@ -0,0 +1,89 @@
...
+    this.currentIsDefaultSms = currentIsDefaultSms;
+    this.future = future;
+    this.contextWeakReference = contextWeakReference;
+    this.activityWeakReference = activityWeakReference;
+    this.tag = tag;
+  }
+
+  @Override
+  protected boolean[] doInBackground(Recipient... params) {
+    Context context = contextWeakReference.get();
```
alex-signal, line 50: alignment:

```
Context   context   = contextWeakReference.get();
Recipient recipient = params[0].resolve()
```

src/org/thoughtcrime/securesms/util/task/InitializeSecurityAsyncTask.java
```diff
@@ -0,0 +1,89 @@
...
+    if (recipient.isPushGroup()) {
+      Log.i(tag, "Push group recipient...");
+      registeredState = RecipientDatabase.RegisteredState.REGISTERED;
+    } else {
+      Log.i(tag, "Checking through resolved recipient");
+      registeredState = recipient.resolve().getRegistered();
+    }
+
+    Log.i(tag, "Resolved registered state: " + registeredState);
+    boolean signalEnabled = TextSecurePreferences.isPushRegistered(context);
```
alex-signal, line 64: This could actually be activity.getApplicationContext() instead of storing a weak reference to context.

alex-signal: Also, if context is null, this will throw a null pointer exception?

src/org/thoughtcrime/securesms/util/task/InitializeSecurityAsyncTask.java
```diff
@@ -0,0 +1,89 @@
...
+    Log.i(tag, "Returning registered state...");
+    return new boolean[]{registeredState == RecipientDatabase.RegisteredState.REGISTERED && signalEnabled,
+        Util.isDefaultSmsProvider(context)};
+  }
+
+  @Override
+  protected void onPostExecute(boolean[] result) {
+    if (result[0] != currentSecureText || result[1] != currentIsDefaultSms) {
+      Log.i(tag, "onPostExecute() handleSecurityChange: " + result[0] + " , " + result[1]);
+      activityWeakReference.get().handleSecurityChange(result[0], result[1]);
```
alex-signal, line 84: If the activity is null, this will throw a null pointer exception

src/org/thoughtcrime/securesms/util/task/SaveDraftAsyncTask.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private Recipient recipient;
+
+  private int thisDistributionType;
+
+  private DraftDatabase.Drafts drafts;
+
+  private SettableFuture<Long> future;
+
+  public SaveDraftAsyncTask(WeakReference<Context> contextWeakReference, WeakReference<ConversationActivity> activityWeakReference, Recipient recipient, int thisDistributionType, DraftDatabase.Drafts drafts, SettableFuture<Long> future) {
```
alex-signal, line 31: Constructors in both classes should have nullability annotations on their parameters.

src/org/thoughtcrime/securesms/util/task/SaveDraftAsyncTask.java
```diff
@@ -0,0 +1,64 @@
...
+    this.contextWeakReference = contextWeakReference;
```
alex-signal, line 32: same comment about context, if it's the application context, we can just pass it in as-is. The memory leak would have been due to keeping the activity in memory.

src/org/thoughtcrime/securesms/util/task/SaveDraftAsyncTask.java
```diff
@@ -0,0 +1,64 @@
...
+    this.recipient = recipient;
+    this.thisDistributionType = thisDistributionType;
+    this.drafts = drafts;
+    this.future = future;
+  }
+
+  @Override
+  protected Long doInBackground(Long... params) {
+    ThreadDatabase threadDatabase = DatabaseFactory.getThreadDatabase(contextWeakReference.get());
```
alex-signal, line 41: Null pointer exception if the weak reference has been cleared.

src/org/thoughtcrime/securesms/util/task/SaveDraftAsyncTask.java
```diff
@@ -0,0 +1,64 @@
...
+    DraftDatabase draftDatabase = DatabaseFactory.getDraftDatabase(contextWeakReference.get());
+    long threadId = params[0];
+
+    if (drafts.size() > 0) {
+      if (threadId == -1)
+        threadId = threadDatabase.getThreadIdFor(recipient, thisDistributionType);
+
+      draftDatabase.insertDrafts(threadId, drafts);
+      threadDatabase.updateSnippet(threadId, drafts.getSnippet(contextWeakReference.get()),
```
alex-signal, line 50: formatting for method calls on multiple lines:

```
threadDatabase.updateSnippet(threadId,
                             drafts...
```

src/org/thoughtcrime/securesms/util/task/InitializeSecurityAsyncTask.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+    future.set(true);
+    activityWeakReference.get().onSecurityUpdated();
```
alex-signal, line 87: This strongly couples this task to the activity. Instead, we should create a callback interface which describes what methods this needs, and pass an implementation of it (which can be the activity, or an inner object in the activity)

```
interface Callback {
    void handleSecurityChange(InitializeSecurityAsyncTaskResult)
    void onSecurityUpdated()
}
```

src/org/thoughtcrime/securesms/util/task/SaveDraftAsyncTask.java
```diff
@@ -0,0 +1,64 @@
...
+    } else if (threadId > 0) {
+      threadDatabase.update(threadId, false);
+    }
+
+    return threadId;
+  }
+
+  @Override
+  protected void onPostExecute(Long result) {
+    future.set(result);
```
alex-signal, line 62: Consider the callback pattern mentioned above, instead of this future (the callback would wrap the future and set the result on it.

This will prevent possible leak if the future outlives the activity, since this class can keep a weak reference to the callback.

Just keep in mind that if your callback implementor is an object in the activity and not the activity itself, the activity needs to keep a strong reference to it, otherwise it'll get GC'd and youi'll never invoke it.

alex-signal (changes requested):

PR #9497, opened 2020-03-22
Added a choice between Google Map types

johanw666:
### Contributor checklist

 * Sony Z3 Compact, Android 6.0.1
 * Sony M4 Aqua Android 5.0

----------

### Description

Signal-maptype

Screenshot_20200322-125505

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/components/location/SignalMapView.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/location/SignalMapView.java
@@ -18,6 +18,8 @@ import com.google.android.gms.maps.OnMapReadyCallback;
 import com.google.android.gms.maps.model.MarkerOptions;
 
 import org.thoughtcrime.securesms.R;
+import org.thoughtcrime.securesms.dependencies.ApplicationDependencies;
+import org.thoughtcrime.securesms.util.TextSecurePreferences;
 import org.thoughtcrime.securesms.util.ViewUtil;
 import org.thoughtcrime.securesms.util.concurrent.ListenableFuture;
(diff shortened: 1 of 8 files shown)


Meteor0id:
As a viewer of a location which was shared to me, I would like to be able to toggle back and forth between satellite and roadmap while viewing the map. I do this often, it often really helps find the right location much faster. A setting under options would not allow me to toggle back and forth. Which map type I want to see depends on the location and circumstances, so that should not be a global setting.

johanw666:
> As a viewer of a location which was shared to me, I would like to be able to toggle back and forth between satellite and roadmap while viewing the map.

Something with toggle buttons on the selection screen? Like this:

Screenshot_20200329-011026

johanw666:
I've aded 3 type selection buttons to the place picker (more looks cramped IMO). They also change the setting so the change is persistent.
Screenshot_20200331-160554

halb9:
Would it be possible to have the layer symbol just like in the google maps app (top right corner)? 

eff01
eff02

johanw666:
> Would it be possible to have the layer symbol just like in the google maps app (top right corner)?

That could be done but you are asking for much more: not only a selectable map type, but only a layer selector like Google Maps has. The advantage of this solution is that quick switching is possible.

Meteor0id:
A choice between layers is still missing in Signal

Meteor0id:
Still missing

greyson-signal:
Hi, thanks for the time ya'll spent on this, but I don't think this is how we'll be doing this.

Meteor0id:
Hope there will be some viewing choices for maps in the future. Sometimes a satelite view is very useful.

PR #9591, opened 2020-05-06
Fix Gradle validation

hannesa2:
When you open project with Android Studio, you run into

image

Now the Gradle verification will be done with a Github action


alan-signal:
1. This is not a fix. The fix is click the "Use 027fdd... as checksum".
2. If you look [here](https://github.com/signalapp/Signal-Android/blob/master/.github/workflows/android.yml#L22) you will see we are already using the gradle verify action.
3. You removed gradle witness too! This is an important plugin.

PR #9592, opened 2020-05-06
Run Espresso tests on pull requests

hannesa2:
The headline says everything

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -4,21 +4,28 @@ on:
   pull_request:
   push:
     branches:
-    - 'master'
-    - '4.**'
+      - 'master'
+      - '4.**'
 
 jobs:
   build:
-
-    runs-on: ubuntu-latest
-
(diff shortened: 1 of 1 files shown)


alan-signal:
Why changed over to mac? That has cost implications.

hannesa2:
only there you have x86 acceleration

alan-signal:
> only there you have x86 acceleration

What is the difference in speed?

hannesa2:
> That has cost implications.

If you think it has some, you should close it

hannesa2:
> What is the difference in speed?

You ask about speed comparing between `arm` and `x86` with hardware acceleration ?

Btw, you can test if this is necessary, or the action already has it

alan-signal:
FYI macos instance costs 10x linux!

You can put the `runs-on` back to linux and we will just see how long the actions take to execute.

hannesa2:
> FYI macos instance costs 10x linux!

Please show me the source of this info

PR #9594, opened 2020-05-06
Remove obsolete buildTools

hannesa2:
The variable was nowhere used


alan-signal:
- It's a method not a variable.
- It's not obsolete, it overrides the default specified by the android gradle plugin https://developer.android.com/studio/releases/build-tools 

I will consider whether we need to be doing that however.

PR #9595, opened 2020-05-06
Update libraries

hannesa2:
Modernize the app by update libs


alan-signal:
Thank you, but each update needs to be done with care and for a particular reason/benefit.

Non-exhaustive examples of cases include: if there is a particular bug with a version we are using that manifests inside Signal, or a new feature requires a later version of a library.

PR #9694, opened 2020-06-04
Replaced a hardcoded string with an existing string id

fumiakiy:

### Contributor checklist

 * Virtual Device Pixel2 API 28 on ADB
 * Essential PH-1, Android 10

----------

### Description
Replaced a hardcoded string with an existing string id that is very similar to the hardcoded one.

I am aware that this pull request may not be what the team and the issue author ultimately want to implement by the following arguments:

1. The author states in the issue that the label may be redundant because the edit text below the label has the same text as the placeholder (aka hint).
2. The original, hardcoded string was `Phone Number` but this pull request will replace it with `Phone number` (lowercase `n`) for users of English locale (other locale users will now see this in their language).

The reasons I did this in this way are:

1. Removing a label seems to me a big decision that I am in no place to make.
1. But showing something in a different language from what are in other places could make the app look broken or immature to some users which I would not like them to see.
1. Adding another string to translate just for the upper/lower casing may not justify the effort by translators. And late translations would result in the same situation I worried in the above bullet point.

But please do let me know if I should have done it in a different way. I am willing to make the change to this PR. I also don't feel bad at all if this PR is closed without being merged because of a different point of view.

### DIFF
--- a/app/src/main/res/layout/fragment_registration_enter_phone_number.xml
+++ b/app/src/main/res/layout/fragment_registration_enter_phone_number.xml
@@ -63,7 +63,7 @@
                 android:layout_height="wrap_content"
                 android:layout_weight="3"
                 app:labeledEditText_background="@color/white"
-                app:labeledEditText_label="Phone Number"
+                app:labeledEditText_label="@string/RegistrationActivity_phone_number_description"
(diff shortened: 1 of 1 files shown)


alan-signal:
Merged in 3bbf0741ee4ddc56311061621f7c8064662610d0

Thank you!

PR #9706, opened 2020-06-06
Avoid unnecessary object allocation upon early receipt cache miss.

2ob:

### Contributor checklist

 * Virtual device Pixel 3a, Android 9

----------

### Description

The `Map<RecipientId, Long>` that is returned by `EarlyReceiptCache.remove(long)` is never written into, so it is safe to replace the `new HashMap` with an immutable, statically allocated, empty map.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/database/EarlyReceiptCache.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/database/EarlyReceiptCache.java
@@ -6,6 +6,7 @@ import org.thoughtcrime.securesms.logging.Log;
 import org.thoughtcrime.securesms.recipients.RecipientId;
 import org.thoughtcrime.securesms.util.LRUCache;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/thoughtcrime/securesms/database/EarlyReceiptCache.java
```diff
@@ -47,6 +48,6 @@ public synchronized void increment(long timestamp, @NonNull RecipientId origin)
     Log.i(TAG, this+"");
     Log.i(TAG, String.format(Locale.US, "Checking early receipts (%d): %d", timestamp, receipts == null ? 0 : receipts.size()));
 
-    return receipts != null ? receipts : new HashMap<>();
+    return receipts != null ? receipts : Collections.emptyMap();
```
alex-signal, line 51: If we're making the null case immutable, we should do the same for the non-null case to be consistent and ideally get some earlier detection of potential bugs. We'll want to wrap `recipients` with `Collections.unmodifiableMap`

2ob:
Updated to wrap the non-empty return value with `Collections.unmodifiableMap` as per the review comments.

PR #9750, opened 2020-06-17
Stop calling deprecated method to update context in newer OS

fumiakiy:
### Contributor checklist

 * Virtual device Pixel 2, Android 28

----------

### Description

Steps to verify the fix
1. Prepare a device that has API 28 or newer.
1. Launch Signal.
1. Tap on your profile picture to open Settings.
1. Tap on Appearance.
1. Tap on Language.
1. Tap on Japanese (or Chinese or any other language that is not the current one).
1. Either tap on the left arrow on the action bar or tap the back button on the device to go back to Settings.
1. Again, either tap on the left arrow on the action bar or tap the back button on the device to go back to the conversation list.
1. Tap on the magnifier icon on the action bar.
1. If you see the hint in the edit text translated in the language you choose at the step above, repeat from the step 3 one more time.
  * I saw sometimes the bug reproduces at the first iteration, but most of the times I see this bug reproduces after the second iteration.

Actual behavior of the master @[dd717b60b8936d9e237ba78b654f99c0777c4f50]:

In most cases, after the second iteration, the hint text in the edit text field is not translated in the current language that you choose.

Expected behavior after this PR:

Every time you choose a language, the hint text in the edit text field is translated in that language.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/util/dynamiclanguage/DynamicLanguageContextWrapper.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/util/dynamiclanguage/DynamicLanguageContextWrapper.java
@@ -20,9 +20,7 @@ public final class DynamicLanguageContextWrapper {
     final Configuration config    = resources.getConfiguration();
     final Configuration newConfig = copyWithNewLocale(config, newLocale);
 
-    resources.updateConfiguration(newConfig, resources.getDisplayMetrics());
-
-    return context;
+    return context.createConfigurationContext(newConfig);
   }
 
   private static Configuration copyWithNewLocale(Configuration config, Locale locale) {


app/src/main/java/org/thoughtcrime/securesms/util/dynamiclanguage/DynamicLanguageContextWrapper.java
```diff
@@ -20,9 +21,12 @@ public static Context updateContext(Context context, String language) {
     final Configuration config    = resources.getConfiguration();
     final Configuration newConfig = copyWithNewLocale(config, newLocale);
 
-    resources.updateConfiguration(newConfig, resources.getDisplayMetrics());
-
-    return context;
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+      return context.createConfigurationContext(newConfig);
+    } else {
+      resources.updateConfiguration(newConfig, resources.getDisplayMetrics());
```
alan-signal, line 27: It's only "deprecated as of API 25":

```
'updateConfiguration(android.content.res.Configuration, android.util.DisplayMetrics)' is deprecated as of API 25: Android 7.1.1 (Nougat) 
```

But you've used 24, so I thought maybe that was the min api level of `createConfigurationContext` but it seems to be <= 19.

So is there a reason we can't always use `createConfigurationContext`?

fumiakiy: > So is there a reason we can't always use createConfigurationContext?

Not really. I was able to test it work as expected in Android 7.0 virtual device, as well as Android 10 device, but was not able to give a shot in other versions. So I thought it would be safer to keep it intact for the older versions. And it was my mistake that it should have either been `> N` or `>= N_MR1`.

Yes I'm all for updating the code which I will do later. Thanks for the review!

alan-signal:
Merged in 5f0d384c9ef65b84ce18c10c994a9f4aca69e23c

Thank you!

Issue #9736, opened 2020-06-14
Language setting isn't reflected


a2y4:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Language setting isn't reflected until Signal will resume.

### Steps to reproduce
- Open Signal
- Go to Settings -> Appearance -> Language
- Change language to something (e.g. English to Japanese)

**Actual result:** Menu is shown in new language, but Search Box and Inbox is shown in old language.
**Expected result:** All text will be shown in new language.
### Screenshots

<img src="https://user-images.githubusercontent.com/53612225/84586032-a8c01a00-ae50-11ea-884f-907c9a840ba0.jpg" width="260"><img src="https://user-images.githubusercontent.com/53612225/84586035-ac53a100-ae50-11ea-9e24-03e40328b13e.jpg" width="260">

### Device info

**Device:** Samsung Galaxy S9 (SCV38)
**Android version:** 10
**Signal version:** 4.63.2

brownstic:
Hey, there are some languages where not all strings have been translated. Also, please attach your debuglogs to make it easier for the developers to find out what's wrong 🙂

fumiakiy:
Fixed by https://github.com/signalapp/Signal-Android/pull/9750

PR #9752, opened 2020-06-17
Don't cache locale in conversation

fumiakiy:
### Contributor checklist

 * Virtual device Pixel 2, Android 9 (API 28)

----------

### Description

Found while debugging #9736. While PR #9750 fixes the issue, I found that the date text in each conversation thread in the list of MainActivity is not translated in the language that was just selected. 

The bug was due to the locale object cached in the list adapter which was set by `Locale.getDefault()` at the moment it was first created, and survives `activity.recreate()`.

### Steps to verify the fix
1. Prepare an account that has one or more conversation.
1. Launch Signal.
1. Tap on your profile picture to open Settings.
1. Tap on Appearance.
1. Tap on Language.
1. Tap on Japanese (or Chinese or any other language that is not the current one).
1. Either tap on the left arrow on the action bar or tap the back button on the device to go back to Settings.
1. Again, either tap on the left arrow on the action bar or tap the back button on the device to go back to the conversation list.

Behavior at master @ [dd717b60b8936d9e237ba78b654f99c0777c4f50]:

Date text is not translated to the language that was selected.

Behavior after the PR:

Every time you choose a language, the date text in each conversation thread in the list is translated in that language.

### Note
This change effectively makes `Conversation` class the same as `ThreadRecord` class. It seems the purpose of two classes are so different, and also to be compliant with making a PR as small as possible, I did not remove Conversation class.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversationlist/model/Conversation.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversationlist/model/Conversation.java
@@ -4,8 +4,6 @@ import androidx.annotation.NonNull;
 
 import org.thoughtcrime.securesms.database.model.ThreadRecord;
 
-import java.util.Objects;
-
 public class Conversation {
   private final ThreadRecord threadRecord;
 
@@ -27,6 +25,6 @@ public class Conversation {
 
   @Override
   public int hashCode() {
-    return Objects.hash(threadRecord);
+    return threadRecord.hashCode();
   }
 }
--- a/app/src/main/java/org/thoughtcrime/securesms/conversationlist/model/Conversation.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversationlist/model/Conversation.java
@@ -4,24 +4,17 @@ import androidx.annotation.NonNull;
 
 import org.thoughtcrime.securesms.database.model.ThreadRecord;
 
-import java.util.Locale;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/thoughtcrime/securesms/conversationlist/model/Conversation.java
```diff
@@ -4,37 +4,29 @@
...
     Conversation that = (Conversation) o;
-    return threadRecord.equals(that.threadRecord) &&
-           locale.equals(that.locale);
+    return threadRecord.equals(that.threadRecord);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(threadRecord, locale);
+    return Objects.hash(threadRecord);
```
alan-signal, line 30: As we're down to one arg, just do:

```
return threadRecord.hashCode();
```

fumiakiy: Of course! My bad. Should I do that now to the branch?

alan-signal: sure

alan-signal:
Merged in 66f26683269c9cbae9ad1c2eb5dc81985af2ef86

Thank you!

PR #9766, opened 2020-06-20
Optimization of runtime in generateWaveForm

Sgn-32:

### Contributor checklist
 * Motorola X Play, Android 7.1.1
 * Virtual Pixel 2, Android 7.1.1
 * Virtual Pixel 2, Android 10.0+

----------

### Description
On some devices, especially my device, it takes a long time to generate the AudioWaveForm/AudioHash of a voice message.
A voice message of 30 minutes has taken ~112sec on stock Signal, ~160sec with self-built Signal.
The cause was the frequent iterative use of extractor.advance(), which seems to be no problem with other devices.
Instead of extractor.advance() to go through each frame, seekTo(...) is used to go directly to the desired frame whose values are to be used for the audio hash.
Two further values were used for this: BAR_FACTOR and barTimer.
BAR_FACTOR contains the time difference between the desired frames.
barTimer contains the time of the actual frame.

In addition, the newer type 
ByteBuffer inputBuffer = codec.getInputBuffer(inputBufferId);
ByteBuffer outputBuffer = codec.getOutputBuffer(outputBufferId);

is used instead of the deprecated processing with buffer arrays
ByteBuffer[] inputBuffers = codec.getInputBuffers();
ByteBuffer[] outputBuffers = codec.getOutputBuffers();

The time for generation is now ~3sec for a 30 minutes voice message with 99%/100% the same values for the AudioHash.

further note:
seekTo() does not seem to recognize whether the entered time goes beyond the stream.
Therefore, when using it, make sure that the entered time remains below the duration of the audio message.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/audio/AudioWaveForm.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/audio/AudioWaveForm.java
@@ -167,8 +167,6 @@ public final class AudioWaveForm {
       boolean               sawInputEOS     = false;
       boolean               sawOutputEOS    = false;
       int                   noOutputCounter = 0;
-      final long            BAR_FACTOR      = (totalDurationUs / (BAR_COUNT * SAMPLES_PER_BAR) + 1);
-      long                  barTimer        = BAR_FACTOR;
 
       while (!sawOutputEOS && noOutputCounter < 50) {
         noOutputCounter++;
@@ -179,13 +177,11 @@ public final class AudioWaveForm {
             int        sampleSize         = extractor.readSampleData(dstBuf, 0);
             long       presentationTimeUs = 0;
 
-            if (sampleSize < 0 || barTimer > totalDurationUs) {
+            if (sampleSize < 0) {
               sawInputEOS = true;
               sampleSize  = 0;
             } else {
               presentationTimeUs = extractor.getSampleTime();
-              extractor.seekTo(barTimer, MediaExtractor.SEEK_TO_NEXT_SYNC);
-              barTimer  += BAR_FACTOR;
             }
 
             codec.queueInputBuffer(
@@ -194,6 +190,18 @@ public final class AudioWaveForm {
               sampleSize,
               presentationTimeUs,
               sawInputEOS ? MediaCodec.BUFFER_FLAG_END_OF_STREAM : 0);
+
+            if (!sawInputEOS) {
+              int barSampleIndex = (int) (SAMPLES_PER_BAR * (wave.length * extractor.getSampleTime()) / totalDurationUs);
+              sawInputEOS = !extractor.advance();
+              int nextBarSampleIndex = (int) (SAMPLES_PER_BAR * (wave.length * extractor.getSampleTime()) / totalDurationUs);
+              while (!sawInputEOS && nextBarSampleIndex == barSampleIndex) {
+                sawInputEOS = !extractor.advance();
+                if (!sawInputEOS) {
+                  nextBarSampleIndex = (int) (SAMPLES_PER_BAR * (wave.length * extractor.getSampleTime()) / totalDurationUs);
+                }
+              }
+            }
           }
         }
 
--- a/app/src/main/java/org/thoughtcrime/securesms/audio/AudioWaveForm.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/audio/AudioWaveForm.java
@@ -167,6 +167,8 @@ public final class AudioWaveForm {
       boolean               sawInputEOS     = false;
       boolean               sawOutputEOS    = false;
       int                   noOutputCounter = 0;
+      final long            BAR_FACTOR      = (totalDurationUs / (BAR_COUNT * SAMPLES_PER_BAR) + 1);
+      long                  barTimer        = BAR_FACTOR;
 
       while (!sawOutputEOS && noOutputCounter < 50) {
         noOutputCounter++;
@@ -177,11 +179,13 @@ public final class AudioWaveForm {
             int        sampleSize         = extractor.readSampleData(dstBuf, 0);
             long       presentationTimeUs = 0;
 
-            if (sampleSize < 0) {
+            if (sampleSize < 0 || barTimer > totalDurationUs) {
               sawInputEOS = true;
               sampleSize  = 0;
             } else {
               presentationTimeUs = extractor.getSampleTime();
+              extractor.seekTo(barTimer, MediaExtractor.SEEK_TO_NEXT_SYNC);
+              barTimer  += BAR_FACTOR;
             }
 
             codec.queueInputBuffer(
@@ -190,18 +194,6 @@ public final class AudioWaveForm {
               sampleSize,
               presentationTimeUs,
               sawInputEOS ? MediaCodec.BUFFER_FLAG_END_OF_STREAM : 0);
-
-            if (!sawInputEOS) {
-              int barSampleIndex = (int) (SAMPLES_PER_BAR * (wave.length * extractor.getSampleTime()) / totalDurationUs);
-              sawInputEOS = !extractor.advance();
-              int nextBarSampleIndex = (int) (SAMPLES_PER_BAR * (wave.length * extractor.getSampleTime()) / totalDurationUs);
-              while (!sawInputEOS && nextBarSampleIndex == barSampleIndex) {
-                sawInputEOS = !extractor.advance();
-                if (!sawInputEOS) {
-                  nextBarSampleIndex = (int) (SAMPLES_PER_BAR * (wave.length * extractor.getSampleTime()) / totalDurationUs);
-                }
-              }
-            }
           }
         }
 
--- a/app/src/main/java/org/thoughtcrime/securesms/audio/AudioWaveForm.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/audio/AudioWaveForm.java
@@ -167,6 +167,8 @@ public final class AudioWaveForm {
       boolean               sawInputEOS     = false;
       boolean               sawOutputEOS    = false;
       int                   noOutputCounter = 0;
+      final long            barFactorUs     = (totalDurationUs / (BAR_COUNT * SAMPLES_PER_BAR)) + 1;
+      long                  barTimeUs       = barFactorUs / 2;
 
       while (!sawOutputEOS && noOutputCounter < 50) {
         noOutputCounter++;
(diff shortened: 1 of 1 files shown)


alan-signal:
Very promising, however the diff is too big. I understand that the efficiency comes from the `seekTo` usage. Try to make the smallest change to the code that introduces that.

Sgn-32:
So, less than now (+5, -13) is not possible.
Duration is now ~5 sec.

app/src/main/java/org/thoughtcrime/securesms/audio/AudioWaveForm.java
```diff
@@ -177,11 +179,13 @@ public void getWaveForm(@NonNull Consumer<AudioFileInfo> onSuccess, @NonNull Run
             int        sampleSize         = extractor.readSampleData(dstBuf, 0);
             long       presentationTimeUs = 0;
 
-            if (sampleSize < 0) {
+            if (sampleSize < 0 || barTimer > totalDurationUs) {
```
alan-signal, line 182: Put `barTimer < totalDurationUs` in the while loop condition. Otherwise you're just skipping the seek and carrying on.

app/src/main/java/org/thoughtcrime/securesms/audio/AudioWaveForm.java
```diff
@@ -167,6 +167,8 @@ public void getWaveForm(@NonNull Consumer<AudioFileInfo> onSuccess, @NonNull Run
       boolean               sawInputEOS     = false;
       boolean               sawOutputEOS    = false;
       int                   noOutputCounter = 0;
+      final long            BAR_FACTOR      = (totalDurationUs / (BAR_COUNT * SAMPLES_PER_BAR) + 1);
+      long                  barTimer        = BAR_FACTOR;
```
alan-signal, line 171: Note that `ALL_CAPS` is reserved for class level `final static` values.

If this was simply:
```
final long barFactor = totalDurationUs / (BAR_COUNT * SAMPLES_PER_BAR);
long barTimer =  0;
```

It would give us this kind of fence post pattern:

```
|__|__|__|
```
With the last one right on `totalDurationUs`.

What you have is:

```
final long barFactor = totalDurationUs / (BAR_COUNT * SAMPLES_PER_BAR);
long barTimer =  barFactor;
```

Which puts the last fence post past `totalDurationUs`:

```
__|__|__|  |
```

What would be better is to have an equal number of fence posts to gaps, (but by creating one more gap between fence posts):

```
final long barFactor = totalDurationUs / (BAR_COUNT * SAMPLES_PER_BAR + 1);
long barTimer =  0;
```

```
|__|__|__|__
```
(I think you tried to do this, but your +1 is outside the brackets.)

And then balance it out like so:

```
final long barFactor = totalDurationUs / (BAR_COUNT * SAMPLES_PER_BAR + 1);
long barTimer =  barFactor / 2; // (note not >>1 😉)
```

```
_|__|__|__|_
```

app/src/main/java/org/thoughtcrime/securesms/audio/AudioWaveForm.java
Sgn-32: The seekTo() is skipped only in the last step, in which sawInputEOS is set and the codec.queueInputBuffer(inputBufIndex, 0, 0 ["= sampleSize"], presentationTimeUs, BUFFER_FLAG_END_OF_STREAM) is called.
It serves as an alternative if readSampleSize should not return -1.
e.g. generating the waveform without the comparison on the virtual device is no problem, on my device a endless problem.

```
2020-06-25 22:47:45.442 31940-31984 presentationTimeUs 1725276000
2020-06-25 22:47:45.467 31940-31984 presentationTimeUs 1734959000
2020-06-25 22:47:45.486 31940-31984 presentationTimeUs 1744665000
2020-06-25 22:47:45.507 31940-31984 presentationTimeUs 1754348000
2020-06-25 22:47:45.527 31940-31984 presentationTimeUs 1764054000
2020-06-25 22:47:45.542 31940-31984 presentationTimeUs 1773737000
2020-06-25 22:47:45.557 31940-31984 presentationTimeUs 1783397000
2020-06-25 22:47:45.573 31940-31984 presentationTimeUs 1783397000
2020-06-25 22:47:45.589 31940-31984 presentationTimeUs 1783397000
...
2020-06-25 22:48:03.311 31940-31984 presentationTimeUs 1783397000
2020-06-25 22:48:03.322 31940-31984 presentationTimeUs 1783397000
2020-06-25 22:48:03.334 31940-31984 presentationTimeUs 1783397000
2020-06-25 22:48:03.346 31940-31984 presentationTimeUs 1783397000
```
furthermore, the number of outputs values for the AudioHash is reduced to 182, if the comparison is in the while loop instead in the if-condition.

app/src/main/java/org/thoughtcrime/securesms/audio/AudioWaveForm.java
Sgn-32: Ok, thought ALL_CAPS would apply to all constants.

It is currently almost the same implementation as the original (extractor.advance).
The readSampleData is called before the first seekTo(barTimer),
so starting at the first frame with presentationTimeUs = 0.
Instead of 184 single values for the HashAudio, the last one has been cut, so only 183 values.
`|__|__|__`

`(totalDurationUs / (BAR_COUNT * SAMPLES_PER_BAR) + 1);`
the +1 is one microsecond, to prevent the unlikely case that the value barFactor should be 0 and the barTimer wouldn't increase. (totalDurationUs < 184 us => probability 0%)

'long barTimer =  barFactor / 2;' that does not hurt, because it divides only once.

I test the source code change furthermore, the number of outputs is reduced to 182.

alan-signal: > Ok, thought ALL_CAPS would apply to all constants.

Yes it does :) `barFactor` is potentially different value each method run, so not a constant like `BAR_COUNT`.

> I test the source code change furthermore, the number of outputs is reduced to 182.

Interesting, be nice to work out why.

Sgn-32:
```
while (...) {
  if (datasource is not EOS) {...}
  do {...} while outputBuffer not empty
}
```

The original while-loop continues to work if the inputstream is EOS, but skipping in total the input-part.
So all remaining data in the outputBuffers can be processed.
The modified while-loop directly ends if barTimer > totalDurationUs, even if data is still available in the output-part.

------------------

The calls of the inputBuffer and outputBuffer are "out of sync". At start the first frames are put in the inputbuffer, some iteration this comes later to the output.

`if (sampleSize < 0 || barTimer > totalDurationUs)`

The inputBuffer and outputBuffer are called in this order:
(presentationtime is the time which comes from extractor.getSampleTime() and put in the queueInputBuffer.
The info.presentationtime is the time from dequeueOutputBuffer)

```
inputBufIndex      1      presentationtime 0
outputBufferIndex -1 info.presentationtime 0
inputBufIndex      2      presentationtime 23220
outputBufferIndex -3 info.presentationtime 0
inputBufIndex      3      presentationtime 69660
outputBufferIndex -2 info.presentationtime 0
output format has changed to {sample-rate=44100, pcm-encoding=2, mime=audio/raw, channel-count=1}
inputBufIndex      0      presentationtime 92880
outputBufferIndex  0 info.presentationtime 0
outputBufferIndex  1 info.presentationtime 23220
outputBufferIndex -1 info.presentationtime 23220
inputBufIndex	   1      presentationtime 116100
outputBufferIndex  2 info.presentationtime 69660
outputBufferIndex -1 info.presentationtime 69660
inputBufIndex      2      presentationtime 162540
outputBufferIndex  3 info.presentationtime 92880
outputBufferIndex -1 info.presentationtime 92880
...
inputBufIndex      2      presentationtime 5503140
outputBufferIndex  3 info.presentationtime 5456700
outputBufferIndex -1 info.presentationtime 5456700
inputBufIndex      3      presentationtime 5549580
outputBufferIndex  0 info.presentationtime 5479920
outputBufferIndex -1 info.presentationtime 5479920
sawInputEOS     sampleSize = 97 barTimeUs 5611198  > totaldurationUs 5596020
inputBufIndex      0      presentationtime 0
outputBufferIndex  1 info.presentationtime 5503140
outputBufferIndex  2 info.presentationtime 5549580
outputBufferIndex -1 info.presentationtime 5549580
```
In total queueInputBuffer is called 184 times (183 times with a frame, 1 time with the EOS-flag), the queueOutputbuffer called 367 times from that 183 times with a decoded frame.

------------

Putting the barTimer < totalDurationUs in the while loop, this changes the end to:

```
inputBufIndex      2      presentationtime 5503140
outputBufferIndex  3 info.presentationtime 5456700
outputBufferIndex -1 info.presentationtime 5456700
inputBufIndex      3      presentationtime 5549580
outputBufferIndex  0 info.presentationtime 5479920
outputBufferIndex -1 info.presentationtime 5479920
inputBufIndex      0      presentationtime 5572800
outputBufferIndex  1 info.presentationtime 5503140
outputBufferIndex -1 info.presentationtime 5503140
```
In total queueInputBuffer is called 184 times (183 times with a frame, 1 time with the EOS-flag), the queueOutputbuffer called 366 times from that 182 times with a decoded frame.


This skips the processing of the last data in the outputbuffer. 
outputBufferIndex  2 info.presentationtime 5549580

------------

I am trying to find a variant in which 184 frames enter and 184 frames leave, with few diff, but only works with more than now.

alan-signal:
Finally got around to testing this and a 3 second clip took 7 seconds to generate the waveform.

I have found that the emulator gives very different results from a real device. Maybe that is the issue.

Sgn-32:
Tested on an emulator a few voice messages of 3sec, generation time is between 1,8 and 2sec.
A 1min clip also 2sec.
On my device the 3sec clips -> 4sec for the generation and a 1min clip 4,5 sec.
My test file of 30mins -> 5,5sec.

What is also a difference, on the emulator with Android 7.1.1 there was no issue with the audio duration.
Shows exactly the 30:00 and not 29:43 (my device) for an aac file.
Found also the reason for that. The sampleTime for a single aac frame is wrong on my device, is only 23000 ms instead of 23220 ms. Seems to be a bad implementation of the aac.decoder.

Would this be ready for merging or do I have to do something else?

alan-signal:
Yeah we can't merge it as it is. The performance of short clips (which I would assume are far more common than 30 minute clips) is much worse now. It should never take longer than the clip to generate the waveform.

It's probably on the right tracks with the seeking approach, but might need some improvements to the decrypted stream code to make seeking less taxing. That's my guess at the problem but I don't expect you to do this. It would be done separately to this PR and when that's done this should be able to come in but I can't give an ETA on that.

Sgn-32:
Poor performance really shouldn't be. Have also checked again, the performance of implementation with advance() is better for voice messages up to ~ 20sec. The processing of voice messages over ~20sec is better with seekTo on my mobile phone. Since every combination of mobile device type and Android version should have this point differently, possibly at ~ 5sec or 1min, it will be difficult to find an optimal solution for everyone.
On solution could be to benchmark the performance, find this "break point" and for each generation for the waveform to decide based on the voice message duration, if a method with advance() or a method with seekTo is called.
This should be related to the execution time of advance() and seekTo(). One call of advance() is faster then one call of seekTo(), but at some numbers of advance() calls, one call of seekTo is faster.

I try to improve the algorithm a bit and see which performance increases are still possible.

Issue #9044, opened 2019-09-23
Signal Android App on F-Droid store in 2020 or have specific wiki page or README section why not.


paulvi:
publish Signal Android App on F-Droid store in 2020  
or have specific wiki page or README section why not.

 - [x] I have searched open and closed issues for duplicates:  
#8469 #7505 #7988 #6292 #5995 #5853 #4430 #3954 #3435 #2858 #525 #281 
and of cause #127 and searched https://community.signalusers.org/search?q=F-droid

 - [ ] I am submitting a bug report for existing functionality that does not work as intended
 - [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
 - [ ] This isn't a feature request or a discussion topic

----------------------------------------

Major reason is that Google Play is not available in China.   
And with all proper warnings direct download https://signal.org/android/apk/ is also not what most users asked for.

There is no single point of reference what is holding publishing on F-Droid.
Some older things may be just outdated.

A wiki page or README section could be good to express authors view.
Making users create such issues again, closing it and referring  to forum is making waste of time. Most developers on GitHub just can wait for an issue to be resolved.

Pieces of good thought, old and newer problems are scattered within many places.
If authors here are busy to summarize situation, let me start the F-Droid page.

Hopefully this issue will be closed, when at least wiki page exists.

greyson-signal:
Many of the reasons provided in the issues you linked to are still valid. 

Here's a simple one: We want to be able to sign releases with our own keys. This would require us to host our own F-Droid repository, at which point people would need to know to add our repository, at which point they might as well download our APK directly, which already has an auto-update mechanism built in, which is all we'd be getting from F-Droid at that point.

But as we've stated previously, we use the forum for discussions, not Github. Github is for tracking bugs. Thank you!

paulvi:
Do you mind me starting wiki page here
https://github.com/signalapp/Signal-Android/wiki/F-Droid
?

eighthave:
@greyson-signal I've been involved in these discussions since the beginning.  The only issue blocking  Signal from F-Droid is that it is not free software.  It includes proprietary Google libraries.  Any app can be signed by the developer's key and included in F-Droid as long as it is 100% free software and can be built reproducibly, like Briar, Bitcoin Wallet, Offi, etc.

paulvi:
@eighthave Hans, could you point where exactly Google libraries are ?
[...]

PR #9802, opened 2020-06-29
F-Droid Store

rmotao:
Upload Signal(4 Android) on **F-Droid** Store!


PiCob:
See here: https://github.com/signalapp/Signal-Android/issues/9044#issuecomment-536159877

PR #9898, opened 2020-08-01
Do not collapse list to hide only one entry

cascheberg:
### Contributor checklist

 * Virtual device, Android 8.1

----------

### Description
Currently if you have a group of exactly 6 members the group member list will show 5 group members and one additional entry to expand the list. So the collapsed list has 6 entries, but the full list has 6 entries as well. It makes more sense to collapse the list only if two or more entries are hidden.

With this PR:
- up to 6 group members: all members are shown immediately
- 7 or more members:  show 5 members and allow to expand

I chose the parameters so that the list has a maximum number of 6 entries, as it is now. Alternatively you could of course use parameters 5/4 to limit the number of initially shown members to a maximum of 5 (as it is now as well).

And obviously instead of introducing a new variable you could just subtract 1, but I thought that this was nicer.


alan-signal:
Merged in 4271700046d4ac2efc4a348fce5f8ff28b611d3d

Thank you!

PR #10132, opened 2020-10-25
Fixes a bug that may generate a broken unicode character when trimming a string

### LINKED ISSUE #10076: Emoji Not Shown as Last Character in Maximum Length Group Name
### Bug description
When trying to rename the group name - to the maximum number of characters allowed - having an emoji as a last character won't work, as it would show as a question mark, like so: �

Note, if the group name is a few characters short of maximum length, then having an emoji as the last character works.

I'm using English on my phone.

Suspect it is something to do with encoding.

### Steps to reproduce
- Go to a group chat and edit its name.
- Type in the name as long as possible to the maximum length.
- Delete the last character.
- Try to type an emoji instead as the last character.

**Actual result:** The last character shows as the question mark.
**Expected result:** The emoji would show.

### Device info
**Device:** Google Pixel 2 XL
**Android version:** 11
**Signal version:** 4.72.5

fumiakiy:
### Contributor checklist

 * AVD, Android 11 (API 30)
 * AVD, Android 10 (API 29)
 * AVD, Android 9 (API 28)
 * AVD, Android 6 (API 23)
 * AVD, Android 5.1 (API 21)

----------

### Description
As is discussed correctly [here](https://github.com/signalapp/Signal-Android/issues/10076#issuecomment-708886589) the current trimToFit implementation uses `substring` method which does not take surrogate pairs and grapheme clusters into account, so that results in a broken unicode character when it is trimmed.

This PR rewrites `trimToFit` method to take unicode characters of a surrogate pair and a grapheme cluster in mind.

Fixes #10076

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/util/StringUtil.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/util/StringUtil.java
@@ -1,17 +1,22 @@
 package org.thoughtcrime.securesms.util;
 
+import android.os.Build;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
 import androidx.core.text.BidiFormatter;
 
 import com.google.android.collect.Sets;
 
+import org.thoughtcrime.securesms.logging.Log;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.Collections;
 import java.util.Set;
 
 public final class StringUtil {
-
+  private static final String         TAG        = StringUtil.class.getSimpleName();
   private static final Set<Character> WHITESPACE = Sets.newHashSet('\u200E',  // left-to-right mark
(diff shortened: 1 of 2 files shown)


alan-signal:
@fumiakiy Thanks for this, I've been able to use some of this in 2d39e436770f9d3cc053d1491cb00e6c3f1758cf and put you down as co-author, I'd appreciate your review of that commit.

The approach has changed and we're now limiting group name length as 32 grapheme clusters, so it's the same apparent length no matter what your language or emoji usage is.

I'll be adding same approach to profiles next.

fumiakiy:
@alan-signal The commit looks good to me. I am glad to see that it's now the length of the "words" not bytes any more. Using EditText's input filter is great too and the iterator implementation looks better than mine. 👍 

I think I should just close this PR at this point but let me know if you don't want me too.

alan-signal:
@fumiakiy thanks! I still need to put a byte limit on there as well, so I still might yet grab more from here, if only the unit tests.

PR #10245, opened 2020-12-03
Fix prettyPrint/prettyPrintFormat for e164 == null

Sgn-32:
### Contributor checklist

 * Virtual device Pixel 2, Android 11.0

----------

### Description

Prevent to try to pretty print format of phone numbers when e164 == null.

According to the @NonNull-annotation, both functions (PrettyPrint, PrettyPrintFormat) should receive non-null values, which is not the case for groups.

As an example a comment from the 4.76 beta thread:
https://community.signalusers.org/t/beta-feedback-for-the-upcoming-android-4-76-release/18126/3

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/phonenumbers/PhoneNumberFormatter.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/phonenumbers/PhoneNumberFormatter.java
@@ -87,11 +87,14 @@ public class PhoneNumberFormatter {
     this.localCountryCode = localCountryCode;
   }
 
-  public static @NonNull String prettyPrint(@NonNull String e164) {
+  public static String prettyPrint(String e164) {
     return get(ApplicationDependencies.getApplication()).prettyPrintFormat(e164);
   }
 
-  public @NonNull String prettyPrintFormat(@NonNull String e164) {
+  public String prettyPrintFormat(String e164) {
+    if (e164 == null) {
+      return e164;
+    }
     try {
       Phonenumber.PhoneNumber parsedNumber = phoneNumberUtil.parse(e164, localCountryCode);
 


greyson-signal:
In what situations are we calling `PhoneNumberFormatter.prettyPrint(null)`?

Sgn-32:
I haven't been able to track everything yet, but this could be a situation:
line 441 of ConversationFragment: String subtitle = recipient.getE164().transform(PhoneNumberFormatter::prettyPrint).orNull();

Here is advanced debug log with the example of issue #10246.
For the debug log i added a Log.w(...) at the beginning of the prettyPrintFormat.

  public String prettyPrintFormat(String e164) {
    Log.w(TAG, "prettyPrintFormat " + e164);

https://debuglogs.org/64716e278d8d00946910a4f7a1b9ffe059010654fa6213363175eb7ab85278ac

greyson-signal:
@Sgn-32 Ah, yep, it's the display name stuff. I saw the issue you opened, I appreciate that you tracked this down! I've fixed it for the next major release, which will go out in the next few days. Thank you very much!

Sgn-32:
I wasn't fast enough. I've tried something that seems to work so far.

```
  public @NonNull String getDisplayName(@NonNull Context context) {
    if (this.displayName == null) {
      updateDisplayName(context);
    }
    return StringUtil.isolateBidi(this.displayName);
  }

  public void updateDisplayName(@NonNull Context context) {
    if ((this.displayName = getName(context))                       != null) return;
    if ((this.displayName = getProfileName().toString())            != null) return;
    if ((this.displayName = PhoneNumberFormatter.prettyPrint(e164)) != null) return;
    if ((this.displayName = email)                                  != null) return;
    this.displayName = context.getString(R.string.Recipient_unknown);
    return;
  }
```

But I'm looking forward to your solution.

PR #10283, opened 2020-12-12
Fix to a crash bug happens when saving audio, video and other files in devices whose API is 28 or lower

### LINKED ISSUE #10159: crash saving audio files
Signal crash if i try following:

1. Open signal 
2. Settings
3. Storage
4. Select all audio files
5. Save

https://debuglogs.org/71bb604fbef289795ca6cff4b2e2c2cc9fee8f10f12b9b0748c4a7fa618aa119

Screenshot_20201102-073825
Screenshot_20201102-073919
Screenshot_20201102-073946

fumiakiy:
### Contributor checklist

 * AVD Pixel 3a API 29
 * AVD Pixel 2 API 28
 * AVD Nexus 6 API 23
 * AVD Nexus 5X API 21

----------

### Description
Fixes a crash bug happens when a file with the same name already exists in the specified content uri by checking if the content uri already exists and rename the file until it's valid to insert. Fixes #10159

The commit [3f2b4d60fd00fc53a714e6d1f1017e71ee7627fa] fixes the issue for audio files because it changed the url to be of `file://` so that the logic that checks if the file exists and keep changing the file name until there is no more conflict, kicks in. Unfortunately that fix works only for audio files, and more importantly it kind of downgrades what the updated code was supposed to do, which was to use content uri instead of file uri when saving a file.

This PR effectively reverts the commit [3f2b4d60fd00fc53a714e6d1f1017e71ee7627fa] and adds conflict checking of the file names to content uri too when the OS is older.

## Steps to very the PR

1. Prepare a chat that has a voice note, a picture, a video and a pdf (and maybe another generic file) attachments.
1. Long tap each of those attachments and download it.
1. After downloading all of them, do that again for each.
1. Just to make sure, do that yet agin.

**Actual behavior at [42b0fe78537cc4d6fdd6fb02cec6167c1e4fe9ce]**:

Either app crashes or a toast with an error message.

**Expected behavior after the PR**:

App doesn't crash. The files are saved.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/util/SaveAttachmentTask.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/util/SaveAttachmentTask.java
@@ -4,9 +4,11 @@ import android.content.ContentResolver;
 import android.content.ContentValues;
 import android.content.Context;
 import android.content.DialogInterface.OnClickListener;
+import android.database.Cursor;
 import android.media.MediaScannerConnection;
 import android.net.Uri;
 import android.os.Build;
+import android.os.Environment;
 import android.provider.MediaStore;
 import android.text.TextUtils;
 import android.webkit.MimeTypeMap;
@@ -97,7 +99,7 @@ public class SaveAttachmentTask extends ProgressDialogAsyncTask<SaveAttachmentTa
     fileName = sanitizeOutputFileName(fileName);
 
     Uri outputUri = getMediaStoreContentUriForType(contentType);
-    Uri mediaUri  = createOutputUri(outputUri, fileName);
+    Uri mediaUri  = createOutputUri(outputUri, contentType, fileName);
 
     try (InputStream inputStream = PartAuthority.getAttachmentStream(context, attachment.uri)) {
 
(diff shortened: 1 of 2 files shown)


Xashyar:
Hey @fumiakiy, can no longer reproduce this (on API 23), can't find the relevant commit but I think it's already got addressed since https://github.com/signalapp/Signal-Android/issues/10159 is also closed by the devs.

Xashyar:
Hey @fumiakiy, can no longer reproduce this (on API 23), can't find the relevant commit but I think it's already got addressed since https://github.com/signalapp/Signal-Android/issues/10159 is also closed by the devs.

fumiakiy:
[The commit that meant to fix](https://github.com/signalapp/Signal-Android/commit/3f2b4d60fd00fc53a714e6d1f1017e71ee7627fa) #10159 was to patch the code specifically for audio files, and imho in a way that should not be. This PR [reverts the commit](https://github.com/signalapp/Signal-Android/pull/10283/files#diff-f8b99f692df14bc7e460f5371b0917822decde1781024e9e1b07a0631143134aL133) and attempts to fix the root cause in a more robust way, for all types of files.

cody-signal:
Merged via 2cca6a5afb7138aa1eee1d30aea811529dd0091d

PR #10331, opened 2020-12-23
Remove unused GMS inclusion in SQLiteDatabase.java

tw-hx:
### Description

`import com.google.android.gms.vision.Tracker;` isn't needed and seems to have snuck in there. Can be deleted.


tw-hx:
So it turns out the Signal devs themselves fixed this in https://github.com/signalapp/Signal-Android/commit/4b8d02fdba244982780b0d2cee546b867e3c786b . Closing.

PR #10357, opened 2021-01-06
Add SHIFT6mq to hardware AEC blacklist

billie80:

### Contributor checklist

 * Device A, Android X.Y.Z
 * Device B, Android Z.Y

----------

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
@@ -331,6 +331,7 @@ public class ApplicationContext extends MultiDexApplication implements DefaultLi
         add("Redmi Note 5");
         add("FP2"); // Fairphone FP2
         add("MI 5");
+        add("SHIFT6mq");
       }};
 
       Set<String> OPEN_SL_ES_WHITELIST = new HashSet<String>() {{


alan-signal:
You say you didn't test this?

Can you confirm that a) you hear an echo when calling into this device and that b) you no longer hear the echo after this change.

billie80:
Hello Alan!
Thanks for getting back to me!

Regarding a) No I can not confirm hearing an echo myself but all people calling me on Signal are hearing an echo! This was also confirmed for other Apps relying on Hardware AEC on the Manufacturers official forum. Apps using Software AEC seem to work fine. Also this problem is acknowledged by the Manufacturer. They are working on a solution but they can not give a timeframe for the fix. At the moment Signal is not usable with activated speakers due to this. Especially if both partners have this problem. Since half a year I am living with this issue and I must say it is quite annoying. Maybe the next update fixes it maybe not but in the meantime it would be great having this workaround. Even if it does not help it is doing no harm.

Regarding b) No I can not confirm that adding the device to the list fixes the issue as I have no means to test it. I thought the purpose of bug #7635 was to report issues with AEC to add affected devices to the list. No offence but with this logic no device will be added to the list until someone steps up and tests the change. The majority of users however do not have this capability also. They only can report the issue (if at all, mostly they just complain silently and look for alternatives) and hope you will fix it.

I understand that this is now your responsibility but reading through the bug Moxie and even Greyson were not so restrictive when it came to adding new devices to the list.

I also understand that you don't want to add further options or settings but in this case it probably makes sense to make this configurable instead of maintaining this Blacklist. There are new devices coming up all the time, maybe there are issues in the beginning but are fixed afterwards. Chances however are high the device stays on the blacklist even if not needed anymore. With an option this can be tested by users easily and will make your life easier because you don't have to maintain the list.

I think Signal is about securely connecting people. Being able to call someone is one of it's main features. Having an option to opt out from HAEC would probably reduce friction for a lot of users. In my case I simply could disable it until there is a fix from the Manufacturer.

Thanks for your time and reading all this.

billie80:
Phone Manufacturer fixed the Hardware AEC issue.

Issue #10409, opened 2021-01-09
Contacts not sorted correctly 


McTonderski:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Language specific letters are last in contact sorted list

### Steps to reproduce
- have contact with i.e. polish letter 

If for example name Is "Łukasz" it should be listed after contacts starting with "L" not at the end

### Device info

**Device:** KB2003
**Android version:** 11
**Signal version:** 5.0.8

bkqtnte10:
Sorting related issue: #10384

Caerbannog:
No, PR #10202 will not fix this issue, it is only concerned with finding search results.
Example result of the PR : searching for "e" will show "Mélissa" and "Éric". But "Éric" will still be sorted last, at "É" (I have tested this).

The two issues are related though.

Caerbannog:
Yes, this issue is still relevant and irritating for lots of languages others than English.
Issue #10635 is a duplicate.

There are a few connected problems in Signal caused by _glyph similary_. *Contact sorting* is related but different from:
- Contact search (addressed in #10202 and subsequent improvements).
- Group name search
- Message content search

What's blocking is a design decision by the Signal team for how to solve these problems without duplicating code.

Issue #9341, opened 2020-01-15
Upper- and lowercase accentuated characters are considered different when searching for a contact


breversa:
- [x] I have searched open and closed issues for duplicates
  

---

### Bug Description

As in the title : when either searching for a term in a conversation, or a recipient to send a message to, accentuated uppercase charactres (e.g. "É") are considered different than the same accentuated but lowercase character (e.g. "é").

Language is french.

### Steps to Reproduce

1.  Open Signal
2.  Tap the "Search field"
3.  Type (for instance) "éric"

Actual Result:

My contact "Éric" is not shown.

Expected Result:

My contact "Éric" is shown.
I need to explicitely type "Éric" (or "ÉRIC", or any other upper/lowercase combination except for "É")
Also, typing "eric" or "Eric" or "ERIC" (etc.) should also display "Éric".

### Platform Info

Signal Version:

Signal 4.52.4
Not sure about Signal Desktop.

Operating System:

Android 9 (LineageOS 16)

Linked Device Version:

pericleskb:
This must have to do with SQLite. As it says here https://www.sqlite.org/lang_expr.html#like

> SQLite only understands upper/lower case for ASCII characters by default. The LIKE operator is case sensitive by default for unicode characters that are beyond the ASCII range. For example, the expression 'a' LIKE 'A' is TRUE but 'æ' LIKE 'Æ' is FALSE.

i7u:
This bug occurs for any non-ASCII search term, breaking search in most non-English languages. Seeing the same with non-accented characters.

second-reality:
I tried to notify dev who closed this MR so he can integrate this as well:

https://github.com/signalapp/Signal-Android/pull/9692#issuecomment-727223772

Indeed, that would make contact search much more usable in french (and any language with accented names).

chtixof:
I confirm that in French, characters with accents should be sorted in the same place as normal characters and, sure, be found if searched without accent. The correction of this bug will make the app a lot easier for French users. Thanks.

Booteille:
Is there any plan to improve usability of search, in French and other languages having a lot of accents, by considering accentuated characters as the same as non-accentuated characters?
It's always a bit frustrating to searching for "etienne" and that Signal is not finding "étienne"...

Caerbannog:
This issue was solved by https://github.com/signalapp/Signal-Android/pull/10202 and released in 5.9.1.
It can be closed.

Booteille:
[...]

PR #10446, opened 2021-01-10
Ensuring case-insensitive sorting for contact names

### LINKED ISSUE #10384: Contacts sorted case sensitive

### Bug description
Contacts are sorted according to name. Contacts that start with a lower case letter are not in normal order but appear at the end.

### Steps to reproduce
- have a few contacts in your phone that also have signal
- rename one of these contacts with lowercase letters
- refresh contacts list
- observe that lower case contacts are listed at the end of the alphabet

**Actual result:** lower case contacts are listed at the end of the alphabet
**Expected result:** sorting should only consider the alphabet, not the capitalisation of letters

### Device info
**Signal version:** 5.0.8

ChewTeaYeah:
Fixes #10384

### Contributor checklist

 * Device Pixel 2XL, Android 11

----------

### Description

This PR fixes a sorting issue with contacts retrieved from the on-device SQLite database. Contact sorting via ORDER BY was previously not requesting case-insensitive sorting, so contacts with lowercase names were appearing at the end of the contact list rather than collated with other contacts starting with the same letter in uppercase.

This fix was tested by first reproducing the issue on a Pixel 2XL running Android 11, changing the order by clause, and then re-installing Signal and refreshing contacts to confirm that a contact with a lowercase name was located with other contacts beginning with the same letter in uppercase rather than at the end of the contacts list

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/database/RecipientDatabase.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/database/RecipientDatabase.java
@@ -134,6 +134,8 @@ public class RecipientDatabase extends Database {
   private static final String IDENTITY_STATUS          = "identity_status";
   private static final String IDENTITY_KEY             = "identity_key";
 
+  private static final String SQLITE_COLLATE_NOCASE = "COLLATE NOCASE";
+
   private static final class Capabilities {
     static final int BIT_LENGTH = 2;
 
@@ -2093,7 +2095,9 @@ public class RecipientDatabase extends Database {
       args       = new String[] { "0", String.valueOf(RegisteredState.REGISTERED.getId()), "1", Recipient.self().getId().serialize() };
     }
 
-    String   orderBy   = SORT_NAME + ", " + SYSTEM_DISPLAY_NAME + ", " + SEARCH_PROFILE_NAME + ", " + USERNAME + ", " + PHONE;
+    String   orderBy   = SORT_NAME + " " + SQLITE_COLLATE_NOCASE + ", " + SYSTEM_DISPLAY_NAME + " "
+            + SQLITE_COLLATE_NOCASE + ", " + SEARCH_PROFILE_NAME + " " + SQLITE_COLLATE_NOCASE
+            + ", " + USERNAME + " " + SQLITE_COLLATE_NOCASE + ", " + PHONE;
 
     return databaseHelper.getReadableDatabase().query(TABLE_NAME, SEARCH_PROJECTION, selection, args, null, null, orderBy);
   }
@@ -2119,7 +2123,9 @@ public class RecipientDatabase extends Database {
       args       = new String[] { "0", String.valueOf(RegisteredState.REGISTERED.getId()), "1", query, query, query, String.valueOf(Recipient.self().getId().toLong()) };
     }
 
-    String   orderBy   = SORT_NAME + ", " + SYSTEM_DISPLAY_NAME + ", " + SEARCH_PROFILE_NAME + ", " + PHONE;
+    String   orderBy   = SORT_NAME + " " + SQLITE_COLLATE_NOCASE + ", " + SYSTEM_DISPLAY_NAME + " "
+            + SQLITE_COLLATE_NOCASE + ", " + SEARCH_PROFILE_NAME + " " + SQLITE_COLLATE_NOCASE
+            + ", " + PHONE;
 
     return databaseHelper.getReadableDatabase().query(TABLE_NAME, SEARCH_PROJECTION, selection, args, null, null, orderBy);
   }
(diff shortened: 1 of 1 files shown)


g-andrade:
Well done! I was looking forward to fixing the issue myself when I noticed your PR.

⚠️ Unfortunately, sqlite's `NOCASE` is limited to Latin alphabet's 26 characters, [according to documentation](https://www.sqlite.org/datatype3.html#collating_sequences). This means the fix you propose **won't work for other scripts** (Cyrillic, CJK, Greek, ...) - nor will it work with common extensions of the Latin alphabet (e.g. **diacritics**) as seen in many European languages or Vietnamese.

A quick look around The Search Engine™ suggests that setting up proper Unicode collation rules on top of sqlite is [rather difficult](https://stackoverflow.com/questions/4285396/case-insensitive-utf8-select).

So, perhaps a different solution is warranted: would sorting the contact list in memory (rather than at the database) be feasible? But I wonder if it would also require ICU - I've had bad experiences with it, in the past.

Anyway, something to keep in mind.

Cheers!

ChewTeaYeah:
It turns out there are two other issues related to common extensions of the Latin alphabet (#9341 and #10409). Maybe doing an in-memory Locale based sort would help close out all 3 of these issues, but this leaves unsolved the case where a user's locale is English but they have non-English contact names. Will see if there is a sort order that would work here for in-memory sorting

ChewTeaYeah:
I think perhaps we're conflating issues here - this PR can fix the issue of lowercase vs uppercase names. Sorting Latin extension characters or non-Latin characters is a separate issue, they will still appear at the end of the contacts list as they do in the current implementation and we can fix that issue with another PR

Would love to get a maintainer's thoughts on this...

g-andrade:
Ohh, please don't take me wrong - your PR is definitely an improvement of the current state of things.

But it's limited in scope, and it would be nice to fix the issue for most languages / scripts / locales - it's an Unicode world and the capabilities are there, just hard to apply to this particular situation (speaking, up to a point, for myself here, as my native language [Portuguese] has diacritics and specific collation rules.)

In any case, maintainers' thoughts on this would be very welcome, for sure. 

In the mean time, I'll keep trying to get locale-specific sorting working. I'd be glad if your PR, as-is, was merged. Gradual improvement is better than no improvement at all :-)

williamspacefire (approved):

Relluem94 (approved):

ChewTeaYeah:
@williamspacefire @Relluem94 thank you for approving these changes. Is there anything else I need to do to get this PR merged in?

williamspacefire:
@ChewTeaYeah to your PR gets merged needs that someone with write permission to review and approve, i approved the changes because i have tested in my branch and it's working.

g-andrade:
As a bystander (and so my opinion is... just that), this PR is fine as an improvement but I don't think it fixes #10384 (albeit it improves the situation.)

My argument, in objective points:
- "Питер" will still be grouped separately from (and inconsistently sorted relative to) "питер" (with 'П' being uppercase for 'п')
- "Σωκράτης" will still be grouped separately from (and ...) "σωκράτης" (with 'Σ' being uppercase for 'σ')
- "Åbjørn" will still be grouped separately from (and ...) "åbjørn" (with ...)

And many other cases, including subtler equivalence rules where, depending on the user's locale, "É" may be equivalent to "E" - or not.

I don't want to be the annoying guy here, but words have exact meanings... this PR solves ASCII case insensitiveness issues, not case insensitiveness in general. It only accounts for the main 26 letters in the Latin alphabet. Should this still be the case in 2020, given the available technology?

Sure, it's an improvement - and it's better than not having at all. But please keep issue #10446 open. This PR doesn't solve it.

With that said - and I've mentioned it before - I'd be glad to see this change make its way into Signal. But please keep in mind the underlying issue will still be there.

ChewTeaYeah:
@williamspacefire thanks for your reply, I guess I'll just wait until someone with write access sees this PR :-(

@g-andrade I symapthize with what you're saying, but I think the cases you addressed are covered by issues #9341 and #10409. I agree in general that we shouldn't assume an ASCII only world, internationalization is important

The main obstacle seems to be that Signal uses SQLCipher rather than directly accessing SQLite on Android. While Android supports COLLATE LOCALIZED and COLLATE UNICODE in it's implementation of SQLite, SQLCipher does not and we cannot use them. We'd have to write an extension that performs the collation for us, which is a much more involved change than what I have here. Since the other two issues I mentioned cover the cases of non-ASCII characters, I think closing the linked issue ought to be fine

williamspacefire:
We are all waiting @ChewTeaYeah 😅

ChewTeaYeah:
@alan-signal any chance we can get some eyes on this PR? It seems to be good to go and already has a couple of approvals

35niavlys:
Similar problem on the search functionality, maybe a global solution is needed to abstract the limitation of sqlite

Caerbannog:
This comment summarizes a few related problems (glyph-similarity):
https://github.com/signalapp/Signal-Android/issues/10409#issuecomment-1023469987

Issue #10428, opened 2021-01-10
Possible to (accidentally) send a message to a blocked contact


rbtr:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
When a user is blocked, we're prevented from sending them a message via the normal routes - the text field and send button are hidden behind a dialogue with "Delete" or "Unblock" buttons.

However, it's possible to _share_ to that blocked contact, allowing us to (accidentally?) send them a message (or photo, or link, etc), and the thread opens without the Delete/Unblock overlay. It's also possible to _forward_ them a message from within the app. Leaving a draft and re-entering the thread with your blocked contact will show both the Delete/Unblock dialogue and enough of the text entry field and Send button to finish that mistake. 

### Steps to reproduce
- User A blocks user B
- User A shares via an Android Intent to: 
  - Signal, and selects B's thread
  - or directly to B's thread if that appears in their sharing intents

**Actual result:** The thread opens with the shared content (link, image, etc) and an editable text field for the message, which can then be sent.
**Expected result:** It should not be possible to send to a blocked contact? Unclear.

### Screenshots

<image src=https://user-images.githubusercontent.com/2940321/104115718-3fb65d00-52e0-11eb-91b2-ef968a079f69.png width="300">

### Device info

**Device:** Pixel 5
**Android version:** Android 11 
**Signal version:** 5.1.1

rbtr:
Extra note: read and delivery reports from the blocked user _are_ delivered and processed by the blocker's client, and the blocker will see delivered/read status of the message they have sent to the blocked user.

UserX404:
Featur requests are handlen within the [community forum](https://community.signalusers.org/) and Github is for bugs only.

rbtr:
@UserX404 I consider this a bug - I expect to be prevented from sending a message to a blocked contact, and I am usually. But if I enter the conversation with that blocked user via a share intent, the overlay that would prevent me from sending a message to them is not shown.

PR #10520, opened 2021-01-12
Fixes #10426: send delivery receipt even for ignored message 

### LINKED ISSUE #10426: Blocked user stops getting delivery reports

### Bug description
Per [the blocking documentation](https://support.signal.org/hc/en-us/articles/360007060072-Block-numbers-or-groups#block)
> Your blocked contacts will not know that you have blocked them. 

As far as I understand it this means that the blocked user should still get delivery reports, while the blocker never actually gets those messages. Between Android->iOS, and between either and Desktop, this is what I observe.

When sending to an Android user, the blocked user immediately stops getting delivery reports - only one check. This is inconsistent with the other platforms that I have tested. It seems like a data leak to the blocked user that they are blocked (also implicitly the platform that the blocker is using).

~Interestingly, with a Desktop client is in the mix: if the blocked user sends a message via Desktop it _will_ indicate that the message is delivered. When that message is synced to the Android client, it syncs the delivery status also.~ I can't reproduce this reliably but I have screenshots of it. Not sure what the deal is, but it smells like a data race - [related?](https://github.com/signalapp/Signal-Android/issues/8843)

### Steps to reproduce
- Android user A blocks <any platform> user B
- User B sends message to user A

**Actual result:** User B does **not** get the second check for "delivered" on their message, potentially implying they are blocked
**Expected result:** User B gets the second check for "delivered" on their message, consistent with the other Signal platforms

### Device info

**Device:** Various, Pixel 5, 4XL, 3A
**Android version:** Latest Android 11 1/2021 patch (RQ1D.210105.003 on Pixel 5)
**Signal version:** 5.1.1

### Link to debug log

Blocked (User B) log: https://debuglogs.org/0dac4bad4a1179586cdc97225a003eeecb3cee15399d3eaf60f94ee912cb3291
Blocker (User A) log: https://debuglogs.org/1c8c16fa607521b0b893d169f7d5ee0831290651bda28ffa025514af949443eb

---
I've been ramping up on the Android source to try and find where things are going wrong and contrib the fix, but maybe someone could point me in the right direction to speed that up? A clarification on what the expected behavior is across all Signal platforms would be a great start 🙂

rbtr:

### Contributor checklist

 * Pixel 5, Android 11
 * Virtual device Pixel 3A, Android 11

----------

### Description

Fixes #10426.

Sends a delivery report even when a message is ignored, to be consistent with other Signal platforms.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/jobs/PushProcessMessageJob.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/jobs/PushProcessMessageJob.java
@@ -340,6 +340,10 @@ public final class PushProcessMessageJob extends BaseJob {
 
       if (content == null || shouldIgnore(content)) {
         log(TAG, content != null ? String.valueOf(content.getTimestamp()) : "null", "Ignoring message.");
+        if (content.getDataMessage().isPresent() && content.isNeedsReceipt()) {
+          log(TAG, content != null ? String.valueOf(content.getTimestamp()) : "null", "Marking ignored message as delivered.");
+          handleNeedsDeliveryReceipt(content, content.getDataMessage().get());
+        }
         return;
       }
 


williamspacefire:
Show to the sender that the message is delivered is wrong, the user actually don't receive the message, as the sender it's blocked from send messages to this user.

grandchild:
@williamspacefire but the sender shouldn't know that they're blocked. to them it should seem as if nothing is different. you can imagine why, if you think about the various (sad) reasons why you'd block someone.

rbtr:
I see valid arguments for doing it either way way. I think there's a (slightly) stronger argument for _continuing_ to send delivery reports to a blocked contact, depending on the exact intent of:
> Your blocked contacts will not know that you have blocked them.

As I mentioned in the linked [issue  (#10426)](https://github.com/signalapp/Signal-Android/issues/10426) I think the current behavior in Android is a data leak back to the blocked user that they have been blocked (and that the blocker uses Android).

In any case, iOS and Android behave differently right now, so one of them should be changed.

williamspacefire:
I see your point, but look at the perspective that when you block someone the text says that the user won't be able to send you messages, so if the sender it's not able to send messages to that user why it's getting delivered receipts? That's why i think it's wrong to tell the sender that the message is delivered when it wasn't, the message will be send and the sender will not receive the delivered receipt for many other reasons, and not just because it's blocked

williamspacefire:
> In any case, iOS and Android behave differently right now, so one of them should be changed.

I agree with that, let's see what the devs has to say about it

grandchild:
> why it's getting delivered receipts?

Yeah that would be illogical right? But it's not about consistency of behavior here, it's about not making people that bother you to the point that you have to block them on messengers (think choleric ex, or worse) _even more angry_. I think this is so important that I'd throw app consistency overboard a hundred times for that.

And btw, as far as I know, _technically_, the message _is_ being delivered to the receiver's device (just not notified about, nor shown, nor saved, just thrown away).

williamspacefire:
Ok, i see your point, but in this scenario what would be worst, the person think that you ignored it's message or think that you didn't received the message? In my point of view I think that let the person think that it's ignored is worst, because the sender will think that the message was seeing in the notification bar or somewhere else and was ignored and this will make it angry. You prefer to be ignored or think that the message wasn't delivered because you was blocked, the person uninstalled Signal, lost the phone, don't have internet...?

The message is delivered to the phone, not the user, the delivered receipts confirms that the user received the message, but in this case it didn't

rbtr:
If the intent is to communicate the least amount of information back to the blocked user about their potentially blocked status, we should continue sending delivery reports. 

In support of continuing to send delivery reports: when a new contact (not already in your phone) messages you for the first time, you can preview the thread (without sending read receipts) and accept/block. Until you accept/block those delivery receipts are sent, and then, if you block, they stop being sent. This clearly indicates to the blocked user that they have been blocked, or that you have abandoned Signal entirely - which they could test by repeating this with a new burner number.

rbtr:
There are other conversations to be had about whether this is truly blocking (or should be called something like "ignore"), whether delivery receipts should also be optional like read receipts, whether there should be a "burn bridge" option which really prevents you from sending (you CAN currently [send to a blocked contact](https://github.com/signalapp/Signal-Android/issues/10428)) or receiving any data from a certain user...but we start to drift away from the intent of this patch.

ktravis:
> The message is delivered to the phone, not the user, the delivered receipts confirms that the user received the message, but in this case it didn't

I would argue that the delivered receipt confirms that the *device* received the message. The **read** receipt confirms that the *user* received the message.

It could be interpreted that "blocking" implies not sending information to the blocked contact, which I think is really more a symptom of unclear terminology given the _explicit_ detail that blocked contacts will not know that they are blocked.

Caveat: if there was some way to disable sending delivery receipts outside of blocking an individual contact, it would make sense to do so when the contact is blocked as well. I would still prefer finer-grained control of "ignore" (don't show me the messages, keep sending delivery notifications/following global preference) vs. "burn bridge" (send nothing to this contact).

tanujkucp:
I think, if we send delivery reports to the sender, then sender thinks that messages are delivered and didn't know that he/she has been blocked.
In that case it creates a lot of confusion for the sender. Sender might think that the messages have been successfully delivered and recipient might be offline so read receipt was not shown.
Now think what if sender has sent some important information and now thinks that info is delivered. This will create confusions.
Its better to not send delivery reports to blocked contacts , so that they don't think that messages are delivered.

rbtr:
@tanujkucp 
> sender thinks that messages are delivered and **didn't know that he/she has been blocked**

According to the docs (["Your blocked contacts will not know that you have blocked them"](https://support.signal.org/hc/en-us/articles/360007060072-Block-numbers-or-groups#block)), this is exactly the goal so we should continue sending delivery reports? 🙂

williamspacefire:
@rbtr According to the docs ("[Blocks the person from seeing your profile name and photo.](https://support.signal.org/hc/en-us/articles/360007060072-Block-numbers-or-groups#block)"), so if you block someone it stops from seen your photo and name, don't you think that's enoth to tell the person that's it has been blocked? Cause if you stops seen the persons photo out of nothing you'll think that you have been blocked, right?

rbtr:
In my testing, you don't suddenly stop seeing the persons photo, or stop them from seeing yours, when you block them - you can verify this yourself and let me know if it works that way for you. _Maybe_ if you blocked the initial contact from a new person, before your profile was synced with them, you would prevent them from _ever_ seeing your profile, but it doesn't remove it later. 

But that's fine, since the intent is actually that "your blocked contacts will not know that you've blocked them".

tanujkucp:
Also consider this- when you blocked someone, and if they still receive delivery reports for their messages to you, then - youe private info is being exposed to them.
The private info is the times when you are using internet and online. They can message you all day and then figure out your schedule of using internet and being online. Isn't that an important privacy issue. If you have blocked someone than they shouldn't get this private info of you through delivery reports.

Also if they don't get delivery reports after being blocked, then they'll not know they have been blocked straight away because message may not have been delivered for so many reasons like they are not using internet, they stopped using signal etc

williamspacefire:
@rbtr you'r right, i tested here and doesn't work as intended, i still see the photo and name of the person who blocked me, so this is a miss behavior of Signal, as pointed in the docs when you block someone it shouldn't be able to see your photo and name, i didn't tested it in other platforms, but I'll do later, i think this is a issue, the app isn't working as described in the docs, this is a privacy concern. I'll do more tests before report an issue. And as @tanujkucp said this is a leak of information, your device is telling the blocked person when you'r online. When i block someone, i don't want that my device keeps in touch with this person, telling it if I'm online, change my photo or my name, i want this person to have zero information about me. If someone is stalking me i want it to stop, if this person is getting every new profile photo i set, or when i changes my name or know when I'm online, this is not good, this isn't right, don't you think?

williamspacefire:
When you block someone, this creates a barrier between you both, since there's a barrier the blocked person shouldn't be able to reach you out and you shouldn't reach it too, that's why you can't send messages to blocked contacts, you have to unblock them first, i think that the message the blocked person sends reach your phone it's because Signal server doesn't know if you block someone, this happens in your phone because Signal keep really little information in their servers, i think that's is the only reason why the message reach your phone.

rbtr:
@williamspacefire @tanujkucp I appreciate your opinions and engagement here, but I think we're going in circles and starting to lose the point of this fix.

The question is not whether you feel that blocking someone should completely prevent **any** data being received/sent from that contact (but this is a good general design thing to debate, I agree with many of your points, and I [want to improve that experience](https://github.com/signalapp/Signal-Android/issues/10428) too!). 

The question is, given:
- the stated intent that "Your blocked contacts will not know that you have blocked them.",
- the behavior of the iOS app is to continue to send delivery reports, and
- new contacts get delivery reports until you accept/block their conversation request,

does it communicate the **least** contextual information (to satisfy point 1) to the blocked user to continue to send delivery reports?

My assumption is that "your blocked contacts will not know that you have blocked them" means that we give the blocked user the least amount of data possible to determine if they've been blocked. I could be wrong, and that could just mean something like "we don't send them a message *telling* them that they've been blocked". If that's case, [like I've brought up above](https://github.com/signalapp/Signal-Android/pull/10520#issuecomment-758832527), there are other discussions to be had about what exactly "blocking" means and how thorough it should be, but that's out of scope here.

PR #8846, opened 2019-05-31
rate limit group chat notifications

afflux:
### First time contributor checklist

- [x] I have read [how to contribute](https://github.com/signalapp/Signal-Android/blob/master/CONTRIBUTING.md) to this project
- [x] I have signed the [Contributor License Agreement](https://whispersystems.org/cla/)

### Contributor checklist

- [x] I am following the [Code Style Guidelines](https://github.com/signalapp/Signal-Android/wiki/Code-Style-Guidelines)
- [x] I have tested my contribution on these devices:
 * Virtual device Nexus 5X, Android 9 (api 28)
 * Virtual device Nexus 5X, Android 5.1 (api 22)
- [x] My contribution is fully baked and ready to be merged as is
- [x] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the `Fixes #1234` [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)

----------

### Description

When discussion starts in medium-size group chats, the current behavior is to notify and vibrate up to every two seconds. This pull request adds rate-limiting on a per thread basis for group chats, so that subsequent messages will not trigger notification alarms for three minutes if the user did not interact with the thread in the meantime.

Issue #6759, opened 2017-06-23
Throttle notifications


GSPP:
Notifications for incoming messages are not throttled at the moment. This means that the phone makes a sound or vibration for each and every incoming message, even if they are just seconds apart. I feel like a slave to the phone because it grabs my attention constantly when I'm in the middle of something else. It breaks my concentration.

At the same time, I do want to receive notifications. Just not so many.

I propose:

1. Notify the user at most once every X minutes (default: 15 minutes?) after the last notification or after the last time the user opened Signal. There is no point in telling the user many times that messages are available.
2. Notify about an incoming message not immediately but after 1 minute delay. That way multiple incoming messages can be read in one go instead of being constantly forced to either monitor the chat or be distracted by the notifications. I think this feature must be default off because it adds a delay that users might not expect. I'm not sure.

The delay constants should be configurable to adapt the app to users diverse needs. I hope this feature request makes sense for the product. I'm using Google Android 7.1 and Windows 7 with Signal Desktop.

p4nci:
related to https://github.com/WhisperSystems/Signal-Android/issues/3165, https://github.com/WhisperSystems/Signal-Android/issues/5625, https://github.com/WhisperSystems/Signal-Android/issues/5945

Be-ing:
My idea for dealing with this: #6288

GSPP:
I'm in a Signal Desktop chat right now. Literally, every few seconds my phone buzzes *immediately* when Desktop receives a message. It also buzzes for each and every message. I basically want to be notified by new messages meaning I can't just disable notifications. I have no mitigation against my phone constantly buzzing.

@Signal Team: Can you prioritize this? You have built a very fine messenger, and it's free, but this notification spam is quite a problem.

nrizzio:
@GSPP Please don't bump issues. If the issue is open it's tracked.

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

Issue #6288, opened 2017-02-22
feature request: intelligent temporary muting for active conversations


Be-ing:
I have:
- [x] searched open and closed issues for duplicates
- [x] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

The muting feature is already decent, but it could be better. I don't want to mute a conversation then forget I had muted it and miss messages in the future. I want to be notified of new messages, but not be annoyed by torrents of new message notifications coming in quickly.

I think this could be possible with a relatively simple feature that I'll call "intelligent muting" for now, unless someone comes up with a better name. When enabled, conversations would automatically be muted for a short, configurable amount of time (like 5 minutes) upon receipt of a new message. New messages would not trigger a new notification until that amount of time elapsed without receiving a new message. Thus, I could be notified that there is new activity, but not have my notification sound go off 5 times within a few minutes. This would be especially helpful for large group conversations.

cascheberg:
Related commit: 3787551878473c703a21210aa24d39080ad57ad4 "limit audible notifications to once every 2 seconds" (does not have a preference or so)

Be-ing:
That's a good feature to have and does not need any configuration. But it addresses a different issue. The issue I am concerned with is active conversations that have new messages coming in not necessarily within a few seconds of each other, but within the amount of time it takes someone to read and respond to a new message.

GSPP:
Reposting a relevant comment here:

I think the delay strategy needs to be configurable. With most friends I'd be OK with receiving messages only after quite some time. When my boss writes me I want my phone to buzz immediately and possibly for every message also.

deutrino:
Very closely related to #5945

automated-signal:
GitHub Issue Cleanup:
See #7598 for more information.

Issue #5945, opened 2016-12-10
Exponential backoff for vibrations


ruuda:
I have:
- [x] Searched open and closed issues for duplicates. #3165 and #5813 are related.
- [x] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports

----------------------------------------

### Bug description
When a contact is rambling at you for say ~20 messages in short succession, the phone will vibrate every time. This is annoying. It looks like 3787551878473c703a21210aa24d39080ad57ad4 introduced a limit of once ever two seconds, but that will not help a lot when a contact keeps sending messages for a few minutes.

When my phone vibrates, I have to make a decision:
- Stop what I was doing and look at the message right now.
- Continue what I was doing, remember that somebody sent me a message, and take a look when I have the time.

Most of the time, I’ll opt for the second option. Then again, there are two cases:
- After a few minutes I remember to take a look.
- I forget that I received a notification and I don’t look at my phone until it starts vibrating again.

If my phone vibrates twice within a few seconds, that is not useful, it is only annoying. But even with a two-second delay, if I decide to not look *right now*, a vibration after 2, 3, 6, ... seconds is still annoying. I already made a mental note to take a look soon, please stop bothering me. On the other hand, if after say ~15 minutes I receive another message, then that is probably not part of the same discussion, and if I haven’t checked by then, I do want a notification.

I think that an exponential delay might work well here. When a message is received, vibrate only if the last vibration was more than *x* seconds ago. Start with *x* = 2 seconds, and double (or maybe even triple) it at every vibration, up to a maximum of e.g. 5 minutes. When I read a message, reset *x* to its initial value.

### Steps to reproduce
- Have a contact send you a message every few seconds.
- Put the phone on the table, screen locked.

**Actual result:** The phone keeps on vibrating while the contact is sending its monologue. Currently (without the 2-second delay) this means the phone can be vibrating continuously for minutes.
**Expected result:** The phone should vibrate once, maybe twice, and then remain silent for a while.

### Device info
**Device:** Nexus 6P
**Android version:** 7.0, 5 November 2016 patch level
**Signal version:** 3.24.1

jeremymasters:
One thing that could reduce it would be to turn off in-conversation notifications in the settings.

ruuda:
Thank you for the suggestion. I had already turned that setting off, but in the scenario where this is an issue, I am not in a conversation; this is about notifications/vibrations when the phone is locked.

[...]

PR #10619, opened 2021-01-14
Rate limit on notifications unique to each conversation

misterPaul0:
Modify notifications to only send you a new notification from each chat once every 10 minutes, or after opening the chat

### Contributor checklist

  Moto Power G, Android 11.0

----------

### Description
This is a modified version of #8846 which attempted to add a rate limit to group conversations. I have not restricted mine to group notifications, as a rate limit is needed for all chats to stop notification spam. This existing PR is over a year old and far too out of date with the codebase to be adapted, so I've redone it.

I've made the notification rate limit once every 10 minutes, but if you open the chat it resets the rate limit allowing another notification through. I feel this strikes a balance between letting people know a conversation is ongoing without nagging them, but happy to take suggestions on modifying the 10 minute restriction.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/notifications/DefaultMessageNotifier.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/notifications/DefaultMessageNotifier.java
@@ -71,10 +71,12 @@ import org.thoughtcrime.securesms.util.MediaUtil;
 import org.thoughtcrime.securesms.util.MessageRecordUtil;
 import org.thoughtcrime.securesms.util.ServiceUtil;
 import org.thoughtcrime.securesms.util.SpanUtil;
+import org.thoughtcrime.securesms.util.SoftHashMap;
 import org.thoughtcrime.securesms.util.TextSecurePreferences;
 import org.thoughtcrime.securesms.webrtc.CallNotificationBuilder;
 import org.whispersystems.signalservice.internal.util.Util;
 
+import java.util.Map;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


jamesdangercarpenter (approved):

tomudding:
Just a thought, but I feel like this needs to be setting instead of something that is forced upon every chat. I can understand why you'd want it for a group chat, but there are instances where for certain private chats I want to get each and every notification.

williamspacefire:
Agree with @tomudding it would be great if the user could customize the timer or deactivate this feature for each chat or global

misterPaul0:
After further investigation I've reverted this change and swapped it with a more simple change - switching the group alert behaviour from "GROUP_ALERT_CHILDREN" to "GROUP_ALERT_SUMMARY". I see no reason we should be using GROUP_ALERT_CHILDREN for a group of message notification, this just leads to spam. Interested in thoughts.

wolf99:
Looks like this would do as wanted.
Would be awesome to see this merged as I've got contacts that have left Signal again due to noisy group chats.

wolf99:
Yes, this is still relevant.

Why this hasn't happened yet though is beyond me. Perhaps a maintainer could comment?

peteruithoven:
This has been a pain for years, for groups and when you have friends that send bursts of messages. 
There have been multiple issues about this: 
- #6759
- #5945
- #3165
- #5625
- #6288

I hate to mention people directly but @greyson-signal @cody-signal @alex-signal @rashad-signal could one of you please have a look? jamesdangercarpenter has already reviewed an earlier version.

wolf99:
Yes this is still relevant.

I'm informed that the developers don't use GitHub issues/PRs for feature selection. However the feature request has been in place for so long, both here and in other request avenues, that at this point any way of signalling (unintentional pun!) to the developers that this is wanted is useful.

cody-signal:
Hi all, we are tracking this internally. We have it planned as part of some other work. We'll likely be using a custom implementation for the timing mostly because OEMs can't be trusted to behave the same. Appreciate the PR but we likely won't be using this approach. If we do, we'll pull it in then, but closing for now.

PR #10630, opened 2021-01-14
Disable MMS Option

smrt28:

### Contributor checklist

 * OnePlus Pro 8, Android 11
- [] My contribution is fully baked and ready to be merged as is - not sure, how is the localization

----------
### Description

I would like to see this option in the Signal app. My mobile operator doesn’t allow me to deactivate the MMS service and it happened to me several times that I accidentally sent the MMS by Signal being charged an immoral fee.

It’s quite easy to implement, but I’m not an Android/iOS developer. Please, let me know if something was wrong with the patch, I'll fix it promply.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/sms/MessageSender.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/sms/MessageSender.java
@@ -359,7 +359,9 @@ public class MessageSender {
     } else if (!forceSms && isPushMediaSend(context, recipient)) {
       sendMediaPush(context, recipient, messageId, uploadJobIds);
     } else {
-      sendMms(context, messageId);
+        if (!TextSecurePreferences.isMMSDisabled(context)) {
+              sendMms(context, messageId);
+        }
     }
   }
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/thoughtcrime/securesms/sms/MessageSender.java
```diff
@@ -359,7 +359,9 @@ private static void sendMediaMessage(Context context, Recipient recipient, boole
     } else if (!forceSms && isPushMediaSend(context, recipient)) {
       sendMediaPush(context, recipient, messageId, uploadJobIds);
     } else {
-      sendMms(context, messageId);
+        if (!TextSecurePreferences.isMMSDisabled(context)) {
```
manuel-91, line 362: How do i 'accidentaly' send an MMS in the old code? By sending media to a non-signal user? If I do so in the new code with `MMSDisabled` it would be great to have a hint then. More specific: This `if` has no `else`. So if it is disabled just nothing happens which might irritate the user. It would be more user-friendly to have an explaining popup.

Relluem94 (approved):

Cleymax (approved):

app/src/main/java/org/thoughtcrime/securesms/sms/MessageSender.java
oxygi: Yes, this is a must-have to approve this change.

PR #10676, opened 2021-01-16
Fix to the issue that a chat jumps to the latest message suddenly when reading older messages

### LINKED ISSUE #10332: Chat keeps jumping to new/latest messages while I am still reading old messages

### Bug description
Chat keeps jumping to new/latest messages while I am still reading old messages

### Steps to reproduce
- receive several (e.g. 50)  messages from a group chat while you don't have the app open (e.g. device locked)
- click on the notification for the chat from android notification shade
- start reading messages
- someone in group sends new message

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)
- signal jumps to the bottom/new/latest message

**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)
- scroll position should not change
### Device info

**Device:** Nokia 5.3
**Android version:** 10
**Signal version:** 5.0.8

### Link to debug log

https://debuglogs.org/2d0b55995068e5c8e91ece4efc4230fcaf2cc13e59e1b2fb41aa9f90e8b08edc

I believe the issue only happens after opening signal from a notification, rather than opening the app from the app drawer. Similarly, if I have induced the bug, when I switch to the app via the recent apps switcher, the issue also persists.

fumiakiy:
### Contributor checklist

 * AVD Pixel 3a API 30
 * AVD Nexus 6P API 27 (w/o Google API)

----------

#### Tl;dr;
Fixes the issue of a chat jumping to the more recent message, by ignoring an execution that would have loaded data in a range that does not include the index that are needed to be loaded by the method call. Fixes #10332 

#### Longer description of the issue and the fix
It happens in v5.2.1 and older when you are scrolling down and reading older messages in a fairly long conversation, the view suddenly scrolls to the latest message. See the screen record below how it looks like.

The issue is because of some threading glitches; when we scroll RecyclerView, at some point it calls [onDataNeededAroundIndex(int aroundIndex)](https://github.com/signalapp/Signal-Android/blob/4b2366e537ef25d3489c08b1e9a508ef9fd37074/paging/lib/src/main/java/org/signal/paging/FixedSizePagingController.java#L55) to load data to add to the list. In the method it determines the range of the data to be loaded and first [mark the range as read](https://github.com/signalapp/Signal-Android/blob/4b2366e537ef25d3489c08b1e9a508ef9fd37074/paging/lib/src/main/java/org/signal/paging/FixedSizePagingController.java#L89), then [spawn a thread to actually load the data in the range](https://github.com/signalapp/Signal-Android/blob/4b2366e537ef25d3489c08b1e9a508ef9fd37074/paging/lib/src/main/java/org/signal/paging/FixedSizePagingController.java#L93). In there it loads data, and stuff them in to the (possibly) existing data according to the index of the data and post the array of data to UI to show them.

The problem here is that `onDataNeededAroundIndex` is called a lot of times when we scroll and read messages. Sometimes two or more calls to `onDataNeededAroundIndex` enter in the method before a thread that loads the data starts to execute. Because we use LIFO strategy to spawn threads and execute code, the later call to `onDataNeededAroundIndex` is executed before earlier calls. But the later one sometimes has very narrow range to load data that does not even include the `aroundIndex`, because the range around `aroundIndex` was already marked as read by the earlier calls (that have not been executed to load data yet). The result of the execution is an array that consist mostly of nulls.

When the array is posted to update the UI, RecyclerView adapter gets confused because most data is null but at the end of the array there are a few items. It calls `onItemRangeInserted` that calls [`snapToTopIfNecessary`](https://github.com/signalapp/Signal-Android/blob/2c1c6fab3560ef12996030cf7c84cc0d895e329e/app/src/main/java/org/thoughtcrime/securesms/util/SnapToTopDataObserver.java#L113) and we see the chat suddenly jumps to the more recent message all of the sudden.

The code in this PR fixes the issue simply by ignoring the execution that would have loaded data of a range that does not include `aroundIndex`. This works because this situation happens when two or more calls to `onDataNeededAroundIndex` happen very close to each other and one of those would likely have the data at the index we just ignored.

### Screen record
I'll attach screen records in comments below.

### DIFF
--- a/paging/lib/src/main/java/org/signal/paging/FixedSizePagingController.java
+++ b/paging/lib/src/main/java/org/signal/paging/FixedSizePagingController.java
@@ -84,6 +84,11 @@ class FixedSizePagingController<E> implements PagingController {
       return;
     }
 
+    if (loadStart > aroundIndex || loadEnd < aroundIndex) {
+      if (DEBUG) Log.i(TAG, buildLog(aroundIndex, "loadStart > aroundIndex || loadEnd < aroundIndex, loadStart: " + loadStart + ", aroundIndex: " + aroundIndex + ", loadEnd: " + loadEnd));
+      return;
+    }
+
     int totalSize = loadState.size();
 
     loadState.markRange(loadStart, loadEnd);


fumiakiy:
Chat jumping to the latest message

https://user-images.githubusercontent.com/28482/104817790-79f89080-57f1-11eb-8ccd-15f48aedfbb0.mp4

fumiakiy:
Fixed

https://user-images.githubusercontent.com/28482/104817791-7fee7180-57f1-11eb-9bb9-ec9a7ff6b800.mp4

Relluem94 (approved):

johnnycondev:
@greyson-signal Is it possible to merge this commit? This is a daily struggle for me and my friends in the group chat.

jcul:
Thanks for fixing this.

Having seen #10332 during the week, I just forked and checked out Signal-Android to have a go at fixing it myself, went to find #10332 and spotted this pull request already fixing it :) too slow!

It's funny I barely noticed it myself but of all people I managed to convert to Signal, this was the thing that was making them all want to switch our groups back to Whatsapp.

Will try build your branch and test it out myself.

joaosa:
> Thanks for fixing this.
>
> It's funny I barely noticed it myself but of all people I managed to convert to Signal, this was the thing that was making them all want to switch our groups back to Whatsapp.
> 
> Will try build your branch and test it out myself.

I second this. I have been experiencing exactly the same issue with new convertees. This would definitely be a game changer 😄 🥇

greyson-signal:
@fumiakiy Thank you so much for doing the research here! We used to have this bug with our previous paging library, but the cause was totally different, and I thought I merely introduced a regression in the same place, so I wasn't even thinking about looking where you looked. I greatly appreciated your in-depth description of the bug, and it was essentially spot-on.

That being said, I think I'm going to go with a more direct solution, which is to switch from a LIFO to a traditional FIFO serial executor. There's some pros and cons to consider, but I do think it's the simplest to grasp, and should definitely resolve the issues we're seeing. I'd like to eventually go back to the LIFO behavior, but I think doing so will take a lot more thinking.

I've applied the executor change to our 5.4 candidate branch, and I've given you a shout-out in the commit message. Thanks again!

error401de:
@greyson-signal Is there any chance you can give an ETA when this fix will be released? I'm wondering as this is by far the biggest complaint from the contacts I have, which is holding them back from fully move to Signal - and the current version is 5.2.3

gmsa:
Where is that candidate branch? I'm willing to make a build for myself just to get rid of this bug!

greyson-signal:
The fix is in master. It's currentin in 5.3.8, which is presently in beta.

Caerbannog:
This was commit 95978f1

PR #10734, opened 2021-01-19
Improve Search functionality

farisshajahan:

### Contributor checklist

 * Virtual device Pixel 4, Android 8.1.0

----------

### Description
The previous implementation of the search function matched only single quotes among the special characters by replacing them with a single space. However the same can be done for other special characters as well. This PR addresses the same and additionally only displays the relevant result for a query made. For instance, if there are two messages test'com and test&com, searching for test'com returns both the texts whereas, with this PR, only test'com is returned.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/search/SearchRepository.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/search/SearchRepository.java
@@ -108,7 +108,7 @@ public class SearchRepository {
 
       Future<List<Recipient>>     contacts        = parallelExecutor.submit(() -> queryContacts(cleanQuery));
       Future<List<ThreadRecord>>  conversations   = parallelExecutor.submit(() -> queryConversations(cleanQuery));
-      Future<List<MessageResult>> messages        = parallelExecutor.submit(() -> queryMessages(cleanQuery));
+      Future<List<MessageResult>> messages        = parallelExecutor.submit(() -> queryMessages(query));
       Future<List<MessageResult>> mentionMessages = parallelExecutor.submit(() -> queryMentions(sanitizeQueryAsTokens(query)));
 
       try {
@@ -133,7 +133,7 @@ public class SearchRepository {
 
     serialExecutor.execute(() -> {
       long                startTime       = System.currentTimeMillis();
-      List<MessageResult> messages        = queryMessages(sanitizeQuery(query), threadId);
+      List<MessageResult> messages        = queryMessages(query, threadId);
       List<MessageResult> mentionMessages = queryMentions(sanitizeQueryAsTokens(query), threadId);
 
       Log.d(TAG, "[ConversationQuery] " + (System.currentTimeMillis() - startTime) + " ms");
@@ -169,9 +169,15 @@ public class SearchRepository {
   }
 
   private @NonNull List<MessageResult> queryMessages(@NonNull String query) {
-    List<MessageResult> results;
-    try (Cursor cursor = searchDatabase.queryMessages(query)) {
-      results = readToList(cursor, new MessageModelBuilder());
+    List<MessageResult> results = new ArrayList<>(), tempResults;
+    String cleanQuery = sanitizeQuery(query);
+    try (Cursor cursor = searchDatabase.queryMessages(cleanQuery)) {
+      tempResults = readToList(cursor, new MessageModelBuilder());
+      for (MessageResult result: tempResults) {
+        if (result.body.toLowerCase().contains(query.toLowerCase())) {
+          results.add(result);
+        }
+      }
     }
 
     List<Long> messageIds = new LinkedList<>();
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/thoughtcrime/securesms/search/SearchRepository.java
```diff
@@ -133,7 +133,7 @@ public void query(@NonNull String query, long threadId, @NonNull Callback<List<M
 
     serialExecutor.execute(() -> {
       long                startTime       = System.currentTimeMillis();
-      List<MessageResult> messages        = queryMessages(sanitizeQuery(query), threadId);
+      List<MessageResult> messages        = queryMessages(query, threadId);
```
jpventura, line 136: Doesn't the code look inconsistent now? :thinking: You've removed `sanitizeQuery` but not `sanitizeQueryAsTokens`.

Maybe it would be better remove both or none. What you think @farisshajahan?

app/src/main/java/org/thoughtcrime/securesms/search/SearchRepository.java
```diff
@@ -360,7 +372,7 @@ private String sanitizeQuery(@NonNull String query) {
       char c = query.charAt(i);
       if (!BANNED_CHARACTERS.contains(c)) {
         out.append(c);
-      } else if (c == '\'') {
+      } else {
```
jpventura, line 375: @farisshajahan This clause specifically double-check `'` because it is a `BANNED_CHARACTERS` according to the [ASCII table][ascii-url].

This patch now replace **all** banned characters are now replaced by space.

[ascii-url]: http://www.asciitable.com/

jpventura (changes requested):

app/src/main/java/org/thoughtcrime/securesms/search/SearchRepository.java
```diff
@@ -169,9 +169,15 @@ public void query(@NonNull String query, long threadId, @NonNull Callback<List<M
...
-      results = readToList(cursor, new MessageModelBuilder());
+    List<MessageResult> results = new ArrayList<>(), tempResults;
+    String cleanQuery = sanitizeQuery(query);
+    try (Cursor cursor = searchDatabase.queryMessages(cleanQuery)) {
+      tempResults = readToList(cursor, new MessageModelBuilder());
+      for (MessageResult result: tempResults) {
+        if (result.body.toLowerCase().contains(query.toLowerCase())) {
+          results.add(result);
+        }
+      }
```
ankit167, line 180: May be we can extract this piece of code as a private helper method, as it is redundant.

farisshajahan: Yes. Makes sense.

app/src/main/java/org/thoughtcrime/securesms/search/SearchRepository.java
farisshajahan: Yes. That is the intention as mentioned in the description :)

app/src/main/java/org/thoughtcrime/securesms/search/SearchRepository.java
farisshajahan: Might need some help with this. Can you tell me what conversations and mentions are?
I assume mentions mean the @Name feature but what about conversations?

app/src/main/java/org/thoughtcrime/securesms/search/SearchRepository.java
farisshajahan: On second thoughts, I feel this is too small to be refactored for now. Hence I will let it be.

app/src/main/java/org/thoughtcrime/securesms/search/SearchRepository.java
jpventura: @farisshajahan what I meant was that I don't know if this `'` exception was intentional or just a bug, because it could be a [SQL injection double-check][https://stackoverflow.com/questions/14370670].

Do you know about the internals of the app? Because if someone just forgot, the patch totally makes sense.

farisshajahan: @jpventura I think you meant to reply on the other thread. The `'` exception was intentional I believe but if I am correct, the author must have thought his hack worked only for `'` when in reality the same hack can work for other special characters as well. Also, there is no question of an injection attack here since we are still "sanitizing" the input query by replacing the special characters with a space.

farisshajahan: Also with regard to your first question on this thread, I have not "removed" sanitizeQuery(), just moved it inside the queryMessages method so that I can compare the results from the DB query using the cleanQuery and the original query in order to return relevant results alone.

farisshajahan:
@alan-signal @greyson-signal 
Hi. Could you please take a look at this?

Issue #10424, opened 2021-01-10
Screen lock problem


akshayd002:
We need to give lock signal every time if we want to lock the signal application. So it should be enabled immediately, or should be given timings. So this fix will be good

Sgn-32:
It's possible in the Signal settings -> Privacy -> Screen lock

![Screenshot_1610280640](https://user-images.githubusercontent.com/49990901/104122474-7a9eac00-5345-11eb-9626-0bd34044ab31.png)


https://support.signal.org/hc/en-us/articles/360007059572-Screen-Lock

ayaankhan98:
i think this is not working as expected. i set the screen lock inactivity timeout to 1 second but this is not automatically locking the app. i do have to manually lock the app from the notification bar everytime.

Sgn-32:
Minimum screen lock time is 1 min, also when is possible to enter only few seconds.

akshayd002:
Everytime I exit the app I should give lock signal manually. So what iam
telling is don't make user to give manually. Give immediately option to
lock when user exit the app.

ayaankhan98:
> Minimum screen lock time is 1 min, also when is possible to enter only few seconds.

yes, we can set it to 1 sec but as soon as we press OK after setting it to 1 sec it does set to 1 minute instead of 1 sec. Moreover, after 1 minute it is not working as expected.

AliabbasMerchant:
A proper, more detailed bug report is available in #10208  

Basically, the changes required are:
1. Allow immediate locking (0 seconds)
2. Allow minimum inactivity time in seconds also (currently the minimum is 1 min. Lower values can be entered, but the value shifts to a minimum of 1 minute) [Note: Seconds level granularity is accepted. i.e., a value of 1min30sec is accepted]
3. Ensure that the inactivity timeout feature works as intended (@ayaankhan98 has reported that it was not working for him)

Sgn-32:
@ayaankhan98 please post a debug log, if the screen lock with 1 min doesn't work.

ayaankhan98:
> A proper, more detailed bug report is available in #10208
> 
> Basically, the changes required are:
> 
> 1. Allow immediate locking (0 seconds)
> 2. Allow minimum inactivity time in seconds also (currently the minimum is 1 min. Lower values can be entered, but the value shifts to a minimum of 1 minute) [Note: Seconds level granularity is accepted. i.e., a value of 1min30sec is accepted]
[...]

PR #10802, opened 2021-01-24
Remove seconds from screen lock timeout input for coherence

### LINKED ISSUE #10788: Screen lock inactivity timeout
https://github.com/signalapp/Signal-Android/blob/234e4be92413b6d47d54af5d01ebc993bf3a960b/app/src/main/java/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java#L252

Time is always one minute even if a second is entered

Caerbannog:
### Contributor checklist

 * Samsung Grand Prime, Android 5.1.1

----------

### Description

Configure the TimeDurationPickerDialog in the preferences to hide seconds.
Seconds were already ignored below 1min. This avoids the user expecting it to work.

Feature regression: after this change, seconds above 1min will also be impossible to input (ex: 1m30s).
But it makes little sense anyway to allow it: they are even less useful for longer durations.

Another possibility to reach a point where eveything is coherent would have been to just remove the Math.max(..., 60) that ignored seconds.

The duration will be displayed as "xx:xx:00" to make it clear that xx:xx represents minutes.

Fixes #10788.

Another evolution that would please some users: accept "0min" to mean "lock as soon as possible". It currently removes the autolock.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/preferences/AppProtectionPreferenceFragment.java
@@ -70,6 +70,7 @@ import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
 import mobi.upod.timedurationpicker.TimeDurationPickerDialog;
+import mobi.upod.timedurationpicker.TimeDurationPicker;
 
 public class AppProtectionPreferenceFragment extends CorrectedPreferenceFragment {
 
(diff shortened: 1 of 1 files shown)


Relluem94 (approved):

Caerbannog:
**Before**

 

**After**

ayaankhan98 (commented):
@Caerbannog it will not allow immediate locking of app and locking the app in less than 1 minute as mentioned in #10424

greyson-signal:
I think this is probably what we want for now, so I've merged it into 5.9.1, thank you! 
@ayaankhan98 I'd want to think about the UX consequences of some of the other options first. For now, this PR removes one element of confusion around setting the timeout.

ayaankhan98:
duplicate #11133

PR #10841, opened 2021-01-28
Automatically retry when there is connection issue

### LINKED ISSUE #7733: Calls always income as missed ones without ringing and Signal doesn't report the call failed
- [X] I have searched open and closed issues for duplicates
- [X] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports

When Signal is opened in front of me on my phone and via Hangouts PC I hear the other party while his Signal is actually ringing me. My Signal shows no sign that I'm being called. At the end of ringing, when the network automatically disconnects, I immediately receive a "You have a missed call" notification in the chat.

Reproduced on three different WiFi networks with the same person.

The time difference between the two phones is approximately 5 seconds. Both phones are set to synch to carrier time.

### Steps to reproduce

- Make phone A call phone B in Signal
- Observe that phone B gets no notification about the incoming call just the missed one.

**Actual result:** Call arrives as missed one.
**Expected result:** Phone B rings.

### Device info

Device: Sony F5121
Android version: 7.1.1
Signal version: 4.14.10

### Link to debug log
Callee debug log: https://gist.github.com/anonymous/9984ab543d85a181b06b02954564c060
Caller debug log: https://gist.github.com/anonymous/5eae662fee114e5e07a72dbd02c0e92d

Original report: #7314

@moxie0:
> According to the logs in the issue, the devices are not able to find a valid ICE candidate.

Please, unambiguously tell there the user there is an error happening during the call and the call has never been initiated.

References: #7603

ascendingSun:
Fixes #7733 

Always retry when connection fails. PushServiceSocket -> buildOkHttpClient call is made by getServiceConnection() and submitServiceRequest(). There is a scenario where app is in background and when a incoming call notification is received, user gets a missed call notification instead of call ringing

### Contributor checklist

 * OnePlus 6: Android 10.0 
 * Virtual device: Pixel XL API 30,Android 11.0

----------

### Description
 This PR attempts to fix the issue where a user receives a missed call notification instead of the call ringing. I am aware this is a very old issue with multiple variations/scenarios reported. 
Following is a list of steps for repro:
For this test FCM should be enabled.(All notifications are enabled for app)

1. Debug the latest code in AVD(Android Virtual Device) and wait for app launch
**Keep the Debugger Console open**
2. Make a test call from a phone to AVD. This call should ring on the AVD as expected.
3. Press the power button on AVD to push app to background
4. After some time websockets are disconnected (since FCM is enabled). This usually takes ~ 1 min. 
**Look out for this log in Console to confirm connections have been torn down:**

> I/ApplicationContext: App is no longer visible.
D/WebSocketConnection: Sending keep alive...
D/WebSocketConnection: Sending keep alive...
W/IncomingMessageObserver: Application level read timeout...
D/IncomingMessageObserver: Network: true, Foreground: false, FCM: true, Censored: false, Registered: true, Websocket Registered: true
W/IncomingMessageObserver: Shutting down pipe...
I/WebSocketConnection: disconnect()
    disconnect()
I/IncomingMessageObserver: Looping...
    Waiting for websocket state change....
D/IncomingMessageObserver: Network: true, Foreground: false, FCM: true, Censored: false, Registered: true, Websocket Registered: true
I/WebSocketConnection: onClosing()
    onClose()
W/ApplicationDependencyPr: onDisconnected()
I/WebSocketConnection: onClosing()
    onClose()
W/ApplicationDependencyPr: onDisconnected()

5. Now immediately make a test call from a phone to AVD. You will see 'Network Failed!'  on your phone
6. Now try to call again. A connection error gets thrown :
org.whispersystems.signalservice.api.push.exceptions.PushNetworkException: java.io.IOException: unexpected end of stream on Connection{textsecure-service.whispersystems.org:443, proxy=DIRECT hostAddress=textsecure-service.whispersystems.org/<some ip>:443 cipherSuite=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 protocol=http/1.1}
in RestStrategy -> execute call

The fix was to enable automatic retry in buildOkHttpClient. From what I could find in the code, this call is only made from submitServiceRequest and getServiceConnection. Based on https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html, retries should not be an issue for idempotent verbs. 

After my fix, I still get 'Network Failed!' on first call attempt, But the AVD rings on second call attempt

### DIFF
--- a/libsignal/service/src/main/java/org/whispersystems/signalservice/internal/push/PushServiceSocket.java
+++ b/libsignal/service/src/main/java/org/whispersystems/signalservice/internal/push/PushServiceSocket.java
@@ -1480,7 +1480,7 @@ public class PushServiceSocket {
     return baseClient.newBuilder()
                      .connectTimeout(soTimeoutMillis, TimeUnit.MILLISECONDS)
                      .readTimeout(soTimeoutMillis, TimeUnit.MILLISECONDS)
-                     .retryOnConnectionFailure(automaticNetworkRetry)
+                     .retryOnConnectionFailure(true)
                      .build();
   }
 


afmbsr:
I hope that the efforts of the developers will join forces to fix this critical problem as soon as possible and completely. I installed the application two weeks ago and until now I have been unable to receive calls

erik55:
Did you also check if this fixes the missed calls, when the clocks on the phone are not perfectly synchronized?

ascendingSun:
@erik55 No, I tested it only for the scenario described above

AsamK:
@ascendingSun I think the underlying cause is the okhttp connection pool and not related to the websocket connection.
Can you check if the PR #10874 fixes the issue for you as well?

ascendingSun:
> @ascendingSun I think the underlying cause is the okhttp connection pool and not related to the websocket connection.
> Can you check if the PR #10874 fixes the issue for you as well?

My fix adds retries for the okhttp connections.

Increasing the timeout was something I thought about earlier. But there are 2 issues with it:

- Even with an extended timeout, there might a be unexpected network issues. So I think we still need retries 
- There might be wider implications for server/client protocol by increasing timeout. 

A potential third fix is to keep the websocket always connected. This is the least ideal since it will drain battery.

So in summary, we might need both increased timeouts and retry logic. I hope the devs can comment. This is a very old issue and it affects the critical calling workflow.

mikel2000:
Sounds great! I tried to understand, if the problem is caused on the calling or the receiving end, but still not sure about it. Would be great, if you could let me know. 

I'm asking bcs I have the same problem on my iphone when called by an android user and I wonder, if this fix would also help in this scenario. Thanks.

libsignal/service/src/main/java/org/whispersystems/signalservice/internal/push/PushServiceSocket.java
```diff
@@ -1480,7 +1480,7 @@ private OkHttpClient buildOkHttpClient(boolean unidentified) {
     return baseClient.newBuilder()
                      .connectTimeout(soTimeoutMillis, TimeUnit.MILLISECONDS)
                      .readTimeout(soTimeoutMillis, TimeUnit.MILLISECONDS)
-                     .retryOnConnectionFailure(automaticNetworkRetry)
```
greyson-signal, line 1483: `automaticNetworkRetry` is configured via a remote feature flag, and has been set to `true` for everybody for at least a few weeks now, probably longer.

greyson-signal:
Hi there! Thank you for the PR, but in effect what you've suggested has already been done. But thanks!

mikel2000:
@greyson-signal Is this problem caused on the calling or the receiving side? Thanks.

PR #10861, opened 2021-01-30
Fixed - Typing indicator showing as unread message #10833

### LINKED ISSUE #10833: Typing indicator counted as unread message + a wallpaper issue

### Bug description
The typing indicator is counted as an unread mesage. The message is not even received still but its showed in the unread count. See picture.

Also look at the black line above unread messages label. Its an issue with custo wallpaper.

### Steps to reproduce
- let someone message you
- open their chat
- when they start typing something , you can see that the unread counter shows an extra unread message counting the typing indicator

**Actual result:** 1. Typing indicator counted as unread message.
2.  Black line above unread messages label
**Expected result:** 1. Typing indicator should not be counted in unread messages
2. There should be no black line above unread messages label

### Screenshots
Screenshot_20210127-200131_Signal~2

### Device info

**Device:** Redmi Note 5 Derpfest Custom ROM
**Android version:** 11
**Signal version:** 5.3.5

### Link to debug log
Not needed

ShivanshGoel221B:
### Contributor checklist

 * Nokia 5.3, Android 10
 * Realme C2, Android 9
 *  Samsung M01 Core, Android 10
 * Virtual device, Android 6.0
- [x] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the `Fixes #10833`

----------

### Description

I created a static reference in ConversationFragment and used it to check whether the typing indicator is showing or not using an if block and handled the number of unread message header accordingly in ConversationAdapter. About the wallpaper issue stated above, I believe that behaviour is intentional. That line has been intentionally put there to distinguish between read and unread section. So, that is not an issue.
Fixes #10833

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationAdapter.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationAdapter.java
@@ -359,7 +359,10 @@ public class ConversationAdapter
   }
 
   void onBindLastSeenViewHolder(StickyHeaderViewHolder viewHolder, int position) {
-    viewHolder.setText(viewHolder.itemView.getContext().getResources().getQuantityString(R.plurals.ConversationAdapter_n_unread_messages, (position + 1), (position + 1)));
+    if (ConversationFragment.staticReference.isTypingIndicatorShowing())
+      viewHolder.setText(viewHolder.itemView.getContext().getResources().getQuantityString(R.plurals.ConversationAdapter_n_unread_messages, (position), (position)));
+    else
+      viewHolder.setText(viewHolder.itemView.getContext().getResources().getQuantityString(R.plurals.ConversationAdapter_n_unread_messages, (position + 1), (position + 1)));
 
     if (hasWallpaper) {
       viewHolder.setBackgroundRes(R.drawable.wallpaper_bubble_background_8);
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
@@ -165,6 +165,7 @@ public class ConversationFragment extends LoggingFragment {
 
   private ConversationFragmentListener listener;
 
+  public static ConversationFragment  staticReference;
   private LiveRecipient               recipient;
   private long                        threadId;
   private boolean                     isReacting;
@@ -1006,7 +1007,8 @@ public class ConversationFragment extends LoggingFragment {
     return firstVisiblePosition == 0 && list.getChildAt(0).getBottom() <= list.getHeight();
   }
 
-  private boolean isTypingIndicatorShowing() {
+  public boolean isTypingIndicatorShowing() {
+    staticReference = this;
     return getListAdapter().getHeaderView() == typingView;
   }
 


app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationAdapter.java
```diff
@@ -359,7 +359,10 @@ public void setPagingController(@Nullable PagingController pagingController) {
   }
 
   void onBindLastSeenViewHolder(StickyHeaderViewHolder viewHolder, int position) {
-    viewHolder.setText(viewHolder.itemView.getContext().getResources().getQuantityString(R.plurals.ConversationAdapter_n_unread_messages, (position + 1), (position + 1)));
+    if (ConversationFragment.staticReference.isTypingIndicatorShowing())
```
thgoebel, line 362: Does this work?

As far as I understand it, `staticReference` is still `null` at this point (since it is only initialised inside the body of `isTypingIndicatorShowing`)?

Caerbannog:
Using a static reference for UI state is probably a bad idea in a fragment. Some reasons: https://stackoverflow.com/a/11908685/1933135

I suspect that the architecture should change a bit. Though that is something that the Signal team does not accept in Pull Requests.
Currently it's the position of the `StickyHeaderDecoration` in the RecyclerView that drives the unread count. Every item in the view counts, including messages like "The Group picture was changed". Suppose that more requirements on the unread count are later introduced that decorrelate it from the position (ex: the unread count should not increase for a group picture change), or other RecyclerView item types are introduced that shouldn't coun't as unread (ex: the timezone of one of the recipients has changed). Each of these changes would cause a similar change in `onBindLastSeenViewHolder`.
Would it make more sens to find a more explicit solution like computing the unread count at the view level?

Alternatively, maybe we could change the layout of the typing indicator so that it does not influence the position. Somehow extract it from the RecyclerView. Or move it to a `RecyclerView.ItemDecoration` if they don't count in the position?

app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationAdapter.java
ShivanshGoel221B: Well, I have checked by running in the debug mode and the reference get initialized before this returns true or false.

ShivanshGoel221B:
> Using a static reference for UI state is probably a bad idea in a fragment. Some reasons: https://stackoverflow.com/a/11908685/1933135
> 
> I suspect that the architecture should change a bit. Though that is something that the Signal team does not accept in Pull Requests.
> Currently it's the position of the `StickyHeaderDecoration` in the RecyclerView that drives the unread count. Every item in the view counts, including messages like "The Group picture was changed". Suppose that more requirements on the unread count are later introduced that decorrelate it from the position (ex: the unread count should not increase for a group picture change), or other RecyclerView item types are introduced that shouldn't coun't as unread (ex: the timezone of one of the recipients has changed). Each of these changes would cause a similar change in `onBindLastSeenViewHolder`.
> Would it make more sens to find a more explicit solution like computing the unread count at the view level?
> 
> Alternatively, maybe we could change the layout of the typing indicator so that it does not influence the position. Somehow extract it from the RecyclerView. Or move it to a `RecyclerView.ItemDecoration` if they don't count in the position?

You're right sir, a static reference may not be a good idea. But, I believe that the method onBindLastSeenViewHolder does not count the unread messages, it simply checks the the position of last item counting from the sticky header. So, the only alternative to what I have done is to move the typing indicator outside the recyclerView handling the conversations, which I believe is not currently an acceptable change.
Thank you for the feedback👍

Caerbannog:
Hi @ShivanshGoel221B,
After looking at the code some more, I realised that the only `ConversationAdapter.headerView` that can exist in is a `ConversationTypingView`.

And if you look at this method, it already assumes that the header does not count as a message:
https://github.com/signalapp/Signal-Android/blob/f42ec5318f14e5a9aac529c85e728aa5b6348939/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationAdapter.java#L377-L384

So the solution is even simpler than we thought:
```java
  void onBindLastSeenViewHolder(StickyHeaderViewHolder viewHolder, int position) {
    int messagePosition = hasHeader() ? position - 1 : position;
    int count = messagePosition + 1;
    viewHolder.setText(viewHolder.itemView.getContext().getResources().getQuantityString(R.plurals.ConversationAdapter_n_unread_messages, count, count));
  ...
```

ShivanshGoel221B:
> Hi @ShivanshGoel221B,
> After looking at the code some more, I realised that the only `ConversationAdapter.headerView` that can exist is a `ConversationTypingView`.
> 
> And if you look at this method, it already assumes that the header does not count as a message:
> https://github.com/signalapp/Signal-Android/blob/f42ec5318f14e5a9aac529c85e728aa5b6348939/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationAdapter.java#L377-L384
> 
> So the solution is even simpler than we thought:
> ```java
>   void onBindLastSeenViewHolder(StickyHeaderViewHolder viewHolder, int position) {
>     int messagePosition = hasHeader() ? position - 1 : position;
>     int count = messagePosition + 1;
>     viewHolder.setText(viewHolder.itemView.getContext().getResources().getQuantityString(R.plurals.ConversationAdapter_n_unread_messages, count, count));
>   ...
> ```

You're right sir, this worked perfectly, thank you so much.
So now, should I add one more commit in this PR or create a new PR and close this one.

ShivanshGoel221B:
> > Hi @ShivanshGoel221B,
> > After looking at the code some more, I realised that the only `ConversationAdapter.headerView` that can exist is a `ConversationTypingView`.
> > 
> > And if you look at this method, it already assumes that the header does not count as a message:
> > https://github.com/signalapp/Signal-Android/blob/f42ec5318f14e5a9aac529c85e728aa5b6348939/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationAdapter.java#L377-L384
> > 
> > So the solution is even simpler than we thought:
> > ```java
> >   void onBindLastSeenViewHolder(StickyHeaderViewHolder viewHolder, int position) {
> >     int messagePosition = hasHeader() ? position - 1 : position;
> >     int count = messagePosition + 1;
> >     viewHolder.setText(viewHolder.itemView.getContext().getResources().getQuantityString(R.plurals.ConversationAdapter_n_unread_messages, count, count));
> >   ...
> > ```
> 
> You're right sir, this worked perfectly, thank you so much.
> Now I'll close this PR and create a new one.
Thank you so much again, sir.

Issue #10787, opened 2021-01-24
Cannot receive calls on Android anymore


petcap:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Since the downtime last week, I am unable to reliably receive calls on my Android phone. When someone calls me, I instantly get the "missed call" notification on my phone, even while Signal-desktop is still ringing. This problem does not seem to be restricted to Android, as I have friends and family who are on iPhone and are experiencing the same issue.

Things I've tried:
- Reset the secure session, both ways
- Turn on and off call relaying via the Signal servers
- Different network combinations, both WiFi and LTE
- Having Signal open on the called phone, with the screen on when placing a test call

### Screenshots
![Screenshot_20210124-101345](https://user-images.githubusercontent.com/8736529/105626073-eac12d80-5e2d-11eb-9252-685e56335f98.png)

### Link to debug log
This debug log is from the phone that is being called.

https://debuglogs.org/ccef93ca0b3dbfe8dc6f2e4cb5e182c58ef193631d54e405df2f85bff58f5d5a

leonidax:
This is a duplicated ticket:

https://github.com/signalapp/Signal-Android/issues/7733

valldrac:
t could be the old known issue that the app failed to retrieve the turn servers list, but this time has gotten worse.

Seems that in v5.2.2 a new feature flag was introduced to prevent okhttp from retrying network requests. This affects negatively the call service when it retrieves the TURN servers from Signal. It happens because there is no retry policy, like tasks run via JobManager.

If the websocket connection to the Signal server is stale and it has not yet restarted, the function "retrieveTurnServers" will instantly fail with "java.io.IOException: unexpected end of stream" raised by "getServiceConnection" and the call will show the network-failed popup.

afmbsr:
Isn't there a great developer fixing this problem?! Since installing the application a week ago and until now, I have not been able to receive any calls at all

AsamK:
The `retrieveTurnServers` method doesn't use the websocket connection, but just sends an HTTP request.
It seems the issue is the okhttp connection pool. Which by default keeps inactive connections around for 5 minutes, but the signal server closes inactive connections already after 60 seconds.

[...]

PR #10874, opened 2021-01-31
Configure keep alive duration for okhttp connection pool to 1 minute

AsamK:
### Contributor checklist
 * OnePlus 8T, Android 11
 * Honor 5C, Android 7

----------

### Description
The signal http server supports http keep alive, but seems to close idle connections after 1 minute.
The default OkHttp connection pool will keep idle connections in the pool for 5 minutes and doesn't notice it when the server closes connections.
As currently the automatic okhttp retries are disabled, reusing such a stale connection will be fatal.

This issue is especially severe for incoming calls, which fail because the request to retrieve the turn servers fails and isn't retried, see for example #10787

This PR configures a custom okhttp connection pool with keep alive of 1 minute, so no stale connections should be reused, under stable network conditions.
Adding some additional retry logic, which might make sense in case of other network issues, is out of scope for this PR.

### DIFF
--- a/libsignal/service/src/main/java/org/whispersystems/signalservice/internal/push/PushServiceSocket.java
+++ b/libsignal/service/src/main/java/org/whispersystems/signalservice/internal/push/PushServiceSocket.java
@@ -135,6 +135,7 @@ import javax.net.ssl.X509TrustManager;
 
 import okhttp3.Call;
 import okhttp3.Callback;
+import okhttp3.ConnectionPool;
 import okhttp3.ConnectionSpec;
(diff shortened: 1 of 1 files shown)


ascendingSun:
Nice!
I think this combined with https://github.com/signalapp/Signal-Android/pull/10841 will inch us closer to a solution.

cody-signal:
Thanks for this change, it's been merged via https://github.com/signalapp/Signal-Android/commit/4bb214cb2af6a10091438067202138ccc61aed59

PR #10897, opened 2021-02-03
Fix crash when changing contact color without contacts permission

AsamK:
### Contributor checklist
 * OnePlus 8T, Android 11

On a multi-device app, after the user changes a contact color, a multi-device contact update job is triggered, which tries to access the contact's system avatar.
This causes a crash if the contact is a system contact and the user has revoked the contacts permission.

This PR fixes the crash by checking the permission before trying to read the system contact.
An alternative solution might be to use the `Recipient.getContactPhoto()` method here, but that would be a larger change.

reported here: https://community.signalusers.org/t/bugs-in-video-call/27119


Sgn-32:
And reported here: https://community.signalusers.org/t/changing-the-chat-colour-crashes-the-app/26713

AsamK:
merged as e0128e7e31f2cca46ff28473477ebb32223f58e9

Issue #10514, opened 2021-01-11
No alerts/notifications/rings received for group video call


krishnav888:
- [X] I have searched open and closed issues for duplicates
- [X] I am submitting a bug report for existing functionality that does not work as intended
- [X] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [X] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Signal for Android. Group with 3 members. When a group video call is initiated by 1 person, the other 2 do not receive any alert/notification/ring/etc. 

One to one phone calls work fine.

### Steps to reproduce
- Create a group
- Add 3 members to group
- Initiate video call to group members
**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)
No alert/notification/rings/sounds are received by the other 2 members of the group to tell them that one of the 3 people is trying to call them.

**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)
When a group video call is initiated, the other 2 members would receive a notification/alert/ring/something to tell them that a member of the group is trying to initiate a video call with them.

### Device info

**Device:** Samsung S10e
**Android version:** 10
**Signal version:** 5.1.1

### Link to debug log

https://debuglogs.org/4939a8ef224d5f97a274d133dd934fda8393314e6e46c439af578cb04e4faaa3

dakontiva:
I understand that the not having call notifcation makes sense for large public rooms. But for small groups with friends and family this is an important feature. Maybe make this a group level setting so the admins can decide if it makes sense to their room or not.

krishnav888:
I believe that groups should be able to just have a group video/(or audio only) call. This should not be an option. It should just be function available to groups. This bug however is specifically referring to group members not receiving notifications/alerts/ring tones when a group video call is initiated by one of the group members. 

If a group member's phone does not ring when a video call is initiated by another group member, then having group video call as a feature is not as useful.

dakontiva:
Not having a ring/notification/alert is useful for groups with many members. For example, imagine a group for an opensource project there will be many members in it and not everyone would join the call or want to get a ring when someone group call is in place. However, for small groups that most people have with friends and family having a calling notification when the group call is made makes complete sense. Since there are two conflicting use cases for this, I think providing whether group calls send a call notfication or not as a configurable setting to the group admin would be a reasonable solution.

krishnav888:
Agreed thanks!

[...]

PR #10921, opened 2021-02-06
Adapt maxInstancesForQueue to only consider instances of the same job

AsamK:
### Contributor checklist
 * OnePlus 8T, Android 11

----------

### Description
Currently in the `JobController` the `maxInstancesForQueue` limit checks the count of all jobs in a
given queue. If there are already too many jobs, the new job is discarded.

However this is not the expected behavior for the two jobs where it is currently used:
`GroupCallPeekWorkerJob` and `AutomaticSessionResetJob`
For both the expected behavior is that there aren't too many jobs of them
queued, but that there will be at least one instance of them queued.
Both of them use the same queue as the `PushProcessMessageJob` and the `MarkerJob`.
Those two jobs are often in the queue at the same time, effectively preventing
the `GroupCallPeekWorkerJob` and `AutomaticSessionResetJob` from being enqueued.

This fixes the second issue with group call started notifications not appearing, mentioned here: https://github.com/signalapp/Signal-Android/issues/10514#issuecomment-774492142

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/jobmanager/Job.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/jobmanager/Job.java
@@ -407,8 +407,8 @@ public abstract class Job {
 
       /**
        * Specify the maximum number of instances you'd want of this job at any given time, as
-       * determined by the job's queue key. If enqueueing this job would put it over that limit,
-       * it will be ignored.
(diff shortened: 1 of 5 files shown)


AsamK:
merged as 8f51bdcb786b0e8e4ffc259ff22d6c3ce0a4254c

PR #10923, opened 2021-02-06
Downscale link preview thumbnails to a maximum of 512x512 px

ittner:
### Contributor checklist
 * Xiaomi Redmi 8, Android 9
 * Moto C Plus, Android 7
 * Virtual device, Android 9
----------

### Description

This commit will make Signal scale large thumbnails generated for link previews to a size no larger than 512x512 pixels before saving or sending. 

Currently, when generating a link preview, Signal will load an image provided by a remote website and, if under a simple 2 MB file size limit, accept, recompress, store on local chat history and then send it to the receivers. These images can be much larger than actually required by the intended application, wasting storage and bandwidth. The problem is exacerbated when sending link previews to large groups due to amplification effects.

Limiting the size of these images is also important as they will accumulate on receiver's chat history and is not unrealistic to expect an user to collect a few thousand link previews on long lived chat sessions or, more commonly, on group chats. Deleting them from the media list, which a typical user may be tempted to do after seeing several large but apparently unused files there, will cause the original message to be deleted too, so this approach may not be always desired.

Taking the page https://commons.wikimedia.org/wiki/File:Oxford_Tolkien.JPG as an example: current stable version (v5.3.12) generates a 1200x1800px thumbnail of 359 KB. With the changes proposed here, this thumbnail is scaled down to 341x512px, resulting in a JPEG file of only 38 KB and still with enough quality to be used in the preview.

The 512x512 px limit is mostly based on the current recommendation for stickers and seems acceptable for this context. It gives enough resolution for a good presentation, while a larger image won't necessarily give more information to the user. Opinions on this are welcome, of course.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/linkpreview/LinkPreviewRepository.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/linkpreview/LinkPreviewRepository.java
@@ -72,6 +72,7 @@ public class LinkPreviewRepository {
 
   private static final long FAILSAFE_MAX_TEXT_SIZE  = ByteUnit.MEGABYTES.toBytes(2);
   private static final long FAILSAFE_MAX_IMAGE_SIZE = ByteUnit.MEGABYTES.toBytes(2);
+  private static final int MAX_THUMBNAIL_CANVAS_SIZE = 512;
 
   private final OkHttpClient client;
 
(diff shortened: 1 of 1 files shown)


ittner:
Yes, this is still valid and the only reason there is no activity is because it was not reviewed yet, despite being in queue for almost a year.

PS. Using stalebot to handle issues is harmful: https://drewdevault.com/2021/10/26/stalebot.html

ittner:
Updating commit after a change that replaced the implementation of optional types in commit 057231b9c3d49340fbeae520dc15606fb6f26915 broke it.

Thrry (approved):

PR #11107, opened 2021-03-15
Add an additional button to take another picture (shortcut)

yan796:
### Contributor checklist

 * Device Huawei p20, Android 10.0.0

----------

### Description

To take multiples pictures one after the other, the current flow involve:
1. Take picture
2. click "+"
3. click up-right camera icon

This flow is tedious because 1) there are three steps although 2 should be enough, and 2) the step 3. is using a camera button which is diagonally opposed to the capture button and thus really hard to reach with one hand
This feature was suggested here:
https://community.signalusers.org/t/ui-improvement-for-adding-multiple-camera-pictures/9200/8

before this PR, 2. and 3. look like that:
device-2021-03-15-145113
device-2021-03-15-150052

With this PR: 2 looks like that (and 3 is not needed in this flow)
device-2021-03-15-145440

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/mediapreview/MediaRailAdapter.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/mediapreview/MediaRailAdapter.java
@@ -21,6 +21,7 @@ public class MediaRailAdapter extends RecyclerView.Adapter<MediaRailAdapter.Medi
 
   private static final int TYPE_MEDIA  = 1;
   private static final int TYPE_BUTTON = 2;
+  private static final int TYPE_CAMERA = 3;
 
   private final GlideRequests            glideRequests;
   private final List<Media>              media;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/thoughtcrime/securesms/mediapreview/MediaRailAdapter.java
```diff
@@ -206,6 +220,6 @@ void recycle() {
   }
 
   public interface RailItemAddListener {
-    void onRailItemAddClicked();
+    void onRailItemAddClicked(boolean startCamera);
```
Caerbannog, line 223: I'm not an Android expert, and I'm not familiar with this bind/listen pattern, but I find this new boolean less explicit than creating a new `onRailItemCameraClicked` method. Would that also work?

Caerbannog:
### About the code
I'm not a member of the Signal team. Feel free to ignore the above comments, I have no more clue than you. I just tried to give you an outside impression to help your PR along.

The main point that I'm not sure about is reusing the `onRailItemAddClicked` and `onCameraSelected` plumbing. It's certainly smart. Maybe something more explicit with more code change would make future changes easier.

When the camera is clicked, this PR currently opens the file picker then navigates to the camera. Could the camera be opened directly? When navigating back from the camera, the file picker is currently opened. This change would go back to the media preview.

### About the feature as it stands
I have tested this PR on a Samsung Grand Prime, Android 5.1.1.
The change works as expected.

And it is great!

thebigalwisi:
Hi guys,

I make a lot of fotos in Signal and i hate the way it works now... 
Please make it real!

PR #11157, opened 2021-03-31
Fixes #11156 When videos are played pause Voice Memos

### LINKED ISSUE #11156: Voice Memos Continue to Play When you Open a Video

### Bug description
When you are playing a voice message, and then you play a video, the voice message still plays such that the user hears both the voice message **and** the video's audio. 

### Steps to reproduce
- be in a conversation with a voice message and video 
- play the voice message 
- play the video 

**Actual result:** The video plays with the voice message continuing to play in the background
**Expected result:** The voice message should pause before the video player opens 

### Device info

**Signal version:** 5.6.0

bitmold:
### Contributor checklist

 * Virtual Nexus 5X , Android 29

----------

### Description

When the user opens a video message while a voice memo/message is playing, the `ConversationFragment` will handle pausing it..

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
@@ -431,6 +431,7 @@ public class ConversationFragment extends LoggingFragment {
     }
   }
 
+
   private int getStartPosition() {
     return conversationViewModel.getArgs().getStartingPosition();
   }
@@ -1469,6 +1470,12 @@ public class ConversationFragment extends LoggingFragment {
     @Override
     public void onVoiceNotePause(@NonNull Uri uri) {
       voiceNoteMediaController.pausePlayback(uri);
+    }
+
+    @Override
+    public void onVideoMessagePlay() {
+      Uri voiceNoteUri = voiceNoteMediaController.getVoiceNotePlaybackState().getValue().getUri();
+      voiceNoteMediaController.pausePlayback(voiceNoteUri);
     }
 
     @Override
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationItem.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationItem.java
@@ -1528,6 +1528,10 @@ public final class ConversationItem extends RelativeLayout implements BindableCo
         intent.putExtra(MediaPreviewActivity.CAPTION_EXTRA, slide.getCaption().orNull());
         intent.putExtra(MediaPreviewActivity.LEFT_IS_RECENT_EXTRA, false);
 
+        if (slide.hasVideo()) {
+          eventListener.onVideoMessagePlay();
+        }
+
         context.startActivity(intent);
       } else if (slide.getUri() != null) {
         Log.i(TAG, "Clicked: " + slide.getUri() + " , " + slide.getContentType());
--- a/app/src/main/java/org/thoughtcrime/securesms/BindableConversationItem.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/BindableConversationItem.java
@@ -68,6 +68,7 @@ public interface BindableConversationItem extends Unbindable {
     void onJoinGroupCallClicked();
     void onInviteFriendsToGroupClicked(@NonNull GroupId.V2 groupId);
     void onEnableCallNotificationsClicked();
+    void onVideoMessagePlay();
 
     /** @return true if handled, false if you want to let the normal url handling continue */
     boolean onUrlClicked(@NonNull String url);


app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
```diff
@@ -431,6 +431,7 @@ public void onWallpaperChanged(@Nullable ChatWallpaper wallpaper) {
     }
   }
 
+
```
alex-signal, line 434: Extra space.

app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationItem.java
```diff
@@ -1528,6 +1528,10 @@ public void onClick(final View v, final Slide slide) {
         intent.putExtra(MediaPreviewActivity.CAPTION_EXTRA, slide.getCaption().orNull());
         intent.putExtra(MediaPreviewActivity.LEFT_IS_RECENT_EXTRA, false);
 
+        if (slide.hasVideo()) {
+          eventListener.onVideoMessagePlay();
```
alex-signal, line 1532: I'd rather we do this within the preview fragment itself, rather than in response to clicking an item in the conversation. What happens when we select a video from "All Media" ?

alex-signal (changes requested):
I think this only covers the case where a user selects a video within a conversation.

app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationItem.java
bitmold: This makes sense, I'll make those changes soon - thank you

bitmold:
This seems trickier to do than I thought... @alex-signal  I'm able to crudely fix this bug with the code in this pull request and by writing similar additional code to handle the "All Media" case you described above.

However, I can't do the more elegant thing you had mentioned of stopping whatever might be playing from the `MediaPreviewFragment`. 

For starters, I am creating an instance of `VoiceNoteMediaController` in `MediaPreviewFragment`'s `onActivityCreated` similar to what is done in `ConversationFragment` and `MediaOverviewPageFragment`. 

I am unable to hook into what the current voice note URI is in `MediaPreviewFragment`. Further, invoking `MediaControllerCompat.getMediaController(activity)` from the `VoiceNoteMediaController` object in the `MediaPreviewFragment` returns null. Whereas invoking this method from the `VoiceNoteMediaController` in `ConversationFragment` or `MediaOverviewPageFragment` returns a real object. 

When I run this method from the Android SDK (in `MediaControllerCompat`) in the debugger: 

```java
public static MediaControllerCompat getMediaController(@NonNull Activity activity) {
        if (activity instanceof ComponentActivity) {
            MediaControllerExtraData extraData =
                    ((ComponentActivity) activity).getExtraData(MediaControllerExtraData.class);
            return extraData != null ? extraData.getMediaController() : null;
...
```
I notice that `extraData` is not null when the Activity is `ConversationActivity` or `MediaOverviewActivity`, but is null if the Activity is a `MediaPreviewActivity`.

I feel like I'm missing something simple here, but it's not clear to me right now. Do you have any suggestions as to what I should be looking into? I'm going to return to this soon...

alex-signal:
@bitmold You shouldn't need the Uri, you can just create a new method in the VoiceNoteMediaController that skips the URI check and just calls stop() (see the existing method, you'd basically be creating a duplicate method without the if check)

As for the MediaController, you're approach is correct. However, the MediaController is not set on the Activity until after `connected` occurs. (See VoiceNoteMediaController line 199).

One approach would be to listen to the `VoiceNotePlaybackState` LiveData. This will only fire after a connection has been made, and has the Uri in question you were looking for. However, from what I remember this will serve back a lot of updates, as it's used to update seek bar position as well. So, you will want to do some filtering to only respond in situations that actually matter. Perhaps something like:

```
LiveData<VoiceNotePlaybackState> state = voiceNoteMediaController.getVoiceNotePlaybackState();
LiveData<Uri> uri = Transformations.distinctUntilChanged(Transformations.map(state, VoiceNotePlaybackState::getUri));

uri.observe(getViewLifecycleObserver(), uri -> {
  if (uri != Uri.EMPTY) {
      voiceNoteMediaController.stopPlaybackAndReset(uri);
  }
});
```

bitmold:
I apologize if I'm missing something here, but as I understand it now, it wouldn't make sense to use `VoiceNoteMediaController` in the `MediaPreviewFragment` at all.

This is because Android's `MediaController` is associated to an Activity that maintains some kind of media UI view (which would be contained within the previous Activity prior to the media preview being started). The VNMC interfaces with the and syncs with an instance of the `VoiceNotePlaybackService`. So I'm inclined to think it would be appropriate to send a message to the `VoiceNotePlaybackService` if it's running and tell it to pause. Pausing also makes sense because if the user is within a long voice note they can return to it after playing a video without having to scrub to try and recover their position. 

> you can just create a new method in the VoiceNoteMediaController that skips the URI check and just calls stop() (see the existing method, you'd basically be creating a duplicate method without the if check) 

In line with this understanding, this wouldn't work with a VNMC spun up in `MediaPreviewFragment`. The reason the if check crashes is because it invokes a call to `getMediaController()` which returns null since `MediaPreviewFragment/Activity` has no media controller; therefore skipping the if check and saying `getMediaController().getTransportControls().stop()` would produce the same null pointer exception I was getting earlier. 

Currently experimenting with this approach now...

alex-signal:
Sending a message to the service could work. The crash you noted is due to a race condition (trying to get the controller too early). I think the method with the livedata I mentioned would work fine, since it would not emit a value until the media controller had been set on the activity, but I understand the hesitancy to go that route.

bitmold:
Still happens

bitmold:
Still happens

alex-signal:
@bitmold Last I saw you were experimenting with some things. Did that yield a fix?

PR #11223, opened 2021-04-20
Add option to block unknown senders

valldrac:
### Contributor checklist

 * Device Pixel 3, Android 10
 * Virtual device, Android 9
 * Virtual device, Android 11

----------

### Description

It adds a new option to block messages from unknown senders. This option is useful for users who keep in touch only with a set of known contacts, and don't want to receive any messages from someone else. It's a sort of allow-list for security, anti-spam, or just convenience for small teams or communities. Many users have requested this feature in the forum.

The incoming messages from unknown senders are just ignored when this option is enabled, following the same rules as if the recipient was already blocked by the user. An "unknown sender" is a recipient that it's not present in the system contacts, or the user doesn't share their profile with.

I also added two more commits to fix some inconsistencies in the blocked-users preference UI. The goal is to make the fragment more like the rest of the preferences of the app (colors, font size, ripple, and padding).

### Screenshots

| Signal v5.9.0 | This PR |
|--|--|
| Screenshot_1618940431 | Screenshot_1618939778 |

| Signal v5.9.0 (Dark) | This PR (Dark) |
|--|--|
| Screenshot_1618940049 | Screenshot_1618940262 |

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/blocked/BlockedUsersActivity.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/blocked/BlockedUsersActivity.java
@@ -164,6 +164,8 @@ public class BlockedUsersActivity extends PassphraseRequiredActivity implements 
         throw new IllegalArgumentException("Unsupported event type " + event);
     }
 
-    Snackbar.make(view, getString(messageResId, displayName), Snackbar.LENGTH_SHORT).show();
+    Snackbar.make(view, getString(messageResId, displayName), Snackbar.LENGTH_SHORT)
+            .setTextColor(Color.WHITE)
+            .show();
   }
 }
--- a/app/src/main/java/org/thoughtcrime/securesms/blocked/BlockedUsersFragment.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/blocked/BlockedUsersFragment.java
@@ -11,12 +11,14 @@ import android.view.ViewGroup;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.appcompat.widget.SwitchCompat;
 import androidx.fragment.app.Fragment;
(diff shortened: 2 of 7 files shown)


JordyEGNL:
Yes! This could be very useful against all those spam messages people get. And this is an improvement on privacy and security.

JordyEGNL (approved):

soladev (approved):
Finally ! I would say that at least 90% of people around me don't use Signal anymore because of spams (In France it seems to be a far greater problem than anywhere else). So nice work and I hope this option can bring back some users on Signal

valldrac:
> Is this still relevant? If so, what is blocking it? Is there anything you can do to help move it forward?

Don't suffer dear bot looking for answers that will never come. Let me close this by myself.

dsshad:
Hello,
how can such an important feature get marked as "wontfix" and get closed, even though someone has already put in all the work and has everything ready? Thanks valldrac for the work!

We’ve installed Signal for an 8 years old on an relatively locked older iPhone, so that he can chat or do a video call with family members. For this we used a new SIM-Card. We tried to set up everything in Signal as secure as possible, i.e. also turned off the setting to allow messages from everyone, though it has a different purpose apparently. Nevertheless the kid received a message from a foreign person 1-2 days later…

From the sound of the text the person must have had that phone number from some of the previous owners, so probabgly got notified by Signal, once we used the numer for registration.

There were options to allow or to block the message / the sender, but it was already one step too far, because the name, number of the other person was visible and also the text of the message.

That’s far from being optimal and the kid could have clicked “Accept” etc. It was somewhat shocking that something like this was possible with Signal, considering it's dedication to privacy etc.

So a feature like the one above or some kind or parential control would definitely make sense.

If kids are old enough and want to find ways around it, they probably will, but that’s by far not the only use case…

PR #11253, opened 2021-05-02
Switch to updating keyboard state in onSizeChanged.

### LINKED ISSUE #11252: Back button doesn't work in conversation activity after lock/unlock cycle

### Bug description
I've noticed on my device that the conversation activity will occasionally get into a state where the back button doesn't do anything. I'm not sure why this only happens on my device, but I think I found the root cause.

### Steps to reproduce
- Open a conversation (I have 100% repro rate in 1:1 conversations but groups don't seem to repro for some reason)
- Begin composing a message to open the keyboard
- Lock the screen
- Unlock the screen (critically, on my device the keyboard is closed by the system during this lock/unlock cycle and the bug does not repro on my friends' devices where this is not true)
- Press the back button

**Actual result:** 
The app will be in a state where the back button has no effect regardless of how many times it's pressed.

**Expected result:** 
The back button should pop the conversation activity and bring up the conversation list.

### Screenshots
Not comfy sharing screenshots since I don't have a burner phone sorry :(

### Device info

**Device:** Pixel 4A
**Android version:** 11 (running GrapheneOS not the stock rom)
**Signal version:** 5.8.x (Going from memory here) and a build built from the current master branch.

### Link to debug log

Let me know if you need this.

### Probable root cause

For some reason the system chooses to close the soft keyboard during the lock/unlock cycle. I'm not entirely sure why this happens. When the system is unlocked, `onMeasure` is called in `KeyboardAwareLinearLayout`, but I don't think the resize has actually completed yet. My hypothesis is that `getAvailableHeight` returns the old height, which causes the keyboard state to not be updated during the device unlock. Since the old state sticks around, the layout continues to believe that the keyboard is open. That bad state in `KeyboardAwareLinearLayout` causes the `onBackPressed` logic in `ConversationActivity` to trigger the keyboard closure flow instead of the `super.onBackPressed` flow. Since the keyboard isn't actually open, it can't be closed so the entire back button press is a no-op.

It seems like triggering `updateKeyboardState` and `onRotation` in `KeyboardAwareLinearLayout.onSizeChanged` instead of `KeyboardAwareLinearLayout.onMeasure` fixes this issue, which supports the theory that onMeasure is called too early in the resize operation for `getAvailableHeight` to return the correct (new) value. That could have ripple effects elsewhere in the codebase but it seems like a fairly safe change overall, so I'm going to send a PR for it.

ellenhp:

### Contributor checklist

 * Pixel 4A, Android 11 (GrapheneOS)

----------

### Description

This fixes #11252. I'm not sure if this is the best way to fix that bug, but it makes sense to me based on my understanding of what's going on this code. I've tested by following the repro steps in #11252 and ensuring that the bug is fixed. Probably worth doing some additional testing though since I don't have other devices readily available to test on. :)

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/components/KeyboardAwareLinearLayout.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/KeyboardAwareLinearLayout.java
@@ -81,10 +81,10 @@ public class KeyboardAwareLinearLayout extends LinearLayoutCompat {
   }
 
   @Override
-  protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
+  protected void onSizeChanged(int w, int h, int oldW, int oldH) {
     updateRotation();
     updateKeyboardState();
-    super.onMeasure(widthMeasureSpec, heightMeasureSpec);
+    super.onSizeChanged(w, h, oldW, oldH);
   }
 
   private void updateRotation() {


ellenhp:
This doesn't seem to completely solve #11252 but it does make it much more difficult to repro (20 repros/20 attempts -> approx 1/20). Switching the PR to a draft and will continue looking into what's going on.

PR #11292, opened 2021-05-12
set contact avatar for notification previews of media messages pre API28

clauz9:

### Contributor checklist

 * Samsung J5 Android 6.01
 * Samsung S5 mini Android 6.0.1
 * Philips S326 Android 5.0.1

----------

### Description

@cody-signal I think this time I did it correctly


cody-signal:
@clauz9  I think we want to not use big picture style whenever possible. Can we add a check to bypass it if the API supports messaging style, I believe that's API24+ but you'll need to double check. Additionally, do you think we should show the conversation icon (which could be a group avatar) or the person who's sending the image (would need to look at the message notification directly). Right now with your fix, it would be the group avatar. What do you think?

clauz9:
> @clauz9 I think we want to not use big picture style whenever possible. Can we add a check to bypass it if the API supports messaging style, I believe that's API24+ but you'll need to double check. 

you are correct, "added API level 24" 👍 

image

I will refactor it to use bigPictureStyle only for API 23 and lower, 

> Additionally, do you think we should show the conversation icon (which could be a group avatar) or the person who's sending the image (would need to look at the message notification directly). Right now with your fix, it would be the group avatar. What do you think?

I'm not really sure about this, my first instinct is to say it should show the group avatar, because that's the most prominent feature and it will allow the user to distinguish a direct message from a group message at a glance, but on the other hand unlike android 10, notifications, the group name in android 6.0.1 notifications is big enough to be visible, but the contact name is not that big, I think ultimately, if we could make the group name as prominent as it is on android 6.0.1 we could use the persons, avatar for all group notifications on all android versions, otherwise, I would suggest, group avatar for Message style notification, and the contact avatar for the rest, but this introduces inconsistencies across devices with regards to group notifications....

you know what, I'm gonna work on this more, and will submit some commits to this branch to make it how I think it should work, fell free to ignore this pull request or close it if you want, and put in whatever fix you consider, for the avatar issue if it can't wait.

PR #11372, opened 2021-06-06
Fixes #10940 (rotation not updated on phones using Legacy-API)

### LINKED ISSUE #10940: Devices with Legacy level Camera2 API show badly rotated video preview, when changing phone orientation
- [X] I have searched open and closed issues for duplicates
- [X] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports

### Bug description
Opening the camera in Signal in portrait mode and rotating it to landscape, makes the preview window display the video rotated 90 degrees opposite to the direction in which you've rotated the phone. This also happens if your start the camera in landscape mode and rotate to portrait. I've tested this on 3 phones, 2 have this bug,  but the third does not. a Samsung SM-A105FN. 
The pictures get saved correctly. 

### Steps to reproduce
- Open Signal
- Open the camera, in portrait or landscape
- Rotate to landscape or portrait

**Actual result:** The preview window shows the video rotated opposite to the orientation of the phone. 
**Expected result:** I expect it to work the same way, most default camera apps work, keeping the video stream locked in the preview window and just rotate the overlay icons in place around their axis to correspond to gravity. 

### Screenshots
- Signal landscape preview window

pic

- default camera app landscape preview (overlay controls rotated only) 

image

- default camera app portrait preview 

image

### Device info

**Device and Android version:** Nokia 3.1, with Android 10 
**Device and Android version:** Philips S326, with Android 5.1

**Signal version:**  5.3.12 but it also happens in 5.3.5 

### Link to debug log
- Nokia with Signal v5.3.5 
log: https://debuglogs.org/fce7da1c7e47bcbc947014f41b67b5f1a56fb8b08d7bd0613c55dec180755b78

- Philips with Signal v5.3.5 
log: https://debuglogs.org/a917cbbcfb28140af44b1e9e460c2a5426d3e4d80557799e5343e2df22ee73cb

- Nokia with Signal 5.3.12
log: https://debuglogs.org/ba751ebace4a90e449fb6c009d42d402dbab1460e19a130ed58ed4b5e8372d98

- Philips with Signal v5.3.12
log: https://debuglogs.org/89c74b9dae7897a5125f9bd00a0b6c0712b368ff7d851667f53e09ddbb968371

rainerm77:
### Contributor checklist

 * MotoG5, Android 8.1.0
 * A few more devices tested and confirmed by @clauz9

----------

### Description
Fixes #10940
See details in the #10940 issue description (by @clauz9) and additionally discussion in the Signal community-forum here:
https://community.signalusers.org/t/starting-built-in-camera-in-portrait-and-then-turning-phone-to-landscape-mode-shows-the-preview-video-90degrees-wrong/28603?u=rainerm77

Signal-community user "clauz" was very helpful and tested this fix on 4 devices, one of which was the one that worked before, and they all rotate correctly.

Remark to @devs:
Please note that this fix uses `Preview.Builder.setTargetRotation()` which in `Preview.java` is marked as 'ExperimentalUseCaseGroup' (I'm not sure if this fact is seen as problematic. If yes, we would need to find an alternative solution)...

### DIFF
--- a/app/src/main/java/androidx/camera/view/SignalCameraXModule.java
+++ b/app/src/main/java/androidx/camera/view/SignalCameraXModule.java
@@ -513,6 +513,10 @@ final class SignalCameraXModule {
   }
 
   public void invalidateView() {
+    if (mPreview != null) {
+      mPreview.setTargetRotation(getDisplaySurfaceRotation()); // Fixes issue #10940 (rotation not updated on phones using "Legacy API")
+    }
+
     updateViewInfo();
   }
 
--- a/app/src/main/java/androidx/camera/view/SignalCameraXModule.java
+++ b/app/src/main/java/androidx/camera/view/SignalCameraXModule.java
@@ -512,6 +512,7 @@ final class SignalCameraXModule {
     return rotationDegrees;
   }
 
+  @SuppressLint("UnsafeExperimentalUsageError")
   public void invalidateView() {
     if (mPreview != null) {
       mPreview.setTargetRotation(getDisplaySurfaceRotation()); // Fixes issue #10940 (rotation not updated on phones using "Legacy API")


clauz9:
I'm confirming that this has been tested and works correctly on the following devices,

Samsung SM-G800F, with Android 6.0.1
Samsung SM-J500F, with LineageOS 14.1
Samsung A10, with Android 10 - did not suffer from this issue 
Nokia 3.1, with Android 10

rainerm77:
Hello @cody-signal 

May I kindly ask you just to give a short feedback:
- if the use of SuppressLint(“UnsafeExperimentalUsageError”) would be ok for certain fixes, such as this one?
- or, if otherwise, I would have to find an alternative way, to finally fix this rotation-bug?

Thanks in advance, kind regards,
Rainer

cody-signal:
Hi @rainerm77, thanks for the PR.

I'm curious how this is different than the default behavior and guessing it's a bug from us having to use an older camerax dependency.

https://developer.android.com/reference/androidx/camera/core/Preview.Builder#setTargetRotation(int) docs say it'll use Display.getRotation if nothing is specified which is the same that `getDisplaySurfaceRotation` returns.

Since what you are doing is what the behavior claims to do by default, this should be fine to include.

cody-signal:
Hi @rainerm77, thanks for the PR.

I'm curious how this is different than the default behavior and guessing it's a bug from us having to use an older camerax dependency.

https://developer.android.com/reference/androidx/camera/core/Preview.Builder#setTargetRotation(int) docs say it'll use Display.getRotation if nothing is specified which is the same that `getDisplaySurfaceRotation` returns.

Since what you are doing is what the behavior claims to do by default, this should be fine to include.

rainerm77:
Thanks for the response. I also think, for (newer) phones which are not affected my fix is somewhat redundant - so it should not have negative effects hopefully. On my phone the bug occured, and it annoyed me quite a while ;-)

Remark on hardware-dependency: @clauz9 came to the assumption, that it depends on the camera-API of the phone. He observed that it seems to occur only on phones with hardware support level "Legacy", see his remarks:
https://community.signalusers.org/t/starting-built-in-camera-in-portrait-and-then-turning-phone-to-landscape-mode-shows-the-preview-video-90degrees-wrong/28603/15?u=rainerm77
 
At least, his tests of my bugfix confirmed, that it fixed the issue for 3 of his phones using "Legacy", and showed no negative side-effect on his 1 phone  with has a higher level.

To be on the save, maybe in your dev-team you could cross-check if this fix works / applies also to your test-phones?
And if it works, I would be happy to help with my PR :-)

rainerm77:
Hello @cody-signal , maybe I've misunderstood your previous message regarding "default behavior".

- Regarding Preview.java: Although the doc of `Preview.setTargetRotation` says "If not set, the target rotation will default to the value of Display.getRotation()", there is no overload method existing without this param (int). Thus, I had to use it this way  passing the rotation.

- The trick of my fix is that I additionally/explicitly call `setTargetRotation` within `invalidateView`. This seems to fix the issue for affected phones, and for not-affected phones, it might be redundant but seems to have no negative side effects.

cody-signal:
We'll be including the fix in 5.16 (next beta)

rainerm77:
Cool!
I guess this gets me motivated to try to fix maybe similar things next months 🙂 (if I find enough spare time besides my job)... Greets, Rainer

PR #11381, opened 2021-06-11
Add check for activity in Bubble for Android 12

### LINKED ISSUE #11224: Crash when opening bubbled conversation

### Bug description
I bubbled a conversation on my Android phone. When I received a message in that conversation and attempted to open the bubble by clicking on the bubble (not the notification), the bubble disappeared and the message never opened. I had to open the app and click on the conversation to view the message.

### Steps to reproduce
The issue is intermittent and I haven't been able to reliably reproduce it.

### Device info
**Device:** Google Pixel 5
**Android version:** 11 (security update on April 5th; Play system update on March 1)
**Signal version:** 5.7.6

### Link to debug log
https://debuglogs.org/f1f6873688219b2af2b5d8b4e543d8b7336b17e493fc1058dba579a698341b4e

alabiaga:
Fixes #11224 

@cody-signal  This is the fix for #11224. Note that I did not actually test this by building Signal, as I am having issues compiling the project, so please apply a similar change as you see fit.

Please find the relevant API change documentation here: https://developer.android.com/reference/android/app/Activity#isLaunchedFromBubble()

Additional sample: https://github.com/android/user-interface-samples/tree/main/People#bubbles-1

Also note that an appcompat lib (ActivityCompat) is in the works and will be available in the future

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationActivity.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationActivity.java
@@ -2082,12 +2082,23 @@ public class ConversationActivity extends PassphraseRequiredActivity
 
   private boolean isInBubble() {
     if (Build.VERSION.SDK_INT >= ConversationUtil.CONVERSATION_SUPPORT_VERSION) {
-      Display display = getDisplay();
-
-      return display != null && display.getDisplayId() != Display.DEFAULT_DISPLAY;
-    } else {
-      return false;
+      return isLaunchedFromBubbleApi30();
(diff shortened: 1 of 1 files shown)


cody-signal:
Thanks @alabiaga - This code is a bit problematic since it relies on APIs from Android 12's beta sdk. I've pulled the intent of your fix (change how we check isInBubble) but implemented in a way that doesn't require knowledge of API31 APIs. Fix will be apart of 5.15.

alabiaga:
Correct, this fix would only be applicable once you compile against the
Android 12 sdk. You can you apply then. Also note that your issue with
bubbles is only applicable in Android 12. Thanks Cody.

Issue #11537, opened 2021-08-11
Sharing shortcuts for Signal conversations do not show up in the Android Sharesheet


nielthiart:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [ ] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

#### Bug description
Signal is by far my most used messaging app, but it never shows up on [Android Sharesheet](https://developer.android.com/training/sharing/send#why-to-use-system-sharesheet) on my phone.

Nor does it show up for these users:
- https://www.reddit.com/r/signal/comments/obavzq/direct_share_not_working/
- https://www.reddit.com/r/signal/comments/khb3r4/signal_why_dont_you_support_androidios_share/
- https://www.reddit.com/r/signal/comments/o4ctkv/top_contacts_not_suggested_by_android_when_doing/
- https://community.signalusers.org/t/support-new-android-share-drawer/15527

I think this happens because Signal does not declare a share target.

### Steps to reproduce

- Install Signal as the only messaging app and send a message to any contact.
- Open a different app on the phone and share a photo.
- You should see the Android Sharesheet

**Actual result:** No Signal Shortcuts among the four suggested share targets.
**Expected result:** Signal shortcuts among the suggested share targets.

### Device info

Model: LM-G710EM
Android Version: 10
SDK Version: 29
Signal version: 5.19.4

### Proposed fix

1.  [Declare a share target](https://developer.android.com/training/sharing/receive#declare-share-target)
    ```
    <?xml version="1.0" encoding="utf-8"?>
    <shortcuts xmlns:android="http://schemas.android.com/apk/res/android">
        <share-target android:targetClass="org.thoughtcrime.securesms.sharing.ShareActivity">
            <data android:mimeType="*/*" />
            <category android:name="org.thoughtcrime.securesms.sharing.CATEGORY_SHARE_TARGET" />
        </share-target>
    </shortcuts>
    ```

2.  When either `android.intent.action.SEND` or `android.intent.action.SEND_MULTIPLE` creates `org.thoughtcrime.securesms.sharing.ShareActivity`, before initiating the search/chooser interface, check for the following extra intent field and handle a direct share immediately if it is present: `EXTRA_SHORTCUT_ID`.

[...]

PR #11539, opened 2021-08-11
Fix Signal Direct Share Shortcuts not appearing in Android Sharesheet

nielthiart:
1.  [Declare a share target](https://developer.android.com/training/sharing/receive#declare-share-target)

2.  In `org.thoughtcrime.securesms.sharing.ShareActivity`, before initiating the search/chooser interface, check for the following extra intent field and handle a direct share immediately if it is present: `EXTRA_SHORTCUT_ID`.

    Since a direct share action does not include intent as originally added when building the dynamic shortcut, we need to find the shortcut and get extra fields (threadId, distributionType) from the shortcut.

3.  If direct share fails, show search/chooser UI as before.

### Contributor checklist

 * LM-G710EM, Android 10
 * Virtual device Nexus 5X, Android 11

----------

### Description

Signal does not show up on Android Sharesheet, because it:
1. does not declare a [share target](https://developer.android.com/training/sharing/receive#declare-share-target),
2. does not handle androidx.core.content.pm.ShortcutManagerCompat.EXTRA_SHORTCUT_ID at all, and
3. relies on the original shortcut intent when it isn't present on direct share.

This pull request intends to fix these three issues.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
@@ -283,7 +283,7 @@ public class ShareActivity extends PassphraseRequiredActivity
    * @param extraShortcutId EXTRA_SHORTCUT_ID string as included in direct share intent
    * @return shortcut extras or null
    */
-  @Nullable private Bundle getShortcutExtrasFor(@NonNull String extraShortcutId) {
+  private @Nullable Bundle getShortcutExtrasFor(@NonNull String extraShortcutId) {
     List<ShortcutInfoCompat> shortcuts = ShortcutManagerCompat.getDynamicShortcuts(this);
     for (ShortcutInfoCompat shortcutInfo : shortcuts) {
       if (extraShortcutId.equals(shortcutInfo.getId())) {
--- a/app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
@@ -287,6 +287,7 @@ public class ShareActivity extends PassphraseRequiredActivity
    * @param extraShortcutId EXTRA_SHORTCUT_ID String as included in direct share intent
    * @return shortcutExtras or null
    */
+  @WorkerThread
   private @Nullable Bundle getDirectShareExtras(@NonNull String extraShortcutId) {
     Bundle shortcutExtras = getShortcutExtrasFor(extraShortcutId);
     if (shortcutExtras == null) {
@@ -301,6 +302,7 @@ public class ShareActivity extends PassphraseRequiredActivity
    * @param extraShortcutId EXTRA_SHORTCUT_ID String as included in direct share intent
    * @return shortcutExtras or null
    */
+  @WorkerThread
   private @Nullable Bundle getShortcutExtrasFor(@NonNull String extraShortcutId) {
     List<ShortcutInfoCompat> shortcuts = ShortcutManagerCompat.getDynamicShortcuts(this);
     for (ShortcutInfoCompat shortcutInfo : shortcuts) {
@@ -314,6 +316,7 @@ public class ShareActivity extends PassphraseRequiredActivity
   /**
    * @param extraShortcutId EXTRA_SHORTCUT_ID string as included in direct share intent
    */
+  @WorkerThread
   private @Nullable Bundle createExtrasFromExtraShortcutId(@NonNull String extraShortcutId) {
     Bundle      extras           = new Bundle();
     RecipientId recipientId      = ConversationUtil.getRecipientId(extraShortcutId);
--- a/app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
@@ -428,7 +428,7 @@ public class ShareActivity extends PassphraseRequiredActivity
 
   private void handleDirectShare() {
     Intent      intent           = getIntent();
-    String extra_shortcut_id     = intent.getStringExtra(ShortcutManagerCompat.EXTRA_SHORTCUT_ID);
+    String      extraShortcutId  = intent.getStringExtra(ShortcutManagerCompat.EXTRA_SHORTCUT_ID);
     long        threadId         = -1;
     int         distributionType = -1;
     RecipientId recipientId      = null;
@@ -436,9 +436,9 @@ public class ShareActivity extends PassphraseRequiredActivity
     List<ShortcutInfoCompat> shortcuts = ShortcutManagerCompat.getDynamicShortcuts(this);
 
     for (int a = 0, N = shortcuts.size(); a < N; a++) {
-      ShortcutInfoCompat shortcut_info = shortcuts.get(a);
-      if (extra_shortcut_id.equals(shortcut_info.getId())) {
-        Bundle extras = shortcut_info.getIntent().getExtras();
+      ShortcutInfoCompat shortcutInfo = shortcuts.get(a);
+      if (extraShortcutId.equals(shortcutInfo.getId())) {
+        Bundle extras = shortcutInfo.getIntent().getExtras();
         recipientId = RecipientId.from(extras.getString(EXTRA_RECIPIENT_ID, null));
         threadId = extras.getLong(EXTRA_THREAD_ID, -1);
         distributionType = extras.getInt(EXTRA_DISTRIBUTION_TYPE, -1);
--- a/app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
@@ -55,6 +55,7 @@ import org.thoughtcrime.securesms.mediasend.MediaSendActivity;
 import org.thoughtcrime.securesms.recipients.Recipient;
 import org.thoughtcrime.securesms.recipients.RecipientId;
 import org.thoughtcrime.securesms.sharing.interstitial.ShareInterstitialActivity;
+import org.thoughtcrime.securesms.util.ConversationUtil;
 import org.thoughtcrime.securesms.util.DynamicLanguage;
 import org.thoughtcrime.securesms.util.DynamicNoActionBarTheme;
 import org.thoughtcrime.securesms.util.DynamicTheme;
@@ -277,8 +278,7 @@ public class ShareActivity extends PassphraseRequiredActivity
   }
 
   /**
-   * Search for dynamic shortcut originally declared in
-   * {@link org.thoughtcrime.securesms.util.ConversationUtil} and return extras
+   * Search for dynamic shortcut originally declared in {@link ConversationUtil} and return extras
    *
    * @param extraShortcutId EXTRA_SHORTCUT_ID string as included in direct share intent
    * @return shortcut extras or null
@@ -307,12 +307,15 @@ public class ShareActivity extends PassphraseRequiredActivity
    * @param extraShortcutId EXTRA_SHORTCUT_ID string as included in direct share intent
    */
   private void createShortcutIntentFromExtraShortcutId(@NonNull String extraShortcutId) {
-    RecipientId recipientId = RecipientId.from(extraShortcutId);
-    Long        threadId    = DatabaseFactory.getThreadDatabase(this).getThreadIdFor(recipientId);
+    RecipientId recipientId = ConversationUtil.getRecipientId(extraShortcutId);
+    Long        threadId    = null;
 
-    getIntent().putExtra(EXTRA_RECIPIENT_ID, recipientId);
-    getIntent().putExtra(EXTRA_THREAD_ID, threadId != null ? threadId : -1);
-    getIntent().putExtra(EXTRA_DISTRIBUTION_TYPE, ThreadDatabase.DistributionTypes.DEFAULT);
+    if (recipientId != null) {
+      threadId = DatabaseFactory.getThreadDatabase(this).getThreadIdFor(recipientId);
+      getIntent().putExtra(EXTRA_RECIPIENT_ID, recipientId.serialize());
+      getIntent().putExtra(EXTRA_THREAD_ID, threadId != null ? threadId : -1);
+      getIntent().putExtra(EXTRA_DISTRIBUTION_TYPE, ThreadDatabase.DistributionTypes.DEFAULT);
+    }
   }
 
   private void initializeToolbar() {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
```diff
@@ -249,6 +252,21 @@ private void initializeIntent() {
       getIntent().putExtra(ContactSelectionListFragment.DISPLAY_MODE, mode);
     }
 
+    boolean isDirectShare = getIntent().hasExtra(ShortcutManagerCompat.EXTRA_SHORTCUT_ID);
+    boolean intentHasRecipient = getIntent().hasExtra(EXTRA_RECIPIENT_ID);
```
nielthiart, line 256: This check for `getIntent().hasExtra(EXTRA_RECIPIENT_ID)` might not be necessary, I couldn't find a valid path to get to this activity with this extra field set.

Android `ACTION_SEND` and `ACTION_SEND_MULTIPLE` only include on extra field we can use: `ShortcutManagerCompat.EXTRA_SHORTCUT_ID`

nielthiart:
I refactored my first change to app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java to rather update only the `initializeIntent` method without changing any other existing methods.

Per Android docs [Receiving simple data from other apps](https://developer.android.com/training/sharing/receive#declare-share-target), Direct Share intent does not include the intent and extras we declared when adding dynamic shortcuts, we only get `EXTRA_SHORTCUT_ID`, which in Signal should be a 'recipientId'.

My last commit includes a fallback to create intent incase the shortcuts have changed before we handle direct share.

Fixes #11537

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
alex-signal, line 256: Align `=`

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
```diff
@@ -249,6 +252,21 @@ private void initializeIntent() {
...
+
+    if (isDirectShare && !intentHasRecipient) {
+      String extraShortcutId = getIntent().getStringExtra(ShortcutManagerCompat.EXTRA_SHORTCUT_ID);
+      Bundle shortcutExtras = getShortcutExtrasFor(extraShortcutId);
```
alex-signal, line 260: align `=`

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
alex-signal: There might be a path via the media preview activity's forward action. We are moving away from using ShareActivity for internal message forwarding from ConversationActivity.

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
```diff
@@ -249,6 +252,21 @@ private void initializeIntent() {
...
+
+      if (shortcutExtras != null) {
+        moveShortcutExtrasToIntent(shortcutExtras);
+      } else {
+        // In case the shortcut no longer exists, fallback to creating intent
```
alex-signal, line 265: No need to leave comments around like this :)

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
```diff
@@ -259,6 +277,53 @@ private void initializeIntent() {
...
+   * Search for dynamic shortcut originally declared in
+   * {@link org.thoughtcrime.securesms.util.ConversationUtil} and return extras if the shortcut
+   * is found, else return null.
+   *
+   * @param extraShortcutId EXTRA_SHORTCUT_ID string as included in direct share intent
+   * @return shortcut extras or null
+   */
+  @Nullable private Bundle getShortcutExtrasFor(@NonNull String extraShortcutId) {
+    List<ShortcutInfoCompat> shortcuts = ShortcutManagerCompat.getDynamicShortcuts(this);
+    for (int a = 0, N = shortcuts.size(); a < N; a++) {
```
alex-signal, line 290: This pattern isn't really used elsewhere in the app. Also, I believe you could just use a for-each loop here:

```
for (ShortcutInfoCompat shortcut : shortcuts) { ... }
```

alex-signal (changes requested):
Thanks for fixing this! A few review comments but looks pretty good otherwise.

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
nielthiart: Thanks, updated in 559ef6a997ee73888812d36f7d7f145fb7597874

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
nielthiart: Thanks, updated in 559ef6a997ee73888812d36f7d7f145fb7597874

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
nielthiart: This is so much simpler, thanks! Fixed in 39d5db1a9e6dd6ab0549b2cd6a2b144ee06c9836

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
nielthiart: Aligned in 559ef6a997ee73888812d36f7d7f145fb7597874

nielthiart:
Alex, thanks so much for the review. I don't write Java regularly, so if you have any other concerns about this pull request, please don't feel bad to point them out or change it completely - however trivial or big the feedback/change.

I guess there is some opportunity to encapsulate the direct share logic in its own class at some point, but I tried to keep this fix as small and simple as possible.

I've been using this branch for a few days on Android 10, and sharing screenshots/images with regular contacts really is much easier now.

Assumptions I made when responding to your previous requests:

- We don't need comments that only repeat what the method name implies.
- Assignments in different but adjacent code blocks have their own alignment, for example the variable names and `=` in these lines: https://github.com/signalapp/Signal-Android/blob/559ef6a997ee73888812d36f7d7f145fb7597874/app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java#L255-L260
- Since these are small requests, I marked them as resolved - not sure about the preferred workflow there.

Thanks again!

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
```diff
@@ -259,6 +277,47 @@ private void initializeIntent() {
     getIntent().putExtra(ContactSelectionListFragment.RV_PADDING_BOTTOM, ViewUtil.dpToPx(48));
   }
 
+  /**
+   * Search for dynamic shortcut originally declared in {@link ConversationUtil} and return extras
+   *
+   * @param extraShortcutId EXTRA_SHORTCUT_ID string as included in direct share intent
+   * @return shortcut extras or null
+   */
+  @Nullable private Bundle getShortcutExtrasFor(@NonNull String extraShortcutId) {
```
alex-signal, line 286: Signal style is:

```
<override annotation>
<access modifier> <nullability annotation> <return type> <method name>(<params>) { ...
```

So this should be adjusted to `private @Nullable Bundle`

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
```diff
@@ -259,6 +277,47 @@ private void initializeIntent() {
...
+
+  /**
+   * @param extraShortcutId EXTRA_SHORTCUT_ID string as included in direct share intent
+   */
+  private void createShortcutIntentFromExtraShortcutId(@NonNull String extraShortcutId) {
+    RecipientId recipientId = ConversationUtil.getRecipientId(extraShortcutId);
+    Long        threadId    = null;
+
+    if (recipientId != null) {
+      threadId = DatabaseFactory.getThreadDatabase(this).getThreadIdFor(recipientId);
```
alex-signal, line 314: This is an issue. We shouldn't ever access the database (or do any other sort of IO) on the main thread.

Proposed solution here is to use `SimpleTask`:

```
SimpleTask.run(
    getLifecycle(), 
    () -> DatabaseFactory.getThreadDatabase(this).getThreadIdFor(recipientId), 
    threadId -> { ... }
)    
```

This may have other implications as well. I believe it means that `handleDestination` will need to be invoked in the 3rd argument (foreground task) of the `SimpleTask`, instead of in `onCreate` (in the case of this method being invoked).

There's then an argument for UI treatment since we're offloading work, but it should return fast enough that it shouldn't make a difference. We're just very antsy about anything that could cause an ANR (like accessing a database from the main thread)

alex-signal (changes requested):
No problem :) I realize that Signal has some oddities when it comes to how we format our code. There's just one issue now, since it appears the code changes will access the database on the main thread, which we cannot do (since it can slow down the user experience, cause ANRs, and is generally just not a good practice).

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
nielthiart: Thanks, moving IO off the main thread makes complete sense. I'd like to fix this.

1. Should I move IO off the main thread first and then treat fixing the UI as a completely separate issue (outside this pull request)?
2. Does the call to Android's `ShortcutManagerCompat.getDynamicShortcuts()` in `getShortcutExtrasFor` not also potentially block on the main thread?

Regarding UI - at the moment we initialize search before `handleDestination`, so there is already a flash of the search UI before we call `ConversationActivity`, which flashes the conversation view, before we call `MediaSendActivity`. I initially checked for direct share first, before falling back to search if necessary, but in the end I decided to keep my update small because I'm not that familiar with the codebase.

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
nielthiart: Thanks, fixed in https://github.com/signalapp/Signal-Android/pull/11539/commits/2d0ac4f4ddfd1fd2610cc5be80c98f7c88996f80

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
alex-signal: > Does the call to Android's ShortcutManagerCompat.getDynamicShortcuts() in getShortcutExtrasFor not also potentially block on the main thread?

It should be done off-thread as well, yeah. Good catch!

> Should I move IO off the main thread first and then treat fixing the UI as a completely separate issue (outside this pull request)?

Let's not worry about UI, and focus on getting these calls off the main thread.

nielthiart: Blocking call handling updated in https://github.com/signalapp/Signal-Android/pull/11539/commits/a6a57e0de87785165e5bf5b9dc86c5c556bcef43

I moved all Direct Share handling to a dedicated method, and call this in onResume so that we have lifecycle for SimpleTask.

handleDestination is only called when the necessary intent is present, so has been simplified.

I tested these changes on Android 10 only, but I haven't included any new external calls.

I'm not sure I'm going in the right direction with this, feedback welcomed. Thanks for the patience and hints here.

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
```diff
@@ -259,6 +263,83 @@ private void initializeIntent() {
...
+        return shortcutInfo.getIntent().getExtras();
+      }
+    }
+    return null;
+  }
+
+  /**
+   * @param extraShortcutId EXTRA_SHORTCUT_ID string as included in direct share intent
+   */
+  private @Nullable Bundle createExtrasFromExtraShortcutId(@NonNull String extraShortcutId) {
```
alex-signal, line 317:
```suggestion
  @WorkerThread
  private @Nullable Bundle createExtrasFromExtraShortcutId(@NonNull String extraShortcutId) {
```

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
```diff
@@ -259,6 +263,83 @@ private void initializeIntent() {
...
+          }
+      );
+    }
+  }
+
+  /**
+   * @param extraShortcutId EXTRA_SHORTCUT_ID String as included in direct share intent
+   * @return shortcutExtras or null
+   */
+  private @Nullable Bundle getDirectShareExtras(@NonNull String extraShortcutId) {
```
alex-signal, line 290:
```suggestion
  @WorkerThread
  private @Nullable Bundle getDirectShareExtras(@NonNull String extraShortcutId) {
```

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
```diff
@@ -259,6 +263,83 @@ private void initializeIntent() {
...
+    return shortcutExtras;
+  }
+
+  /**
+   * Search for dynamic shortcut originally declared in {@link ConversationUtil} and return extras
+   *
+   * @param extraShortcutId EXTRA_SHORTCUT_ID String as included in direct share intent
+   * @return shortcutExtras or null
+   */
+  private @Nullable Bundle getShortcutExtrasFor(@NonNull String extraShortcutId) {
```
alex-signal, line 304:
```suggestion
  @WorkerThread
  private @Nullable Bundle getShortcutExtrasFor(@NonNull String extraShortcutId) {
```

alex-signal (changes requested):
A few annotations I'd like added just to call these methods out, then it should be all good.

app/src/main/java/org/thoughtcrime/securesms/sharing/ShareActivity.java
alex-signal: > I'm not sure I'm going in the right direction with this, feedback welcomed. Thanks for the patience and hints here.

If we want to get technical and proper, this call would get routed through the view model to a repository. However, this particular activity already has calls elsewhere that do offthread work from within this class, and I'd really rather get it in and into the hands of users (because it's an important change) rather than bikeshed :)

nielthiart:
Thanks for the reviews and your help in getting this done.

alex-signal:
Thank you!

Thanks for sticking through it! This should go out in a build soon.

alex-signal:
This will go out in 5.21.2

Issue #11573, opened 2021-08-28
Sending a text file changes its content


mgerlach-phys:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
When I send a `UTF-8` encoded text file with `JSON` conent the receiver gets a slightly changed one.
Test issue was with the following file: [send.txt](https://github.com/signalapp/Signal-Android/files/7071142/send.txt)

The send file has 531B, the received one has 541B and looks like this: [received.txt](https://github.com/signalapp/Signal-Android/files/7071141/received.txt)

The content is equal except for a second line with some chunk from the end of the original file and no newline at the end.

It also occurred with sender/receiver switched.

### Steps to reproduce
- Send file `send.txt` to another user and noting its size
- Check file on receiving device and see that file size and content differ

**Actual result:** The file I send is not the file others receive.
**Expected result:** The files should (obviously) be the same.

### Screenshots

File name was changed to upload it to github.

#### Sender:
![sender](https://user-images.githubusercontent.com/55398943/131226417-af88dcd3-2192-4b7f-841b-4e7feb328d7d.png)

#### Receiver:
![receiver](https://user-images.githubusercontent.com/55398943/131226409-525547ef-cf5d-4065-a14f-2ed5560dc3f1.png)

#### Sender:
**Device:** OnePlus One
**Android version:** 11
**Signal version:** 5.21.6

#### Receiver:
**Device:** OnePlus 5
**Android version:** 10
**Signal version:** 5.21.6

### Link to debug log

Not Immediately after occurrence but maybe it is sufficient:
https://debuglogs.org/991ef852ac763bb50e1d87e3c82f9eea95ae75868f82a9168f601ebf07a19ea6

Please let me know if you need more details.

Thanks in advance.

Sgn-32:
"Sending" the file in "Note to Self" don't change it.
Receiving it on the device with no auto-download shows first 531B
![Screenshot_1630336654](https://user-images.githubusercontent.com/49990901/131362905-bd95dcc3-79aa-4ea9-b2c3-d241b8587b4b.png)
After downloading 541B
[...]

PR #11645, opened 2021-09-25
Remove padded bytes when downloading attachment files

fumiakiy:
### Contributor checklist

 * AVD Pixel 3a API 39
 * AVD Pixel 3a API 29
 * AVD Pixel API 23

----------

### Description

Some bytes are added when [Signal sends a file attachment in order to obfuscate the length of the file](https://github.com/signalapp/SignalCoreKit/blob/bacb5ab174e8638458ab2032a7a472422efdf165/SignalCoreKit/src/Cryptography.swift#L204-L207) according to a comment in Signal-iOS source code. [Signal iOS removes the padded bytes here](https://github.com/signalapp/SignalCoreKit/blob/bacb5ab174e8638458ab2032a7a472422efdf165/SignalCoreKit/src/Cryptography.swift#L479-L486) and [Signal Desktop does it here](https://github.com/signalapp/Signal-Desktop/blob/1913752fa0a98f4f57d60af9850be0d01dc46044/ts/textsecure/MessageReceiver.ts#L1190-L1201).

Signal-Andorid doesn't do the similar processing, resulting some files of certain length would add some padded trailing bytes when the app downloads the file, as was reported in #11573 . The added test found that it failed at least for files whose size was between 520 and 540. See the screen recordings for how it happens in the app too, as was reported in the aforementioned GH issue.

This PR is to add the similar processing to what iOS and Desktop does to Android, so it ignores the padded bytes in the file if there would be any.

#### Screen Recordings

**Before the fix at [556ca5a5730e57046419fe8805249204c5be237f]**

Note that the size of the file was initially reported as 531B but it becomes 541B after downloading it.

https://user-images.githubusercontent.com/28482/134777454-09a704bb-799e-44ec-abe3-a15a1f79ca26.mp4

**After the fix by the PR**

https://user-images.githubusercontent.com/28482/134777475-c1d1619e-8358-44f9-a71e-e8bc72c257b9.mp4

### DIFF
--- a/libsignal/service/src/main/java/org/whispersystems/signalservice/internal/util/ContentLengthInputStream.java
+++ b/libsignal/service/src/main/java/org/whispersystems/signalservice/internal/util/ContentLengthInputStream.java
@@ -32,10 +32,15 @@ public class ContentLengthInputStream extends FilterInputStream {
   public int read(byte[] buffer, int offset, int length) throws IOException {
     if (bytesRemaining == 0) return -1;
 
-    int result = super.read(buffer, offset, Math.min(length, Util.toIntExact(bytesRemaining)));
+    int lengthToRead = Math.min(length, Util.toIntExact(bytesRemaining));
+    int result       = super.read(buffer, offset, lengthToRead);
 
-    bytesRemaining -= result;
-    return result;
+    if (result <= lengthToRead) {
+      bytesRemaining -= result;
+      return result;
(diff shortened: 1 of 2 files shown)


cody-signal:
Thanks for this, I pulled this and will be part of 5.25. I made a small tweak to the fix, but it's in the same spirit of fixing the length on the final read.

PR #11721, opened 2021-10-31
Use Captcha-URL from BuildConfig

F2theK:
Actually the used URL is read from RegistrationConstants. But in the app's build.gradle we can define the URL for the recaptcha which is than generated in the builds config, but unused here. So it will be nice to really use the defined parameter from that config. Otherwise the definition in build.gradle is useless and can be removed there. I think using the BuildConfig would be the better solution.

- [x] I have signed the [Contributor License Agreement]()

### Contributor checklist

 * Samsung Galaxy S9, Android 10
 * Samsung Galaxy S6 Edge, Android 7.0

----------


F2theK:
I don't know, what is blocking it and what is needed to move it forward.
For me there is no reason to not merging.

cody-signal:
The two urls are not the same, and are used for separate purposes.

PR #11745, opened 2021-11-08
Close okhttp response body for all storage requests and for unsuccessful requests

AsamK:
### Contributor checklist

 * OnePlus 8T, Android 11

----------

### Description
OkHttp response bodies need to be closed to not leak resources.
In PushServiceSocket.java some code paths didn't close the body:
- some storage requests that ignored the response body also didn't close it
- In `makeServiceRequest` and `makeStorageRequestResponse` the body wasn't closed for unsuccessful requests

I fixed the unsuccessful requests by adding a catch and rethrow the exception after closing the body. Maybe there's a nicer solution for that ...

### DIFF
--- a/libsignal/service/src/main/java/org/whispersystems/signalservice/internal/push/PushServiceSocket.java
+++ b/libsignal/service/src/main/java/org/whispersystems/signalservice/internal/push/PushServiceSocket.java
@@ -1087,7 +1087,7 @@ public class PushServiceSocket {
 
   public Optional<StorageManifest> writeStorageContacts(String authToken, WriteOperation writeOperation) throws IOException {
     try {
-      makeStorageRequest(authToken, "/v1/storage", "PUT", protobufRequestBody(writeOperation));
+      makeAndCloseStorageRequest(authToken, "/v1/storage", "PUT", protobufRequestBody(writeOperation));
       return Optional.absent();
     } catch (ContactManifestMismatchException e) {
       return Optional.of(StorageManifest.parseFrom(e.getResponseBody()));
@@ -1095,7 +1095,7 @@ public class PushServiceSocket {
   }
 
   public void pingStorageService() throws IOException {
-    makeStorageRequest(null, "/ping", "GET", null);
+    makeAndCloseStorageRequest(null, "/ping", "GET", null);
   }
 
   public RemoteConfigResponse getRemoteConfig() throws IOException {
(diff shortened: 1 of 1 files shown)


libsignal/service/src/main/java/org/whispersystems/signalservice/internal/push/PushServiceSocket.java
```diff
@@ -1087,15 +1087,21 @@ public StorageItems readStorageItems(String authToken, ReadOperation operation)
 
   public Optional<StorageManifest> writeStorageContacts(String authToken, WriteOperation writeOperation) throws IOException {
     try {
-      makeStorageRequest(authToken, "/v1/storage", "PUT", protobufRequestBody(writeOperation));
+      ResponseBody response = makeStorageRequest(authToken, "/v1/storage", "PUT", protobufRequestBody(writeOperation));
```
alex-signal, line 1090: Alternatively, could we use try-with-resources here?

```
try (ResponseBody response = makeStorageRequest(...)) {
}
```

That would ensure that the body is closed since it implements closeable.

libsignal/service/src/main/java/org/whispersystems/signalservice/internal/push/PushServiceSocket.java
```diff
@@ -1665,9 +1671,17 @@ private Response makeServiceRequest(String urlFragment,
   {
     Response response = getServiceConnection(urlFragment, method, body, headers, unidentifiedAccessKey, doNotAddAuthenticationOrUnidentifiedAccessKey);
 
-    responseCodeHandler.handle(response.code(), response.body());
+    ResponseBody responseBody = response.body();
+    try {
+      responseCodeHandler.handle(response.code(), responseBody);
```
alex-signal, line 1676: Same comment, let's prefer try-with-resources wherever we can.

alex-signal (changes requested):
Thanks for your PR! I think we're better off using try-with-resources if we are able to.

AsamK:
Thanks for reviewing @alex-signal . I've changed the code as requested.
I didn't use try with resource at first because of the empty try block ... but it's probably better this way :)

libsignal/service/src/main/java/org/whispersystems/signalservice/internal/push/PushServiceSocket.java
AsamK: ah, actually here we can't use try-with-resources, as the body should only be closed if an exception is thrown, and not in the success case.

alex-signal: Got it, ok. Let's then just make sure it's closed where it makes sense to be in these cases, and verify calling code (whomever is being passed the resource) also closes things (and ignore me if you already did)

alex-signal:
@AsamK Sorry to be a pain, we discussed it and would prefer instead of an empty response, to implement a method specifically for ignored responses:

`makeAndCloseStorageRequest`

Which takes the same arguments that `makeStorageRequest` takes but also immediately cleans up the response object if it's present.

Furthermore, it was decided that empty try blocks don't read as nice as a simple if != null check, so let's do that instead :)

AsamK:
Alright I've pushed the changes @alex-signal . With a separate `makeAndCloseStorageRequest` it's indeed nicer.
I checked the methods using the response body in the success case and found just one method that didn't close the body: `getGroupsV2GroupHistory`
The other places all call either `.string()` or `.bytes()` which implicitly close the body.

AsamK:
I found two more places, one in `getMessages` and another in the `submitServiceRequest` method that returns a `ListenableFuture`. I've pushed an additional commit.

alex-signal:
This is going out in `5.27` (baring any glaring issues that pop up)

Thanks very much for your PR!

PR #11786, opened 2021-11-24
Fix #11784 Enable Hyphenation on call screen speaker toggle label

### LINKED ISSUE #11784: much too tight character limits have been imposed on a few strings on transifex

### Bug description
The Dutch translation for "speaker" is "luidspreker". This is the only proper translation, used everywhere where speaker is translated. I now see that Signal imposed a limit of 8 characters for the translation of speaker. That's 1 character more than the English word. That's unreasonable. In this case it would make sense to update the UI so it can take a few more characters, rather than imposing a limit which only hurts users in other language regions and calling the matter resolved.

Furthermore we see that we already translated these before the character limit was placed, and thus we can see that longer strings actually do fit quite well, with the exeption of the word "luidspreker" for which the last letter jumps to the next line. Ideally the word would be translated as "`Luid&shy;spreker`" allowing it to word wrap properly, however:
1. Signal does not support rendering &shy; on this string
2. The character limit prevents us from editing this text string to anything longer than 8 char.

Proposed solution:
Leave this to translators discretion. Enable rendering of `&shy;` only on call button labels (performance impact, but only at a critical place in the app) and stop imposing a character limit.
This is less ideal performance wise, but allows for proper translations that make sense to users in other languages. Forcing 8 char simply doesn't cut it here, the label might as well be removed then as only the icon would still make sense.

### Screenshots

Screenshot_20211122-212237

### Device info

**Device:** Motorola Moto G7 Plus
**Android version:** 10.0.0
**Signal version:** 5.27.2

Sgn-32:
### Contributor checklist

 * Moto X Play, Android 7.1.1
 * Virtual Pixel 2, Android 11

----------

### Description
Fix #11784
Enabled the hyphenation for the call screen speaker button label.
The special character "Soft Hyphen" SHY can be used by translators on transifex.

Without hyphenation:

Screenshot_1637753038

With hyphenation:

Screenshot_1637753432

### DIFF
--- a/app/src/main/res/layout/webrtc_call_view.xml
+++ b/app/src/main/res/layout/webrtc_call_view.xml
@@ -238,6 +238,7 @@
             android:textAppearance="@style/TextAppearance.Signal.Subtitle"
             android:textColor="@color/core_white"
             android:visibility="gone"
+            android:hyphenationFrequency="normal"
             app:layout_constrainedHeight="true"
             app:layout_constraintBottom_toTopOf="@id/call_screen_start_call_controls"
             app:layout_constraintEnd_toEndOf="@id/call_screen_speaker_toggle"
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1497,6 +1497,7 @@
     </plurals>
 
     <string name="WebRtcCallView__flip">Flip</string>
+    <!-- Hyphenation enabled for this string. You can use the special character "Soft Hyphen" (SHY) -->
     <string name="WebRtcCallView__speaker">Speaker</string>
     <string name="WebRtcCallView__camera">Camera</string>
     <string name="WebRtcCallView__mute">Mute</string>


major-mayer:
I really like this PR :+1: 
In german the speaker label is abbreviated with "Ltspr." which looks very weird in my opinion.
I would like the version with a hyphen much more.

Screenshot_20211124-214744

major-mayer:
Yes it is still relevant and unfortunately still not merged yet.

Sgn-32:
@major-mayer
With commit https://github.com/signalapp/Signal-Android/commit/210bb23aa4c0bbdedb05725e65c671785b0505fe the buttons have content description which is used on text-to-speech. The same label string is used. Now speaking "Ltspr."
The german translation should be in full length "Lautsprecher", to be fine on text-to-speech.

greyson-signal:
Hi there, it looks like this change has been made already. Thank you!

Issue #11449, opened 2021-07-07
Giphy search exits completely after backing out of fullscreen view


theuser18:
Title: Giphy search exits after backing out of fulscreen view

- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

**Bug Description**
Upon opening a Giphy in the full screen view before sending, pressing back on Android quits the Giphy search query. 

**Steps to reproduce**

1. Open Signal
2. Enter a conversation thread
3. Press the gif button
4. Tap "Search GIPHY"
5. Select a GIF (or enter a search term prior to selection)
6. Press back on Android

**Actual result:** The user gets booted out to the page that occurs after step 3 above. The original one quarter size page.
**Expected result:** After pressing back, the user should return to where they were in the search.

**Device info**

Device: Google Pixel 3a

Android version: Android 11 (CalyxOS)

Signal version: 5.15.6 (auto updating apk)

**App state**
Signal messages and calls registered, SMS enabled, passphrase enabled

**Link to debug log:**
https://debuglogs.org/8de3a952464180367ed3fc4c48a1e04b9d54cbb8fc6939eff4f1fc7ddcd73543

phloose:
At first, thanks for the great Signal app! Much appreciated! 

I can confirm this behaviour and would love to see this optimized, because from a usability perspective the current behaviour is not very optimal. 

Consider the following: 
You search for a gif, want to look at it in fullscreen, decide it doesn't fit and want to search for another from the previous search page. You would **not** expect to go completely back to the emoji section were you would again need to tap that gif icon, type again your search word, wait for the results to pop up and start the next selection round. 

This path takes a long time and does not feel very intuitive and user friendly. Any chance this can be fixed?

Biepa:
The issue still persists in the current version.

phloose:
Yes this is still an issue. I already opened a PR to address this issue (#11851) but this didnt get any attention yet. The problem in the current implementation is that the search view must exit to pass the selection to the message creation window which is suboptimal.

phloose:
This is still relevant and annoyingAm 27.03.2022 12:39 schrieb "stale[bot]" ***@***.***>:
Is this still relevant? If so, what is blocking it? Is there anything you can do to help move it forward?
[...]

PR #11851, opened 2021-12-19
Proposal: Move giphy search into presets bar

phloose:

- [ ] I have signed the [Contributor License Agreement](https://whispersystems.org/cla/)

### Contributor checklist

 * AVD Pixel 4a API 30, Android 11.0

----------

### Description

With regard to #11449 where another user reported that he had issues with the gif selection view that exits completly after looking at a selection i thought it would be better to make the gif selection more intuitive and with less "clicks" for the user. 

Here i propose to use a toggleable search bar where the gif presets are shown to facilitate the selection of a gif as you need much less clicks and are also able to keep the recent search after looking at a selection (in contrast to the GiphyActivity which exited completly after a gif selection making it very user unfriendly). 

After clicking on the magnifying glass icon to the left in what i call the presets bar

presets_bar

the search bar appears in the same spot. The search results appear where the preset gifs would appear. By hitting the back arrow one switches back again to the presets bar.

gif_search

In my eyes the GiphyActivity can be removed completly as with this change it is obsolete. If you can guide me a little how to safely remove it from the code base i can do this.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/keyboard/gif/GifKeyboardPageFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/keyboard/gif/GifKeyboardPageFragment.kt
@@ -3,19 +3,19 @@ package org.thoughtcrime.securesms.keyboard.gif
 import android.net.Uri
 import android.os.Bundle
 import android.view.View
+import android.widget.LinearLayout
 import android.widget.Toast
 import androidx.appcompat.app.AlertDialog
 import androidx.lifecycle.ViewModelProvider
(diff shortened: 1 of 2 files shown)


phloose:
This is still an issue

greyson-signal:
Hi there, thank you for the suggestion and for putting in the time, but this isn't a feature we plan on implementing. We believe having a full-screen search  experience is important. Thank you!

phloose:
Thank you for the response. Then please change the implementation to not close the fullscreen selection after looking at a gif in detail. The usability is really bad because sometimes you don't like the first selection and would have to go for a long process for choosing another possible candidate...

PR #11891, opened 2022-01-09
Do not allow forwarding a link preview when its image is not downloaded yet

fumiakiy:
### Contributor checklist

 * Samsung Galaxy S6, Android 7
 * AVD Pixel 4, API 30

----------

### Description
Link preview can have images and those images can be pending download depending on the media settings. Images and movies cannot be forwarded before they are downloaded but forwarding button is available on the toolbar if you select a link preview whose image is not downloaded yet. But forwarding ultimately fails if the image was not available yet, making the behavior looks confusing to users.

The solution implemented in this PR is to set hasPendingMedia for link previews so that actions that should not be available would not become available.

The bug was [reported in the beta forum by Blue42](https://community.signalusers.org/t/beta-feedback-for-the-upcoming-android-5-27-release/38993/202).

### Screen recordings

**bug**

https://user-images.githubusercontent.com/28482/148703023-90d7b055-ee3c-4ec8-b1cc-71fb54e1d6b5.mp4

**fixed**

https://user-images.githubusercontent.com/28482/148703045-531372d6-b2a3-4845-a400-d51c0181859d.mp4

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/mms/Slide.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/mms/Slide.java
@@ -40,10 +40,8 @@ import java.security.SecureRandom;
 public abstract class Slide {
 
   protected final Attachment attachment;
-  protected final Context    context;
 
-  public Slide(@NonNull Context context, @NonNull Attachment attachment) {
-    this.context    = context;
+  public Slide(@NonNull Attachment attachment) {
     this.attachment = attachment;
   }
 
@@ -122,7 +120,7 @@ public abstract class Slide {
     return hasVideo() && attachment.isVideoGif();
   }
 
-  public @NonNull String getContentDescription() { return ""; }
+  public @NonNull String getContentDescription(@NonNull Context context) { return ""; }
 
   public @NonNull Attachment asAttachment() {
     return attachment;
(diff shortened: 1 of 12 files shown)


teynav:
Won't it better to still allow forwarding link previews but if image is not downloaded then download it first?

fumiakiy:
> Won't it better to still allow forwarding link previews but if image is not downloaded then download it first?

Could be. But it would become a new feature because Signal doesn't behave that way when forwarding other media.

fumiakiy:
Yes it's still relevant.

app/src/main/java/org/thoughtcrime/securesms/mms/Slide.java
```diff
@@ -122,7 +120,7 @@ public boolean isVideoGif() {
     return hasVideo() && attachment.isVideoGif();
   }
 
-  public @NonNull String getContentDescription() { return ""; }
+  public @NonNull String getContentDescription(@NonNull Context context) { return ""; }
```
greyson-signal, line 123: Seems like this (and resulting changes to subclass) is a separate thing that doesn't need to be in this PR.

greyson-signal (changes requested):

app/src/main/java/org/thoughtcrime/securesms/mms/Slide.java
fumiakiy: It's necessary to not require `Context` in Slide constructor to make [this line](https://github.com/signalapp/Signal-Android/pull/11891/files#diff-3da0cade0ad78366f5959ea36324f43bca9614878e47ff5a932fa13cd068d995R78) possible there.

I could remove the `@Nullable Context context` parameter in the constructor altogether but I didn"t because I thought it would make this PR look doing something that is not absolutely necessary for the PR to be self contained.

Issue #11509, opened 2021-08-01
Chat Backups are inflated in size


SebiderSushi:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
When creating backups using the "Chat backups" feature the resulting backup files don't appear to be deduplicated. They can be vastly bigger than the actual data used by the Signal app. Forwarding messages inflates backup size while Signal App Data Size does not change.

### Steps to reproduce
- Send a large-ish file or media (a couple megabytes)
- Create a backup and note down: Backup size & Signal app data usage
- Forward the file one or more times
- Create another backup, then compare backup sizes & Signap app data usage

**Actual result:** The Backup file size increases by the size of the forwarded file while Signal App usage does not. For example, in the case of another persons Signal i observed 400MB of app data usage while the backup file size was about 1GB.
**Expected result:** If Signal deduplicates forwarded messages internally they should also be deduplicated in backup files. The backup file should not be considerably larger than Signals app data.

### Device info
**Device:** Xiaomi Redmi 4X
**Android version:** LineageOS 15.1 based on AOSP 8.1.0
**Signal version:** 5.18.5

### Link to debug log
https://debuglogs.org/bb268c5fb8102c6c3430235813b64721c100dec8c3ac8dfe71885073a2133fc1

SebiderSushi:
Replying to a message of a large file/media has the same effect.

SebiderSushi:
This is reproducable with the current Signal version `5.29.7`
This issue is being blocked by the fact that no one considered this issue when implementing backups and even now there is no one available to review it.
There is nothing any user can do to help it move forward.

Thanks for trying though, stale bot. /s

SebiderSushi:
This is reproducable with the current Signal version `5.34.5`

SebiderSushi:
i guess it could be argued that this should not be discussed as a github issue as per this rule from #7598
> If you disagree with the way that a feature is working (but it's working!) that's not a bug.

[...]

PR #11900, opened 2022-01-15
introduce chunked backups

Roghetti:

### Contributor checklist

 * Oneplus 3, Android 9
 * Virtual device, Android 9
 * Virtual device, Android 11

----------

### Description

I work for bevuta IT GmbH, a company based in Germany.  Our boss is a Signal user on Android. He has the problem that he can't backup his Signal app data to the SD card and despite 128 GB of internal memory,  he no longer has free internal storage, mainly because of the signal backups.  The most reasonable explanation was that his Android version doesn't support file systems other than FAT32 on SD cards.  FAT32 has some limitations, such as a maximum file size of ~4 GB.  His Signal installation uses more than 4 GB.  To us, the use of a sd card for backups also seems advantageous, as it is easier to get at the data if the device breaks down.

We discussed several solutions.  The best solution we came up with, is to just chunk the Signal backup into parts smaller than 4 GB, if desired.  I was tasked to implement it.  To use this code in a real Signal installation, it has to be in the official release.  Now, here is my MR! :-)

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/backup/FullBackupBase.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/backup/FullBackupBase.java
@@ -15,9 +15,9 @@ public abstract class FullBackupBase {
   private static final int DIGEST_ROUNDS = 250_000;
 
   static class BackupStream {
-    static @NonNull byte[] getBackupKey(@NonNull String passphrase, @Nullable byte[] salt) {
+    static @NonNull byte[] getBackupKey(@NonNull String passphrase, @Nullable byte[] salt, boolean isChunked) {
       try {
-        EventBus.getDefault().post(new BackupEvent(BackupEvent.Type.PROGRESS, 0, 0));
+        EventBus.getDefault().post(new BackupEvent(BackupEvent.Type.PROGRESS, isChunked, 0, 0));
 
         MessageDigest digest = MessageDigest.getInstance("SHA-512");
         byte[]        input  = passphrase.replace(" ", "").getBytes();
@@ -26,7 +26,7 @@ public abstract class FullBackupBase {
         if (salt != null) digest.update(salt);
 
         for (int i = 0; i < DIGEST_ROUNDS; i++) {
-          if (i % 1000 == 0) EventBus.getDefault().post(new BackupEvent(BackupEvent.Type.PROGRESS, 0, 0));
+          if (i % 1000 == 0) EventBus.getDefault().post(new BackupEvent(BackupEvent.Type.PROGRESS, isChunked, 0, 0));
           digest.update(hash);
           hash = digest.digest(input);
         }
@@ -45,17 +45,23 @@ public abstract class FullBackupBase {
     }
 
     private final Type type;
+    private final boolean isChunked;
     private final long count;
     private final long estimatedTotalCount;
 
-    BackupEvent(Type type, long count, long estimatedTotalCount) {
+    BackupEvent(Type type, boolean isChunked, long count, long estimatedTotalCount) {
       this.type                = type;
+      this.isChunked           = isChunked;
       this.count               = count;
       this.estimatedTotalCount = estimatedTotalCount;
     }
 
     public Type getType() {
       return type;
+    }
+
+    public boolean isChunked() {
+      return isChunked;
     }
 
     public long getCount() {
--- a/app/src/main/java/org/thoughtcrime/securesms/backup/FullBackupExporter.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/backup/FullBackupExporter.java
@@ -106,7 +106,23 @@ public class FullBackupExporter extends FullBackupBase {
       throws IOException
   {
     try (OutputStream outputStream = new FileOutputStream(output)) {
-      internalExport(context, attachmentSecret, input, outputStream, passphrase, true, cancellationSignal);
+      internalExport(context, attachmentSecret, input, outputStream, passphrase, true, false, cancellationSignal);
+    }
+  }
+
+  public static List<File> exportChunked(@NonNull Context context,
+                                         @NonNull AttachmentSecret attachmentSecret,
+                                         @NonNull SQLiteDatabase input,
+                                         @NonNull File backupDir,
+                                         @NonNull String backupPrefix,
+                                         @NonNull String backupSuffix,
+                                         @NonNull String passphrase,
+                                         @NonNull BackupCancellationSignal cancellationSignal)
+      throws IOException
+  {
+    try (MultiFileOutputStream outputStream = new MultiFileOutputStream(backupDir, backupPrefix, backupSuffix)) {
+      internalExport(context, attachmentSecret, input, outputStream, passphrase, true, true, cancellationSignal);
+      return outputStream.getFiles();
     }
   }
 
@@ -120,7 +136,24 @@ public class FullBackupExporter extends FullBackupBase {
       throws IOException
   {
     try (OutputStream outputStream = Objects.requireNonNull(context.getContentResolver().openOutputStream(output.getUri()))) {
-      internalExport(context, attachmentSecret, input, outputStream, passphrase, true, cancellationSignal);
+      internalExport(context, attachmentSecret, input, outputStream, passphrase, true, false, cancellationSignal);
+    }
+  }
+
+  @RequiresApi(29)
+  public static List<DocumentFile> exportChunked(@NonNull Context context,
+                                                 @NonNull AttachmentSecret attachmentSecret,
+                                                 @NonNull SQLiteDatabase input,
+                                                 @NonNull DocumentFile backupDir,
+                                                 @NonNull String backupPrefix,
+                                                 @NonNull String backupSuffix,
+                                                 @NonNull String passphrase,
+                                                 @NonNull BackupCancellationSignal cancellationSignal)
+      throws IOException
+  {
+    try (MultiFileOutputStream29 outputStream = new MultiFileOutputStream29(context, backupDir, backupPrefix, backupSuffix)) {
+      internalExport(context, attachmentSecret, input, outputStream, passphrase, true, true, cancellationSignal);
+      return outputStream.getFiles();
     }
   }
 
@@ -131,7 +164,7 @@ public class FullBackupExporter extends FullBackupBase {
                               @NonNull String passphrase)
       throws IOException
   {
-    internalExport(context, attachmentSecret, input, outputStream, passphrase, false, () -> false);
+    internalExport(context, attachmentSecret, input, outputStream, passphrase, false, false, () -> false);
   }
 
   private static void internalExport(@NonNull Context context,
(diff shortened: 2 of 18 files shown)


newhinton:
@Roghetti This sounds amazing, one feature i am really looking forward to! 

I do have a question, could you explain how it is beeing chunked?
I had the idea some time ago that it would be amazing to create "yearly" backups to achieve the same. Reason for that is it would allow for a partial restore of a time range (without loosing the rest, if the key does not change)

Also do the chunks change between runs? Eg. does only the last chunk change because it is the one containing the new data?

Roghetti:
> @Roghetti This sounds amazing, one feature i am really looking forward to!
Thanks! :-)

> I do have a question, could you explain how it is beeing chunked?
It's just like a normal backup, but split into multiple files, if a single file would exceed a certain size. E.g., if your normal backup has 1.5 GB, the chunked backup would create two files, one with 1GB and one with 0.5GB as 1GB is the threshold I chose.
If you combine these files, e.g. with `cat` on the command line, you would get the same file as the 'normal' backup mechanism would give you.

> I had the idea some time ago that it would be amazing to create "yearly" backups to achieve the same. Reason for that is it would allow for a partial restore of a time range (without loosing the rest, if the key does not change)
> Also do the chunks change between runs? Eg. does only the last chunk change because it is the one containing the new data?
All chunks are affected when certain data has changed.
As this chunking is like the normal backup, but with multiple files, it depends in which order the data is exported. The current code exports each table of the sqlite-database, one by one, e.g., first MMS, then SMS, then Reactions and so on. If you get an MMS, all chunks (at least at the point with the new MMS) will change.
I like your idea of incremental backups, but it has to be addressed in a separate Pull Request and would affect the 'normal' and the chunked backup mechanism.

newhinton:
Ahhh thats a shame, i had hoped that this could reduce the data transferred each night and reduce wear on my flash-storage. Thanks anyway!

cody-signal:
Hey there, appreciate the PR. I'm not sure if we'll pull this in or not. I'll bring it up with the team.

newhinton:
@cody-signal There is also an [discussion](https://community.signalusers.org/t/some-form-of-incremental-backup/12047) in the forum regarding this. I think we need to somehow adress this, because backups can easily grow beyond 4Gb, which is problematic for different reasons mentioned in the discussion

Roghetti:
The following feature request is related: [Backup to FAT card](https://community.signalusers.org/t/backup-to-fat-card/12620)

HyperCriSiS:
Will the backups also be incremental?
The actual implementation  which forces full backup every day is really dumb :slightly_frowning_face:

newhinton:
@HyperCriSiS Sadly no, this will only 'split' the backup-file into parts. I hope that they will switch to incremental backups eventually. There is a linked discussion.

However, changing the backup-format should not be done very often, so we should do it properly. I think this pr should not be merged, since it fixes the symptom (backups beeing over 4Gb) and not the issue (missing segmentation/incremental backups) of ballooning backup-files

Smojo:
> I think this pr should not be merged, since it fixes the symptom (backups beeing over 4Gb) and not the issue (missing segmentation/incremental backups) of ballooning backup-files.

I would not agree to that. Chunking would help people with devices with small internal storage but an SD-Card slot, if the Signal container size growth over 4GB and so most probably also the backup file size. Without junking it need 3 times >4GB (2 x existing backups + temporarily other copy until the new backup finished). The backup is encrypted so no problem to put it there. That is also the reason this issue is described initially.

Looking forward that it gets merged.

newhinton:
@Smojo I dont disagree with you that the growing filesize is a problem (it is, and will get even worse with time)

However, the underlying issue is not that some filesystem has limitations that chunking needs to work around, but that signal-backup fikes grow in size indefinitively.

Chunking DOES fix that, but only the symptom, not the problem. 

I would like to see a solution where a single backup is split by year, and then only made 'on demand' (do we ALWAYS have to store chats again for 2016?)

This would solve the described issue AND the problem (as long as someone does not add more thab 4Gb per year, but i deem that unlikely)

Smojo:
Yeah, this might be something, which is needed and helps as well. 

Right (not sure if I'm wrong) the only possibility is to shorten chats by message-count (which also somehow results in a smaller container and so also smaller backup-file[s]), but nobody can really tell what this means in regards of "message retention" of a single chat. Some chats will not even go back a year or so if they are very talkative.

So I would agree if a chunking by time which results in a chunked container AND backup-files (which might also have a file-size limit of 4GB in addition to the split by time) will also solve the users problem described here. (without further thinking about symptom and problem ^^)

lieblingsnerd:
> I would like to see a solution where a single backup is split by year, and then only made 'on demand' (do we ALWAYS have to store chats again for 2016?)

I think what the problem is is in the eye of the beholder. Incremental backups have the risk that an increment will break unnoticed. If, like me, you want to keep all your messages, the solution implemented here is generally reasonable and complete in itself. If you don't want to have old messages anymore, you could delete them in the application, then they wouldn't be part of a full backup anymore. Different people, different needs. Either way, this solution should be better than none. If you implement something better that would be all the cooler of course. But no reason to wait right?

I would be very happy if this PR would be merged. Then I wouldn't have to delete backups by hand every day 😃

HyperCriSiS:
I see it not as breaking change. The current status of the backup system is really bad because it is incomplete. 
Since some days Signal even looses the permission to the folder. I have to re-enable it every few days.

newhinton:
@Smojo I am not against merging this pr. I am just advocating on doing it properly the first time, and make it work properly for everyone without stopgap measures. We should not change such important parts of an application more often than we need to.

In my opinion, just chunking only fixes a minor part of the issue, because big files (chunked or not) still cannot properly be moved to cloud-storage or remote storage in general.

In the end the dev's need to find out how they would like to adress this issue, and i can only hope that it will be a solution that also satisfies my usecase, aswell as yours. If it only helps people with a 4GB limit, i will still be happy for you ;)

newhinton:
There is also this comment:

https://github.com/signalapp/Signal-Android/issues/11509#issuecomment-1097142650

Smojo:
> big files (chunked or not) still cannot properly be moved to cloud-storage or remote storage in general.

Isn't a chunking every 1GB (@Roghetti explained that already) at least one way to deal with that? It is an imho okayish size-limit for the chunks and/or for "bigger" files on Smartpones + also to copy where-ever you want to copy. Or in other words: I would not say that the chunks are big files ;-) - the current backups are.

> what exactly is the 'root' cause? is it FAT with it's limitation, or is it the files that are to big? This is not necessarily a technical question but an user experience one, and that depends on your view

I can just repeat myself (and close with the same sentence). There is for sure other stuff to be solved, but it feels like this PR should not be merged (and/or discussed forever) just because it does not address all the problems and/or potential root cause(s) out there. FAT is the default of Android for portable storage. We don't need to discuss it's limit for files bigger than 4GB or push the problem to it.

Also I don't see that the discussion that it is bad that the signal message container is big
and/or the backup is even bigger (for whatever reason - missing dedup for media in backup like explained in the comment etc.) 
and/or why people want to keep messages 
and/or on which ways messages can be deleted
and/or how long older backups should be supported
leads to something.

Just the opinion of one user who is struggling with this. (next to @Roghetti / @HyperCriSiS / and @lieblingsnerd )

Roghetti:
To clarify a bit:
The backup-chunking is an additional backup option, introduced with this Pull Request.
So, you can make traditional single-file backups and chunked backups – as you wish.
The backup mechanism is already the same and uses the same code: I.e. it isn't much of a burden to maintain this.
Technically we use a custom Outputstream class there which does the chunking instead of the normal FileOutputStream.
This leads also to a compatibility in another way: (1) You can take a normal backup, split it up on the command line and import it in Signal (when this PR is merged). (2) You can take a chunked backup, cat it on the command line into a single file and import it on a regular Signal installation.

klorydryk:
I don't see any option for this right now. Is this coming in a futur release?

Diapolo:
That pull would make my life so much easier, as I currently can use just FAT32 for Signal backups on my external SDCard on Android + the backup exeeds a 4 GiB file size. It would be great if the devs could consider merging this one 👍 .

Thank you!

janvlug:
The is still a need for this functionality.

klorydryk:
Please someone to merge this! <3

greyson-signal:
We won't be merging this as-is, but we do have some plans for trying to do some of this work in the coming weeks. I'll leave this open until we get around to it. Thanks!

klorydryk:
Good news, @greyson-signal <3
Please keep us informed of the progress and the way you want to do it!

dev-aniketj (approved):

Grunthos:
If there is any plan in this process to allow even basic incremental backups (eg. chunking + rotating selection from tables based on time periods), then I would be happy to help on any aspect of the Android/common coding or testing side of things.

I have not looked at the data side of things, but a naive approach would seem to me to:

- find all records < (year1 = min(year))
- backup via chunking
- find all records from (year-1 to (year2=(year1-1))
- backup to new base name via chunking
- Finish the backup and start a 'new' one for the next year
- ditto all the way to yearN and year(N+1)
- subsequent backups read old chunks and write new temporary chunks, discarding the temp chunks if the old chunk matches. Discard the old chunk if a discrepency is found. 

This should result in a single current and consistent backup, and only result in 'new' files for the recent year. One could also chunk by month or some other user-defined interval.

Grunthos:
Chunked backups of some kind are an important feature.

Grunthos:
It would be great to have SOME solution along these lines!

janvlug:
Once I lost all history because backup failed of the file being too large for the file system. This is very important and highly needed functionality.

janvlug:
There is still an urgent need for this functionality.

Grunthos:
This issue is not stale! It's just receiving the attention it deserves. The problem it addresses is not going away and is getting worse. It has resulted in me largely giving up on using Signal except in rare circumstances.

Grunthos:
`wontfix` tag was added by the bot and should probably be removed since that represents policy by stagnation vs deliberative decision.

Grunthos:
This still remains a crucial feature. The lack of it means that I now use signal for far fewer people and purposes.

Xyaren:
bump

Grunthos:
And I second that bump. The current situation is untenable.

Smojo:
@greyson-signal over a year ago you commented ( https://github.com/signalapp/Signal-Android/pull/11900#issuecomment-1369013837 ): 

> We won't be merging this as-is, but we do have some plans for trying to do some of this work in the coming weeks. I'll leave this open until we get around to it. Thanks!

What happened since then?

LordBurtz:
what is the status for incremental backups?
as my backup is growing to 8gb I never want to transfer a full backup to the cloud as it takes _a lot_ of time, incremental backups would be perfect or at least the option to just backup a certain timeframe and then splice them together on restore..

gerum100:
I was tired waiting for smaller backup for years, I have implemented it as an external tool. So if anybody want to use my [Signal Backup Deduplicator](https://gitlab.com/gerum/signal-backup-deduplicator), it is available.

klorydryk:
Thanks for this! So nice, I'll test it!

Xyaren:
bump

cody-signal:
We took this as feedback as part of designing our new local backup system that will uses multiple smaller files and deduping. This will mean more files but none likely to exceed a 100mb. Take a look at our commit history if you're interested. As such, we won't be taking this work into the app but appreciate the time and effort put into it.

PR #11984, opened 2022-02-10
Show your groups in Note to Self.

Sgn-32:
### Contributor checklist

 * Device A, Android X.Y.Z
 * Moto X Play, Android 7.1.1
 * Virtual Pixel 2, Android 11

----------

### Description

Currently there is no listing of all groups a user is a member of.
With this PR the listing is shown in the conversation settings.

----------------------------
Screenshot_1644524031

-----------------------------------

Screenshot_1644524131

------------------------------------
signal-2022-02-10-212345

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -835,6 +835,11 @@
         <item quantity="other">%d groups in common</item>
     </plurals>
 
+    <plurals name="ManageRecipientActivity_in_d_groups">
+        <item quantity="one">You are in %d group</item>
+        <item quantity="other">You are in %d groups</item>
+    </plurals>
+
     <plurals name="GroupMemberList_invited">
         <item quantity="one">%1$s invited 1 person</item>
         <item quantity="other">%1$s invited %2$d people</item>
--- a/app/src/main/java/org/thoughtcrime/securesms/components/settings/conversation/ConversationSettingsFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/settings/conversation/ConversationSettingsFragment.kt
@@ -467,7 +467,7 @@ class ConversationSettingsFragment : DSLSettingsFragment(
           }
         }
 
-        if (recipientState.identityRecord != null) {
+        if (recipientState.identityRecord != null && !state.recipient.isSelf) {
           clickPref(
             title = DSLSettingsText.from(R.string.ConversationSettingsFragment__view_safety_number),
             icon = DSLSettingsIcon.from(R.drawable.ic_safety_number_24),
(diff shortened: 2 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -824,6 +824,12 @@
         <item quantity="other">%d groups in common</item>
     </plurals>
 
+    <string name="ManageRecipientActivity_in_no_group">You are in no group</string>
```
alex-signal, line 827: "You are in 0 groups" is fine, no need for this string.

Sgn-32: @alex-signal 
What about the string name?
Should I leave it as `name="ManageRecipientActivity_in_d_groups"` or change it to
`name="ConversationSettingsFragment__note_to_self_in_d_groups"` and paste it at line 3866?

Sgn-32:
The string `You are in no group` is deleted. Now using only the plural strings.

What about the string name?
Should I leave it as `name="ManageRecipientActivity_in_d_groups"` or change it to
`name="ConversationSettingsFragment__note_to_self_in_d_groups"` and paste it at line 3866?

Screenshot_1644601192

------------------------

Short Summary of PR:

Normally getGroupsInCommon is used to get all common groups with your contact.
Now using it to get all common groups with yourself.

The getGroupsInCommon in ConversationSettingsViewModel.kt isn't executed due to this if-condition:
https://github.com/signalapp/Signal-Android/blob/6f5475fc94507618f1c38226e7084365eabddf4d/app/src/main/java/org/thoughtcrime/securesms/components/settings/conversation/ConversationSettingsViewModel.kt#L158

So deleting this if-condition, gives groupsInCommon, hasGroups and identityRecord also for yourself.

In ConversationSettingsFragment.kt:

Adding `&& !state.recipient.isSelf`, because yourself has now a identityRecord. So "View safety number" would be shown in "Note to Self".
https://github.com/signalapp/Signal-Android/blob/6f5475fc94507618f1c38226e7084365eabddf4d/app/src/main/java/org/thoughtcrime/securesms/components/settings/conversation/ConversationSettingsFragment.kt#L466

Adding `|| state.recipient.isSelf`, to show "Your are in 0 groups". The original condition checks if yourself are in groups. So when you are in no group this code part wouldn't considered.

https://github.com/signalapp/Signal-Android/blob/6f5475fc94507618f1c38226e7084365eabddf4d/app/src/main/java/org/thoughtcrime/securesms/components/settings/conversation/ConversationSettingsFragment.kt#L519

clauz9:
If this gets added in, could we have the group list here as well ? 

<details>
<summary>Screenshot</summary>

image
</details>

in the menu that pops up when you click the avatar of note to self, in the conversation list. 
That menu is currently pretty useless, also, 

I think it would be useful to be able to tap on the groups avatar and get a quick menu for the most common group actions like "Open group", "Copy group Link" (if it's enabled)
 "Leave Group", "Block group" 

I know currently any tap on the group name or avatar opens the group, but I feel like having a quick menu to the most common actions is a good idea.

greyson-signal:
Hi there, thanks for taking the time to build it, but we talked with our design team and we're going to pass on this for now. Thanks!

PR #12027, opened 2022-02-28
Fix #12017 pulse to position not happening 

### LINKED ISSUE #12017: Jumping to a message outside the viewport does not highlight it immediately

### Bug description
When jumping to a message that is initially outside the viewport the message does not get highlighted ("pulse") after the jump. You can for example jump to a quoted message, a search result in a conversation etc.

The highlight functionality does work for jumps to messages that are already visible before the jump.

Apart from that, I think that the highlight should be more noticeable, e.g. make it pulse twice, extend the highlight to the whole message instead of just the border.

### Steps to reproduce
1.  find or create a message that is replying to another message further up the conversation, so that the quoted message is not visible while viewing the replying message
2. tap the quote that is embedded in the replying message to jump to the quoted message
3. after the jump, scroll a little bit up or down

**Actual result:** Highlight is executed too late, it's only shown after the manual scroll event at step 3 
**Expected result:** Highlight should be executed automatically after the jump at step 2

### Device info
**Device:** Pixel 4a
**Android version:** 12
**Signal version:** 5.32.7

clauz9:

### Contributor checklist

 * Nokia 3.1, Android 10
 * Huawei L29, Android 8

----------

### Description

This pull request Fixes #12017 the issue was happening because according to the documentation [scrollToPositionWithOffset() doesn't reflect the scroll position change until the next layout call](https://developer.android.com/reference/androidx/recyclerview/widget/LinearLayoutManager#scrollToPositionWithOffset(int,%20int)) and also, the debugger never stopped at the [break point I placed here](https://github.com/signalapp/Signal-Android/blob/1314b0499436dd721c7525dbf26c12a09e33c7ee/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java#L1382) when I clicked on a quoted message or a search result for the first time, 

which means the code was in this state, [setting the pulsePosition variable to the current position, and scrolling to position with offset,](https://github.com/signalapp/Signal-Android/blob/1314b0499436dd721c7525dbf26c12a09e33c7ee/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java#L1195-L1198) then waiting for [onScrollStateChange to run, and pulse at the newly set pulsePosition](https://github.com/signalapp/Signal-Android/blob/1314b0499436dd721c7525dbf26c12a09e33c7ee/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java#L1378-L1384)

which doesn't happen, so I changed it to make sure pulseAtPosition runs after scrollToPositionWithOffset no matter what path the code takes

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
@@ -1192,10 +1192,9 @@ public class ConversationFragment extends LoggingFragment implements Multiselect
                                  if (child != null && layoutManager.isViewPartiallyVisible(child, true, false)) {
                                    getListAdapter().pulseAtPosition(position);
                                  } else {
-                                   pulsePosition = position;
+                                   layoutManager.scrollToPositionWithOffset(p, list.getHeight() / 4);
+                                   getListAdapter().pulseAtPosition(position);
                                  }
-
-                                 layoutManager.scrollToPositionWithOffset(p, list.getHeight() / 4);
                                } else {
                                  layoutManager.scrollToPositionWithOffset(p, list.getHeight() / 4);
                                  getListAdapter().pulseAtPosition(position);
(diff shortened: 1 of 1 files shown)


clauz9:
This has been merged 👍

PR #12035, opened 2022-03-03
Support MicroG disabling FCM

Doomsdayrs:

### Contributor checklist

----------

### Description
FCM can be disabled by MicroG.

This causes IOException to be thrown by getToken().
In order to handle this, each instance calling FcmUtil.getToken is wrapped in a try catch. Upon receiving FCMDisabledException, the caller will set AccountValues.fcmEnabled to false to prevent this issue from happening again.

I have no interest in letting a "secure" messaging application on my device use google's APIs.

---

The application `Element` provides a GPS free version on F-Droid, Which Signal does not provide, nor is there any clear indication anymore about downloading Signal in its standalone form without using the google play store to install.


app/src/main/java/org/thoughtcrime/securesms/gcm/FcmUtil.java
```diff
@@ -20,16 +21,31 @@
...
   @WorkerThread
-  public static Optional<String> getToken() {
+  public static Optional<String> getToken() throws FCMDisabledException {
     String token = null;
     try {
       token = Tasks.await(FirebaseMessaging.getInstance().getToken());
     } catch (InterruptedException e) {
       Log.w(TAG, "Was interrupted while waiting for the token.");
+    } catch (IOException e) { // IGNORE THIS WARNING
+      throw new FCMDisabledException();
```
greyson-signal, line 31: These methods flat-out don't throw IOExceptions. The app wouldn't compile.

Doomsdayrs: 74caf10 solves this annoying java issue.

greyson-signal: Sorry, but while changing from java to kotlin will let you catch whatever exception you want, it doesn't change the fact that the exception isn't actually thrown.

greyson-signal:
Thank you for the time you've put in, but this code change doesn't do what you think it does, and has a lot of other potential behavior problems. If you don't want to use Google Services, you should uninstall MicroG.

Doomsdayrs:
> Thank you for the time you've put in, but this code change doesn't do what you think it does, and has a lot of other potential behavior problems. 

May you list these problems? I'm happy to learn and try and patch any possible problems.

> If you don't want to use Google Services, you should uninstall MicroG.

No one can do a full stallman, what you're doing here is essentially telling me to fuck off.

app/src/main/java/org/thoughtcrime/securesms/gcm/FcmUtil.java
Doomsdayrs: Did you not read the description?

The exception is thrown when using MicroG. 

Do I need to provide you log files to prove it?

greyson-signal: If that is the case, the problem is that MicroG is breaking the API contract of the thing they're trying to mimic. You should file an issue with them.

greyson-signal:
> No one can do a full stallman, what you're doing here is essentially telling me to fuck off.

I'm saying that Signal supports people *with* Play Services, and we support people *without* Play Services. We do not support people with *possibly altered maybe broken* Play Services. Putting people in and out of Play Services mode is very sensitive and can have very bad consequences if done incorrectly (mostly in the form of broken notifications). Supporting the multitude of Play Services replacements is a bit of a fool's errand, especially if they are doing what you claim (breaking the API contracts of the interfaces they're mocking).

PR #12085, opened 2022-03-21
Fix wrong subjectView text for conversation list item when resuming

clauz9:
### Contributor checklist

 * Samsung s5 mini, Android 5.1
 * Nokia 3.1, Android 10
 * Huawei L29, Android 8.0.1

----------

### Description

This fixes the following issue:

<details>
  <summary>Click for GIF of issue</summary>

GIF 21 03 2022 22-12-43
</details>

The onChanged() for ConversationListItem was being triggered by [activeAdaptor.notifyDataSetChanged()](https://github.com/signalapp/Signal-Android/blob/5e46e1e3d9a0fb9b228ab6f10402f5a9d7f39c1d/app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java#L354-L356) in [reverse order](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:recyclerview/recyclerview/src/main/java/androidx/recyclerview/widget/RecyclerView.java;l=12934-12942) of the conversation list items. 

which was added in [this commit to fix an issue with conversation list times](https://github.com/signalapp/Signal-Android/commit/68f718a210a78513ae18bf0368a9c16905e88c3d) 

so I had to use something that would also update conversation list times, but would not call the onChanged() method.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
@@ -352,7 +352,7 @@ public class ConversationListFragment extends MainFragment implements ActionMode
     }
 
     if (activeAdapter != null) {
-      activeAdapter.notifyDataSetChanged();
+      activeAdapter.notifyItemRangeChanged(0, activeAdapter.getItemCount());
     }
 
     SignalProxyUtil.startListeningToWebsocket();


clauz9:
I just discovered there's a ConversationUpdateTick, and I plan that use that to update the timestamp of the conversation item in the conversation list, which will allow me to remove the call to update from onResume()

clauz9:
the initial commit https://github.com/signalapp/Signal-Android/pull/12085/commits/fa545e7918b64151610da46305841e44542027fc fixes the issue from the initial PR description, 

the subsequent commits have been added because the app didn't used to update the conversation timestamps while being visible and idle, and it seems like it should. 

Feel free to cherry pick only the fix for the issue. 👍

clauz9:
I noticed the search results timestamps weren't updating, so I made the ConversationUpdateTick trigger a timestamp update for the search results if the search toolbar is visible,

I've also noticed there is some text jitter when the search query is changed. and the view is refreshed.. so I'm turning this into a draft again... 

feel free to chime in if this is of Interest 🙂

greyson-signal:
Hey! We're pulling in your first commit because we think it resolves the flicker you brought up. Thank you :) We'd probably want to go on own way on live-updating timestamps though. Hard to say until we start looking at it in more depth.

clauz9:
> yeah .. the first commit will fix the flicker on the ConversationList, but unfortunately `ConversationListSearchAdapter`, uses the same `notifyDataSetChanged()` in `UpdateResults()`, which triggers the same flicker but only when search results are visible and you tap on a result, and then return to the results list.

@greyson-signal sorry for the ping but I think the above is information you might want to know

clauz9:
> `ConversationListSearchAdapter`, uses the same `notifyDataSetChanged()` in `UpdateResults()`, which triggers the same flicker but only when search results are visible and you tap on a result, and then return to the results list.

I just realized something....
since `notifyDataSetChanged()` is called on every query change, because of `UpdateResults()` but the glitch doesn't happen when the query changes, and it does happen when `onResume()` is called from `ConversationListFragment` meaning when you return to the fragment, either from homes screen, or a search result's conversation.. 

it must have something to do with the lifecycle change coupled with the use of `notifyDataSetChanged()` 
and a closer look at the glitch on the search screen, shows that only Conversation search results suffer from the glitch, 


<details>
<summary>GIF Recording</summary>

GIF 25 03 2022 10-57-40

</details>

for Conversation Search results `bindThread()` is used to update the `ConversationListItem` and `bindThread()` is the only method that sets a ForeverObserver for displayBody... 🤔

greyson-signal:
There's multiple things going on here, and at the end of the day there was too much async stuff going on, and `notifyDataSetChanged()` is bad in general. I've updated things to be less async and I've also switched over to `notifyItemRangedChanged()` where appropriate, which is a bit smarter in it's rebind calls. I think in general we got this settled. I pulled in your first commit and gave you a much-deserved shoutout in my other change set for your help in researching this. Thanks!

PR #12127, opened 2022-04-07
Optimized PNGs

VadimMuhtarov:
Optimized PNGs via oxipng


clauz9:
This looks like a good way to introduce some arbitrary binary code into an open source app 😆 
And since these PNG's have been optimized,  I assume there is no way to validate that these are in fact the same files, is there ? 🤔 

https://github.com/signalapp/Signal-iOS/pull/5306
I see you made a similar PR for Signal-iOS 

An actual PR description would have been nice, to explain the need for this a bit 🤔

VadimMuhtarov:
I just want to make it as optimized as posible. You cat use oxipng and with "-o 6 --strip safe" parameters and compare hashes

teynav:
I am also using oxipng for my project too , I think it's nice and fast.

clauz9:
> I am also using oxipng for my project too , I think it's nice and fast.

I'm not doubting oxipng, I'm sure it's a great tool 👍 

I was just saying that the only way to be sure the files are actually the same PNG's but optimized is for the developers themselves to use oxipng to optimize them, and not pull PNGs in from someone's PR  🤔

VadimMuhtarov:
I didn't hear about any attacks using compromised png files. Yes, devs can do that themselves, but this project is open source and users can help with such tasks.

Issue #11458, opened 2021-07-11
.onion addresses are not links


jtracey:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
URLs that convey [onion addresses](https://en.wikipedia.org/wiki/.onion) are not links when viewed in Signal Android. Signal Desktop behaves correctly here. I suspect this problem was introduced in 3f7dd2118654516aea26cd87e095a4ad87d2a606, which I believe is preventing the URL from being treated as a link at all, rather than the intended effect of preventing link previews. Since onion addresses (especially now that only the v3 format is supported) are intended to be used as non-human readable identifiers, similar to e.g. IPv6 addresses, that are conveyed via links rather than typing in the domain, this winds up being a pretty annoying behavior to work around.

### Steps to reproduce
- send or receive a message with a .onion URL in it (e.g., https://dns4torpnlfs2ifuz2s2yf3fc7rdmsbhm6rw75euj35pac6ap25zgqad.onion/)
- attempt to tap on URL to open (presumably in Tor Browser, though others exist as well)

**Actual result:** URL is displayed and treated as unformatted text
**Expected result:** URL functions as a hyperlink, similar to other URLs, albeit without preview functionality

### Device info

**Device:** Samsung Galaxy A5 (2017)
**Android version:** 10 (LineageOS 17.1)
**Signal version:** 5.15.6

jtracey:
This is still an issue.

jtracey:
Is there something I need to do to keep this open?

cody-signal:
We use Android's linkify methods to turn text into links. We currently don't have any plans to write our own linkify framework, sorry.

jtracey:
You don't need to write your own linkify framework, as I said, you appear to have broken Android's built-in one, and I'm asking you to undo that. If you need me to write a patch, feel free to ask.

cody-signal:
Have at it.

PR #12164, opened 2022-04-14
fix broken .onion links

jtracey:
### Contributor checklist

 * AVD emulator, Android API 30 

----------

### Description

As described in #11458, 3f7dd2118654516aea26cd87e095a4ad87d2a606, which intended to merely disable link previews for .onion addresses, instead disabled them as links entirely. This commit fixes that. The other obvious way of resolving this is to move some of the `isLegalUrl` logic into `isValidPreviewUrl`. I opted for the boolean arg so the URL only needs to be parsed once, but it's a matter of taste, and if you want me to switch it feel free to ask, or to just write the commit on your own directly.


cody-signal (changes requested):
Thanks for the PR. Can you update to either create a new utility class that others and link preview can use with properly named methods for caring or not caring about TLDs instead of adding the extra boolean everywhere.

jtracey:
@cody-signal How about this? Rather than breaking it out into its own standalone class, I made a private subclass, and `isLegalUrl` now just works everywhere as expected, with the TLD logic moved into the `isValidPreviewUrl` method. I can create a standalone `LinkUtil`, it just gets a little ugly because again, I'd like to make sure we're only parsing once, so we'd either have to leak the list of non-preview TLDs into the `LinkUtil` class, or the internal URL representation of the parsed domains out of it.

cody-signal:
Pulled in, tweaked it a bit, but still attributed to you and will be in the next beta.

PR #12284, opened 2022-06-07
Changed LICENSE to WTFPL because GPL sucksGPL is a horrible license because it doesn't allow proprietary forks.That's why I have now relicensed this repository under WTFPL.

ColinTheContributor:


clauz9:
@ColinTheContributor so let me get this straight 🤔.... 
the problem your PR is trying to solve can be summarized as 

"I can't be bothered to use my brain and create something of my own, so I'm going to change the license of someone else's work, so I can fork it into a closed source clone, slap a different name on it, and sell it as my own" 

...right ? 🤔  would you say I captured the essence of your PR ? 🤣 
I'm asking because you can't even be bothered to write an actual description for your PR 🤦

PR #12312, opened 2022-06-26
Add "existing group" picker dialog to new group flow

brian-mcnamara:

### Contributor checklist

 * Pixel 5 Android 12
 * Virtual device Android 9

----------

### Description
This change adds a new dialog as part of the new group flow between the user selection and the group naming. The dialog only shows if there exists groups locally that match the added members. The dialog also allows the user to continue to create a new group.

The inspiration behind this is I often forget the name I set for old groups and occasionally create new groups when I would have preferred to find the old group.
Member selection screen

New Dialog after the user presses next

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/database/GroupDatabase.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/database/GroupDatabase.java
@@ -483,6 +483,29 @@ public class GroupDatabase extends Database implements RecipientIdDatabaseRefere
     return groups;
   }
 
+  @WorkerThread
+  public @NonNull List<GroupRecord> getGroupsWithMembers(@NonNull List<RecipientId> recipientId) {
+    List<RecipientId> recipientListCopy = new ArrayList<>(recipientId);
+    recipientListCopy.add(Recipient.self().getId());
+    Collections.sort(recipientListCopy);
+
+    SQLiteDatabase database   = databaseHelper.getSignalReadableDatabase();
+    String         table      = TABLE_NAME + " INNER JOIN " + ThreadDatabase.TABLE_NAME + " ON " + TABLE_NAME + "." + RECIPIENT_ID + " = " + ThreadDatabase.TABLE_NAME + "." + ThreadDatabase.RECIPIENT_ID;
+    String         query      = MEMBERS + " = ? AND " + ACTIVE + " = 1";
+    String[]       args       = SqlUtil.buildArgs(RecipientId.toSerializedList(recipientListCopy));
+    String         orderBy    = ThreadDatabase.TABLE_NAME + "." + ThreadDatabase.DATE + " DESC";
+
+    List<GroupRecord> groups = new LinkedList<>();
+
+    try (Cursor cursor = database.query(table, null, query, args, null, null, orderBy)) {
+      while (cursor != null && cursor.moveToNext()) {
+        groups.add(new Reader(cursor).getCurrent());
+      }
+    }
+
+    return groups;
+  }
+
   public Reader getGroups() {
     @SuppressLint("Recycle")
     Cursor cursor = databaseHelper.getSignalReadableDatabase().query(TABLE_NAME, null, null, null, null, null, null);
--- a/app/src/main/java/org/thoughtcrime/securesms/groups/ui/creategroup/CreateGroupActivity.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/groups/ui/creategroup/CreateGroupActivity.java
@@ -19,7 +19,9 @@ import org.thoughtcrime.securesms.ContactSelectionListFragment;
 import org.thoughtcrime.securesms.R;
 import org.thoughtcrime.securesms.contacts.ContactsCursorLoader;
 import org.thoughtcrime.securesms.contacts.sync.ContactDiscovery;
+import org.thoughtcrime.securesms.database.GroupDatabase;
 import org.thoughtcrime.securesms.database.RecipientDatabase;
+import org.thoughtcrime.securesms.database.SignalDatabase;
 import org.thoughtcrime.securesms.groups.ui.creategroup.details.AddGroupDetailsActivity;
 import org.thoughtcrime.securesms.recipients.Recipient;
 import org.thoughtcrime.securesms.recipients.RecipientId;
@@ -166,9 +168,17 @@ public class CreateGroupActivity extends ContactSelectionActivity {
 
       return ids;
     }, recipientIds -> {
+      List<GroupDatabase.GroupRecord> existingGroups = SignalDatabase.groups().getGroupsWithMembers(recipientIds);
+      stopwatch.split("group-search");
+
       dismissibleDialog.dismiss();
       stopwatch.stop(TAG);
-      startActivityForResult(AddGroupDetailsActivity.newIntent(this, recipientIds), REQUEST_CODE_ADD_DETAILS);
+      if (existingGroups.isEmpty()) {
+        startActivityForResult(AddGroupDetailsActivity.newIntent(this, recipientIds), REQUEST_CODE_ADD_DETAILS);
+      } else {
+        ExistingGroupChooserDialog existingGroupChooserDialog = new ExistingGroupChooserDialog(existingGroups, recipientIds);
+        existingGroupChooserDialog.show(getSupportFragmentManager(), "ExistingGroupChooserFragment");
+      }
     });
   }
 }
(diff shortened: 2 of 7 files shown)


app/src/main/java/org/thoughtcrime/securesms/database/GroupDatabase.java
```diff
@@ -412,6 +412,29 @@ public List<String> getPushGroupNamesContainingMember(@NonNull RecipientId recip
     return groups;
   }
 
+  @WorkerThread
+  public @NonNull List<GroupRecord> getGroupsWithMembers(@NonNull List<RecipientId> recipientId) {
+    List<RecipientId> recipientListCopy = new ArrayList<>(recipientId);
```
brian-mcnamara, line 417: Let me know if adding the current user here makes sense or if it should be the callers responsibility. My thought was the current user should be part of all local groups.

app/src/main/java/org/thoughtcrime/securesms/groups/ui/creategroup/ExistingGroupChooserDialog.java
```diff
@@ -0,0 +1,79 @@
...
+import org.thoughtcrime.securesms.groups.ui.creategroup.details.AddGroupDetailsActivity;
+import org.thoughtcrime.securesms.recipients.Recipient;
+import org.thoughtcrime.securesms.recipients.RecipientId;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ExistingGroupChooserDialog extends DialogFragment {
+
+  private final int REQUEST_CODE = 2840;
```
brian-mcnamara, line 29: Random number used for the "Create group" button, unsure if there is a convention.

app/src/main/res/layout/choose_existing_group_dialog.xml
```diff
@@ -0,0 +1,50 @@
...
+        android:gravity="center"
+        android:text="@string/ChooseExistingGroup__choose"
+        android:textAppearance="@style/Signal.Text.Body"
+        android:textStyle="bold"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintRight_toRightOf="parent"
+        app:layout_constraintLeft_toLeftOf="parent" />
+
+    <androidx.recyclerview.widget.RecyclerView
+        android:id="@+id/selected_list"
```
brian-mcnamara, line 24: I reused a bunch of ids. let me know if I should create unique ones.

app/src/main/res/layout/choose_existing_group_dialog.xml
```diff
@@ -0,0 +1,50 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:orientation="vertical"
+        android:paddingHorizontal="@dimen/dsl_settings_gutter"
```
brian-mcnamara, line 28: Margins and paddings were either copied or set based on what I thought looked best. Will not claim to have a keen eye for UI

brian-mcnamara:
I have tested scrolling of groups (the recyclerView has a max height), text overflow and normal operations (no group existing so dialog does not show, and dialog finds expected groups). The one thing I could not test was the min api (19 as of this writing). Unfortunately when I attempted to use an avd, signal crashed.

Tursko:
Does this also work for MMS groups? You can only have 1 working MMS thread with a set of users, and it would be nice if Signal would open the pre-existing thread instead of creating a new dummy thread.

brian-mcnamara:
Hey @Tursko , Ive actually never used MMS/SMS on signal. I did just try it just to check and it does show the existing MMS group I created from another phone between two other numbers I have.

Tursko:
Nice! If you wanted to go above and beyond you would basically force the user to reuse that group.

I think it should work a bit differently but still you seem to have laid the framework. 

If a person tries to create a new mms group with the same set of people that already exist in an mms thread it should simply open that thread. Just like it does when you select a single user.

brian-mcnamara:
Yea, thats not a bad suggestion. Ill take a look. Looks like I need to rebase anyway.

Tursko:
Here's my FR for it if you want more details. 

https://community.signalusers.org/t/sms-mms-respect-existing-mms-groups-when-creating-a-new-mms-group-with-the-same-list-of-users/15077

brian-mcnamara:
Well that was not too hard... Added support to skip the dialog and send you right into the existing group.

Tursko:
So what about the scenario where the user has more than 1 MMS group with the same set of users (like they haven't yet deleted the other group). Maybe you should just take the first MMS group result intead. Otherwise, you would just get in a loop of creating new MMS groups until you clean everything up.

app/src/main/java/org/thoughtcrime/securesms/groups/ui/creategroup/CreateGroupActivity.java
```diff
@@ -176,6 +178,19 @@ private void handleNextPressed() {
       if (existingGroups.isEmpty()) {
         startActivityForResult(AddGroupDetailsActivity.newIntent(this, recipientIds), REQUEST_CODE_ADD_DETAILS);
       } else {
+        //If there is only one group and its an MMS, there is should not be an option to create a new group
+        if (existingGroups.size() == 1 && existingGroups.get(0).isMms()) {
```
Tursko, line 182: Could be more than 1 MMS group in your local database with the same set of users (until this change of course).

app/src/main/java/org/thoughtcrime/securesms/groups/ui/creategroup/CreateGroupActivity.java
```diff
@@ -176,6 +178,19 @@ private void handleNextPressed() {
...
+          GroupDatabase.GroupRecord group = existingGroups.get(0);
```
Tursko, line 183: If that's the case, probably should take the last group.

GroupDatabase.GroupRecord group = existingGroups.get(existingGroups.size() - 1); // or whatever will grab the last index, not sure if just '-1' will work.

app/src/main/java/org/thoughtcrime/securesms/groups/ui/creategroup/CreateGroupActivity.java
brian-mcnamara: Yes, but I am not confident that this is always the case. I would have to dig further into the code to figure out how it handles receiving messages, and check if it simply chooses the oldest by date.

brian-mcnamara: My 2 cents on this are its a workaround for an existing bug. In an ideal world the user would clean up the existing messages and the workaround no loner is ever needed. By showing the user the existing groups with the same contacts, its easier for them to find the groups and delete them vs brushing it under the rug.

Tursko: The problem with cleaning them up automatically is that you have the possibility of losing messages that you've sent. 

Give it a try. Create a second mms group with the same set of users. Send a message in that new one. 

When they respond to the thread, you will only get the messages in the original thread. The first MMS thread that was created. But the message you sent still exists in the secondary thread. 

It's odd, but it's a bug because it allows you to create more groups.

brian-mcnamara: Right, I'm not saying they should be automatically cleaned up. I'm just stating that if there are more then 1 mms in the list, I do not believe logic should exist which chooses the last one and open it.

Tursko: Gotcha. Well we would have to open the one that is active. I was just assuming that it's the oldest thread is the correct thread. You wouldn't want to open one of the newer threads since they cannot receive messages in the first place.

Something that we'd have to test.

Tursko:
Might as well revert the mms group chat change since Signal is officially dropping support.

greyson-signal:
Hey there, just wanted to chime in.
1. I like this feature in general, and it's been something I've wanted.
2. If we end up building this, the UX will likely be different.

Which is my way of saying: we likely will not be merging this PR. But I do appreciate the time you've spent on it, and my intention is that if we end up building this, we will re-use the bits of your code that are possible to re-use, or have some general way for you to get authorship in the history. Thanks!

brian-mcnamara:
Thanks for looking Grayson, somehow missed the notification you commented on it. 
I did just update the branch on my fork (since my device was nearing the incompatibility window) with the migration to kotlin and group_membership table introduction. Had to rework the query quite a bit, and probably a much more elegant sql query can be used, but I keep this branch on my device.

Anyway thanks again for the comments, looking forward to the feature one day :)

PR #12496, opened 2022-10-06
Fix sending normal group messages when falling back to socket

AsamK:
In the sendGroupMessage message the socket fallback for sending normal group messages always set the story parameter to true. This causes the message to be discarded by the receivers, because it has a story envelope, but no story content
> Envelope was flagged as a story, but it did not have any story-related content! Dropping.

Issue was introduced in https://github.com/signalapp/Signal-Android/commit/3895578d5188c7d2ef3378d61d7a9aa6efa03928#diff-6820409682d75df5f8d2fe5804e63c2712ccab77e99d0f987a40e50646fcddacR1957

@greyson-signal I think this is a rather severe issue, that could cause group messages to go missing, if the websocket is not connected.


greyson-signal:
Thanks for finding this! I'll get it out quickly.

PR #12545, opened 2022-10-21
Expanded functionality for private group chats

unex2022:
Please expand private group chats by:
1.
the possibility of having more than one person with administrative right theres. Its now the group owner who has them.
The new role maybe called moderator and should contain adding and removing members.
Adding new members to a private chat should be removed from normal group members.
2.
the possibility to upload a profile picture for a private group chat


unex2022:
This should be for session.

clauz9:
> This should be for session.

You should close this .. and open the PR on their Repo..

PR #12563, opened 2022-10-29
Fix back button behaviour in bubbles

orange-elephant:
### Contributor checklist

 * Device A, Graphene OS A13
 * Stock A12 emulator

----------

### Description

Due to the onBackPressed implementation in ConversationParentFragment pressing back to exit a bubble dismisses it entirely in the best case,which already doesn't seem desirable, but on my grapheneos device it launches the conversation list which entirely defeats the point of bubbles.

Falling back to the default onBackPressed behaviour specifically when we are expecting to leave a bubble fixes this.


alex-signal:
Thank you!

This should go out in 6.1

Issue #10810, opened 2021-01-25
When You Delete An Audio Message Whle It's Playing it will Sitll Play


bitmold:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
If you delete an audio message while it is playing, the message will continue to play. 

### Steps to reproduce
- Record an audio message, easier to see this bug with a longer one
- Play it
- Delete the message while it is playing without stopping the message

**Actual result:** The message continues to play, there is still a notification for the ongoing message as well. If you click the the notification, the app will open and select the previous message in the conversation. If you have that notification disabled, you have to take drastic measures to stop that message from playing since you can not stop the audio from the notification. 
**Expected result:** The message should immediately cease playing and the notification associated with it should also be gone.

### Device info
**Signal version:** 5.2.3

bitmold:
Ya

bitmold:
Can confirm still happens

PR #12611, opened 2022-11-17
When You Delete An Audio Message While It's Playing it will Still Plays

gitstart:

### Contributor checklist

 * Pixel 6, Android 13
 * Virtual device Pixel 4a, Android 12

----------

### Description

fixes https://github.com/signalapp/Signal-Android/issues/10810

Bug -> https://www.loom.com/share/28c2baea553b461098886f1444f6a744
Fix-> https://www.loom.com/share/56675f70de564dc6a7bbf110c0ac37e4

when a voice note is playing and user attempt to delete the voice note, the voice not continues playing in the background. 

After this changes when the voice is deleted, it no longer plays in the background


app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
```diff
@@ -1096,11 +1099,24 @@ private static boolean isNoteToSelfDelete(Set<MessageRecord> messageRecords) {
...
+    }
+  }
+
   private void handleDeleteForEveryone(Set<MessageRecord> messageRecords) {
     Runnable deleteForEveryone = () -> {
       SignalExecutors.BOUNDED.execute(() -> {
         for (MessageRecord message : messageRecords) {
           MessageSender.sendRemoteDelete(message.getId(), message.isMms());
+          if (message.isMms()){
+            handleStopAudioMessageAfterDeletion((MediaMmsMessageRecord)message);
```
alex-signal, line 1118: `(MediaMmsMessageRecord)message`

Please add a space between cast and object

app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
```diff
@@ -1065,6 +1067,7 @@ protected Void doInBackground(Void... voids) {
 
             if (messageRecord.isMms()) {
               threadDeleted = SignalDatabase.mms().deleteMessage(messageRecord.getId());
+              handleStopAudioMessageAfterDeletion((MediaMmsMessageRecord)messageRecord);
```
alex-signal, line 1070: `(MediaMmsMessageRecord)message`

Please add a space between cast and object

app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
```diff
@@ -1046,6 +1047,7 @@ private void handleDeleteMessages(final Set<MultiselectPart> multiselectParts) {
   }
 
   private AlertDialog.Builder buildRemoteDeleteConfirmationDialog(Set<MessageRecord> messageRecords) {
+    Context             context       = requireActivity();
```
alex-signal, line 1050: Where is this used?

app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
```diff
@@ -1096,11 +1099,24 @@ private static boolean isNoteToSelfDelete(Set<MessageRecord> messageRecords) {
     return messageRecords.stream().allMatch(messageRecord -> messageRecord.isOutgoing() && messageRecord.getRecipient().isSelf());
   }
 
+  private void handleStopAudioMessageAfterDeletion(@NonNull MediaMmsMessageRecord message){
```
alex-signal, line 1102: Something about this fix makes me wonder if we shouldn't be handling message deletion in the `VoiceNotePlaybackService`.

We could add a database notifier in there that lasts for the lifetime of the service, and if we see that the current playing item is deleted, we can stop playback at that point. That would cover things like expiring messages and would also cover remote deletion by another user.

alex-signal (changes requested):
A few nitpicks, but I'm going to share this with the team and see what they think of the approach.

app/src/main/java/org/thoughtcrime/securesms/conversation/ConversationFragment.java
gitstart: Handling message deletion seems to be outside the scope of this PR, what do you think?

alex-signal: This PR is all about message deletion, and this feels like a partial fix to me. Having the service handle the deletion would make more sense, you wouldn't need any of this code here, you'd just need to handle it properly in the service itself.

Issue #12557, opened 2022-10-27
An empty line should have no background - Media editing


Meteor0id:
> 

- [ ] I have searched open and closed issues for duplicates
- [ ] I am submitting a bug report for existing functionality that does not work as intended
- [ ] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [ ] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Describe here the issue that you are experiencing.

### Steps to reproduce
- using hyphens as bullet points
- list the steps
- that reproduce the bug

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)
**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)

### Screenshots

![Screenshot_20221028-005107](https://user-images.githubusercontent.com/34976212/198412818-1d99c191-1fdb-45c8-a547-2665cfbb4b13.png)

### Device info

**Device:** Manufacturer Model XVI
**Android version:** 0.0.0
**Signal version:** 0.0.0

Meteor0id:
Get lost bot

greyson-signal:
This should be fixed? Are you still able to reproduce it?

PR #12612, opened 2022-11-17
[SIG-3] An empty line should have no background - Media editing #12557

gitstart:

### Contributor checklist

 * Samsung Galaxy M02, Android 11
 * Pixel 3, Android 10

----------

#### Base for the fix:
The target bound for rendering a text effect (including HIGHLIGHT) is represented by a `RectF` object calculated on top of the text block's intrinsic property. This calculation, and everything related to actual rendering/drawing happens inside `Line` object's `render()` function.

#### The issue:
The `render()` function calculates the render target bound without actually checking if the text block is empty.

The issue is more clear if we look at how the render function calculates the target, but can be generalised as:

> `targetBound = textBlockProperty + additionalProperty`

Even if `textBlockProperty` is null/empty, `targetBound` will always have `additionalProperty` - which is why the issue exists.

#### The fix:
In case the text block is empty (which is currently the case for newlines), then the bound for render target is cleared.

```
  if (mode == Mode.HIGHLIGHT) {
        if(text.isEmpty()){
          modeBounds.setEmpty(); 
        }else{
          modeBounds.set(textBounds.left - HIGHLIGHT_HORIZONTAL_PADDING,
                  selectionBounds.top - HIGHLIGHT_TOP_PADDING,
                  textBounds.right + HIGHLIGHT_HORIZONTAL_PADDING,
                  selectionBounds.bottom + HIGHLIGHT_BOTTOM_PADDING);
        }
```

### Recordings

- Issue

https://user-images.githubusercontent.com/13991373/201327739-d258b1d5-cb31-4eb9-85a7-906550ccb9eb.mov

- Fix

https://user-images.githubusercontent.com/13991373/201327819-d7f54a65-a71c-4f7f-b7b4-74494260bca4.mov

### DIFF
--- a/image-editor/lib/src/main/java/org/signal/imageeditor/core/renderers/MultiLineTextRenderer.java
+++ b/image-editor/lib/src/main/java/org/signal/imageeditor/core/renderers/MultiLineTextRenderer.java
@@ -320,10 +320,14 @@ public final class MultiLineTextRenderer extends InvalidateableRenderer implemen
       rendererContext.canvasMatrix.concat(projectionMatrix);
 
       if (mode == Mode.HIGHLIGHT) {
-        modeBounds.set(textBounds.left - HIGHLIGHT_HORIZONTAL_PADDING,
-                       selectionBounds.top - HIGHLIGHT_TOP_PADDING,
-                       textBounds.right + HIGHLIGHT_HORIZONTAL_PADDING,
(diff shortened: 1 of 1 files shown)


alex-signal:
Thanks for your contribution! Unless there are issues, this will go into `6.4`

alex-signal:
@gitstart Should've let it auto-close, but I've brought this into 6.4 on our end. When we publish 6.4 (probably after the holiday) you'll see this work in `main`

PR #12616, opened 2022-11-17
Make it clear that payment notes have limited length

orange-elephant:
### Contributor checklist

 * Device A, Android 13

----------

### Description

Payment notes are limited in length to 40 characters, but previously there was absolutely no indication of this, which is very confusing.

This commit adds a simple character counter, and a toast when the EditText is full, to make it clear how much you can write. A missing accessibility description has also been added to the save note button.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/payments/create/EditNoteFragment.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/payments/create/EditNoteFragment.java
@@ -4,6 +4,8 @@ import android.os.Bundle;
 import android.text.TextUtils;
 import android.view.View;
 import android.view.inputmethod.EditorInfo;
+import android.widget.TextView;
+import android.widget.Toast;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -15,11 +17,16 @@ import org.thoughtcrime.securesms.LoggingFragment;
 import org.thoughtcrime.securesms.R;
 import org.thoughtcrime.securesms.components.emoji.EmojiEditText;
 import org.thoughtcrime.securesms.util.ViewUtil;
+import org.thoughtcrime.securesms.util.text.AfterTextChanged;
+
+import java.util.Locale;
 
 public class EditNoteFragment extends LoggingFragment {
 
   private CreatePaymentViewModel viewModel;
   private EmojiEditText          noteEditText;
+
+  private static final int NOTE_MAX_LENGTH = 40;
 
   public EditNoteFragment() {
     super(R.layout.edit_note_fragment);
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -3051,6 +3051,8 @@
 
     <!-- EditNoteFragment -->
     <string name="EditNoteFragment_note">Note</string>
+    <string name="EditNoteFragment__max_note_length_reached">Maximum note length reached</string>
```
alex-signal, line 3054: Please add a comment to each of these strings describing what they are used for.

ex.

```
<!-- Note displayed in a toast informing user that max note length has been reached -->
```

Also, this should probably be __maximum_note_length_reached to be consistent with other strings.

app/src/main/java/org/thoughtcrime/securesms/payments/create/EditNoteFragment.java
```diff
@@ -48,6 +59,22 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
       return false;
     });
 
+    TextView lengthIndicator = view.findViewById(R.id.character_count);
+
+    noteEditText.addTextChangedListener(new TextWatcher() {
```
alex-signal, line 64: We have an `AfterTextChanged` class you can use here.

app/src/main/java/org/thoughtcrime/securesms/payments/create/EditNoteFragment.java
```diff
@@ -33,6 +42,8 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
     toolbar.setNavigationOnClickListener(v -> Navigation.findNavController(v).popBackStack());
 
     noteEditText = view.findViewById(R.id.edit_note_fragment_edit_text);
+    noteEditText.setFilters(new InputFilter[] { new InputFilter.LengthFilter(NOTE_MAX_LENGTH) });
```
alex-signal, line 45: Can we not set this directly in xml?

alex-signal (changes requested):

app/src/main/java/org/thoughtcrime/securesms/payments/create/EditNoteFragment.java
orange-elephant: Thanks!

app/src/main/java/org/thoughtcrime/securesms/payments/create/EditNoteFragment.java
orange-elephant: Yeah, the only reason I did this was to have the limit defined in a single place since there is no easy way to get the xml defined value apart from by defining it in a resource integers file or iterating an array of InputFilters. I've now just set this in the XML and the Fragment separately.

dev-aniketj (approved):

greyson-signal:
This is in 6.16, thanks :)

Issue #10324, opened 2020-12-20
Panning sideways over a zoomed image causes switching to other image


Meteor0id:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Describe here the issue that you are experiencing.

### Steps to reproduce
- Full screen some image
- Zoom in with pinch zoom
- Pan around all you like
- Pan all the way to the right till you can go no further
- Attempt to pan back left

**Actual result:** You switch to another image
**Expected result:** Since you are still zoomed, swiping in any direction should only let you pan. You should be able to pan left.

### Device info

**Device:** Motorola Moto G7 Plus
**Android version:** 10.0
**Signal version:** 50.0.8

mdatla:
I'm seeing this issue as well.
Android, Pixel 7 Pro

mdatla:
What's the process to reopen this ticket?

zeyadtamimi:
Hey,

I saw the comments on #12629 and I see where you are coming from, allowing paging while zoomed in does have benefits for some people.  However, in trying to get my own circle of friends to use signal I have encountered multiple complains regarding this behavior.  In fact there are times where we download images so we can use a separate photo viewing application.

Given the fact that there will always be two differing opinions on this behavior, wouldn't  it be better to have a setting so that the user can chose which behavior best suites their use case? 

If you are open to the idea then I can try reviving the changes with a new "Image Viewer" setting under which will allow people to switch to a mode where the image viewer does not page while panning a zoomed in photo. 

Lastly, i am not really sure if this is good etiquette so I apologize in advance. However, since @greyson-signal was the last person to comment on the PR I am `@` ing him for some continuity.

Meteor0id:
These days Signal had a team of designers  every change of the app goes through them and they rarely take feedback like this seriously. Thanks for trying but good luck.

Starcommander:
I think, this is a bit confusing.

I also have sometimes the issue:
You zoom in, and try to pan left.
But Signal switches the image instead allowing to pan left.
Switching the image should only happen, when reaching the border of the image.

Unfortunately, I was not able to reproduce and record this behaviour.

Starcommander:
@Meteor0id I would suggest to update the "**expected result**" in this issue.
[...]

PR #12629, opened 2022-11-22
[SIG-12] Panning sideways over a zoomed image causes switching to other image

gitstart:

### Contributor checklist

 * Samsung Galaxy M02, Android 11
 * Pixel 3 Emulator, Android 10

----------

#### Issue
- When a zoomed image is swiped left or right on the edge, the image is switched to the next image. (Issue [#10324](https://github.com/signalapp/Signal-Android/issues/10324))

#### Expected Behaviour:
- When a zoomed image is swiped left or right on the edge, the image should not be switched to the next image.
- Swiping to the next image should only be possible if images are not zoomed.

#### Description of the fix:
`PhotoView` class is an extension of `ImageView` class that handles functions related to displaying an image and adding zoom functionality on top of that.
This fix uses `PhotoView.setAllowParentInterceptOnEdge(false)` to disable parent view (in this case the `ViewPager2` class) from intercepting touch event on the edge of the image if the image is zoomed (i.e scale factor is not equal to 1.0f)

#### Recordings
- Issue

https://user-images.githubusercontent.com/13991373/202855449-78fa732c-77c3-4945-997a-d87f381610c7.mov

- Fix

https://user-images.githubusercontent.com/13991373/202855504-2fce7f4b-c042-4f67-81fe-4898d14badcc.mov

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/components/ZoomingImageView.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/ZoomingImageView.java
@@ -2,7 +2,6 @@ package org.thoughtcrime.securesms.components;
 
 import android.annotation.SuppressLint;
 import android.content.Context;
-import android.graphics.drawable.Drawable;
 import android.net.Uri;
 import android.util.AttributeSet;
 import android.view.MotionEvent;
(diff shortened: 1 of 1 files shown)


greyson-signal:
Hi there, thanks for the PR, but we're going to keep it the way it is. There's pro's and con's to allowing paging while zoomed in, and we prefer keeping it the same.

gregreen:
@greyson-signal The current behavior makes it almost impossible to pan within a zoomed-in image. It's honestly extremely frustrating to deal with, and feels much more like buggy behavior than a feature.

A good compromise would be to only switch to the next image when the edge of the zoomed image is reached. There are no downsides to this change, because I can't imagine anyone who's zoomed into an image expecting the next image to show up when they pan slightly to the left or right.

tabaker99:
@greyson-signal paging should occur when the user reaches the edge of the image. But right now, when trying to pan a zoomed-in image (at the center of the image, nowhere near the edge), paging will sometimes occur and sometimes not occur. It seems to be random. This is not the behavior of any other image viewer I've ever used on a smartphone, and its unpredictability makes me pretty certain it's a bug.

PR #12771, opened 2023-02-03
Fix database creation crash on first launch

orange-elephant:
### Description

Trailing comma in create table command makes the statement invalid


nicholas-signal:
Thank you for your contribution, but this is already fixed in the upcoming release.

PR #12919, opened 2023-04-26
Remove misfeatures of forced upgrades and SMS removal.

jgilmore:
### Description
You're going to reject/ignore/close this pull request.

I'm making the pull request primarily because I want people to find this, even though it's a hatchet job. And I totally suck at android development. As an example, I wasn't able to isolate the nag screens about SMS removal, so they're still there.

So I both would appreciate others contribution to a project focusing on being the best private messaging app (rather than cutting features and controlling people) and don't want other people to go through what I had to do for this hatchet job.

Yes, this is a hatchet job. But when "removing misfeatures" is the goal, it'll do the job. Do I really want my version of signal to EVER expire? No. If for some reason I CANNOT upgrade, it's stupid to disable what I already have. Better to allow continued use if the alternative is NO use. Same thing with SMS. If it works, don't fix it. If it breaks, that's the users problem, especially if we warned them it probably would! Taking blame on yourselves for other people's misinterpretations (however costly to them) is pointless.

You're correct, in that it may not be possible to support SMS in the future. You're incorrect, in that that fact doesn't mean we should remove existing functional features, however buggy they may be under whatever circumstances.

I don't know if I will EVER do anything further with this, as I have very limited focus time and my family suffered to put this out. But if I ever do add features, the first one will be "import old backups, merging with existing messages" as the lack of this feature is why I'm losing my message history with this upgrade. (And why my wife is, which is much more important.)


nicholas-signal:
Interested users can understand our reasoning for SMS removal by reading the Signal Blog: https://www.signal.org/blog/sms-removal-android/

dnkmmr69420:
> Interested parties can understand our reasoning for SMS removal by reading the Signal Blog: https://www.signal.org/blog/sms-removal-android/

but that means the person who you message must have the signal app which would make this app quite useless

PR #12972, opened 2023-05-28
Bail out quickly if WebSocket can't connect

g1a55er:
### Contributor checklist
 * Pixel 6A, running Android 13

----------

### Description

Currently, if the app gets a message from Firebase Cloud Messaging but can't connect to the WebSocket, it will just sit there and burn battery holding the wake lock and a foreground service for up to 90 seconds. 

Instead, it is probably best to bail out early and try again later if the WebSocket doesn't connect. 

Tested via adding this diff to my test code so that WebSocketConnection always times out: 

```
diff --git a/app/src/main/java/org/thoughtcrime/securesms/messages/IncomingMessageObserver.kt b/app/src/main/java/org/thoughtcrime/securesms/messages/IncomingMessageObserver.kt
index 69c5cac44d..38539007ba 100644
--- a/app/src/main/java/org/thoughtcrime/securesms/messages/IncomingMessageObserver.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/messages/IncomingMessageObserver.kt
@@ -412,6 +412,8 @@ class IncomingMessageObserver(private val context: Application) {
           decryptionDrained = false
         }

+        Thread.sleep(10000);
+
         signalWebSocket.connect()
         try {
           while (isConnectionNecessary()) {
```

This is part of a series of things I found when looking at the Android code to try and improve notification delivery.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/messages/WebSocketStrategy.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/messages/WebSocketStrategy.java
@@ -32,6 +32,10 @@ public class WebSocketStrategy extends MessageRetrievalStrategy {
     Stopwatch               stopwatch = new Stopwatch("websocket-strategy");
     IncomingMessageObserver observer  = ApplicationDependencies.getIncomingMessageObserver();
 
+    if (!blockUntilWebsocketConnected(observer)) {
+      return false;
+    }
+
     observer.registerKeepAliveToken(KEEP_ALIVE_TOKEN);
     try {
       JobManager              jobManager    = ApplicationDependencies.getJobManager();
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/thoughtcrime/securesms/messages/WebSocketStrategy.java
```diff
@@ -58,6 +62,29 @@ public boolean execute() {
     }
   }
 
+  private static boolean blockUntilWebsocketConnected(IncomingMessageObserver observer) {
```
g1a55er, line 65: I could factor out these `blockUntil...` methods to share code instead of repeating it, if desired. I'm not sure where the Signal team draws the line on when it's better to have indirection vs repetition.

app/src/main/java/org/thoughtcrime/securesms/messages/WebSocketStrategy.java
```diff
@@ -58,6 +62,29 @@ public boolean execute() {
...
+
+    observer.addServiceConnectedListener(new Runnable() {
+      @Override public void run() {
+        latch.countDown();
+        observer.removeServiceConnectedListener(this);
+      }
+    });
+
+    try {
+      if (!latch.await(5, TimeUnit.SECONDS)) {
```
g1a55er, line 76: I choose this somewhat arbitrarily as an estimate for what a P99 WebSocket connection time might be. If you have a better estimate, please let me know.

cody-signal (commented):
Can you help me understand where the 90s number comes from in your pr description.

cody-signal:
This was addressed in our recent web scoket work. Thanks for the initial PR.

PR #12978, opened 2023-06-02
Handle Network Access Loss in Background Due to Data Saver

g1a55er:
### Contributor checklist
 * Pixel 6A, Android 13, Data Saver On, Wifi set to "Metered"

----------

### Description

Notification delivery is a bit wonky if data saver is on, as this causes the network to be blocked in the background any time the app is not running a foreground service. 

The overall problem here is that the foreground service stops too early, before the delivery receipt makes it out. Then, if more messages come in, FCM will buffer and delay the alerts for them because it just woke the app up right before, and it looks like the app is still running. 

The behavior right now is that if I send two consecutive messages to a device in Data Saver mode, only the first one will make it through. The second one will be delayed for some arbitrary amount of time. Also, if I send a message to a device in Data Saver right after the app is backgrounded, the message will be delayed. 

This is a small incremental change that saves the app from continuing to wait on a WebSocket that is already dead. This resolves some of the delays around messages that arrive right after the app is foregrounded.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/messages/IncomingMessageObserver.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/messages/IncomingMessageObserver.kt
@@ -87,9 +87,9 @@ class IncomingMessageObserver(private val context: Application) {
 
   private val lock: ReentrantLock = ReentrantLock()
   private val connectionNecessarySemaphore = Semaphore(0)
-  private val networkConnectionListener = NetworkConnectionListener(context) { isNetworkAvailable ->
+  private val networkConnectionListener = NetworkConnectionListener(context) { isNetworkLost ->
     lock.withLock {
-      if (isNetworkAvailable()) {
+      if (isNetworkLost()) {
         Log.w(TAG, "Lost network connection. Shutting down our websocket connections and resetting the drained state.")
         decryptionDrained = false
         disconnect()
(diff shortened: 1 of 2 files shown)


g1a55er:
I was reading up on the documentation for [ConnectivityManger.NetworkCallback](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback) and I think there may actually still be a general problem with these callbacks not matching what IncomingMessageObserver wants, as well as a more specific problem with blocked networks. 

The high level overview of the problem is that IncomingMessageObserver actually cares about the status of the network that is currently supporting its WebSockets, but the `NetworkCallback` retrieved with registerDefaultNetworkCallback will actually receive callbacks about the network the OS thinks that new requests should be sent over. Because the WebSocket was created some time ago and cannot be migrated, these may be different networks if the user switches networks. 

This conceptual mismatch makes it difficult to ensure that the blocked status change event refers to the network that supports the WebSocket. It also makes it difficult to ensure we only deliver one event per network connection, because each new network connection will result in one `onAvailable` call followed immediately by one `onBlockedStatusChanged(false)` call. 

I still believe this change produces better behavior than the status quo, so I will leave it open. However, I do admit I no longer believe it is strictly 100% correct.

greyson-signal:
Hi there, it looks like this was independently addressed in c08f1355.

PR #13016, opened 2023-06-21
Fix Do not animate spoilers if system animations are disabled.

Sgn-32:
### Contributor checklist

 * Virtual Pixel XL, Android 12

----------

### Description
The commit https://github.com/signalapp/Signal-Android/commit/272860f07132db05e52e0d8f7f017d78e59052cb in 6.24.1 disables the animation of spoilers if system animations are disabled.
When system animations are disabled the draw function
https://github.com/signalapp/Signal-Android/blob/272860f07132db05e52e0d8f7f017d78e59052cb/app/src/main/java/org/thoughtcrime/securesms/components/spoiler/SpoilerPaint.kt#L107
is executed anyway, although the values of the particles do not change, due to dt = 0.

This PR changes this. The update function is called onetime (in the init with update(true)) to compute a spoiler particles image.
Every other call of update() is directly returned, if system animations are disabled.


Sgn-32:
Checks failed:
`This request was automatically failed because there were no enabled runners online to process the request for more than 1 days.`

PR #13041, opened 2023-07-02
Change SpoilerPaint to avoid small visible patterns

Sgn-32:
### Contributor checklist

 * Moto X Play, Android 7
 * Virtual Pixel XL, Android 12

----------

### Description
The actual implementation of SpoilerPaint generates a small visible pattern on spoiler text. The main cause is the small height of 10dp/20dp and tiling the bitmap repeated. Due to the low height only the particles "lives long" with main direction to the x-axis. This also results in the wave-shaped pattern.

On the virtual device the original values gives a bitmap of (175px, 35px) (low perf.) or (350px, 70px) (high perf.)
With low performance the particle_count is 12, given 36 displayed points (12 * 3 alpha channels).
High performance the particle_count is 98 -> 294 displayed points.

Screenshot_1688294049 Screenshot_1688294127

This commits change the bitmap size to a better height. On low performance the bitmap is a square of 100dp x 100dp, on the test device this is 350px x 350px. High performance uses a bitmap of 120dp x 120 dp (420px x 420px).
With 24 particles (low perf.) or 105 particles (high perf.) This gives 72 or 315 points. Using additional the canvas.drawPoints with a offset of 1 "doubles" the displayed points to 141 or 627.
The offset of 1 gives points (y_i, x_i+1). So the bitmap must be a square.

Screenshot_1688294390 Screenshot_1688294632

For some more performance the frame-rate is reduced to only 64ms. Also the `val change = dt * velocity ` is constant for one call of the draw-method. So no need the recompute the value for each particle.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/components/spoiler/SpoilerPaint.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/spoiler/SpoilerPaint.kt
@@ -27,9 +27,9 @@ object SpoilerPaint {
    */
   var shader: BitmapShader? = null
 
-  private val WIDTH = if (Util.isLowMemory(ApplicationDependencies.getApplication())) 50.dp else 100.dp
-  private val HEIGHT = if (Util.isLowMemory(ApplicationDependencies.getApplication())) 10.dp else 20.dp
-  private val PARTICLES_PER_PIXEL = if (Util.isLowMemory(ApplicationDependencies.getApplication())) 0.002f else 0.004f
+  private val WIDTH = if (Util.isLowMemory(ApplicationDependencies.getApplication())) 100.dp else 120.dp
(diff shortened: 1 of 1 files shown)


Sgn-32:
https://github.com/signalapp/Signal-Android/assets/49990901/d0af1999-f500-46d5-b87d-9cee365fc767

cody-signal:
Thanks for taking a swing at this. However, I do have some concerns about the changes. Mostly around the 3 extra calls to drawPoints. Consistently in our testing, painting was the slowest part of update not computation.

I've taken inspiration from your changes and made a few tweaks to the class that'll be out in 6.26.

1. Change height to 20/40 and particles per pixel to 0.001/0.002 to compensate and keep the same particle count as before
2. Remove the bounds check for determining particle death

PR #13199, opened 2023-10-02
Add setting for users to determine frequency of backup creation

farewelltospring:

### Contributor checklist

 * Sony Xperia 5iii, Android 13
 * Virtual device Pixel 3, Android 13

----------

### Description

This pull request allows users change the frequency that backups are made to a selected interval (weekly, monthly, quarterly, every 6 months, annually).

Currently, when backups are enabled, Signal will create a fresh backup every 24-hours. This causes unnecessary write cycles on the device's internal storage or SD card.

- Making backups is a power-intensive process which drains battery and uses CPU cycles which can put a strain on daily usage when users are out and about, and slow down other apps on the device while the backups are being created.
- Different users have different needs from backups. Some users message on Signal all day long, and others message on Signal infrequently. A 24-hour backup schedule is useful for the first group, but is of dubious value for the second group. It will write an identical file over and over.
- Low-end devices are wearing out prematurely. Many low-end Android phones have low-durability eMMC chips. A 24-hour backup schedule makes their already-short lifespan even shorter. eMMC chips will fail catastrophically when they reach end-of-life. The device bricks, and the user must replace the whole thing. **This disproportionately hurts Signal's poor users and produces environmental pollution.**

A setting to control the backup frequency would also give people more control over their data, which is an important value to cultivate in today's society.

Verified that this behaviour works by setting backup frequency to 30 days and then waiting that long to see if a backup got created (I restarted the device several times during that time period to verify that it persists).

https://github.com/signalapp/Signal-Android/assets/108375980/13f7643f-0d26-4ecb-b4a3-0b428c40ac07


farewelltospring:
Fixed the localisation issue. I have only added a string for English but at least there are resources now.

farewelltospring:
Is there anything I can do to make this PR better?

LordBurtz:
@farewelltospring one suggestions, could you please add the option to put the number of days as an int (with sensible limits) as having to choose from a list feels arbitrary and might not cover all use cases, but I can totally understand if you keep it as is

+ i really like that feature

farewelltospring:
Just caught wind of this commit - https://github.com/signalapp/Signal-Android/commit/49ba83dda8ec7e9457dc620c960693f91590161e - is this PR able to be resolved?

EmoBurrito:
Bumping because I'm genuinely surprised this functionality isn't present and I don't want this to be forgotten. I'm not copying my backup elsewhere daily, so there's no point in wasting power and slowing down my phone that frequently.

tkrah:
Found this PR as *this* is the feature I really miss - what needs to be done to get this merged? This https://github.com/signalapp/Signal-Android/commit/49ba83dda8ec7e9457dc620c960693f91590161e is not be able to replace this PR because it is meant to be used in the upcoming remote chat backup - if you *WANT* a local backup, we still need this PR.

farewelltospring:
By "manual" you mean backups will only be created when the user presses the create button? I can get behind that.

farewelltospring:
Ok I uploaded some code. Full disclosure I haven't tested it yet because I gotta head out now and it'll take too long to compile, but I'll test it and take a video when I get back.

farewelltospring:
Demo recording attached.

https://github.com/user-attachments/assets/b953ddcd-a6d8-4a43-a658-236e1284acc1

farewelltospring:
Played around with it a little and I think found an edge case where if you change the frequency from a schedule to `NEVER`, then the original scheduled backup never gets canceled. So it'll create one more backup according to the previous frequency, and then stop afterwards. Do you think this is acceptable behaviour?

FWIW, I tried to dig into the code to figure out how to avoid this. There's a lot of abstraction and general cruft in the job management system (whose files are also not very well documented) and I can't immediately see a clean way to cancel a scheduled backup. **Could you give me a code pointer to cancel a scheduled job?** Or I can dig deeper myself at the cost of more time. Having said that, doesn't seem to persist across reboots, so there's that.

Curiously, one thing that *does* persist across reboots is in-progress backups. This seems to be a built-in feature to the JobRunner/JobDatabase/etc files, because logcat shows that it only tracks when the job was submitted and doesn't seem to respect the backup frequency setting when I switch it to `NEVER`.

IMO a good UX would be for switching the frequency setting to `NEVER` to:
1. Cancel future scheduled backups without relying on the user to reboot to clear it out
2. Cancel in-progress backups

Please let me know how you would like to move this PR forward.

farewelltospring:
Come through

alex-signal:
Cancellation methods are on JobManager, which you can access via `AppDependencies.getJobManager()`

gabefair:
Thank you for all the updates to make this feature possible

farewelltospring:
I've been daily driving this for the past month and the new "never" setting has been working as expected.

- Setting backup frequency to `NEVER` cancels a scheduled job without a need for reboot, confirmed by having a latest backup from April, setting the frequency to `DAILY`, and then setting it to `NEVER`, and confirming that a backup never begins getting created. :ballot_box_with_check: 
- Setting backup frequency to `NEVER`cancels an in-progress job, confirmed by checking logcat for `W LocalBackupJobApi29: Backup cancelled`. :ballot_box_with_check: 

If there are no other comments, I believe that this PR is ready to be reviewed for full merge.

farewelltospring:
:slightly_smiling_face:

farewelltospring:
🙂

farewelltospring:
. :pleading_face:
:point_right: :point_left:

farewelltospring:
@greyson-signal Is there anything I can do to make this easier for the team to review and merge?

gabefair:
Hi, this fork is a bit behind main, could you update your branch?
@greyson-signal could this feature be merged?

farewelltospring:
image

greyson-signal:
So we're actually in a weird spot now with adding something like this. We're in the midst of stabilizing our [new backup system](https://community.signalusers.org/t/public-signal-backups-testing/69984?u=greyson-signal), and with that comes a new local backup format that will likely solve every complaint you have in your initial post. Specifically, it'll be a differential backup, likely resulting in tens of MB's of data written per day (instead of GB's like it is now), and the entire backup will likely take seconds, not minutes.

For the new Signal-hosted backups, there's some important cryptography-related reasons why we'll be creating backups every day, and while those same reasons don't apply to the new local backups, I think it's gonna be weird if we allow scheduling for local and not Signal-hosted ones. And so I think for now I'm gonna officially "pause" this until we think about it some more.

tkrah:
> So we're actually in a weird spot now with adding something like this. We're in the midst of stabilizing our [new backup system](https://community.signalusers.org/t/public-signal-backups-testing/69984?u=greyson-signal), and with that comes a new local backup format that will likely solve every complaint you have in your initial post. Specifically, it'll be a differential backup, likely resulting in tens of MB's of data written per day (instead of GB's like it is now), and the entire backup will likely take seconds, not minutes.
> 
> For the new Signal-hosted backups, there's some important cryptography-related reasons why we'll be creating backups every day, and while those same reasons don't apply to the new local backups, I think it's gonna be weird if we allow scheduling for local and not Signal-hosted ones. And so I think for now I'm gonna officially "pause" this until we think about it some more.

Imho "we" thought about that more than 2 years now and I and I guess others too still miss this PR for my local backups. We need to setup + disable backups every week now for example and it is a hassle to do that, because I can't choose my passphrase, so every week I get a new one added to my list ... yeah.

Why it is weird to have that option only for scheduled backups, I don't see a problem with that difference - it is a different type of backup after all, local vs. signal hosted, so it can have different options imho, it already has, e.g. you choose a local folder where to save them.

+1 to move forward here and get this merged, please.

bmondream:
> I think it's gonna be weird if we allow scheduling for local and not Signal-hosted ones.

Scheduling could be added to local backups first and to Signal-hosted backups later. That's also an option.

I, too, would like to see local backup scheduling implemented.

HyperCriSiS:
I hope this still gets integrated because:
When local backups finally also have the incremental feature, I want to be able to decide, when there will be an incremental one and when there will be a full one. 
So for example:
- Incremental backup: everyday
- Full Backup: once a week

IMO this is really necessary.

bmondream:
I just tried it after figuring out how to build the project, and I noticed some to me unexpected behavior. As soon as I change the time of the scheduled backup, a backup is immediately created. I haven't yet tested if, in case of daily backups, it's created on time the next time (~~will report tomorrow~~ _it worked just fine the next time_).

bmondream:
I'm not sure this pull request is still necessary since the current 8.3.x version has new on-device backups (introduced possibly sooner than in 8.3.x). It's still a daily backup, but with differential changes from what I've gathered.

e5697768-98b8-4bb0-b123-0b8ec6d1ad57

91f2813c-449a-4673-be0f-1776ea50dc98

PR #13217, opened 2023-10-10
[maintenance] eventbus

GrzegorzBobryk:

### Contributor checklist
 * Pixel 3a, API 33

----------

### Description
Eventbus renovation:
- bump,
- logging exceptions only in debug,
- eventBusIndex generated from kapt.

It will add new fixes to this lib. Also it will use generated indexes instead of using reflection. In places with private access or in anonymous classes it will work in old way. Documentation for this feature [here](https://greenrobot.org/eventbus/documentation/subscriber-index/), if you want to check it out.


app/build.gradle
```diff
@@ -11,6 +11,7 @@ plugins {
     id 'com.squareup.wire'
     id 'translations'
     id 'licenses'
+    id 'kotlin-kapt'
```
alex-signal, line 14: Questioning whether we shouldn't be using KSP https://developer.android.com/build/migrate-to-ksp

I'll bring this up with the team... I'm apprehensive of adding kapt.

alex-signal (commented):
I'm going to run this by the team, because bringing in an annotation processor is setting off a flag in my brain.

alex-signal:
Pending any issues, this should go out in `6.37.x`

GrzegorzBobryk:
> Pending any issues, this should go out in `6.37.x`

So anything to fix in this pr? I see that version `6.38.2` is released.

greyson-signal:
Unfortunately while this build is functionally fine, it broke our reproducible builds. We're not sure why, but we're sure this change was it. Must be something non-deterministic with how kapt is used?

GrzegorzBobryk:
Could I read some logs or anything?

greyson-signal:
It's not like a functional thing where there's a bug with logs. The issue is that if you build two copies of the app, they are not identical. The dex's don't match. Our reproducible build instructions are [here](https://github.com/signalapp/Signal-Android/tree/main/reproducible-builds).

You can test it yourself by using our docker image (which is what we use to build release version of the app) to build multiple copies, and then you can compare them with the python script and it'll present you with the data that doesn't match. In fact, that very process is what we have run in a nightly cron job, and that's how we knew it broke :)

Issue #12904, opened 2023-04-16
None stop "Checking for messages..." notification 


h-n-a:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
This issue has been reported 2 years ago and it started happening again 2 releases back:
https://github.com/signalapp/Signal-Android/issues/11071

I get bombarded none stop while listening to music, browsing the internet, typing a note. basically, anything.

### Steps to reproduce
The issue happens after upgrading to the latest signal app (since the last 2-3 releases)
No specific steps to reproduce it.

### Screenshots

![signalapp](https://user-images.githubusercontent.com/126036770/232262685-d0f5075c-2e1b-4121-8c3c-6a3872364dfa.PNG)

### Device info

**Device:** Oneplus Nord N10 5G (BE2029)
**Android version:** 11.0.12.BE86AA
**Signal version:** 6.17.3

### Link to debug log

[signal-log-1681608303986.zip](https://github.com/signalapp/Signal-Android/files/11241076/signal-log-1681608303986.zip)

h-n-a:
bueller?

h-n-a:
@cody-signal only that message. sometimes it's about an actual notification. like someone adding emoji to a comment or just "checking for messages..." without an actual new message or interaction.

it's very inconsistent and becoming a nuisance.

cody-signal:
> it's very inconsistent and becoming a nuisance.

I feel ya, unfortunately this is currently by design to ensure proper message delivery. Android requires us to post a notification in order to get foreground execution time when we receive a push notification.

However, you should almost always see a message notification afterwards unless most of the messages are in muted/blocked chats.

h-n-a:
lol, what? why did this only happen recently?
If you don't want to address the issue. that's fine. but this is something recent, not "by design". 
Your team forces users to update the software from working versions to untested code just to bloat the app with unnecessary features. trying to make it like whatsapp.

neirbowj:
@cody-signal said,
> Android requires us to post a notification in order to get foreground execution time when we receive a push notification.

Ok, so why does the notification persist for minutes at a time? ... while another person on an iPhone has already received the common message? What are you doing during that "foreground execution time"? Mining cryptocurrency!?

[...]

PR #13218, opened 2023-10-11
Use timestamp of last fetch instead of counting them

p1gp1g:

### Contributor checklist

 * Device Pixel 6A, Android 14 (GOS)

----------

### Description

Before, `activeCount` was used to know whether a new fetch was enqueued while the messages were being retrieved. But if somehow, a first fetch is being idle or the count is not correctly decreased, the service is not properly stopped. Previous fetches should be ignored when one succeed.

Previous behavior (activeCount) may result in :
- the "Checking for messages..." notification lasts for a long time (maybe part of #12904)
- Increased battery drain (useless foreground service)

This is part of the log I've seen when the foreground service's notification was lasting

```
Incrementing active count to 1
No more active. Stopping.
Incrementing active count to 1
No more active. Stopping.
Incrementing active count to 1
Incrementing active count to 2
Incrementing active count to 2
Incrementing active count to 2
Incrementing active count to 2
Incrementing active count to 2
```

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/gcm/FcmFetchManager.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/gcm/FcmFetchManager.kt
@@ -48,7 +48,7 @@ object FcmFetchManager {
   val WEBSOCKET_DRAIN_TIMEOUT = 5.minutes.inWholeMilliseconds
 
   @Volatile
-  private var activeCount = 0
+  private var last = 0L
 
   @Volatile
   private var highPriority = false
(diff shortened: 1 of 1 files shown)


p1gp1g:
I've updated the commit to use System.nanotime() which is monotonic (since device's boot).

Also, activeCount seems to be a remnant of the old REST strategy to fetch messages.

cody-signal:
Thanks for the contribution @p1gp1g . Can you share the full debuglog for the snippet in the description?

p1gp1g:
I unfortunately didn't save it. I'll try to save it next time it appears.

Nevertheless, even without logs, it makes more sense to rely on the last event timestamp than counting pending requests since this commit : https://github.com/signalapp/Signal-Android/commit/33828439fbe95bcda3a1cf058ce4093058d41b4e

----

Also, we should probably change this `if (hasHighPriorityContext)` to `if (hasHighPriorityContext && last <= now)` : https://github.com/signalapp/Signal-Android/blob/6d7d5280d6996a5b41f2801d890e1699275faff0/app/src/main/java/org/thoughtcrime/securesms/gcm/FcmFetchManager.kt#L105 

It would avoid some false positive where the message has been fetch with a more recent request. I'll open a new PR for this later if you want to

PR #13357, opened 2024-01-10
Improved  function performance in base83

MahdiRahmani80:

### Contributor checklist

 * Device A, Android X.Y.Z
 * Device B, Android Z.Y

----------

### Description

The time complexity of linear search is O(n), whereas binary search operates at O(log n). This means that in the worst-case scenario, the 'indexOf' function should perform (83+1) checks for a linear search, while a binary search would only take floor(log(83) + 1) cycles, resulting just 7 cycles at worst.

However, one challenge I encountered was related to modifying the 'ALPHABET' array, primarily due to manually sorting it using ASCII codes. Upon initial assessment, it seems acceptable, but further scrutiny might be necessary.

References:

ASCII Code: https://www.ascii-code.com/
Time Complexity: https://en.wikipedia.org/wiki/Time_complexity
Binary Search Algorithm: https://en.wikipedia.org/wiki/Binary_search_algorithm

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/blurhash/Base83.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/blurhash/Base83.java
@@ -1,17 +1,17 @@
 /**
  * Source: https://github.com/hsch/blurhash-java
- *
+ * <p>
  * Copyright (c) 2019 Hendrik Schnepel
- *
+ * <p>
  * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
  * associated documentation files (the "Software"), to deal in the Software without restriction,
  * including without limitation the rights to use, copy, modify, merge, publish, distribute,
  * sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is
  * furnished to do so, subject to the following conditions:
- *
+ * <p>
  * The above copyright notice and this permission notice shall be included in all copies or
  * substantial portions of the Software.
- *
+ * <p>
  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT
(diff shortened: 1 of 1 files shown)


alex-signal:
Could you please perform some testing say, between Android and Desktop or Android and iOS and confirm the blurhashes appear to be appropriate?

MahdiRahmani80:
## Screenshots

"  **Android && Desktop**  "
<p>



<br><br>

</p>
<br>
<hr></hr>

### Description 
I'm a newbie to the Signal codebase, and I'm unsure where to find certain elements, but I've noticed that these pictures might be related to the `Base83`.


**Device Information:**
- Device: Xiaomi M2010J19CI
- Android Version: Android 12
- API 31
- Emulator/Physical Device: Physical Device

alex-signal:
Hi, I'm not sure you understood what I asked.

BlurHashes are used in the following case:

Alice sends Bob an Image
The client is going to:
1. Upload the image to a CDN
2. Send Bob a DataMessage denoting where the image is
Inside the DataMessage is a BlurHash, which is calculated by Alice's device using Base83

Bob opens his chat with Alice, and the image begins downloading.
While the image is downloading, the app will use the BlurHash to generate a temporary 20x20pix bitmap that is a blurred represenation of that image.

It's important that the blurhash looks identical on all devices.
My recommendation is to create a group with a single user (yourself) and start a group call, connecting both your desktop and phone. (Be sure to mute your mics!)

Then, send an image from Android to the group. Since a call is in progress, the image will not be autodownloaded, so you should just see the blur and a download icon overlay.

Then, send the same image from Desktop to Android. The blur on Android should look the same as the blur on desktop.

Furthermore, have you run any tests to suggest that the current path is actually negatively impacting performance? What meansurements have you done to suggest that this is a necessary improvement? We generally don't like to rewrite things like this if it isn't causing an issue.

PR #13367, opened 2024-01-17
Added the String Resource Values for the Hardcoded texts.

Ansalpandey:
This PR diligently replaces hardcoded text with String Resource Values.


cody-signal:
Thanks for your time. Two things of note:

1. When creating an issue or PR please fill out the template, do not simply delete it.
2. We do not translate strings using in our test apps so we don't bother externalizing strings. It's unnecessary for test tools and so won't be accepting this PR.

PR #13391, opened 2024-01-28
Cleanup root build.gradle.kts. Remove duplicate dependency

maksym-moroz:

### Contributor checklist

 * Samsung Note 10+, Android 12

----------

### Description

- Register tasks instead of creating them eagerly to improve configuration times 
- Use `onlyIf` predicate block to run `qa` task conditionally
- Remove duplicate dependency on `kotlinx.collections.immutable`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -50,7 +50,7 @@ allprojects {
 }
 
 subprojects {
-  if (JavaVersion.current().isJava8Compatible()) {
+  if (JavaVersion.current().isJava8Compatible) {
     allprojects {
       tasks.withType<Javadoc> {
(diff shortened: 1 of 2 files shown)


greyson-signal:
Mostly good! We can't use `onlyIf` -- seems like the project fails to configure. I'm no gradle expert, but the documentation suggests that it's likely because `onlyIf` prevents execution of the task, but the error itself is happening during configuration because it can't find the lint task on the` benchmark` project. So we need to prevent the task being registered at all if those conditions aren't met.

So I reverted that little bit, but the rest will be in the 6.46 release. Thank you!

maksym-moroz:
Makes sense. I will look into `onlyIf` in more detail later. Thanks for swift response!

Issue #13017, opened 2023-06-21
Video recording on Google Pixel devices uses the wrong camera


tedgravlin:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
When recording video in Signal on a Google Pixel 7a, the wide camera is used instead of the standard camera.

### Steps to reproduce
- Using a Google Pixel device with multiple rear cameras, open the camera in Signal
- Record a video

**Actual result:** Signal uses the phone's wide camera to record video.
**Expected result:** 
Signal uses the standard camera to record video.

Note: 
* The correct camera is used when taking photos. 
* This issue has also been reported by others using Google Pixel phones: https://community.signalusers.org/t/signal-uses-the-wrong-camera-for-video-recording/29241

### Device info

**Device:** Google Pixel 7a
**Android version:** Android 13
**Signal version:** 6.24.1

### Link to debug log

https://debuglogs.org/android/6.24.1/4d302dfb5796e1ad9e9b0f9e244952b6440fcacdd95ce49522e1891670388303

Herohtar:
Just to add some detail, it seems that the wide angle camera is treated as part of the zoom level -- that is, if you drag your finger up while recording, Signal will zoom in and eventually switch to the regular/main camera and continue zooming in. You can zoom back out and it will switch back to the wide camera at the lower zoom levels.

This results in a really bad video experience because the amount of movement required to switch between zoom levels is pretty small and the transition between wide and regular is not smooth.

Furthermore, the preview only uses the regular camera, so it is quite unexpected that you suddenly jump into wide mode when you start recording.

Herohtar:
Still an issue. I think maybe this could be partially resolved by initializing the camera to the correct zoom level so that it doesn't go to wide angle. Assuming such a thing is possible, though it seems likely since Android treats the wide angle as something like 0.9x zoom, so maybe just need to make sure it starts at 1x?

That won't fix the issue with the zoom being really sensitive though.

tedgravlin:
I partially addressed this issue in #13411 but I still don't have it working 100%. If anyone is familiar with CameraX take a look and see if you can figure it out. Thanks :)

tedgravlin:
Fixed with 6a65a1c!

PR #13411, opened 2024-02-09
Fix default video record zoom level (Assistance needed)

tedgravlin:
### Contributor checklist

 * Pixel 7a, Android 14

----------

### Description

This addresses #13017 

When recording video, the zoom ratio is set to minZoomRatio by default. On devices with wide angle cameras, the minZoomRatio is actually less than 1.0x. This means that if a device has a wide angle camera, then Signal will start recording with it instead of the main camera. This change ensures that when recording starts, the zoom level is always at 1.0x.

However, the behavior is still broken when zooming and I haven't been able to figure out how to fix it. The recording now properly starts at 1.0x, but when you start to zoom it jumps back to 0.5x. 

If I change `cameraController.setZoomRatio((range * increment) + zoomState.getMinZoomRatio());` in `onZoomIncremented()` to `cameraController.setZoomRatio((range * increment) + 1.0f);` then zooming works smoothly, but you can no longer zoom out to anything below 1.0x.

I'm not an Android dev, so any help would be greatly appreciated! Thanks :)

---

Before: Recording starts at 0.5x

https://github.com/signalapp/Signal-Android/assets/123981212/df1d005e-dda5-4e37-a3a3-bf18b9aaae65

---

After: Recording starts at 1.0x, but zoom starts at 0.5x

https://github.com/signalapp/Signal-Android/assets/123981212/745018af-58ee-484f-b523-e3e57615a7fd

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/mediasend/CameraXVideoCaptureHelper.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/mediasend/CameraXVideoCaptureHelper.java
@@ -74,7 +74,7 @@ class CameraXVideoCaptureHelper implements CameraButtonView.VideoCaptureListener
         } else {
           try {
             debouncer.clear();
-            cameraController.setZoomRatio(Objects.requireNonNull(cameraController.getZoomState().getValue()).getMinZoomRatio());
+            cameraController.setZoomRatio(1.0f);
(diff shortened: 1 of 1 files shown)


tedgravlin:
This was essentially closed with 6a65a1c. Thanks for the shout-out! I'm glad I could help solve this issue!

PR #13424, opened 2024-02-14
present perfect for some English strings

lpirl:
What would you think of some more present perfect for some English strings where the focus is on the current situation, rather than the past? Simple past triggered me especially in "message was deleted" because I suppose Signal is actually trying to tell users that the message *is deleted*.

Feel free to add comments in the review where my changes are inappropriate.

If this PR is welcome, I'd port it to iOS and desktop.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -320,7 +320,7 @@
     <string name="ConversationItem_read_more">&#160; Read More</string>
     <string name="ConversationItem_download_more">&#160; Download More</string>
     <string name="ConversationItem_pending">&#160; Pending</string>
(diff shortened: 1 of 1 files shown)


lpirl:
I am happy to apply some changes, if desired. :)

alex-signal:
Hi! I'll bring this up with our localisation manager.

lpirl:
That is great, thank you @alexr-signal .

lpirl:
still waiting for response from localisation manager, happy if this would not be closed

greyson-signal:
Hi, I think we'll be keeping the strings as they are, but thank you!

lpirl:
Sad, because IMHO, present perfect would be more correct in some cases.
Anyway, thanks a lot for considering.

PR #13461, opened 2024-03-05
Fix racey dialog dismissal in `PlacePickerActivity`

fynngodau:

### Contributor checklist

 * OnePlus Nord / avicii, Android 12

----------

### Description

The first commit avoids a race condition with closing the location picker map for location sharing where the following exception would occur:

    IllegalArgumentException: View[…] not attached to window manager

This was because `dismiss`, together with its delay, could be executed after the activity that contained the dialog, `PlacePickerActivity`, had already been destroyed. This faulty behavior can be reproduced more easily when artificially increasing the minimum time that the dialog should be shown. Instead of potentially scheduling a dismissal for the future using `dismiss`, `dismissNow` is now called during `onDestroy`.

The second commit avoids similar problems in the future, because it ensures that the delayed dismiss callback is always cancelled when `dismissNow` is called (which it should be whenever the activity containing the progress bar is closed for some reason, which may be in between a `dismiss` call and the `dismiss`-scheduled delayed dismissal.

The second commit could be left out as the first one already fixes the crash I mentioned above, but I would recommend to add it nonetheless.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/components/settings/conversation/ConversationSettingsFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/settings/conversation/ConversationSettingsFragment.kt
@@ -187,7 +187,7 @@ class ConversationSettingsFragment : DSLSettingsFragment(
     when (requestCode) {
       REQUEST_CODE_ADD_MEMBERS_TO_GROUP -> if (data != null) {
         val selected: List<RecipientId> = requireNotNull(data.getParcelableArrayListExtraCompat(PushContactSelectionActivity.KEY_SELECTED_RECIPIENTS, RecipientId::class.java))
-        val progress: SimpleProgressDialog.DismissibleDialog = SimpleProgressDialog.showDelayed(requireContext(), this)
+        val progress: SimpleProgressDialog.DismissibleDialog = SimpleProgressDialog.showDelayed(requireContext(), viewLifecycleOwner)
 
         viewModel.onAddToGroupComplete(selected) {
           progress.dismiss()
--- a/app/src/main/java/org/thoughtcrime/securesms/components/settings/conversation/ConversationSettingsFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/settings/conversation/ConversationSettingsFragment.kt
@@ -187,7 +187,7 @@ class ConversationSettingsFragment : DSLSettingsFragment(
     when (requestCode) {
       REQUEST_CODE_ADD_MEMBERS_TO_GROUP -> if (data != null) {
         val selected: List<RecipientId> = requireNotNull(data.getParcelableArrayListExtraCompat(PushContactSelectionActivity.KEY_SELECTED_RECIPIENTS, RecipientId::class.java))
-        val progress: SimpleProgressDialog.DismissibleDialog = SimpleProgressDialog.showDelayed(requireContext())
+        val progress: SimpleProgressDialog.DismissibleDialog = SimpleProgressDialog.showDelayed(requireContext(), this)
 
         viewModel.onAddToGroupComplete(selected) {
           progress.dismiss()
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/mutiselect/forward/MultiselectForwardFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/mutiselect/forward/MultiselectForwardFragment.kt
@@ -247,7 +247,7 @@ class MultiselectForwardFragment :
         MultiselectForwardState.Stage.SendPending -> {
           handler?.removeCallbacksAndMessages(null)
           dismissibleDialog?.dismiss()
-          dismissibleDialog = SimpleProgressDialog.showDelayed(requireContext(), this)
+          dismissibleDialog = SimpleProgressDialog.showDelayed(requireContext(), viewLifecycleOwner)
         }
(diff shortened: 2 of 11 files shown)


app/src/main/java/org/thoughtcrime/securesms/maps/PlacePickerActivity.java
```diff
@@ -240,6 +240,13 @@ protected void onPause() {
     }
   }
 
+  @Override protected void onDestroy() {
```
alex-signal, line 243: Curious, could we pass a LifecycleOwner to showDelayed instead of writing out this custom destroy override?

That'd let us encapsulate that dismiss now logic inside SimpleProgressDialog, wouldn't require a new field, and be less error prone in the future (for other uses)

Something like

```
// Pass @NonNull LifecycleOwner owner (in the case of this PR, it's the activity)

DismissableDialog dialog = /* what we currently return */

LifecycleObserver observer = new DefaultLifecycleObserver {
  onDestroy(owner) {
    // dismiss dialog logic
  }
}

lifecycleOwner.registerObserver(observer)

return dialog;
```

fynngodau: Many thanks for the suggestion, @alex-signal! I've added a commit that implements your proposal, and in the process, also found some ways to make the `SimpleProgressDialog` code simpler – basically now `tryDismiss` is the runnable that dismisses the dialog if it is not already dismissed, but is allowed to be called multiple times (from the lifecycle observer, and from `dismiss` and `dismissNow`).

app/src/main/java/org/thoughtcrime/securesms/components/settings/conversation/ConversationSettingsFragment.kt
```diff
@@ -185,7 +185,7 @@ class ConversationSettingsFragment : DSLSettingsFragment(
     when (requestCode) {
       REQUEST_CODE_ADD_MEMBERS_TO_GROUP -> if (data != null) {
         val selected: List<RecipientId> = requireNotNull(data.getParcelableArrayListExtraCompat(PushContactSelectionActivity.KEY_SELECTED_RECIPIENTS, RecipientId::class.java))
-        val progress: SimpleProgressDialog.DismissibleDialog = SimpleProgressDialog.showDelayed(requireContext())
+        val progress: SimpleProgressDialog.DismissibleDialog = SimpleProgressDialog.showDelayed(requireContext(), this)
```
alex-signal, line 188: Use `getViewLifecycleOwner()` instead of `this` in fragments.

app/src/main/java/org/thoughtcrime/securesms/conversation/mutiselect/forward/MultiselectForwardFragment.kt
```diff
@@ -249,7 +249,7 @@ class MultiselectForwardFragment :
         MultiselectForwardState.Stage.SendPending -> {
           handler?.removeCallbacksAndMessages(null)
           dismissibleDialog?.dismiss()
-          dismissibleDialog = SimpleProgressDialog.showDelayed(requireContext())
+          dismissibleDialog = SimpleProgressDialog.showDelayed(requireContext(), this)
```
alex-signal, line 252: Use `getViewLifecycleOwner()` instead of `this` in fragments.

app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
```diff
@@ -1279,7 +1279,7 @@ private void handleUnmute(@NonNull Collection<Conversation> conversations) {
   }
 
   private void updateMute(@NonNull Collection<Conversation> conversations, long until) {
-    SimpleProgressDialog.DismissibleDialog dialog = SimpleProgressDialog.showDelayed(requireContext(), 250, 250);
+    SimpleProgressDialog.DismissibleDialog dialog = SimpleProgressDialog.showDelayed(requireContext(), this, 250, 250);
```
alex-signal, line 1282: Use `getViewLifecycleOwner()` instead of `this` in fragments.

app/src/main/java/org/thoughtcrime/securesms/groups/ui/invitesandrequests/invite/GroupLinkInviteFriendsBottomSheetDialogFragment.java
```diff
@@ -154,7 +154,7 @@ protected GroupId.V2 getGroupId() {
   private void setBusy(boolean isBusy) {
     if (isBusy) {
       if (busyDialog == null) {
-        busyDialog = SimpleProgressDialog.showDelayed(requireContext());
+        busyDialog = SimpleProgressDialog.showDelayed(requireContext(), this);
```
alex-signal, line 157: Use `getViewLifecycleOwner()` instead of `this` in fragments.

fynngodau:
@alex-signal I added another commit with your requested changes.

fynngodau:
Still current, please reopen.

fynngodau:
In the meantime, @greyson-signal fixed the crash that this PR also fixes with c6f4a010014ce04d143ba59f37e83eca42350658, but the dialog view still leaks. With the implementation provided in this PR, the view is not leaked as the dialog is destroyed as soon as the lifecycle owner that it is attached to is destroyed.

Therefore I consider reversing c6f4a010014ce04d143ba59f37e83eca42350658 and applying the changes from this PR to be an improvement; I updated the branch to reflect these suggested changes.

Issue #13572, opened 2024-05-23
Reduced camera resolution, forced portrait mode


timpoiko:
- [X] I have searched open and closed issues for duplicates
- [X] I am submitting a bug report for existing functionality that does not work as intended
- [X] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [X] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
I have a Cosmo Communicator from Planet Computers. It's cameras resolution is 5632x4224. After a forced update of Signal, camera's mode is forced to portrait instead of its natural landscape mode.

### Steps to reproduce
- Open Signal App
- Try to take image and send

**Actual result:** Small portrait mode picture in middle of screen, wasting camera's resolution.

**Expected result:** Signal uses full capacity of my device's camera

### Screenshots
![signal-screenshot](https://github.com/signalapp/Signal-Android/assets/86437306/c9be1e38-b842-4604-80ee-63f868b22eb1)

### Device info

**Device:** Cosmo Communicator  (Manufacturer: Planet Computers)
**Android version:** 9, API28
**Signal version:** 7.6.2

alex-signal:
Hello! Could you please include a debug log? That activity should be fixed to portrait only 🤔 the image will automatically be properly oriented after capturing a photo.

PR #13591, opened 2024-06-06
(#13572) Resolving this issue

sounar97:
Where it does'nt have to shift to portrait mode when camera opens


greyson-signal:
Hi there, we've changed how rotation is handled in calls, but thank you!

PR #13746, opened 2024-10-26
intpref: remove force emoji and search index download summaries unnecessary back slash symbol

rottenwheel:
Note: original file contains two (2) \\ yet last release only shows one (1) in UI; I am removing both with this PR, if at least one is needed, let me know so I can adjust properly, otherwise, feel free to merge as this seems to be just a UI thing, doesn't touch any actual code.


greyson-signal:
Hey there! Don't worry, you still have authorship :) As this screenshot says, you authored the commit, but I committed it.
image

We prepare releases in private branches. In order to make sure the commit links back to the PR, I put "Resolves #<PR number>" in the commit body which links the commit back and auto-closes the issue.

PR #13791, opened 2024-11-16
Remove unreachable dependencies from verification-metadata.xml

jamesonwilliams:
### Contributor checklist

 * Build only change

----------

### Description

Currently the `verification-metadata.xml` drifts out of date over time as people update dependencies but forget to clean out unused stuff from this XML file. This PR just does some routine manual maintenance.

It can be updated ad-hoc this way:

1. Clearing the `<components>` entries from the file:
```bash
perl -i -0777 -pe 's/<components>.*?<\/components>//gs'  gradle/verification-metadata.xml
```
2. And then regenerating them:
```bash
./gradlew --write-verification-metadata sha256 qa --rerun-tasks
```

This might be a nice thing to validate on every PR android in `android.yml`?


jamesonwilliams:
> Also, some of the dependencies are actually OS-specific, so we likely wouldn't get an exact match when running on a linux box.

Ahh, shoot. I wonder if this PR will pass the CI checks In that case 🤔  I did only run it on a Mac

PR #13808, opened 2024-11-21
Import correct InvalidKeyException in LinkDeviceRepository

valldrac:
### Description

Updated the import statement to catch the correct `InvalidKeyException` in `Curve.decodePoint` call.


greyson-signal:
Ah, thanks for catching this!

PR #13809, opened 2024-11-21
Fix creating message_fts table during database migration.

bepaald:
### First time contributor checklist

- [x] I have read [how to contribute](https://github.com/signalapp/Signal-Android/blob/main/CONTRIBUTING.md) to this project
- [x] I have signed the [Contributor License Agreement](https://signal.org/cla/)

### Contributor checklist

- [x] I am following the [Code Style Guidelines](https://github.com/signalapp/Signal-Android/wiki/Code-Style-Guidelines)
- [x] I have tested my contribution on these devices:
 * ZTE ZMax Pro (Z981), Android 6.0.1
 * Custom software
- [x] My contribution is fully baked and ready to be merged as is
- [x] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the `Fixes #1234` [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)

----------

### Description

In database migration [175](https://github.com/signalapp/Signal-Android/blob/main/app/src/main/java/org/thoughtcrime/securesms/database/helpers/migration/V175_FixFullTextSearchLink.kt), the full text search table is dropped and recreated with a new name (`mms` -> `message`). First the old table and triggers are dropped:
https://github.com/signalapp/Signal-Android/blob/49eb80b4404762ec509925cc02dda2e67bcd3468/app/src/main/java/org/thoughtcrime/securesms/database/helpers/migration/V175_FixFullTextSearchLink.kt#L14-L17

Then the new one is created:
https://github.com/signalapp/Signal-Android/blob/49eb80b4404762ec509925cc02dda2e67bcd3468/app/src/main/java/org/thoughtcrime/securesms/database/helpers/migration/V175_FixFullTextSearchLink.kt#L24

Because no data is actually migrated from the old to the new search table (the old one doesn't even exist anymore), the new one is empty. However, the content table is (likely) not: `content=message`. This leaves the new table in an inconsistent state.

[...]

Issue #13506, opened 2024-04-07
Signal crashes on restoring backup


asdfjkl:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description

Signal crashes on restoring a (valid) backup file. In particular when starting restoring from a file backup, messages seems to be processed and imported correctly. However afterwards, Signal crashes repeatedly on startup and cannot be started anymore, i.e. Android reports that the app is crashing repeatedly responding.

I've verified that the backup is intact with 3rd party tools (signal-backup-tools). I've also verified that the passphrase is correct; in particular I was able to extract all messages with signal-backup-tools (https://github.com/bepaald/signalbackup-tools) to text format and verified that all information is present. I've also tried to (re)write the backup using backup-tools in order to check that the backup file is not corrupted.

The database version as reported from signal-backup-tools: 136
Signal-Version: 7.2.4

The backup is from 2022; so I suspect that there is some issues when handling such "old" backups.
The backup contains personal information, but I am willing to help w.r.t. debugging this issue in any way I can

### Steps to reproduce
- Install Signal from scratch from the Play Store
- Choose restore account from file backup
- Messages are processed, but afterwards Signal won't start anymore. Only wiping/re-installing Signal helkps

**Actual result:** Signal crashes and can't be started anymore
**Expected result:** Should restore the backup and start as expected

### Screenshots
unfortunately I have no screenshot

### Device info
**Device:** Pixel 6a
**Android version:** 14
**Signal version:** 7.2.4

### Link to debug log
[signal-log-1712506609190.zip](https://github.com/signalapp/Signal-Android/files/14898681/signal-log-1712506609190.zip)
attached to this report

asdfjkl:
any news on this? I think this should not be closed, as there is apparently an issue with old backups, likely affecting other users as well.

itsluminous:
Hey I am also having this issue.
Device : Samsung Galaxy S21 FE.
Backup Date : June 2022

I decided to give Signal one more try after a couple of years, and now restore won't work.
Immediately after restore the app crashes.
Tried restoring different backups I had, same result.

itsluminous:
[...]

PR #13810, opened 2024-11-21
More fts table fixes

bepaald:

### Contributor checklist

----------

### Description

This is strongly related to #13809. The exact same FTS-table-drop-and-recreate pattern is used in these two migrations (239, and 242). These will certainly again cause database corruptions. For more information, please see #13809.

I am not currently aware of open issues caused by this problem (I only found them now just by `grep`ping for "message_fts" through the migration-helpers), and I have not tested these two 1-liners, but both the problem and solution are identical to #13809 (which was thoroughly investigated and tested).

If i'm understanding everything right, some people may already have inconsistent FTS tables due to these migrations, so I strongly suggest a new database version and a migration like this:
```
object V2XX_EnsureSearchTableConsistency : SignalDatabaseMigration {

  private const val FTS_TABLE_NAME = "message_fts"
  db.execSQL("INSERT INTO $FTS_TABLE_NAME ($FTS_TABLE_NAME) VALUES('rebuild')")

}
```
I'd be willing to write a patch like this if requested. Then after that, it should just be a matter of _keeping_ the table consistent.

Thanks!

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/database/helpers/migration/V239_MessageFullTextSearchEmojiSupport.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/database/helpers/migration/V239_MessageFullTextSearchEmojiSupport.kt
@@ -43,6 +43,8 @@ object V239_MessageFullTextSearchEmojiSupport : SignalDatabaseMigration {
 
     db.execSQL("""CREATE VIRTUAL TABLE message_fts USING fts5(body, thread_id UNINDEXED, content=message, content_rowid=_id, tokenize = "unicode61 categories 'L* N* Co Sc So'")""")
 
+    db.execSQL("INSERT INTO message_fts(message_fts) VALUES ('rebuild')")
+    
     db.execSQL(
       """
       CREATE TRIGGER message_ai AFTER INSERT ON message BEGIN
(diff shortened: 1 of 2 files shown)


greyson-signal:
I think these changes are reasonable, thank you! They'll be in 7.27.

I wanna hold off on a new migration that just does a rebuild, though. On large installs this can take 7-10 seconds and can be very disruptive, so I want to take more of a "wait and see" approach.

bepaald:
Thanks!

In regards to a new migration: it is possible to do the actual rebuild conditionally on whether or not the fts table is good. For this use `INSERT INTO message_fts(message_fts, rank) VALUES('integrity-check', 1);` and check the return code. (reference: https://www.sqlite.org/fts5.html#the_integrity_check_command). This way there is no penalty for users if the fts table is good (most users I would think). For others, indeed it could take some time, but it could be argued that it _needs_ to be done in their case regardless.

Note also that if it's not done, similar problems like #13034 and #13506 could easily happen again if any future migration touches the `message` table again. (at least assuming the separate job that exists for rebuilding does not run _during_ the migrations.)

But I'll leave that up to you (though I'm still willing to create a pr if desired). Thank you for all your hard work on this!

PR #13820, opened 2024-11-26
Inject SignalWebSocket into IncomingMessageObserver

valldrac:
### Description

This PR updates `IncomingMessageObserver` to accept `SignalWebSocket` as a constructor parameter. This change ensures each IncomingMessageObserver is explicitly tied to the SignalWebSocket instance owned by its `NetworkDependenciesModule`, resolving a race condition that could occur during a network reset.

Before this, `IncomingMessageObserver` got its `SignalWebSocket` from `AppDependencies`. During network resets (`AppDependencies.resetNetwork()`), NetworkDependenciesModule and its SignalWebSocket were recreated. If the old IncomingMessageObserver was still terminating asynchronously (`terminateAsync()`), it could accidentally reference and manipulate the new SignalWebSocket. So when `disconnect()` is called, it may disconnect the current SignalWebSocket, even if it is intended to disconnect the previous one. This led to issues like the old observer disconnecting the new socket, and the old socket remaining connected, appearing as if two observers were running simultaneously in the logs.

By directly injecting the corresponding SignalWebSocket, we eliminate the shared dependency and resolve the race condition.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/dependencies/AppDependencies.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/dependencies/AppDependencies.kt
@@ -339,7 +339,7 @@ object AppDependencies {
     fun provideMegaphoneRepository(): MegaphoneRepository
     fun provideEarlyMessageCache(): EarlyMessageCache
     fun provideMessageNotifier(): MessageNotifier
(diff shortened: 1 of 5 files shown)


greyson-signal:
This seems very reasonable! I'll run it on my phone for a bit, and if all is well we'll get it into the next release :)

PR #13890, opened 2025-01-08
Create enabledAndNotLoading state and don't set enabled in parameter

Sagar0-0:

### Contributor checklist

 * Device Realme GT NET 3T, Android 14
 * Device Infinix Hot 20 Pro, Android 14
 * Virtual device Pixel 8 Pro, Android 14

----------

### Description

An important thing to note here is,
NEVER set a param default value to another parameter(If they are interdependent).
This can lead to unexpected behavior from the call site.

For example in this case:
If the TabRow is also dependent on any other condition to enable the Toggle, and it is set to true at the call site. Then the default value of `enabled = !isLoading` will be overridden to always true, which leads to inconsistent enabled state behavior for all elements.

So, a better way is to keep the default values in params as it is, and then abstract out the logic for handling the "interdependent" states as I did here with `enabledAndNotLoading`. And now the component's developer is assured to have the states consistent.

Let's make sure other places are also following the same pattern.

### DIFF
--- a/core-ui/src/main/java/org/signal/core/ui/Rows.kt
+++ b/core-ui/src/main/java/org/signal/core/ui/Rows.kt
@@ -27,6 +27,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment.Companion.CenterVertically
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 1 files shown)


greyson-signal:
Just checking, is this fixing a specific bug?

Sagar0-0:
There is no existing bug.

However there can be some uses in future which can produce bugs with the existing implementation. I hope I was able to clearly explain the problem with current implementation.

greyson-signal:
I don't disagree that having the fields be co-dependent creates the potential for odd results, but your solution does the same -- it creates a situation where a user has explicitly passed enabled = true, but the field is not, in fact, enabled. 

Turns out no one has actually used the explicit `enabled` field, so I've simply removed it in 7.30. Thanks!

Sagar0-0:
@greyson-signal Removing the param works for now. But, The default values of `isLoading` as `false` and `enabled` as `true` work for all standard use cases where we have the disable making changes when the state is loading.

Considering the case where we have to explicitly depend on the `enabled` state and not disable the items when `isLoading` is true. (Which is not covered by my implementation).

The solution is to have two different implementations of these behaviors where in one, `isLoading` affects the enabled state and not in the other. But, I think not to complicate things, a flexible solution is to add the param of `shouldLoadingDisableItems`. 

I agree this might be overengineering for now, but if we have to add this in the future, what would you choose? Just wanted to know your perspective. Thanks

greyson-signal:
In general, if a function takes two independent booleans that can produce invalid combinations, your options are usually to either (1) accept one of the invalid combinations, (2) assert that the user didn't pick one of the invalid combinations, or (3) switch to using an enum that only includes the valid states. One could prefer (1) or (2) if it's expected that callers wouldn't ever make the mistake in practice, but (3) is the most robust. Which I choose in this case would depend on what the first usecase was for setting them independently -- right now that case doesn't exist :)

PR #13918, opened 2025-01-17
Remove Hamcrest from project files

jamesonwilliams:
### Contributor checklist

 * MacBook

----------

### Description

We recently removed references to Hamcrest in the test files. This is the last bit of cleanup to remove it from the project files.

Hamcrest is still brought into this project as a transitive dependency of JUnit 4.x. So I've retained the licensing notice since it is in the classpath closure. But, it is no longer in any `build.gradle`.

To update the `verification-metadata.xml`, I did 
```bash
perl -0777 -pi -e 's|<components>.*?</components>|<components>\n</components>|gs' gradle/verification-metadata.xml 
./gradlew qa --write-verification-metadata sha256 
```

(And then added back the mysterious `junit-bom`...)


greyson-signal:
Awesome, thank you! That `verification-metadata.xml` update script is also clever, I'll stash that away somewhere :) Only downside is that there's one dependency (aapt) that has specific osx/linux/windows variants, so gotta add those back (already did it), but that's not that bad.

PR #13920, opened 2025-01-17
Show message while Adding a Already Invited member

Sagar0-0:

### Contributor checklist

 * Device A, Android X.Y.Z
 * Device B, Android Z.Y

----------

### Description

Show toast if recipient is already invited to a specific group


greyson-signal:
I actually think there's a valid usecase where you would add someone who is pending. For instance, originally you might not have had the person's profile key, so they're pending, but then you get their key, and now you can add them. Interesting question, and one I'm not sure I know the answer to atm.

Sagar0-0:
@XSm2i the logic only applies to the same user who's already invited and tries to invite again. 

We can add an additional check "If the user is already invited && we still don't have profile key" then we can show the toast, else we can allow them to add.

PR #13948, opened 2025-01-28
Fix textStoryToggle visibility when hasSelectedMedia

Sagar0-0:

### Contributor checklist

 * Device A, Android X.Y.Z
 * Device B, Android Z.Y

----------

### Description

Before:

https://github.com/user-attachments/assets/7c63adce-f2b2-43ae-abc9-0bdccdda49c5

After:
Observing to the latest state of selectedMedia count and updating the visibility

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/mediasend/v2/MediaSelectionActivity.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/mediasend/v2/MediaSelectionActivity.kt
@@ -153,6 +153,14 @@ class MediaSelectionActivity :
       textViewModel.restoreFromInstanceState(savedInstanceState)
     }
 
+    viewModel.hasSelectedMedia.observe(this) { hasSelectedMedia->
+      if(hasSelectedMedia) {
+        textStoryToggle.visible = false
+      } else {
+        textStoryToggle.visible = canDisplayStorySwitch()
+      }
+    }
+
     (supportFragmentManager.findFragmentByTag(NAV_HOST_TAG) as NavHostFragment).navController.addOnDestinationChangedListener { _, d, _ ->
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/thoughtcrime/securesms/mediasend/v2/MediaSelectionActivity.kt
```diff
@@ -233,7 +241,6 @@ class MediaSelectionActivity :
   private fun canDisplayStorySwitch(): Boolean {
```
greyson-signal, line 241: This is called from multiple places, and now those call sites no longer take into account whether there's selected media. Are those call sites still necesessary?

greyson-signal (commented):
I'm interested in how you're reproducing this bug, it doesn't personally happen to me.

Sagar0-0:
> I'm interested in how you're reproducing this bug, it doesn't personally happen to me.

Capture a media, Click anywhere on the Preview screen and wait for 1-2 seconds, press back button. See video. This might not show sometimes

app/src/main/java/org/thoughtcrime/securesms/mediasend/v2/MediaSelectionActivity.kt
Sagar0-0: However, this might create a problem, as now there are two sources for textStoryToggle visibility state(hasSelectedMedia and DestinationChangedListener) and it might create difficulty in testing and finding bugs. A better approach can be to move the logic to ViewModel and check the combinations there. If destinations are mediaCapture or textStore then handle the logic, else false. This way we will only expose a single LiveData of isTextStoryToggleVisible as a single source of truth.

PR #13957, opened 2025-01-31
Support Scheduled Send for Recorded Audio

Sagar0-0:

### Contributor checklist

 * Device A, Android X.Y.Z
 * Device B, Android Z.Y

----------

### Description

1. Hold the mic and start record the audio. 
2., Lock the Recorder
3. Currently the user can only send the audio
4. After this change, the user can long press send and can schedule the audio.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
@@ -1830,7 +1830,8 @@ class ConversationFragment :
     slide: Slide? = null,
     contacts: List<Contact> = emptyList(),
     quote: QuoteModel? = null,
-    clearCompose: Boolean = true
+    clearCompose: Boolean = true,
+    scheduledDate: Long = -1,
   ) {
     sendMessage(
       slideDeck = slide?.let { SlideDeck().apply { addSlide(slide) } },
@@ -1842,7 +1843,8 @@ class ConversationFragment :
       messageToEdit = null,
       quote = quote,
       linkPreviews = emptyList(),
-      bypassPreSendSafetyNumberCheck = true
+      bypassPreSendSafetyNumberCheck = true,
+      scheduledDate = scheduledDate
     )
   }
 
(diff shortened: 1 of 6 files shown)


greyson-signal:
I like this idea, but I think the UX could be improved. As-written, the app is actually still recording while you're picking the scheduled time. That could be a non-trivial amount of awkward silence at the end :)

I think we could fix it though by stopping recording when the user long-presses the send button. When the long-press goes through, we could stop the recording at render it like we render voice note drafts (to see it, lock the recorder, then leave the chat and come back -- it's a special UI that shows a pending voice note). That way the recording is finalized when you're picking it.

app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
```diff
@@ -3956,6 +3963,10 @@ class ConversationFragment :
     }
 
     override fun onSendScheduled() {
+      if(inputPanel.isRecordingInLockedMode) {
```
Sagar0-0, line 3966: I could have used the `inputPanel.onPause()` call, which calls the `cancelAction` which then calls the onRecorderCanceled` with `byUser = false`, which also saves the audio draft and gives same results.

But, I decided to use new functions `inputPanel.onSaveRecordDraft`,  `onRecorderSaved`, etc.so that these will make sense with the call site and if anyone who have not worked in this flow can easily understand what is going on by just the function names.

What would you generally recommend? @greyson-signal 
Some code duplicacy but more sensible names. Like I have created `onSaveRecordDraft` which works exactly same as `onRecordCanceled(false)`
or
Less Easy to understand code with no duplicacy? Like calling the `inputPanel.onPause()` to save audio draft, or calling `onRecordCanceled(false)` to save audio draft.


I believe, a good code is self-explanatory with no need for documenting every function.

app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
```diff
@@ -3956,6 +3963,10 @@ class ConversationFragment :
...
+        inputPanel.onSaveRecordDraft()
```
Sagar0-0, line 3967: I could have used the `inputPanel.onPause()` call, which calls the `cancelAction` which then calls the `onRecorderCanceled` with `byUser = false`, which saves the audio draft and gives same results.

But, I decided to use new functions `inputPanel.onSaveRecordDraft`,  `onRecorderSaved`, etc. so that these will make sense with the call site and anyone who has not worked in this flow can easily understand what is going on by just the function names.

What would you generally recommend? @greyson-signal 
Some code duplicacy but more sensible names. Like I have created `onSaveRecordDraft` which works exactly same as `onRecordCanceled(false)`
**or**
Less Easy to understand code with no duplicacy? Like calling the `inputPanel.onPause()` to save audio draft, or calling `onRecordCanceled(false)` to save audio draft.


I believe, a good code is self-explanatory with no need for documenting every function.

greyson-signal (approved):
This is fantastic, thank you! It'll go out in 7.38

PR #13960, opened 2025-02-01
Highlight all the message search matches

abdullahemad12:
It is expected that when a user search for a message, that all the matches within the same messages are highlighted (if there are more than one match) not just the first match. This is also consistent with how the search in the desktop app is implemented

- [X] I have read [how to contribute](https://github.com/signalapp/Signal-Android/blob/main/CONTRIBUTING.md) to this project
- [X] I have signed the [Contributor License Agreement](https://signal.org/cla/)

### Contributor checklist

 * Pixel 2, API 35
 * Pixel tablet

----------

### Description

This commit changes the highlight strategy for message search from STRICT to MATCH_ALL. As a user, it is expected that if a message contains more than one occurrence of the key I am searching for, that all the found keys be highlighted not only the first one (which is indeed the behavior of the signal desktop app)

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/items/V2ConversationItemTextOnlyViewHolder.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/items/V2ConversationItemTextOnlyViewHolder.kt
@@ -406,7 +406,7 @@ open class V2ConversationItemTextOnlyViewHolder<Model : MappingModel<Model>>(
       linkifyMessageBody(styledText)
     }
 
-    styledText = SearchUtil.getHighlightedSpan(Locale.getDefault(), STYLE_FACTORY, styledText, conversationContext.searchQuery, SearchUtil.STRICT)
+    styledText = SearchUtil.getHighlightedSpan(Locale.getDefault(), STYLE_FACTORY, styledText, conversationContext.searchQuery, SearchUtil.MATCH_ALL)
     if (record.hasExtraText()) {
       binding.body.setOverflowText(getLongMessageSpan())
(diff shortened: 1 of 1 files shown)


ImAwsumm (approved):

abdullahemad12:
Hey @greyson-signal :) is there any thing I could do to move the merging of this PR forward?

Zyphrenn (approved):

gabefair:
I am curious about the value add of this feature for mobile. 
The search feature is to surface a text message. Once the search finds the first result search has fulfilled its job.
By adding the Match all option, would it increase search result's memory footprint? Cell phones have an average of less than 4gb of RAM, and a foreground app has an expectation (As of May 2025) to use 200-500mb of RAM.

abdullahemad12:
I forgot that I opened this one line PR 😅

@gabefair I am not sure that highlighting multiple words in a single message as opposed to highlighting one word in a single message puts that much pressure on memory. Even for mobile devices. Why do you think this additional highlights will consume that much memory? 

The reason I added this change is that this is how other Apps like whatsapp and telegram do it (and I would assume if this produces memory problems and was insignificant UX change they wouldn't have done it :thinking:

But aside from the other Apps, this is also how signal on the desktop does it and this change will make it consistent with how it is on the desktop app

JR-git-hub:
I was surprised it didn't work this way when trying to find/jump to something specific... While I found what was looking for, this pull request would have spead that up...

greyson-signal:
My apologies, PR's sometimes get lost as we triage a lot of feedback on a daily basis. This looks reasonable to me, and it'll go into 7.49. Thanks!

PR #13962, opened 2025-02-02
Fixes Overlapping Conversation Banner Position for Hidden Toolbar

Sagar0-0:

### Contributor checklist

 * Device A, Android X.Y.Z
 * Device B, Android Z.Y

----------

### Description

Before:
photo_6122833905083075353_y

After:
Store the Toolbar height in a variable and use it in MultiSelect Mode.
photo_6122833905083075352_y

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
@@ -54,6 +54,7 @@ import androidx.core.content.ContextCompat
 import androidx.core.content.pm.ShortcutManagerCompat
 import androidx.core.view.ViewCompat
 import androidx.core.view.doOnPreDraw
+import androidx.core.view.isInvisible
 import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentResultListener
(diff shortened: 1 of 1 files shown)


greyson-signal (commented):
Ah, good catch on the bug. I agree it should be fixed, but in general, I try to avoid manually setting things like margins and such dynamically whenever possible. Is there something in the view hierarchy that could be fixed so that this overlap doesn't happen?

Sagar0-0:
> Is there something in the view hierarchy that could be fixed 

I'll explore the possibilities and update it here

Sagar0-0:
@greyson-signal Only setting the isInvisible can work instead of setting visible = false. Are there any cons for using this?

greyson-signal:
Changing visibility is fine, I would just like to avoid having to manually change margins in code

Sagar0-0:
> Changing visibility is fine, I would just like to avoid having to manually change margins in code

This change I've proposed won't be required anymore, we can only change the visible= false to isInvisible = true. And everything will work as expected.

greyson-signal:
That sounds much better, will TAL when the PR is updated

Sagar0-0:
@greyson-signal there we go!
Also, I will slow down any new contributions for now and will try to make sure all my opened PRs are getting merged.

greyson-signal (approved):
Looks great, thanks! Will go in 7.34

PR #13965, opened 2025-02-03
Prevent Setup Profile Screen when account already exists

Sagar0-0:

### Contributor checklist

 * Device A, Android X.Y.Z
 * Device B, Android Z.Y

----------

### Description

When an existing user comes back and log in again, we should prevent showing the Setup-profile screen for the following reasons:
1. "setup" generally refers to the act of setting something up for the *first time*, meaning to prepare, arrange, or establish something for use or operation
2. It might mislead the returning user that they have lost their account and now need to setup again.
3. As Avatar is optional, there is no need to go to this screen just for Avatar. Users should enter the app as quickly as possible.


greyson-signal (commented):
The avatar check is intentional. We want to give users an opportunity to set an avatar if one wasn't restored.

Sagar0-0:
@greyson-signal Maybe we can update the Screen title to something like "Welcome back", so that my point 2 isn't applicable.

greyson-signal:
The profile data is pre-populated, so they shouldn't think something is lost. I don't think we need more changes, but thank you!

PR #13970, opened 2025-02-06
Double-tap quick react

zadeviggers:

### Contributor checklist

 * Physical Samsung S21FE, Android 14
 * Virtual Pixel 5, Android 8.1

----------

### Description

Allows the user to quickly add a reaction by double-tapping a message, akin to Instagram and Facebook Messenger. The first of the user's list of reaction emojis is used, matching Instagram's behaviour. Double tapping again removes the reaction. If the user attempts to "quick-react" when they have already reacted with a different emoji, nothing will happen.  

This doesn't replace the holding down to react function. This only works for messages sent by other people; the user double-tapping their own message maintains the quick editing functionality.

This had been mentioned on the ideas forum a couple of times, and it seemed trivial to implement for a large UX gain.

I'm uncertain if the `disposables` / `.subscribe()` I used was correct, but it matches other behaviour for adding reactions.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
@@ -2857,7 +2857,7 @@ class ConversationFragment :
     override fun onItemDoubleClick(item: MultiselectPart) {
       val edited = onDoubleTapToEdit(item.conversationMessage)
       if (!edited) {
-        onDoubleTapToQuickReact(item.conversationMessage);
+        onDoubleTapToQuickReact(item.conversationMessage)
       }
     }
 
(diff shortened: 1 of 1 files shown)


greyson-signal:
I think we're probably gonna pass on this for now, but thank you! I feel like people often have issues with false-positives with the double-tap gesture, and I think the long-press-release gesture we currently have is pretty good. But thank you!

zadeviggers:
That's understandable - hopefully once those issues get sorted out it can be added in the future.
I did notice that the double tap event doesn't fire when the user's tap hits the sent indicator icons to the timestamp, but the rest of the message bubble works fine, so that could be part of the problem?

banocean:
> I think we're probably gonna pass on this for now, but thank you! I feel like people often have issues with false-positives with the double-tap gesture, and I think the long-press-release gesture we currently have is pretty good. But thank you!

It's kinda standard in many apps, making it opt-in feature wouldn't be a good idea?

kieraneglin:
My friends and I are moving a few different group chats away from Meta products, but for whatever reason the double tap to react is a real sticking point for us. I can’t put it to words without sounding silly, but there’s something about the social aspect of ACK-ing a message in a way that’s super fast while your focus remains on the chat as a whole.

I definitely understand your decision here, but I wanted to post my anecdote in case it helps explain the desire/use-case for this feature!

PR #13973, opened 2025-02-06
Support AddToFolder for Multiple Selected Threads

Sagar0-0:

### Contributor checklist

 * Device A, Android X.Y.Z
 * Device B, Android Z.Y

----------

### Description

Reason:
A user may want to add multiple chats in an existing or new folder, which was impossible YET.
I tested it thoroughly, previously I was having some FOREIGN KEY CONFLICT exception which crashed the app everytime when I try to  "select all" chats and add it to an existing folder. This is not reproducing it for me. Let me know if anyone is able to reproduce it. 

Next plans:
Remove from Folder for multi-select. Let me know if this sounds good. 

I have also observed many initialization improvements in the code that we can improve to make the app opening time faster. Like Lazily adding the callbacks and more. Let me know if those PRs are welcomed.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/chats/folders/ChatFoldersViewModel.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/settings/app/chats/folders/ChatFoldersViewModel.kt
@@ -225,21 +225,26 @@ class ChatFoldersViewModel : ViewModel() {
     }
   }
 
-  fun addThreadToIncludedChat(threadId: Long?) {
-    if (threadId == null || threadId == -1L) {
+  fun addThreadsToFolder(threadIds: LongArray?) {
+    if (threadIds == null || threadIds.isEmpty()) {
       return
     }
     viewModelScope.launch {
       val updatedFolder = internalState.value.currentFolder
-      val recipient = SignalDatabase.threads.getRecipientForThreadId(threadId)
-      if (recipient != null) {
-        internalState.update {
-          it.copy(
-            currentFolder = updatedFolder.copy(
-              includedRecipients = setOf(recipient)
-            )
-          )
+      val includedRecipients = mutableSetOf<Recipient>()
+      threadIds.forEach { threadId ->
+        val recipient = SignalDatabase.threads.getRecipientForThreadId(threadId)
+        if (recipient != null) {
+          includedRecipients.add(recipient)
         }
+      }
+
+      internalState.update {
+        it.copy(
+          currentFolder = updatedFolder.copy(
+            includedRecipients = includedRecipients
+          )
+        )
       }
     }
   }
--- a/app/src/main/java/org/thoughtcrime/securesms/conversationlist/AddToFolderBottomSheet.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversationlist/AddToFolderBottomSheet.kt
@@ -45,31 +45,41 @@ import org.thoughtcrime.securesms.util.viewModel
 /**
  * Bottom sheet shown when choosing to add a chat to a folder
  */
-class AddToFolderBottomSheet private constructor() : ComposeBottomSheetDialogFragment() {
+class AddToFolderBottomSheet private constructor(private val onDismissListener: OnDismissListener) : ComposeBottomSheetDialogFragment() {
 
   override val peekHeightPercentage: Float = 1f
+
+  interface OnDismissListener {
+    fun onDismiss()
+  }
+
+  enum class ThreadType(val value: Int) {
+    INDIVIDUAL(1),
+    GROUP(2),
+    OTHER(3)
+  }
 
   private val viewModel by viewModel { ConversationListViewModel(isArchived = false) }
 
   companion object {
     private const val ARG_FOLDERS = "argument.folders"
-    private const val ARG_THREAD_ID = "argument.thread.id"
-    private const val ARG_IS_INDIVIDUAL_CHAT = "argument.is.individual.chat"
+    private const val ARG_THREAD_IDS = "argument.thread.ids"
+    private const val ARG_THREAD_TYPES = "argument.thread.types"
 
     /**
      * Shows a bottom sheet that allows a thread to be added to a folder.
      *
      * @param folders list of available folders to add a thread to
-     * @param threadId the thread that is going to be added
-     * @param isIndividualChat whether the thread is an individual/1:1 chat as opposed to a group chat
+     * @param threadIds list of threads that are going to be added
+     * @param threadTypes list of ThreadType of threads present in threadIds
      */
     @JvmStatic
-    fun showChatFolderSheet(folders: List<ChatFolderRecord>, threadId: Long, isIndividualChat: Boolean): ComposeBottomSheetDialogFragment {
-      return AddToFolderBottomSheet().apply {
+    fun showChatFolderSheet(folders: List<ChatFolderRecord>, threadIds: List<Long>, threadTypes: List<Int>, onDismissListener: OnDismissListener): ComposeBottomSheetDialogFragment {
+      return AddToFolderBottomSheet(onDismissListener).apply {
         arguments = bundleOf(
           ARG_FOLDERS to folders,
-          ARG_THREAD_ID to threadId,
-          ARG_IS_INDIVIDUAL_CHAT to isIndividualChat
+          ARG_THREAD_IDS to threadIds.toLongArray(),
+          ARG_THREAD_TYPES to threadTypes.toIntArray(),
         )
       }
     }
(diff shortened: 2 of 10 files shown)


Sagar0-0:
The issue was caused by allowing to add an archived chat to a folder. If an archived chat was added to any folder and then we tried to "select all" chats and then add them to any folder, the app crashed with `android.database.sqlite.SQLiteConstraintException: FullCode: 787 | ErrorCode: 19 | ExtendedErrorCode: 3 | Message: FOREIGN KEY constraint failed | ExtraMessage: null`.
Fixed this by not allowing "add to folder" from archived chats screen.

app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
```diff
@@ -1580,6 +1581,27 @@ public void onEvent(MessageSender.MessageSentEvent event) {
     closeSearchIfOpen();
   }
 
+  private void showAddToFolderBottomSheet(List<Conversation> conversations) {
```
mtang-signal, line 1584: Please correct me if I'm wrong - it looks like this only gets called once with the singular conversation. If so, you can drop the size check and stream map, but I'd actually consider changing the parameter for this to just `Conversation` and then wrapping it in a list right before the call to `showAddToFolderBottomSheet(List<Long> threadIds, Boolean isIndividual)` since there's no benefit in having it be a list at this point.

app/src/main/java/org/thoughtcrime/securesms/conversationlist/AddToFolderBottomSheet.kt
```diff
@@ -131,8 +137,8 @@ private fun AddToChatFolderSheetContent(
     ) {
       items(folders) { folder ->
         val isIncludedViaChatType = (isIndividualChat && folder.showIndividualChats) || (!isIndividualChat && folder.showGroupChats)
-        val isIncludedExplicitly = folder.includedChats.contains(threadId)
-        val isExcludedExplicitly = folder.excludedChats.contains(threadId)
+        val isIncludedExplicitly = folder.includedChats.containsAll(threadIds)
+        val isExcludedExplicitly = folder.excludedChats.containsAll(threadIds)
```
mtang-signal, line 141: The first check won't fully work because in ConversationListFragment Line 1653, you're always passing false in. This means if you have a folder that is only "All 1:1 chats", `isAlreadyAdded` will always be false if you try to add multiple 1:1 chats. To fix, you can do what you did for thread -> threads, but there might be a cleaner way too.

app/src/main/java/org/thoughtcrime/securesms/conversationlist/AddToFolderBottomSheet.kt
```diff
@@ -78,33 +82,35 @@ class AddToFolderBottomSheet private constructor() : ComposeBottomSheetDialogFra
...
 
     AddToChatFolderSheetContent(
-      threadId = threadId,
+      threadIds = threadIds,
       isIndividualChat = isIndividualChat,
       folders = remember { folders ?: emptyList() },
       onClick = { folder, isAlreadyAdded ->
         if (isAlreadyAdded) {
           Toast.makeText(context, requireContext().getString(R.string.AddToFolderBottomSheet_this_chat_is_already, folder.name), Toast.LENGTH_SHORT).show()
         } else {
```
mtang-signal, line 95: Small thing - this is a singular `This chat is already in the folder x`, but in cases where threadIds is plural, we'll want to change it to `These chats are already in the folder x`

mtang-signal (commented):
Hi, I think this would be an amazing thing to add to chat folders :). Added some comments below, let me know if you have any questions!

app/src/main/java/org/thoughtcrime/securesms/conversationlist/AddToFolderBottomSheet.kt
Sagar0-0: I agree, there is a problem with `isIncludedViaChatType`, which doesn't make sense anymore with multiple chats, to make sense for it, we can have `areChatsIndividuals` (stating if all selected are individuals) `areChatsGroups` (stating if all selected are groups) instead of only relying on `isIndividualChat` and using these two new fields for `isIncludedViaChatType` will give the correct output, which will then solve the issue with `isAlreadyAdded`.

I believe this approach is cleaner and will make sense to all the existing checks.

mtang-signal: Definitely solves the old issues, but there will be a new issue when your selection is not just all groups or all 1:1 chats. Example would be Folder A containing the all groups flag and a chat with Alice. Then, if you try to add Alice and any group to Folder A, it won't mark Folder A as already included because your selection mix is no longer strictly all 1:1 chats/all groups chats.

app/src/main/res/values/strings.xml
```diff
@@ -760,6 +760,10 @@
     <string name="AddToFolderBottomSheet_choose_a_folder">Choose a folder</string>
     <!-- Toast shown when a chat has been added to a folder, where %s is the name of the folder -->
     <string name="AddToFolderBottomSheet_added_to_s">Added to \"%1$s\"</string>
+    <plurals name="AddToFolderBottomSheet_chats_added_to_s">
```
mtang-signal, line 763: Thanks for adding the string! Looks like you can remove the old string `AddToFolderBottomSheet_this_chat_is_already` (line 768) since it's no longer in use. I'd also move the comment of that string (line 767) to this string. One thing when working with quotations is that they have to be escaped so the strings should be: `\"%1$s\"` instead of `"%1$s"`

app/src/main/java/org/thoughtcrime/securesms/components/settings/app/chats/folders/ChatFoldersViewModel.kt
```diff
@@ -225,22 +225,27 @@ class ChatFoldersViewModel : ViewModel() {
...
-        internalState.update {
-          it.copy(
-            currentFolder = updatedFolder.copy(
-              includedRecipients = setOf(recipient)
-            )
-          )
+      val includedRecipients = mutableSetOf<Recipient>()
+      threadIds.forEach { threadId ->
+        val recipient = SignalDatabase.threads.getRecipientForThreadId(threadId)
+        if(recipient != null) {
```
mtang-signal, line 237: nit: Not a big deal, but I'm surprised the ktlint didn't catch this and add a space.

app/src/main/java/org/thoughtcrime/securesms/conversationlist/AddToFolderBottomSheet.kt
Sagar0-0: I misunderstood the concept that adding the group type will consider all group chats are Included, but one group chat can be explicitly included in a Folder with the Group type. Now I get it.

The problem you mentioned is because the `areAllGroupChats` is false, and now we are checking if all selected chats are Explicitly added, which also returns false, as groups are not explicitly added.
This also means there is another problem with this logic, for an explicitly removed group chats, if we try to add multiple group chats in a group-type folder, this will mark it as "Already Included" even when one group chat was excluded.

One solution can be to send `groupThreadIds` and `individualThreadIds` as only 2 params. Then we can check separately for chatTypes and add only the threads that are NOT explicitly included. This way we can avoid adding group chats to group-type folders.

We need to check one more thing, if for example a group chat was excluded from a group-type folder, and then we add it back, then the group should not be added in the `includedChats` list(Which it does in the existing codebase). We can ONLY remove it from the `excludedchats` list and this will work. To fix this, we can check the threads which are of the Folder's chat-type but are excluded, and we can only remove them from excluded.

I will make the mentioned changes.

mtang-signal: > We need to check one more thing, if for example a group chat was excluded from a group-type folder, and then we add it back, then the group should not be added in the includedChats list(Which it does in the existing codebase). We can ONLY remove it from the excludedchats list and this will work. To fix this, we can check the threads which are of the Folder's chat-type but are excluded, and we can only remove them from excluded.

I would actually keep this behavior as is. You're right that it's a little redundant in cases when a chat was only explicitly excluded but it's a much cleaner/simpler check (also in both cases, you'll still need to write to the db). I also think it's more in line with user expectations - if they manually choose to add a chat, then it shows up as explicitly included and won't change if they ever change the 1:1-type/group-type of a folder.

Sagar0-0: The solution is a little different from what I said, and the reason is to provide a more scalable solution. This is better than passing 2 lists, groupThreadIds and IndividualThreadIds. If we decide to add a new ThreadType(apart from Individual and Group), then this can be easily handled without much refactoring.

The `isThreadListAlreadyAddedis` is self-explanatory.

This should now cover all the edge cases.

Sagar0-0: The solution is a little different from what I said, and the reason is to provide a more scalable solution. This is better than passing 2 lists, groupThreadIds and IndividualThreadIds. If we decide to add a new ThreadType(apart from Individual and Group), then this can be easily handled without much refactoring.

The `isThreadListAlreadyAddedis` is self-explanatory.

This should now cover all the edge cases.

app/src/main/java/org/thoughtcrime/securesms/conversationlist/AddToFolderBottomSheet.kt
```diff
@@ -45,31 +45,44 @@ import org.thoughtcrime.securesms.util.viewModel
...
-     * @param threadId the thread that is going to be added
-     * @param isIndividualChat whether the thread is an individual/1:1 chat as opposed to a group chat
+     * @param threadIds list of threads that are going to be added
+     * @param threadTypes list of ThreadType of threads present in threadIds
      */
     @JvmStatic
-    fun showChatFolderSheet(folders: List<ChatFolderRecord>, threadId: Long, isIndividualChat: Boolean): ComposeBottomSheetDialogFragment {
-      return AddToFolderBottomSheet().apply {
+    fun showChatFolderSheet(folders: List<ChatFolderRecord>, threadIds: List<Long>, threadTypes: List<Int>, onDismissListener: OnDismissListener): ComposeBottomSheetDialogFragment {
+      assert(threadIds.size==threadTypes.size) {
```
mtang-signal, line 78: I'd remove the assertion here - looking at the calls, it's impossible for the sizes to be different and we try to avoid using assertions unless we need to guarantee the state of a more critical operation.

app/src/main/java/org/thoughtcrime/securesms/conversationlist/AddToFolderBottomSheet.kt
```diff
@@ -210,14 +225,36 @@ private fun AddToChatFolderSheetContent(
   }
 }
 
+fun isThreadListAlreadyAdded(folder: ChatFolderRecord, threadIds: List<Long>, threadTypes: List<Int>): Boolean {
```
mtang-signal, line 228: brilliant work :) . (also please make this function private)

app/src/main/java/org/thoughtcrime/securesms/conversationlist/ConversationListFragment.java
```diff
@@ -1580,6 +1579,58 @@ public void onEvent(MessageSender.MessageSentEvent event) {
     closeSearchIfOpen();
   }
 
+  private void showAddToFolderBottomSheet(Conversation conversation) {
+    boolean isIndividual = conversation.getThreadRecord().getRecipient().isIndividual();
+    boolean isGroup = conversation.getThreadRecord().getRecipient().isGroup();
```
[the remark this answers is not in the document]
Sagar0-0: > just change getRecipient().isGroup to getRecipient().isPushGroup

Ok, I have done it. 
Can you tell me the difference between pushGroup, pushV1Group, and pshV2Group?

>  but I would actually factor out those two into their own function.

You mean adding more ThreadType for SMS Groups and GroupV1? 


This PR took many review cycles, is it common for these types of changes?
I should have done proper testing for the `isAlreadyAdded` check. Anything else I should learn here to make things easier for reviewers?


Also, should I squash all the commits into 1, or is It fine?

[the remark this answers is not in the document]
mtang-signal: Ah, just meant factoring out this: 
```
      boolean isIndividual = conversation.getThreadRecord().getRecipient().isIndividual();
      boolean isGroup = conversation.getThreadRecord().getRecipient().isGroup();
     ...
        type = AddToFolderBottomSheet.ThreadType.GROUP.getValue();
      } else {
        type = AddToFolderBottomSheet.ThreadType.OTHER.getValue();
      }
```
from here and then Line 1605 into its own function since it's the same thing.

> Can you tell me the difference between pushGroup

We used to support SMS groups but don't do that anymore and group V1 was our initial design of groups, but didn't allow for some group operations, so a few years ago, we rewrote a new group V2 that's the standard now.

> This PR took many review cycles

Chat folders is very new and the change you've made is a feature, not a bug, so that's why there's slightly more review cycles than a bug fix. However, I think you've done great work and the complexity around `isAlreadyAdded` is not something you could've easily discovered. Your responses and openness to feedback make this pr very easy to review :).

Eventually, it'll need to be 1 commit. If you want to squash it now, I'd appreciate it, otherwise I'll do it from my end.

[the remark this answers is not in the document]
Sagar0-0: Thanks!! Made the required changes. Let me know if we are good to go.

mtang-signal (approved):
Great work 🥂 . Should be in 7.35

PR #13985, opened 2025-02-13
Enable verified subtitle on group chat. Fixes #13873

### LINKED ISSUE #13873: Groups not showing as verified even when all members are marked as verified
### Guidelines

- [X] I have searched searched open and closed issues for duplicates

### Bug description

None of the groups I am in are showing as verified, even when I am verified with all of the other members of the group.

Just now, I created a group with only one other person with whom I am verified, but there is no checkmark at the top of the group:

image

I've had this issue for several months now. It started on my Pixel 5 and still happening on my Pixel 7 Pro.

However I do still see groups showing up as verified on Signal for Desktop which is synced to my phone.

Pardon me if this feature was removed on purpose. I couldn't find anything about it. Without my laptop, I have no other way of easily checking that I'm still verified with everyone else in a group before sending a sensitive message, which I would like to do as we are an activist organization with some underground work. Thanks! :)

### Screenshots

_No response_

### Device

Google Pixel 7 Pro

### Android version

15

### Signal version

7.28.4

### Link to debug log

_No response_

Ascheidweiler:
The verified subtitle doesn't appear when all group members have verified status, unlike iOS/Desktop apps. Currently existing code looks capable of this feature, but was being explicitly skipped. I'm not seeing an obvious reason why that check was in place. Removing the isGroup check allows the verified badge to display on group chats.

Fixes #13873

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
@@ -1236,9 +1236,7 @@ class ConversationFragment :
   }
 
   private fun presentIdentityRecordsState(identityRecordsState: IdentityRecordsState) {
-    if (!identityRecordsState.isGroup) {
-      binding.conversationTitleView.root.setVerified(identityRecordsState.isVerified)
-    }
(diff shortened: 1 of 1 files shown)


greyson-signal (approved):
This makes sense to me, thanks! Will be in 7.35.

PR #14048, opened 2025-03-15
Add disposable management for message resend operation

Sagar0-0:

- [ ] I have read [how to contribute](https://github.com/signalapp/Signal-Android/blob/main/CONTRIBUTING.md) to this project
- [ ] I have signed the [Contributor License Agreement](https://signal.org/cla/)

### Contributor checklist

- [ ] I am following the [Code Style Guidelines](https://github.com/signalapp/Signal-Android/wiki/Code-Style-Guidelines)
 * Device A, Android X.Y.Z
 * Device B, Android Z.Y

----------

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
@@ -2458,7 +2458,7 @@ class ConversationFragment :
   }
 
   private fun handleResend(conversationMessage: ConversationMessage) {
-    viewModel.resendMessage(conversationMessage).subscribe()
+    viewModel.resendMessage(conversationMessage).subscribe().addTo(disposables)
   }
 
   private fun handleEnterMultiselect(conversationMessage: ConversationMessage) {


ImAwsumm (approved):

app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
```diff
@@ -2458,7 +2458,7 @@ class ConversationFragment :
   }
 
   private fun handleResend(conversationMessage: ConversationMessage) {
-    viewModel.resendMessage(conversationMessage).subscribe()
+    viewModel.resendMessage(conversationMessage).subscribe().addTo(disposables)
```
alex-signal, line 2461: I would argue that we should never cancel a message resend due to the lifecycle going out of scope. Instead, what we likely want to do here is unwrap the operation from Rx, and instead utilize either a coroutine scoped to the ViewModel (performed in the `Dispatchers.IO` context) or directly dispatch it on the `SignalExecutors.BOUNDED_IO` executor to ensure that we do our best to make sure the resend happens.

Sagar0-0: I agree that we should do our best to make sure resend happens. But, for the same reason, I want to understand why we are doing the same for operations like `handleCopyMessage` or `updateReaction`. Ideally, we want these operations to always succeed. 
Does that mean we should be also unwrapping `handleCopyMessage` or `updateReaction` from using Rx?

alex-signal: `handleCopyMessage` can remain attached to the lifecycle. Arguably the Context usage could be changed to `AppDependencies.application` but I'm less worried about that one since it doesn't involve sending a message to another user. `updateReaction` could definitely be unwrapped in a way similar to what I mentioned for `handleResend`.

Sagar0-0: I'm closing this PR as invalid and we can handle unwrapping the mentioned function later sometime

PR #14087, opened 2025-04-11
Log Refactor

kasswag11:
Overview
We propose refactoring the Log utility in org.signal.glide to better align with DDD principles by introducing the following:
A LoggingService interface.
LogTag and LogMessage value objects.
Moving implementation details to the infrastructure layer. 
This will decouple logging from domain logic, improve modularity, and enhance testability. During a deep dive into the Signal Android codebase for a university software architecture course, our team identified a tight coupling between domain logic and infrastructure in the current logging mechanism. 
The static, global access to logging (SignalGlideCodecs.getLogProvider()) and raw strings for tags/messages make logging less flexible, error-prone, and more complicated to test. It is this idea we are targeting for our refactoring.
Problems Identified
Tight Coupling to static global state (SignalGlideCodecs).
There is no Abstraction between domain logic and infrastructure.
Strongly typed metadata, which reduces clarity and consistency.
Logging Output Loss due to silent fallback EMPTY Provider.
Benefits
Improved separation of concerns.
Consistent and error-resistant metadata handling.
Aligns logging infrastructure with DDD best practices.

We’re happy to submit a Draft Pull Request with this implementation for further discussion if the maintainers are interested. We would love to hear your thoughts.

Proposed Refactor 
Abstract Logging Behind a Service Interface
public interface LoggingService {
    void logInfo(LogTag tag, LogMessage message);
    void logError(LogTag tag, LogMessage message, @Nullable Throwable throwable);
}

Introduce Value Objects for Log Metadata

public record LogTag(String value) { }
public record LogMessage(String value) { }

Move the Implementation to the Infrastructure Layer

public class AndroidLoggingService implements LoggingService {
    public void logInfo(LogTag tag, LogMessage message) {
     SignalGlideCodecs.getLogProvider().i(tag.value(), message.value());
}
    public void logError(LogTag tag, LogMessage message, @Nullable Throwable t) {
     SignalGlideCodecs.getLogProvider().e(tag.value(), message.value(), t);
    }
}

Thanks for your time and the great work on Signal
 - Griffin Urban, Seth Clover, Kasson Plummer, Nathan Voung, Lawson Port

### DIFF
--- a/app/src/main/java/org/signal/glide/transforms/LogMessage.java
+++ b/app/src/main/java/org/signal/glide/transforms/LogMessage.java
@@ -1,0 +1,9 @@
+package org.signal.glide.transforms;
+
+public record LogMessage(String value) {
+  public LogMessage {
+    if (value == null || value.isBlank()) {
+      throw new IllegalArgumentException("LogMessage cannot be null or blank");
+    }
+  }
+}
--- a/app/src/main/java/org/signal/glide/transforms/LogTag.java
+++ b/app/src/main/java/org/signal/glide/transforms/LogTag.java
@@ -1,0 +1,9 @@
+package org.signal.glide.transforms;
+
+public record LogTag(String value) {
+  public LogTag {
+    if (value == null || value.isBlank()) {
+      throw new IllegalArgumentException("LogTag cannot be null or blank");
+    }
+  }
+}
--- a/app/src/main/java/org/signal/glide/transforms/LoggingService.java
+++ b/app/src/main/java/org/signal/glide/transforms/LoggingService.java
@@ -1,0 +1,10 @@
+package org.signal.glide.transforms;
+
+import androidx.annotation.Nullable;
+
+public interface LoggingService {
+  void logDebug(LogTag tag, LogMessage message);
+  void logInfo(LogTag tag, LogMessage message);
+  void logWarn(LogTag tag, LogMessage message);
+  void logError(LogTag tag, LogMessage message, @Nullable Throwable throwable);
+}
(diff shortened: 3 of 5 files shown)


app/src/main/java/org/signal/glide/transforms/LogMessage.java
```diff
@@ -0,0 +1,9 @@
+package org.signal.glide.transforms;
+
+public record LogMessage(String value) {
+  public LogMessage {
+    if (value == null || value.isBlank()) {
+      throw new IllegalArgumentException("LogMessage cannot be null or blank");
+    }
+  }
+}
```
shivanandyadav11, line 9: I noticed that this PR introduces a new file in Java, while we're in the midst of migrating our codebase to Kotlin. Is there a particular reason we opted for Java here instead of Kotlin?
Would it be possible to implement this new functionality in Kotlin to maintain consistency with our migration efforts?

app/src/main/java/org/signal/glide/transforms/LogTag.java
```diff
@@ -0,0 +1,9 @@
+package org.signal.glide.transforms;
+
+public record LogTag(String value) {
+  public LogTag {
+    if (value == null || value.isBlank()) {
+      throw new IllegalArgumentException("LogTag cannot be null or blank");
+    }
+  }
+}
```
shivanandyadav11, line 9: Same as above

app/src/main/java/org/signal/glide/transforms/LoggingService.java
```diff
@@ -0,0 +1,10 @@
+package org.signal.glide.transforms;
+
+import androidx.annotation.Nullable;
+
+public interface LoggingService {
+  void logDebug(LogTag tag, LogMessage message);
+  void logInfo(LogTag tag, LogMessage message);
+  void logWarn(LogTag tag, LogMessage message);
+  void logError(LogTag tag, LogMessage message, @Nullable Throwable throwable);
+}
```
shivanandyadav11, line 10: Same as above

PR #14137, opened 2025-05-12
Fix country code filtering by removing "+" prefix handling.

Sagar0-0:
The search query for "+1" won't show anything. So, removing the + prefix. I'll probably add testcases for these classes too.


ankushkushwaha (commented):
Thanks for this fix, Is this PR linked with any issue number / ticket ?

cody-signal (commented):
Will be in 7.43

PR #14145, opened 2025-05-14
Add unit tests for FindByViewModel.

Sagar0-0:
Added 100% test coverage with some refactoring.


app/src/main/java/org/thoughtcrime/securesms/recipients/ui/findby/FindByActivity.kt
```diff
@@ -195,7 +193,7 @@ class FindByActivity : PassphraseRequiredActivity() {
                 val cleansed = state.userEntry.removePrefix(state.selectedCountry.countryCode.toString())
                 E164Util.formatAsE164WithCountryCodeForDisplay(state.selectedCountry.countryCode.toString(), cleansed)
               }
-              stringResource(id = R.string.FindByActivity__s_is_not_a_valid_phone_number, state.userEntry)
```
cody-signal, line 198: I think we want to keep this as the user provided input instead of the formatted input so things match what the user sees/entered.

Sagar0-0: So we can also remove the formattedNumber.

gabefair:
Should this be reopened?

PR #14283, opened 2025-08-08
Fix: Resolve memory leak in sticker keyboard

rishabh-bhatia:

### Contributor checklist

 * Device Pixel 7a, Android 16

### Description

This pull request fixes a significant memory leak in the sticker keyboard, which caused UI lag and system slowdowns, particularly on lower-end devices.

**Testing and Validation:**

The solution was developed through extensive testing on a Pixel 7a using the Android Studio Profiler to monitor memory allocation and CPU usage.

1. Problem Confirmation: Confirmed that rapidly scrolling through animated stickers caused memory usage to climb above 1.5GB, resulting in severe GC pressure and UI jank.
2. Solution Analysis: Tested several potential solutions independently and in combination:
       * A global `GlideModule` cache limit was found to be a necessary safety net but insufficient on its own, as it still suffered from "cache churn" when full.
       * An `onViewRecycled` implementation proved to be the most effective fix for performance during a scrolling session, as it dramatically reduced GC pressure by preventing the memory cache from filling.
       * An `onDestroyView` cleanup was found to be crucial for preventing lag between chat sessions by ensuring the cache was cleared when the keyboard was closed.
3. Final Solution: The final implementation combines all three approaches for maximum performance and stability. It was verified that with this fix, the app remains responsive under normal and heavy use, with lag rerely appearing under deliberate, prolonged stress testing. The fix was also validated against the project's code style guidelines.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/keyboard/sticker/KeyboardStickerListAdapter.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/keyboard/sticker/KeyboardStickerListAdapter.kt
@@ -39,13 +39,14 @@ class KeyboardStickerListAdapter(
     }
   }
 
-  internal inner class StickerViewHolder(itemView: View) : MappingViewHolder<Sticker>(itemView) {
+  private inner class StickerViewHolder(itemView: View) : MappingViewHolder<Sticker>(itemView) {
 
-    internal val image: ImageView = findViewById(R.id.sticker_keyboard_page_image)
+    val image: ImageView = findViewById(R.id.sticker_keyboard_page_image)
 
     override fun bind(model: Sticker) {
       requestManager.load(model.uri)
         .set(ApngOptions.ANIMATE, allowApngAnimation)
+        .skipMemoryCache(true)
         .transition(DrawableTransitionOptions.withCrossFade())
         .into(image)
 
--- a/app/src/main/java/org/thoughtcrime/securesms/keyboard/sticker/StickerKeyboardPageFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/keyboard/sticker/StickerKeyboardPageFragment.kt
@@ -105,7 +105,6 @@ open class StickerKeyboardPageFragment :
   }
 
   override fun onDestroyView() {
-    Glide.get(requireContext()).clearMemory()
     AppDependencies.databaseObserver.unregisterObserver(this)
     requireView().removeOnLayoutChangeListener(this)
     super.onDestroyView()
--- a/glide-config/src/main/java/org/thoughtcrime/securesms/mms/SignalGlideModule.kt
+++ b/glide-config/src/main/java/org/thoughtcrime/securesms/mms/SignalGlideModule.kt
@@ -6,9 +6,6 @@ import com.bumptech.glide.Glide
 import com.bumptech.glide.GlideBuilder
 import com.bumptech.glide.Registry
 import com.bumptech.glide.annotation.GlideModule
-import com.bumptech.glide.load.engine.bitmap_recycle.LruBitmapPool
-import com.bumptech.glide.load.engine.cache.LruResourceCache
-import com.bumptech.glide.load.engine.cache.MemorySizeCalculator
 import com.bumptech.glide.module.AppGlideModule
 
 /**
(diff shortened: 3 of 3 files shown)


app/src/main/java/org/thoughtcrime/securesms/keyboard/sticker/KeyboardStickerListAdapter.kt
```diff
@@ -39,9 +39,9 @@ class KeyboardStickerListAdapter(
     }
   }
 
-  private inner class StickerViewHolder(itemView: View) : MappingViewHolder<Sticker>(itemView) {
+  internal inner class StickerViewHolder(itemView: View) : MappingViewHolder<Sticker>(itemView) {
 
-    private val image: ImageView = findViewById(R.id.sticker_keyboard_page_image)
+    internal val image: ImageView = findViewById(R.id.sticker_keyboard_page_image)
```
jeffrey-signal, line 44: nit: StickerViewHolder should be able to remain private

```
private inner class StickerViewHolder(itemView: View) : MappingViewHolder<Sticker>(itemView) {

  val image: ImageView = findViewById(R.id.sticker_keyboard_page_image)
  ...
}
```

app/src/main/java/org/thoughtcrime/securesms/keyboard/sticker/KeyboardStickerListAdapter.kt
```diff
@@ -62,6 +62,12 @@ class KeyboardStickerListAdapter(
     }
   }
 
+  override fun onViewRecycled(holder: MappingViewHolder<*>) {
+    if (holder is StickerViewHolder) {
+      requestManager.clear(holder.image)
+    }
+  }
```
jeffrey-signal, line 69: 👍 I think this is definitely the right move, since it's the most targeted fix of the three proposed solutions.

glide-config/src/main/java/org/thoughtcrime/securesms/mms/SignalGlideModule.kt
```diff
@@ -16,21 +19,33 @@ import com.bumptech.glide.module.AppGlideModule
...
+
+    val calculator = MemorySizeCalculator.Builder(context).build()
+    val defaultMemoryCacheSize = calculator.memoryCacheSize.toLong()
+    val defaultBitmapPoolSize = calculator.bitmapPoolSize.toLong()
+
+    val customMemoryCacheSize = (defaultMemoryCacheSize * MEMORY_CACHE_SAFETY_FACTOR).toLong()
+    val customBitmapPoolSize = (defaultBitmapPoolSize * MEMORY_CACHE_SAFETY_FACTOR).toLong()
+
+    builder.setMemoryCache(LruResourceCache(customMemoryCacheSize))
+    builder.setBitmapPool(LruBitmapPool(customBitmapPoolSize))
```
jeffrey-signal, line 40: While these smaller cache sizes may help sticker‑heavy screens, globally shrinking the cache may harm performance elsewhere by reducing memory‑cache hit rates and increasing and disk reads.

In my testing, the changes to `KeyboardStickerListAdapter` alone dramatically improve the performance of the sticker keyboard. If possible, I think it makes sense to apply these smaller cache sizes to sticker loads only. Otherwise, I'd prefer to revert the global cache changes.

app/src/main/java/org/thoughtcrime/securesms/keyboard/sticker/StickerKeyboardPageFragment.kt
```diff
@@ -105,6 +105,7 @@ open class StickerKeyboardPageFragment :
   }
 
   override fun onDestroyView() {
+    Glide.get(requireContext()).clearMemory()
```
jeffrey-signal, line 108: Clearing Glide's memory cache each time this fragment is destroyed feels a bit too aggressive and may hurt cache hit rates, reducing the image loading performance on other screens.

jeffrey-signal (changes requested):
Nice work, thanks for the contribution!

app/src/main/java/org/thoughtcrime/securesms/keyboard/sticker/KeyboardStickerListAdapter.kt
rishabh-bhatia: Thank you so much for the insightful feedback! You are absolutely right, the global changes were too aggressive and would have impacted the cache hit rates and other screens.

  I've pushed a new commit to address your points:

   1. I have reverted the changes to `SignalGlideModule.kt` and removed the global memory cache limits, restoring the default behavior.
   2. I have also removed the call to clear cache from `StickerKeyboardPageFragment.kt`.

   3. To address your point about sticker load specific caching, I did some further investigation and found a more stable solution. I've added `skipMemoryCache(true)` to the Glide requests in `KeyboardStickerListAdapter`.
  
This approach prevents the stickers from being held in the shared memory cache, which was the source of the memory pressure, screen freeze and lags. This does mean we are intentionally bypassing the memory cache for stickers, but Glide will still leverage its **disk cache** for them. 

This protects the memory-cache hit rate for all other images in the app and fully resolves the stability issue, with a small trade off of a slightly slower reload for stickers when scrolling very quickly(I have not experienced this but on lower end devices it is a possibility as disk read/write speeds may not be as fast as pixel).

Keen to know what you think about this fix. If you prefer just keeping the `onViewRecycled` fix, happy to remove `skipMemoryCache(true)` from the fix.

PR #14463, opened 2025-11-28
Anti timing attack features

sullystuff:

### Contributor checklist

 * Samsung S24 Ultra - Android 16

----------

### Description

Add privacy features to prevent timing attacks as detailed [here](https://arxiv.org/pdf/2411.11194)
- Add random delay 300-5000ms to delivered/read receipts
- Disable delivered receipts for:
- - Edited messages
- - Reactions (delivered receipts do not show on any UI, this has no reason to exist)
- - Deleted messages (delivered receipts do not show on any UI, this has no reason to exist)
- - blocked users (this is default on iOS but doesn't exist on Android, this can be used to profile users)

Tested with iOS 26.1 on iPhone 16 Pro as well as Windows 11 for the sending client, and aforementioned Android mobile device for the receiving client (with this modified apk)


ghost:
this is good fun!

If the server can see the sender, it will send a receipt.

Ironically, it doesn't send a receipt until the client is online. Exposing you are online. **This is also a problem that should be fixed!!!**

> Disable delivered receipts for:
>
>    Edited messages

These show up in the ui. It's inconsistent to disable them.

> blocked users (this is default on iOS but doesn't exist on Android, this can be used to profile users)

Again, if the server can see the sender, it [sends a receipt](https://github.com/signalapp/Signal-Server/blob/v20251126.0.0/service/src/main/java/org/whispersystems/textsecuregcm/push/ReceiptSender.java). This will commonly be the case with blocked senders. When the server cannot see the sender no receipt is sent to blocked senders on [Android](https://github.com/signalapp/Signal-Android/blob/v7.66.3/app/src/main/java/org/thoughtcrime/securesms/messages/MessageContentProcessor.kt#L146-L192).

You don't need all those options; when might someone prefer instant receipts?

Consider removing the minimum delay so instant can be chosen at random.

sullystuff:
> If the server can see the sender, it will send a receipt.

Are you referring to the Signal servers?
If so, I toggled "disable edit delivered receipts" on the test device, edited a message on Device 1 and it was not marked as delivered. I then interacted with test device, sending and receiving messages to and from another user, and the edited message still has not been marked as delivered. Same goes for all other delivered receipt toggles I added

> These show up in the ui. It's inconsistent to disable them.

Yes, it should be an optional toggle, and it should be clear how it will affect UX for the other users.


I'll consider ways to drop the time add minimum to 0, or perhaps make the delay an optional range.

sullystuff:
Disabled delivered receipts are enabled by default for reactions+deletes as these don't affect UX at all

ghost:
>the edited message still has not been marked as delivered

If you ensure Allow from anyone in Settings>Privacy>Advanced is off for the test device, ensure both devices don't share group membership, block Device 1 and then edit the message you'll have a different result.

Show status icon is on by default I believe in the same section of settings.

On Device 1, If you look at "Message details" for the message to be edited you'll see the icon. Then once you edit the message it'll vanish.

If Device 1 can prove it has the test device's current profile keys, the server cannot see that Device 1 sent a message to the test device.

https://signal.org/blog/sealed-sender#:~:text=Delivery%20tokens

Blocking rotates test device's profile key meaning Device 1 will only receive server receipts, nothing directly from the test device.

if Allow from anyone is on, Device 1 doesn't need test device's profile keys.
On Android, shared group membership means blocking doesn't rotate profile key.

sullystuff:
Interesting, you're right, so this only works when the attacker is using sealed sender, and an attacker is most likely to not use sealed sender lol... so is there no way to patch that on the client end? Is this a protocol issue?

Lama-Thematique:
> Add random delay 300-5000ms to delivered/read receipts

This is good but not perfect, is make the phone delay distribution much less obvious by adding a lot of variance and shifting the mean by 2.65 seconds but with enough data points you could still statistically estimate the mean over a period and compensate for both the added variance and the shift of the distribution.

The new distribution is now a sum of the 'true' distribution (delay from the phone) that we want to hide, and the 'fake' distribution the random 0.3 to 5 second delay.

By modelling the coupled distribution you can get estimations of the real distribution of timings from the device and be able to defeat the random delay mitigation. There is 2 possibilities to fix it:
- Have the range of delay so high that you need way to much datapoint to do analysis making it practically useless but this is unpractical as it would either needs a low rate limit or a long delay for seen indicators to correctly mask the distribution.
- Compensate the fake distribution to hide the true distribution.

How to compensate:
1. Compute the delay (call it 'd') from the moment the signal server received the message (and i hope the is a way to get a timestamp for that) and the moment you are processing the response (just before adding it to queue).
2. get a random value bewteen 0.3 and 5 seconds (call it 'r').
3. delay the message by : r - d

This way it's no longer possible to mesure d and hence do the exploit.

Note: the graphs on the paper observe maximum delays of about 3 seconds, this should become the lower bound of the delay instead of 0.3 sec, so 'r' should be sampled form a uniform distribution ranging from 3 to 5 second.

Note 2: The exploit might still be possible after this if queued responses are also being delayed differently depending the device state (I don't know how if it's the case), in which case estimation of the full delay is needs to be compensated which is harder to do preciselly. My hypothesis (that need to be verified empirically before implementation) for a simple estimation of that full delay could be the double of 'd' initally estimated changing the last step "to delay the message by : r - 2*d".

Lama-Thematique:
Finally I would recommend contacting the original authors of the papers once you have some good mitigations in place so that they can verify it by testing it in practice on the same devices used for their research, as they likelly already have code for that kinds of mesurements.

ghost:
If the intent is protection from traffic analysts @sullystuff this seems effective. Signal might implement it in a more joined-up, rigorous and thorough way.

They should also fix that server-side issue [I mentioned](https://github.com/signalapp/Signal-Android/pull/14463#issuecomment-3590047165#:~:text=This%20is%20also%20a%20problem%20that%20should%20be%20fixed) too.

Lama-Thematique:
What I'd like to suggest is to merge the random delay as is, for 2 reasons:
- test if there is community backlash for the additional delay that it creates 
- have a quick answer to less sophisticated analysis

At the same time, a second cross platform issue/tracker that also include protocol discussion should be created to discuss how to best make the entire attack completely impossible. 
This discussion is complex and likelly it will take a significant amount of time keeping the vulnerability exploitable for long which is why it's better with a temporary "quick fix" merged.

Lama-Thematique:
Finally after thinking it over, it might much easier to mitigate server side. This does place more trust in the server as the server would be able to collect all the data for the exploit. But the server can easily compute the full exact RTT of your device and correct for it when sending the read indicator back to the attacker. 
A server side mitigation also does not care about the wide range of possible devices that could have different behaviours nor the platform used by the user.

sullystuff:
@Lama-Thematique Sure, it's easier to mitigate server-side, but as you stated that places more trust in the server when we're trying to avoid that (imo)


> This is good but not perfect, is make the phone delay distribution much less obvious by adding a lot of variance and shifting the mean by 2.65 seconds but with enough data points you could still statistically estimate the mean over a period and compensate for both the added variance and the shift of the distribution.

Yes, over a long enough period of time an attacker could, in theory, be able to calculate away the noise. I think the best trustless solution here is to adjust the min/max delays on a per-device/OS-version basis, as reported by those researchers. I'll do my best to get in touch with them.

sullystuff:
I've sent an email to one of the researchers, I'll update here if I get a response, and if I don't I'll reach out to a few more.

Lama-Thematique:
> Sure, it's easier to mitigate server-side, but as you stated that places more trust in the server when we're trying to avoid that (imo)

It's tradeoffs as:
- server side mitigation is certainly better than current status as today everyone (server and 3rd party alike) can use this attack.
- the server side mitigations is very easy  to implement code wise
- it makes 100% sure the distribution is hidden to 3rd party (as we can have exact RTT)
- but the negative is that it trust the server

Tho i agree that i would prefer a solid trustless implementation. 

---
Trustless solution are great because it's trustless but could not entirely hide the distributions correctly as it's based on estimations of RTT. Implementation needs to be carefull to have good unbiased estimates of the true RTT distribution in all situations (4g/wifi, different OS and version, different hardwares, different background/foreground tasks, etc...) otherwise better data anlysis could bring back the problem.

> per-device/OS-version basis

This is a bad idea IMO at it requires constant maintenance for the table of devices and OS versions and could still be affected by other factors. For trustless solutions I think heuristic based methods are preferable meaning device estimates it's own RTT by itself.

--- 
In all, my preferences (and this is a opinion so it's debatable) are:
Best = Precise trustless solution
Ok = trusted server side solution
Worse = Naïve trustless solution
Bad = No solutions

sullystuff:
We can probably automate tests on the client to measure the average time to perform return pings (standby, app foreground, app background, active phone call etc) and adjust the delay by the measured average time per state. That will take a while to write though.

sullystuff:
@greyson-signal @alex-signal thoughts?

kernoelpanic:
Hey there,

I’m one of the authors of the [paper](https://arxiv.org/pdf/2411.11194) you referred to. We’re happy to see progress being made to mitigate the issue. We fully agree that having no solution is the worst option, but finding a “real fix” is also quite tricky.

As you already noted, there are things which can be done with relatively little effort, though:
* Prevent sending receipts of any kind for actions that do not trigger observable notifications for the user. This prevents an attacker from gathering a large set of samples in a short amount of time.
* Force a specific distribution of receipts on the client side (as discussed here). This will not be perfect, but it would have an immediate effect by making it much harder  in practice to extract useful information from receipts.

Keep in mind, this will _not_ be a solution against the various "is online" side-channels, but make it harder for an attacker to extract more information from receipts.

In the long run, other solutions can be considered. These might also include a discussion about whether explicit delivery receipts are needed at all, or whether they could be disabled by privacy-sensitive users. From our current understanding, delivery receipts mainly serve the purpose of allowing the sender to clear their send queue and handle certain edge cases. But we would love to have a more detailed discussion about whether it might be possible to get rid of them completely. Maybe it is enough to implicitly acknowledge received messages only when an explicit reply is sent, for example?

greyson-signal:
Thanks for your feedback, @kernoelpanic. It's great to see one of the authors of the paper acknowledging some of the limitations with the approaches that have been presented. The assumption that delivery receipts are mostly unnecessary couldn't be further from the truth. The reality is that they serve several important functions that significantly *enhance* security, and they are a crucial feature that users expect to work as quickly and reliably as possible.

The vast majority of delivery receipts on Signal are sent as end-to-end encrypted messages that are completely opaque to the service. If you know someone's [profile key](https://signal.org/blog/signal-profiles-beta/), and are therefore able to exchange messages with them using [sealed sender](https://signal.org/blog/sealed-sender/), then you can also send each other delivery receipts that look just like normal e2ee message traffic.

One of the hallmark features of the Signal Protocol is that message keys "ratchet" forward, with each message [encrypted by a new and unique key](https://signal.org/docs/specifications/doubleratchet/). Because delivery receipts can be encapsulated in a standard Signal message, just like a handcrafted reply, they can also advance the ratchet. This is a substantial improvement that enables key material to rotate far more frequently than it would otherwise.

In the case of transient errors, Signal clients can also ask each other to re-send a message that wasn't delivered successfully. The delivery receipts act as an important checkpoint in this context. Once a message has been delivered, the client can safely ignore any requests to re-deliver the same message. Dropping delivery receipts "for actions that do not trigger observable notifications" (for example, a reaction to someone else's message in a group chat) would make this entire line of defense far more complicated and error-prone.

In the case of a fast-moving conversation, a delay of up to 5 seconds before delivery receipts are sent would negatively affect the perceived and actual performance of the app in ways that would harm the user experience and that could lead people to choose less secure alternatives.

In order to maximize battery life, mobile platforms place extraordinary limitations on the amount of background processing that apps can perform. Introducing randomized client-side delays could easily result in situations where background processes are killed by the OS before a delivery receipt can be sent. Those receipts could be delayed until additional messages are received (and therefore the app gets another limited execution window) or held up until the app is launched into the foreground. Whenever those situations inevitably occurred, the sender would perceive Signal as being broken and slow. It simply isn't a tenable suggestion, from either a UX or technical perspective.

Introducing these delays on the server-side doesn't address the UX issues at all, and as people have already pointed out in this thread it could increase the scope of the metadata that's stored on the service which is something we don't want to do.

The Signal service strives to retain as little metadata as possible, and also goes to great lengths to avoid having access to any sensitive information in the first place. The only way to do this is for the app to process end-to-end encrypted messages directly on your own device instead of our servers. Regarding potential "Resource Exhaustion Attacks" via delivery receipts in Signal, we think it's important to emphasize this quote from the paper:

> For Signal, we were not able to considerably drain the battery of our testing phone (iPhone 13 Pro). Due to considerably stricter rate limits, it only decreased by 1% after an hour of attack.

Many of our competitors store plaintext messages in a database, and they can send message content directly to Apple and Google for efficient push notification delivery that requires zero background processing on your phone. You don't need to wake up an app to decrypt a message if it wasn't encrypted to begin with. These apps will never be mentioned in any papers discussing end-to-end-encrypted delivery receipts and resource exhaustion, but that doesn't make them more secure. Quite the opposite.

For people who want to restrict delivery receipts, Signal already supports disabling phone number discoverability (Settings > Privacy > Phone Number > Who Can Find Me By Number). With this setting enabled, you can choose a random alphanumeric username and no one will be able to send you any messages (delivery receipts or otherwise) unless you share that username with them.

sullystuff:
@greyson-signal thanks for your thoughts, but what about timing attack vulnerabilities? Resource exhaustion isn't a problem, sure, but timing attacks can presently be used to analyze users' behavior. Is there really absolutely no way to mitigate this?

kernoelpanic:
Thank you, @greyson-signal, for your detailed comment.

I strongly encourage continued discussion on this topic, as it's clear that there is no simple, “silver bullet” solution.

I’d like to break down the necessity of delivery receipts again in more detail to ensure everyone (especially me 🙂) is on the same page.
Please correct me if I’m misunderstanding anything:

**1. First, argument for delivery reciepts:**
> they can also advance the ratchet. This is a substantial improvement that enables key material to rotate far more frequently than it would otherwise.

If we assume that every message triggers exactly one delivery receipt, this increases key-material rotation by a constant factor of two. While this is indeed more frequent, I don't think it represents a substantial improvement, especially not one that would meaningfully degrade protocol security if absent.

Especially, since conversations in which no messages are exchanged will stall key rotation regardless; if no messages are sent, no messages can be acknowledged by delivery receipts. So the lack of key rotation in these “low-volume” conversations should be a greater concern than the incremental improvement gained by doubling rotation frequency in active ones due to delivery receipts.

**2. Second argument for delivery reciepts:**

> In the case of transient errors, Signal clients can also ask each other to re-send a message that wasn't delivered successfully. The delivery receipts act as an important checkpoint in this context. Once a message has been delivered, the client can safely ignore any     requests to re-deliver the same message. Dropping delivery receipts "for actions that do not trigger observable notifications" (for example, a reaction to someone else's message in a group chat) would make this entire line of defense far more complicated and error-prone.

The basic argument here seems to be:
Without delivery receipts, corner cases become more complex.

I agree—this is probably true, and additional complexity is never ideal, especially since complexity is the natural enemy of security. But this is exactly why I appreciate this conversation: perhaps there’s a better solution?

Right now, delivery receipts acknowledge the happy path, where clients confirm successful receipt and decryption. But as you pointed out, what the protocol really needs to handle well are the unhappy path scenarios where receipt or decryption fails.

However, since delivery receipts can simply be withheld by a malicious client, they do not provide strong protection against repeated re-delivery requests. Am I misunderstanding something here? Wouldn’t it make more sense to design robust logic specifically for transient- error handling and re-delivery requests, rather than acknowledging every message in the happy path?

Maybe I am wrong, and delivery receipts are the best solution.
If that’s the case, then the key question becomes: What are the privacy tradeoffs, and are they worth it?

meyou69:
Well, this is sadly the end for Signal, at least from my perspective.

The official statement doesn't address any of the security concerns with regards to this timing attack. Even though advancing the ratchet does improve security, that doesn't change the fact that this also allows for this timing attack to be possible.

While it's absolutely true that there is no easy solution to fix this problem since this is an architecturally difficult problem to solve, this just proves that Signal, by design, is flawed from a security perspective.

That, by itself is not a deal breaker, since I'm convinced that the development team is intelligent enough to eventually fix these issues. However, the fact that they clearly seem intent on not treating this as seriously as it needs, pretending that it's not a security flaw, even though it clearly is, and trying to discredit the research paper by framing it having limitations, and trying to sweep everything under the rug is not an acceptable approach from a team that touts to be security centric on a privacy focused app.

As this issue will likely never get resolved, from the looks of it, I have no other recourse but to uninstall the app from my phone. I urge others to do the same as well.

DirkPitt1:
>I have no other recourse but to uninstall the app from my phone. I urge others to do the same as well.

Sure, you can grab your toys and go, where? You behave as if all Signal messages suddenly turned plaintext, which is very far from true, just because the protocol isn't perfect. What better alternatives do you have?

If you want to migrate to other, possibly less secure messenger just because Signal won't scramble to make some hasty fix "for yesterday" (which is a recipe for security disaster), by all means do it. Just spare us this lame blackmail "if you won't do what I want now, I quit and tell all my friends to do the same".

I certainly hope that Signal will consider potential fixes and come up with some good solution. Preferably sooner than later, but if "sooner" should mean "hasty" and "ill-conceived", then later is much better.

NonSus:
> I strongly encourage continued discussion on this topic, as it's clear that there is no simple, “silver bullet” solution.

Strongly agreed, @kernoelpanic.

But if getting rid of delivery receipts or slowing them down is not a feasible option, why not consider other methods proposed in the beforementioned paper, one of them:

- **Improve Client-side Validation:**
> Many of the presented attacks are not possible when clients properly validate the referenced message IDs and thus discard invalid messages (instead of acknowledging them via a delivery receipt).

Why would you confirm a delivery of a message which is crafted for pinging purposes by a modified, unofficial client? Seems like ignoring those messages would not affect every-day UX.

Furthermore, I suggest creating a way to inform the user of the "pings" that they cannot see. This can be achieved via a recent actions list, which would include all the actions history (add a reaction, remove a reaction, malformed message received, etc.) A global list for all received messages and lists for every contact seem comfortable.

In addition, an algorithm should be developed, quoting idea from the paper (Rate Limiting section of mitigations):
> Moreover, receiving an excessive
amount of messages could also be automatically detected by
the receiver and then trigger a UI notification and (temporarily)
block the corresponding phone number.

greyson-signal:
@kernoelpanic

> If we assume that every message triggers exactly one delivery receipt, this increases key-material rotation by a constant factor of two. While this is indeed more frequent, I don't think it represents a substantial improvement, especially not one that would meaningfully degrade protocol security if absent.

Just some background: the Signal Protocol uses a double-ratchet algorithm (well, now with the post-quantum stuff, more like a triple-ratchet). You can read more about it [here](https://signal.org/docs/specifications/doubleratchet/), but there is a very real benefit to having frequent _bi-directional_ messaging. The TL;DR is that if Alice sends 100 messages in a row with no response, those message all live on the same "chain" and could be decrypted with the same key material in a compromise scenario. But when Bob replies, you start a whole new ratchet, and a whole new chain. This is where Signal's forward-secrecy comes from. This is why it is, indeed, very beneficial to have client-sent delivery receipts. You have much shorter message chains, even when only one side is initiating message sends.

> Especially, since conversations in which no messages are exchanged will stall key rotation regardless; if no messages are sent, no messages can be acknowledged by delivery receipts. So the lack of key rotation in these “low-volume” conversations should be a greater concern than the incremental improvement gained by doubling rotation frequency in active ones due to delivery receipts.

Not exactly clear what you're saying here. If no messages are sent, then there's nothing to be compromised? Like, if two people don't talk for a year, that's ok. You don't need to be ratcheting key material for messages that aren't sent. You'll ratchet the next time you _do_ send.

> However, since delivery receipts can simply be withheld by a malicious client, they do not provide strong protection against repeated re-delivery requests. Am I misunderstanding something here? 

The system for recovering from decryption errors and resending messages is too complex to explain in a single comment, but if a client is configured to never send delivery receipts, sure, they could request the message again later one time (with how the system works). That's not the concern. The concern would be someone getting access to all of a user's key material, but not their messages, and then being able to request all the messages to be resent to them. If the target in question has a normal client with normal delivery receipt behavior, then this greatly mitigates what an attacker could do.

The long and short of it is, that if we decided not to send delivery receipts for those messages, we'd also have to stop making those messages resendable, which would be a UX regression that we don't want to make.

We understand that Signal cannot be everything to everybody. If people are concerned about this, as previously mentioned, you can disable discovery in the settings. If someone's threat model includes concerns about network activity _in general_ (because let's be clear, there are many other potential vectors to consider depending on your specific threat model), we recommend using a VPN.

Hopefully this is helpful. I wanted to respond to your initial concerns and explain why the approach that was suggested in this pull request wouldn't work. But we don't typically use GitHub for these types of discussions.
https://github.com/signalapp/Signal-Android/blob/main/CONTRIBUTING.md#github-is-not-a-generic-discussion-forum

Whatnoww:
Hi Greyson,

It's expected that for Signal Connections, the server wouldn't be able to stop delivery receipts, but for non-Signal connections, where the server is always the one generating the delivery receipt (due to the attacker not having a profile key), there could be some work done to throttle it there, and would probably substantially reduce these type's of attacks, since most Signal users aren't communicating with threat actors that are trying to track device usage via delivery receipts.

Perhaps there is room here for some improvements to prevent someone from blanket scanning clients (not withstanding general server throttling):

https://github.com/signalapp/Signal-Server/blob/main/service/src/main/java/org/whispersystems/textsecuregcm/push/ReceiptSender.java

https://github.com/signalapp/Signal-Server/blob/9c4047a90bee044255fdcf7c5c2e59f89f1ff5e8/service/src/main/java/org/whispersystems/textsecuregcm/WhisperServerService.java#L557

beantaco:
> For people who want to restrict delivery receipts, Signal already supports disabling phone number discoverability (Settings > Privacy > Phone Number > Who Can Find Me By Number). With this setting enabled, you can choose a random alphanumeric username and no one will be able to send you any messages (delivery receipts or otherwise) unless you share that username with them.

[The suggested setting](https://support.signal.org/hc/en-us/articles/6829998083994-Phone-Number-Privacy-and-Usernames-Deeper-Dive) is intended to enable/disable contact discovery in the case someone already knows a Signal user’s phone number. Does disabling this really stop a *spooky stranger* who already knows the target’s phone number (or username) from extracting delivery receipts from their target **using a custom client**?

sullystuff:
> > For people who want to restrict delivery receipts, Signal already supports disabling phone number discoverability (Settings > Privacy > Phone Number > Who Can Find Me By Number). With this setting enabled, you can choose a random alphanumeric username and no one will be able to send you any messages (delivery receipts or otherwise) unless you share that username with them.
> 
> [The suggested setting](https://support.signal.org/hc/en-us/articles/6829998083994-Phone-Number-Privacy-and-Usernames-Deeper-Dive) is intended to enable/disable contact discovery in the case someone already knows a Signal user’s phone number. Does disabling this really stop a _spooky stranger_ who already knows the target’s phone number (or username) from extracting delivery receipts from their target **using a custom client**?

exactly, it does not.. my proposed changes only protect against sealed-sender attackers, but if an attacker is blocked and loses sealed sender they can just send non-sealed-sender messages with the server-provided delivered receipts (so my fixes don't help much). I'm exploring ways to fully reject messages from blocked contacts, but I'm not positive it can be done since we don't know a packet's content until decryption, and the packet acknowledgement is done at the network layer which we don't have control over.

DianaNites:
Wait, you're saying its possible to get somebodies signal account and send it messages, from only their phone number, even if phone number visibility and/or discovery is off?

kernoelpanic:
Thank you for your reply, @greyson-signal.

> TL;DR is that if Alice sends 100 messages in a row with no response, those messages all live on the same "chain" and could be decrypted with the same key material in a compromise scenario. But when Bob replies, you start a whole new ratchet, and a whole new chain. This is where Signal's forward-secrecy comes from. This is why it is, indeed, very beneficial to have client-sent delivery receipts. You have much shorter message chains, even when only one side is initiating message sends.

As stated previously, I agree that there is a benefit. My question concerns the magnitude of this benefit. How many real-world conversations are actually that unidirectional? Could this be mitigated in other ways, such as sending a delivery receipt only after N unidirectional messages (e.g., every 100) to advance the ratchet? If each delivery receipt triggers a visible UI notification, this would reduce the attack surface for misuse of delivery receipts already quite a bit.

More broadly, I recognize that this is a difficult design decision, which is unlikely to have a one-size-fits-all answer. However, the current approach exposes all users to non-disableable delivery receipts, which can be used for tracking purposes and have also been cited as a concern in the context of sealed sender (see: [https://www.cs.umd.edu/~kaptchuk/publications/ndss21.pdf](https://www.cs.umd.edu/~kaptchuk/publications/ndss21.pdf)), versus maintaining the strongest forward-secrecy guarantees, which rely on a complete key compromise - a hopefully rare scenario. My point is that, given the tradeoffs, it may be reasonable to allow expert users to decide per conversation what they value most.

> Not exactly clear what you're saying here. If no messages are sent, then there's nothing to be compromised.

My point is the same as above (as you described) : even when no new messages are sent, the previously sent message(s) (even those deleted via disappearing messages) remain susceptible to recovery if an attacker recorded them in transit and the key material is subsequently compromised.

> The concern would be someone getting access to all of a user's key material, but not their messages, and then being able to request all the messages to be resent to them.

As mentioned in my initial post, this could be mitigated by implicitly acknowledging all previously unacknowledged messages whenever an explicit reply is sent. At that point, all earlier messages can be treated as delivered, and retransmission can be restricted.

If the goal is to eliminate end-to-end delivery receipts entirely, one approach would be to only have delivery receipts between the client and the server. 
The sender would receive a receipt from the server confirming that the message was stored. The server would then retain the message until it receives a delivery receipt from the client that the message was received. Under this approach, messages cannot be silently lost. If a decryption error occurs at the receiver the client has to request retransmission anyway with a dedicated message. 

So from my current perspective, removing end-to-end delivery receipts is technically feasible. The real question is whether this is worthwhile for Signal. I assume the change may be unpopular due to the high implementation effort relative to benefits that are difficult to quantify. Still, I believe an open discussion is necessary to reach an informed decision.

I am happy to continue the discussion on the community forum as suggested. Is there a preferred place to post this, or is there an existing thread on the topic?

greyson-signal:
@Whatnoww 

> It's expected that for Signal Connections, the server wouldn't be able to stop delivery receipts, but for non-Signal connections, where the server is always the one generating the delivery receipt (due to the attacker not having a profile key), there could be some work done to throttle it there, and would probably substantially reduce these type's of attacks, since most Signal users aren't communicating with threat actors that are trying to track device usage via delivery receipts.

You'll be happy to know that _sending_ non-sealed-sender messages is pretty heavily rate-limited already. Has been for the past year or two. I agree that with you that this should already mitigate most of the concern here, for the reasons you've mentioned.

@kernoelpanic 

> even when no new messages are sent, the previously sent message(s) (even those deleted via disappearing messages) remain susceptible to recovery if an attacker recorded them in transit and the key material is subsequently compromised.

Because of the double-ratcheting, only the messages on the most recent chain could be decrypted, and the ratcheting you get from delivery receipts makes these chains 1 message long in the common case. Removing the delivery receipts would substantially increase the length of the average chain.

> As mentioned in my initial post, this could be mitigated by implicitly acknowledging all previously unacknowledged messages whenever an explicit reply is sent. At that point, all earlier messages can be treated as delivered, and retransmission can be restricted.

Just because a user has sent you a message afterwards doesn't mean they successfully-decrypted all previously-received messages. A sender could have sent a single bad message in the past, and there's other catalysts for key rotation that can occur besides receiving replies, resulting in situations where you can decrypt more recent messages from a user but not older ones.

> If the goal is to eliminate end-to-end delivery receipts entirely, one approach would be to only have delivery receipts between the client and the server.

Server-managed delivery receipts are not possible when using [Sealed Sender](https://signal.org/blog/sealed-sender/), which represents the vast majority of message traffic. The server does not know who the sender was, and therefore cannot provide a delivery receipt.

> The sender would receive a receipt from the server confirming that the message was stored.

Maybe the term "receipt" is getting overloaded here -- just getting a 200 from the server for your message send implies that the message was stored, no additional receipt is necessary for that information.

>  The server would then retain the message until it receives a delivery receipt from the client that the message was received. Under this approach, messages cannot be silently lost.

This is effectively what happens today. The receiver pulls messages from their message queue, acking each message they received so that the server knows they can discard it.

> If a decryption error occurs at the receiver the client has to request retransmission anyway with a dedicated message.

Indeed, and they have to request re-encryption from the sender, since whatever was sent to them before could not be decrypted.

At the end of this suggestion I'm left a little confused: this doesn't just eliminate "end-to-end delivery receipts", it eliminates any concept of a delivery receipt as it exists in the app today. Like, the sender can't know when the receiver actually received their message.

We have no intention of removing the concept of delivery receipts from Signal. If the point of this suggestion was just to show that it's theoretically possible to build an app without delivery receipts, sure, not saying it's impossible, just that we would find the compromises unacceptable. I still hold that if people are concerned about these types of issues, they're better off changing Who Can Find Me By Number > Nobody and/or using a VPN.

I do appreciate the suggestions! I don't want people to feel like we're not listening, which is why I'm doing my best to respond point-by-point. But I don't currently see us making any changes in this area right now for the reasons I've given.

> I am happy to continue the discussion on the community forum as suggested. Is there a preferred place to post this, or is there an existing thread on the topic?

You can create a topic here: https://community.signalusers.org/c/general/7

beantaco:
@greyson-signal

*Improving client-side validation* was proposed as a mitigation of abuse of delivery receipts. I have some questions regarding client-side validation. What client-side validation do official Signal clients perform on the messages they receive? If no/minimal validation is performed by official Signal clients at this stage, what is the Signal team's thoughts on introducing/improving client-side validation and discarding (i.e. not sending delivery receipts in response to) received invalid messages?

PR #14529, opened 2026-01-14
Update strings.xml

benediktjohannes:
Fixed typing issue (in German it has to be "bei einer unterstützten" with another "t" (I'm from Germany))


mtang-signal:
Thanks for the pr! We use an external translation management system so we would need to change it there in order for the change to persist. But thanks to you, I have flagged this issue to them and I expect it'll be fixed in the next release.

laidbackindustries:
Fixed, thanks!

PR #14530, opened 2026-01-14
Update strings.xml

benediktjohannes:
Fixed missing "." at the end of the sentence


mtang-signal:
Same comment as in #14529 (we do not change these files directly so we need to change the original source). However, always feel free to open issues in this repo for future string issues.

laidbackindustries:
Fixed, thanks!

PR #14541, opened 2026-01-16
Update strings.xml

benediktjohannes:
Fixes typing: IMPORTANT this is not a translation change, this is an internal change of the value field for the translations

Contributor checklist
[ x] I am following the [Code Style Guidelines](https://github.com/signalapp/Signal-Android/wiki/Code-Style-Guidelines)
[ x] My contribution is fully baked and ready to be merged as is
[ x] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the Fixes #1234 [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)
Description
Fixes typing: Important this is not a translation change, this is an internal change of the value field for the translations - another important thing to mention: This (probably) has to be changed in all language translation files as well and also in the general value context if it's wrong there as well and copied


mtang-signal (commented):
Same thing here - will need to update where in the codebase this is called from (I think a fragment). All the translated files will automatically update so nothing further for you to do on that specific avenue.

Issue #14618, opened 2026-02-21
Unexpected keyboard appearance on repeated back swipe


Eclipse-7:
### Guidelines

- [x] I have searched searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] This isn't a feature request or a discussion topic

### Bug description

While navigating back from a chat, the keyboard appears if the back gesture is held briefly, then released, and the gesture is swiped again. This issue does not occur on the main chats list screen.

### Screenshots

Here’s the screen recording: [Drive](https://drive.proton.me/urls/A0Z0DMMRB8#zqD1PMHVU6nq)

### Device

Pixel 9

### Android version

16

### Signal version

Version 7.73.2

jeffrey-signal:
Fixed in version 8.5 - thanks @DivyaKhunt07 🙂

PR #14633, opened 2026-02-27
Fix unexpected keyboard appearance on repeated back swipe (#14618)

DivyaKhunt07:

### Contributor checklist

Virtual device Pixel 9, Android OS 36
[x] My contribution is fully baked and ready to be merged as is
[x] I ensure that all the open issues my contribution fixes are mentioned in the commit message of my first commit using the `Fixes #1234` [syntax](https://help.github.com/articles/closing-issues-via-commit-messages/)

----------

### Description
**Problem**
When navigating back from a conversation screen using Android's predictive back gesture, if the gesture was held briefly and then cancelled, the keyboard would unexpectedly appear on the subsequent back swipe attempt.

**Root Cause:**
Android's predictive back gesture cancellation internally restores focus to previously focused views (in this case, composeText). The onFocusChange listener was showing the IME whenever focus was gained, so this system-driven focus restoration was unintentionally triggering the keyboard to appear — even when the user had no intent to type.

**Solution**
Decouple IME visibility from focus changes entirely.

- onFocusChange — Removed the IME show call. Focus can be legitimately restored by the system (e.g., during predictive back cancellations) without the user intending to open the keyboard. The keyboard should only appear from explicit user or programmatic actions.
- Popup case — Replaced composeText.requestFocus() with container.showSoftkey(composeText) to explicitly show the keyboard where it is intentionally required, rather than relying on the focus change callback as a side effect.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/conversation/v2/ConversationFragment.kt
@@ -4520,9 +4520,6 @@ class ConversationFragment :
       }
     }
 
-    override fun onFocusChange(v: View, hasFocus: Boolean) {
-    }
-
     override fun onCursorPositionChanged(start: Int, end: Int) {
       linkPreviewViewModel.onTextChanged(composeText.textTrimmed.toString(), start, end)
     }


JesseWeinstein:
Just a small note -- commented out code shouldn't be in a "fully baked" PR. Better to actually remove it; we have the revision history. :-)

DivyaKhunt07:
Thanks for the feedback! I've removed the commented-out code and updated the PR.

JesseWeinstein:
> Thanks for the feedback! 
Sure! Is the override still needed? If not, you can remove the whole method.

DivyaKhunt07:
Hi @alex-signal, I've addressed all the review feedback — removed the commented-out code and the unused override method. The PR should be ready for review whenever you get a chance. Happy to make any further adjustments if needed. Thanks! 🙏

PR #14644, opened 2026-03-04
fix(database): make ThreadTable.getCurrent() return non-null ThreadRecord

### LINKED ISSUE #14641: Nullability conflict between ThreadTable.getCurrent & MessageSearchResult
`getCurrent` returns a nullable `ThreadRecord`: 
https://github.com/signalapp/Signal-Android/blob/66f0470960596850ab4d65d076105aef68be0b3d/app/src/main/java/org/thoughtcrime/securesms/database/ThreadTable.kt#L2260

But `MessageSearchResult` expects a list of non-null `ThreadRecord`s:
https://github.com/signalapp/Signal-Android/blob/8b2aeba3bdda6543798b01cc4739ef26d2a862ee/app/src/main/java/org/thoughtcrime/securesms/search/MessageSearchResult.kt#L3

This hasn't been noticed yet because between them is [SearchRepository](https://github.com/signalapp/Signal-Android/blob/8b2aeba3bdda6543798b01cc4739ef26d2a862ee/app/src/main/java/org/thoughtcrime/securesms/search/SearchRepository.java) which is still in Java and lacks sufficient nullability tracking to show the problem.

`getCurrent` only returns null on errors, so we could fix it by just filtering out the nulls -- but I'm not sure if that's the Right Thing.

MohamedMBG:
Fixes #14641

## Problem

ThreadTable.StaticReader.getCurrent() was declared as returning `ThreadRecord?` after the Kotlin migration.  
However, `SearchRepository` (still written in Java) consumes the value as a non-null `ThreadRecord` through `ThreadModelBuilder.build()`.

Because Java does not enforce Kotlin nullability, this mismatch could allow a `null` value to propagate into the search pipeline. If a parsing error occurred while reading cursor values, it could result in:

- a `NullPointerException` when adding the result to `List<ThreadRecord>`
- downstream crashes in `ThreadSearchResult` or `MessageSearchResult`, which assume non-null records.

## Root Cause

The original Java implementation never returned `null`. Instead, it always constructed a `ThreadRecord` and degraded individual fields (e.g. body or recipient) when parsing failed.

During the Kotlin migration, the method was conservatively typed as `ThreadRecord?`, even though the implementation never explicitly returned `null`. The real risk was uncaught exceptions during cursor parsing.

## Solution

Ensure `getCurrent()` always returns a non-null `ThreadRecord`.

Changes:
- Change return type from `ThreadRecord?` → `ThreadRecord`
- Wrap parsing logic in a `try/catch`
- On failure:
  - log a warning
  - return a fallback `ThreadRecord` with safe defaults:
    - `threadId` from cursor if available, otherwise `0L`
    - empty body `""`
    - `Recipient.UNKNOWN`
    - other fields left as builder defaults

This restores the behavior of the original Java implementation and ensures the search pipeline always receives valid records.

## Why not filter nulls in SearchRepository?

Filtering nulls in `SearchRepository.readToList()` would silently drop search results and hide the underlying parsing issue. Fixing the contract at the source ensures safer and more predictable behavior.

## Testing

Added tests covering:

- valid thread parsing
- reader behavior through cursor iteration
- malformed cursor schemas
- fallback record behavior when parsing fails

These tests ensure `getCurrent()` never returns `null` and that degraded records remain safe to consume.

### DIFF
--- a/app/src/androidTest/java/org/thoughtcrime/securesms/database/ThreadTableTest_getCurrent.kt
+++ b/app/src/androidTest/java/org/thoughtcrime/securesms/database/ThreadTableTest_getCurrent.kt
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2024 Signal Messenger, LLC
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+package org.thoughtcrime.securesms.database
+
+import android.database.MatrixCursor
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.signal.core.models.ServiceId.ACI
+import org.thoughtcrime.securesms.database.model.ThreadRecord
+import org.thoughtcrime.securesms.recipients.Recipient
+import org.thoughtcrime.securesms.testing.SignalDatabaseRule
+import java.util.UUID
+
+/**
+ * Tests for [ThreadTable.StaticReader.getCurrent] nullability contract.
+ *
+ * Validates that getCurrent() always returns a non-null [ThreadRecord],
+ * restoring the behavior expected by [SearchRepository] and [MessageSearchResult].
+ *
+ * See: https://github.com/signalapp/Signal-Android/issues/14641
+ */
+@Suppress("ClassName")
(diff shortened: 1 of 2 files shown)


greyson-signal:
I'm fine with just updating the return type to make it non-nullable, the rest of the change looks to be completely unnecessary. This should be a one-character diff.

JesseWeinstein:
Thanks! Delightful to have someone else work on a fix before I got around to it. But upon further inspection, I misunderstood the code -- it *DOESN'T* actually ever return null; it just had the wrong annotation. It compiles fine without the question mark -- so that's all that should be changed.

PR #14668, opened 2026-03-16
Remove NowhereBufferedSink

JesseWeinstein:
### Contributor checklist

- [ ] I have tested my contribution on these devices
- N/A I ensured that all the open issues my contribution fixes are mentioned in the commit message
----------

### Description
The `NowhereBufferedSink` class violates the closed hierarchy guideline, because `BufferedSink` is a sealed interface. Technically implementing a sealed interface in a different package is allowed (unlike inheriting from a sealed class), but it's a bad idea.

It is replaced by inlining the no-op `OutputStream` directly in `DigestingRequestBody`, accessed via a new method, `writeToNowhere`.

I still need to get my test environment up and running in order to test this -- but it *seems* safe...


lib/libsignal-service/src/main/java/org/whispersystems/signalservice/internal/push/http/DigestingRequestBody.kt
```diff
@@ -40,10 +41,38 @@ class DigestingRequestBody(
     return contentType.toMediaTypeOrNull()
   }
 
+  /**
+   * nowhereOutputStream allows a programmer to write out data into the void. This has no memory
+   * implications, as we don't actually store bytes.
+   */
+  val nowhereOutputStream = object : OutputStream() {
```
jeffrey-signal: for android SDK >= 33, we should be able to use `OutputStream.nullOutputStream()`

jeffrey-signal:
```suggestion
  private val nowhereOutputStream = object : OutputStream() {
```

lib/libsignal-service/src/main/java/org/whispersystems/signalservice/internal/push/http/DigestingRequestBody.kt
```diff
@@ -40,10 +41,38 @@ class DigestingRequestBody(
...
+    @Throws(IOException::class)
+    override fun write(i: Int) {
+    }
+
+    @Throws(IOException::class)
+    override fun write(bytes: ByteArray?) {
```
jeffrey-signal:
```suggestion
    override fun write(bytes: ByteArray) {
```

lib/libsignal-service/src/main/java/org/whispersystems/signalservice/internal/push/http/DigestingRequestBody.kt
```diff
@@ -40,10 +41,38 @@ class DigestingRequestBody(
...
+    }
+
+    @Throws(IOException::class)
+    override fun write(bytes: ByteArray?, i: Int, i1: Int) {
```
jeffrey-signal:
```suggestion
    override fun write(bytes: ByteArray, offset: Int, length: Int) {
```

jeffrey-signal (changes requested):

jeffrey-signal (commented):
> I still need to get my test environment up and running in order to test this -- but it seems safe...

Has this been tested?

lib/libsignal-service/src/main/java/org/whispersystems/signalservice/internal/push/http/DigestingRequestBody.kt
JesseWeinstein: Irrelevant due to using nullOutputStream()

lib/libsignal-service/src/main/java/org/whispersystems/signalservice/internal/push/http/DigestingRequestBody.kt
JesseWeinstein: Irrelevant due to using nullOutputStream()

JesseWeinstein:
> > I still need to get my test environment up and running in order to test this -- but it seems safe...
> 
> Has this been tested?

Looking into it further, I'm not sure how best to test this! It looks like it is used for uploading attachments and backup files, and only when it tries to resume an already uploaded file. Suggestions are welcome.

PR #14695, opened 2026-04-01
Fix duplicate word in bug report template

leoheitmannruiz:
As in signalapp/Signal-Desktop#7571.

I don't intend for this PR to be merged, but for it to notify you of the duplicate word. Thanks for your work!


jeffrey-signal (approved):
Fixed, thanks!

PR #14856, opened 2026-06-27
Move task defaults into build logic plugins

maksym-moroz:
## Summary
- Register build-logic conventions as binary plugins instead of precompiled scripts.
- Add a dedicated `conventionPlugins` version catalog for in-repo convention plugin aliases.
- Move shared Javadoc/Test defaults into `signal-build-task-conventions` and apply that convention through existing conventions or explicit catalog aliases.
- Keep build-logic implementation details internal/private where Gradle allows it.

## Notes
- Existing dependency-verification project traversal is unchanged; this only removes the root Javadoc/Test cross-configuration.


build.gradle.kts
```diff
@@ -48,20 +48,6 @@ tasks.withType<Wrapper> {
...
-        (options as StandardJavadocDocletOptions).addStringOption("Xdoclint:none", "-quiet")
-      }
-    }
-  }
-
-  tasks.withType<Test>().configureEach {
-    maxParallelForks = (Runtime.getRuntime().availableProcessors() / 4).coerceAtLeast(1)
-  }
-}
-
```
maksym-moroz, line 64: This is the main win. There are a lot of cross-configurations in the repo and it's not feasible to address them in one sweep but it's possible to lessen the amount gradually, decouple the structure to get parallel compilations gains and be ready to move to isolated builds upcoming Gradle feature

PR #14867, opened 2026-07-07
Fix black bars around circular notification avatars

### LINKED ISSUE #14811: Black bars around notification images
### Guidelines

### Bug description

There are black bars around contact photos or group images in notifications. This might be related to One UI.

### Screenshots

Notice the black bars around the circular images:
Image
The top notification is from a group, the bottom one from a contact.

### Device

Samsung Galaxy A34

### Android version

16

### Signal version

8.11.5

### Link to debug log

_No response_

dweep-js:
Uses the app's own square contact_photo_target_size dimension for the Glide submit() call instead of the OEM-overridable system dimensions android.R.dimen.notification_large_icon_width/height. Those system dimensions aren't guaranteed to be square, so CircleCrop() could produce a rectangular bitmap with transparent letterbox padding, which some OEM skins render as black bars instead of transparency.

Fixes #14811

 * Realme 15 pro, Android 16

----------

### Description
While I was investigating this, I stepped through the rendering of the notification avatar and noticed Recipient.getContactDrawable() (in NotificationExtensions.kt) issues a Glide request using the system dimensions of android.R.dimen.notificationlargeiconwidth and height. These are dimensions that OEM can override, and there’s nothing to say they are equal. Since CircleCrop() only inscribes a circle inside whatever box is provided, an imperfectly shaped box leaves transparency letterboxed around the circle. This bitmap is then almost entirely unchanged and handed off to setLargeIcon() (the resizing call in toLargeBitmap()/BitmapUtil.createFromDrawable() doesn’t happen for BitmapDrawable, so the padding carries through). On some OEM notification-shade skins, this transparent area seems to display as black, which is a plausible explanation for the black bars in the initial report.
The proposed fix uses the app’s own square contactphototarget_size value (which is already being used later in the notification icon pipeline) instead of the OEM-supplied system values. That way, the image is already square when it’s given to Glide, so CircleCrop() never produces any letterboxing.

Testing status

I’ve built and installed this code after applying the fix onto my Realme device (Android 16) and can see that notification icons are now displayed without the black bars. I don’t know whether the bug would have reproduced on this device prior to the fix though, so I cannot definitively confirm that this device was one affected by the issue – the report mentioned the bug as being on a Samsung Galaxy A34 on One UI. It would be useful if the original reporter or anyone with a Samsung/One UI device could check that this solves the issue for them as well.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/notifications/v2/NotificationExtensions.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/notifications/v2/NotificationExtensions.kt
@@ -49,8 +49,8 @@ fun Recipient.getContactDrawable(context: Context): Drawable? {
         .diskCacheStrategy(DiskCacheStrategy.ALL)
         .transform(MultiTransformation(listOf(CircleCrop())))
         .submit(
-          context.resources.getDimensionPixelSize(android.R.dimen.notification_large_icon_width),
-          context.resources.getDimensionPixelSize(android.R.dimen.notification_large_icon_height)
+          context.resources.getDimensionPixelSize(R.dimen.contact_photo_target_size),
+          context.resources.getDimensionPixelSize(R.dimen.contact_photo_target_size)
         )
         .get(IMAGE_LOAD_TIMEOUT_SECONDS, TimeUnit.SECONDS)
     } catch (e: InterruptedException) {


alex-signal:
Hi! Thanks, this was pulled into 8.19 and should go out to beta this week.

PR #14951, opened 2026-08-18
Render UltraHDR images in the full-screen media preview

MajorMurphy:

### Contributor checklist

 * Samsung Galaxy S26 Ultra, Android 16, HDR 
 * Google Pixel 8 Pro, Android 16, HDR
 * Google Pixel 2, Android 11, SDR

----------

### Issue
Signal Android already sends and receives UltraHDR (gain map) JPEGs intact; the same images render in HDR on Signal Desktop for Windows and macOS. Android displayed only their SDR base image.

### Fix
Decoding needed no change. BitmapFactory and BitmapRegionDecoder already attach a Gainmap on API 34+, and SubsamplingScaleImageView already draws it to the hardware canvas. A gain map is simply not applied unless the activity opts its window in via Window.setColorMode(COLOR_MODE_HDR).

The decoders now report when a decoded bitmap carries a gain map, the preview records to which media that applies, and MediaPreviewActivity raises the window color mode while such an image is the current page, clearing it on stop and on exit. Window.setDesiredHdrHeadroom() caps the headroom so SDR chrome sharing the window does not dim; that API is why the feature requires Android 15 (API 35) rather than 14.

### Notes
Scope is the full-screen preview only. Conversation thumbnails still render SDR -- Glide's transformations copy the bitmap through a Canvas, which drops the gain map -- and would need a gain-map-preserving decode path plus an HDR window for the whole conversation list.

Note that images must be sent at the high-quality setting to retain their gain map; at standard quality they are rescaled during compression and the gain map is lost.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/components/ZoomingImageView.java
+++ b/app/src/main/java/org/thoughtcrime/securesms/components/ZoomingImageView.java
@@ -8,6 +8,7 @@ import android.view.MotionEvent;
 import android.view.View;
 import android.widget.FrameLayout;
 
+import androidx.annotation.MainThread;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.exifinterface.media.ExifInterface;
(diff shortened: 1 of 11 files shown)


cody-signal:
Thanks, we're taking a look

alex-signal:
Hi! this has been merged and will go out in 8.24

Issue #12324, opened 2022-06-30
Sharing contacts file (vcf) does not work: is interpreted, and only very partially.


knarrff:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended
- [x] I have read https://github.com/signalapp/Signal-Android/wiki/Submitting-useful-bug-reports
- [x] This isn't a feature request or a discussion topic

----------------------------------------

### Bug description
Sending a vcf file containing all of my contacts is instead of sending as-is interpreted and only one contact info is used and sent. On the receiving side I could not save that file, just display that one contact.

### Steps to reproduce
- Export your contacts using the Contacts app into a vcf file and save on-device
- Share that file with the Signal app
- Select one receiver, even 'note to self' works for testing

**Actual result:
- Already on the sending side, the file seems to be interpreted instead of being sent as-is.
- Even the interpretation only lists one of the many contacts contained in the file.

**Expected result:
- I expected Signal to simply send that file as-is.
- If Signal cannot deal (import/export) vcf files with multiple contacts inside, it should not try to interpret them, but send them as they are, or maybe give an option to do so. That way, the receiver would be able to save the file and use it for an import in the contacts app. As is, Signal cannot be used to share contact lists.

### Device info
**Device:** Samsung SM-J330F
**Android version:** 9
**Signal version:** 5.41.11

knarrff:
This issue has not seen updates because I had written everything in the original report and no developer reported so far. Why should this be a reason to close this report? Am I supposed to "ping" the issue every now and then to say "still an issue"?

knarrff:
This issue has not seen updates because I had written everything in the original report and no developer reported so far. Why should this be a reason to close this report? Am I supposed to "ping" the issue every now and then to say "still an issue"?

Issue #6520, opened 2017-04-11
send and receive contacts (vCard) properly


Ricain:
I have:
- [X] read https://github.com/WhisperSystems/Signal-Android/wiki/Submitting-useful-bug-reports
- [X] searched open and closed issues for duplicates

----------------------------------------

Hi :smile: 

This is a fresh feature request for #535 and #1407.

|Signal|Android Message|Android Message detailed|
|-----|--------|--------|
|![screenshot_20170411-143016](https://cloud.githubusercontent.com/assets/4070933/24909498/69b92262-1ec4-11e7-8b42-9f66ac8ea3bd.png)|![screenshot_20170411-142238](https://cloud.githubusercontent.com/assets/4070933/24909503/73a02884-1ec4-11e7-9a21-a67a8a5363bb.png)|![screenshot_20170411-142251](https://cloud.githubusercontent.com/assets/4070933/24909799/c4c50eb8-1ec5-11e7-88b9-31c361565416.png)|

Sharing contacts with Signal will extract only phone number and not any other information such as emails, dates, or contact picture. Worse, it won't do anything if a contact does not contain any phone numbers.

**Can you add the ability to share contacts in vCard format with an easy way to import/export them from Signal app?**

A good example of how it could work is Android Message (see screenshots).

Things worth mentioning:

- Looking at privacy, it would be really nice to be able to share sensitive contacts this way with all metadata included.
- vCard format is not something easy: https://tools.ietf.org/html/rfc6350
- Some implementations exist: https://github.com/mangstadt/ez-vcard
- Most SMS/MMS applications support this: Signal is late.
- Opening a text editor and copying or making your own `vcf` file is not a solution.
- This has been requested many times: #535, #1407, #922, #2504, #3808, #4205, #5075, #5612, #499, #2306...

ByteHamster:
You already provide a list of issues that your issue is a duplicate of. Why do you even create a new one then?

VikOlliver:
Just ran into this one myself, while trying to show someone how wonderful Signal is and why they should use it. They decided not to. There's your problem.

travisfw:
I message with an iphone user often, and I can't even copy the message into a text file to make it work. I can import other vcf files, for instance, when sent by email. Don't know why it doesn't work coming through Signal.

This is what comes out of signal and fails to import as a vcf file (anonymized):

    ﻿BEGIN:VCARD
    VERSION:3.0
[...]

PR #14958, opened 2026-08-22
Handle shares of content URIs without permission grants gracefully

### LINKED ISSUE #14617: Sharing a contact via Android share menu results in 'invalid file' on GrapheneOS
### Bug description

When sharing a single contact from the AOSP/GrapheneOS Contacts app to Signal via Android's share menu (ACTION_SEND with `text/x-vcard` MIME type), Signal reports the file as invalid and does not send the contact.

**Steps to reproduce:**
1. Open Contacts app on GrapheneOS (stock AOSP-based contacts app)
2. Open any contact
3. Tap Share > choose Signal
4. Signal shows "invalid file" or silently fails to attach the contact

**Expected:** The contact is shared as a vCard attachment in the Signal conversation.

**Actual:** Signal rejects the file as invalid.

**Key observation:** Sharing the same contact to SMS/MMS apps works perfectly. The vCard data itself is valid. The issue appears to be in how Signal resolves the shared content URI.

**Likely root cause:** The Contacts app shares a `content://com.android.contacts/contacts/as_vcard/` URI. Signal's `ShareRepository.resolve()` calls `contentResolver.openInputStream()` and `contentResolver.query()` on this URI. If Signal does not have `READ_CONTACTS` permission (common on GrapheneOS, where users often deny this or use Contact Scopes), the URI read fails with a `SecurityException` that Signal catches silently, resulting in `ResolvedShareData.Failure`.

Additionally, the ContactsProvider returns `NULL` for the `SIZE` column on vCard URIs (`OpenableColumns.SIZE` maps to `"NULL"` in `sContactsVCardProjectionMap`). If Signal's size validation treats zero/null size as invalid, this would also cause rejection.

**Note:** A complementary fix has been submitted on the GrapheneOS Contacts side to explicitly add `FLAG_GRANT_READ_URI_PERMISSION` to the share intent ([GrapheneOS/platform_packages_apps_Contacts#25](https://github.com/GrapheneOS/platform_packages_apps_Contacts/pull/25)). However, Signal should also handle this case gracefully, as other apps using the same AOSP share pattern will have the same issue.

### Related issues
- #12324 - "Sharing contacts file (vcf) does not work"
- #6520 - "send and receive contacts (vCard) properly"

### Device

Pixel (GrapheneOS)

### Android version

15

### Signal version

Latest from accrescent/GitHub releases

tuncoglu:

### Contributor checklist

 * Virtual device Pixel 8 API 37.1, Android 17

----------

Related issue: #14617

## Problem

When an app shares a `content://` URI **without** `FLAG_GRANT_READ_URI_PERMISSION` (e.g. AOSP/GrapheneOS Contacts sharing a vCard, see #14617), Signal cannot read the URI unless it holds the provider's own permission (e.g. `READ_CONTACTS`).

The failure mode was bad:
- `ShareRepository.getSize()`/`getFileName()`/`getMimeType()` call `ContentResolver.query()` / `getType()` with **no exception handling**, so a `SecurityException` propagated as an RxJava error instead of a share failure.
- Either way, `ShareActivity` just called `finish()` — the user got a share sheet that flashes and closes with **zero feedback**.

## What changed

**ShareRepository** — all `ContentResolver` reads are wrapped:
- `query()` (size/file name), `getType()` (mime type), and `openInputStream()` now catch `SecurityException` / `IOException`.
- `SecurityException` produces `ResolvedShareData.Failure(ShareError.ACCESS_DENIED)`; other failures produce `ShareError.UNKNOWN`.
- Covers both single-share and multi-share resolvers.

**UX** — the silent `finish()` is replaced with a toast before finishing:
- `ACCESS_DENIED` → "Signal couldn't access the shared content. The sending app didn't grant the necessary permission."
- `UNKNOWN` → "Signal couldn't load the shared content."

**Tests** — new `ShareRepositoryTest` (Robolectric) covering:
- `SecurityException` from the resolver → `Failure(ACCESS_DENIED)` (the vCard scenario from #14617)
- `IOException` → `Failure(UNKNOWN)`
- readable vCard URI → resolved `ExternalUri` with correct mime type

## Notes

This is a **defensive/UX fix** on Signal's side. When the sending app doesn't grant URI access and Signal lacks the provider permission, the share physically cannot be read (platform limitation) — the complementary sender-side fix is GrapheneOS/platform_packages_apps_Contacts#25. With this change, users at least learn why instead of seeing a silent flash, and other apps sharing the same way don't produce confusing silent failures.

**Multi-share behavior:** if *any* selected item's provider denies URI access (at `getType()` or `openInputStream()`), the whole share fails with the `ACCESS_DENIED` toast instead of silently sharing the remaining subset.

Fixes #14617 behaviourally: the failure is now graceful and explainable instead of a silent flash. The contact still cannot be shared until the sending app grants URI access (GrapheneOS/platform_packages_apps_Contacts#25). Related: #12324, #6520.

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/sharing/v2/ShareRepository.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/sharing/v2/ShareRepository.kt
@@ -30,10 +30,9 @@ class ShareRepository(context: Context) {
 
   @NonNull
   @WorkerThread
-  @Throws(IOException::class)
   private fun resolve(multiShareExternal: UnresolvedShareData.ExternalSingleShare): ResolvedShareData {
     if (!multiShareExternal.isInternalShare && !UriUtil.isValidExternalUri(appContext, multiShareExternal.uri)) {
-      return ResolvedShareData.Failure
+      return ResolvedShareData.Failure(ShareError.UNKNOWN)
     }
 
     val uri = multiShareExternal.uri
@@ -45,8 +44,11 @@ class ShareRepository(context: Context) {
       appContext.contentResolver.openInputStream(uri)
     } catch (e: SecurityException) {
       Log.w(TAG, "Failed to read stream!", e)
-      null
-    } ?: return ResolvedShareData.Failure
+      return ResolvedShareData.Failure(ShareError.ACCESS_DENIED)
+    } catch (e: IOException) {
+      Log.w(TAG, "Failed to read stream!", e)
+      return ResolvedShareData.Failure(ShareError.UNKNOWN)
+    } ?: return ResolvedShareData.Failure(ShareError.UNKNOWN)
 
     val blobUri: Uri = try {
       AppDependencies.blobs
@@ -56,7 +58,7 @@ class ShareRepository(context: Context) {
         .createForSingleSessionOnDisk(appContext)
     } catch (e: IOException) {
       Log.e(TAG, "Failed to get blob uri")
-      return ResolvedShareData.Failure
+      return ResolvedShareData.Failure(ShareError.UNKNOWN)
     }
 
     return ResolvedShareData.ExternalUri(
@@ -77,14 +79,19 @@ class ShareRepository(context: Context) {
       }
 
     if (mimeTypes.isEmpty()) {
-      return ResolvedShareData.Failure
+      return ResolvedShareData.Failure(ShareError.UNKNOWN)
     }
 
+    var accessBlocked = false
     val media: List<Media> = mimeTypes.toList()
       .take(RemoteConfig.maxAttachmentCount)
       .map { (uri, mimeType) ->
         val stream: InputStream = try {
           appContext.contentResolver.openInputStream(uri)
+        } catch (e: SecurityException) {
+          Log.w(TAG, "Failed to open: $uri", e)
+          accessBlocked = true
+          null
         } catch (e: IOException) {
           Log.w(TAG, "Failed to open: $uri")
           null
@@ -122,8 +129,10 @@ class ShareRepository(context: Context) {
 
     return if (media.isNotEmpty()) {
       ResolvedShareData.Media(media, externalMultiShare.text)
+    } else if (accessBlocked) {
+      ResolvedShareData.Failure(ShareError.ACCESS_DENIED)
     } else {
-      ResolvedShareData.Failure
+      ResolvedShareData.Failure(ShareError.UNKNOWN)
     }
   }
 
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/thoughtcrime/securesms/sharing/v2/ShareRepository.kt
```diff
@@ -69,87 +71,134 @@ class ShareRepository(context: Context) {
...
-          isVideoGif = false,
-          bucketId = Media.ALL_MEDIA_BUCKET_ID,
-          caption = null,
-          transformProperties = null,
-          fileName = null
-        )
-      }.filterNotNull()
+    // Same as above: if URI access was denied for any item, fail the whole share so the user
+    // learns why instead of silently sharing a subset of their selection.
+    if (mediaResults.any { it is MediaResult.AccessDenied }) {
```
tuncoglu, line 100: Now fails the whole multi-share with ACCESS_DENIED if *any* selected item is denied, even when other items resolve fine. Previously `accessBlocked` was only consulted when the media list was empty, which silently shared a subset — the same silent-loss problem this PR exists to fix.

Non-permission failures (`IOException`) keep their per-item skip semantics, and an all-skipped multi-share still surfaces UNKNOWN. The check intentionally runs before `take(maxAttachmentCount)`: if a URI the user selected is denied, we surface the error rather than treating the denial as part of truncation. Happy to scope it to the first N items if preferred.

app/src/main/java/org/thoughtcrime/securesms/sharing/v2/ShareRepository.kt
```diff
@@ -69,87 +71,134 @@ class ShareRepository(context: Context) {
...
+    val mimeTypeResolutions: List<Pair<Uri, MimeTypeResolution>> = externalMultiShare.uris
       .filter { externalMultiShare.isInternalShare || UriUtil.isValidExternalUri(appContext, it) }
-      .associateWith { uri -> getMimeType(appContext, uri, null) }
-      .filterValues {
-        MediaUtil.isImageType(it) || MediaUtil.isVideoType(it)
+      .map { uri -> uri to getMimeType(appContext, uri, null) }
+
+    // If any URI's provider denied the metadata query, the share cannot be completed: fail loudly
+    // rather than silently dropping the user's selection.
+    if (mimeTypeResolutions.any { it.second.accessDenied }) {
```
tuncoglu, line 80: Denied mime lookups are now classified as ACCESS_DENIED rather than UNKNOWN. Previously `getMimeType()` swallowed the SecurityException and returned UNKNOWN, so a multi-share where every provider rejected `getType()` collapsed to `Failure(UNKNOWN)` — the permission toast never appeared. Single-share is unchanged: it prefers the intent's mime type as a fallback and only classifies at `openInputStream()`.

app/src/main/java/org/thoughtcrime/securesms/sharing/v2/ShareViewModel.kt
```diff
@@ -75,9 +79,25 @@ class ShareViewModel(
...
-    store.update { it.copy(loadState = ShareState.ShareDataLoadState.Failed) }
+  private fun moveToFailedState(error: ShareError) {
+    store.update { it.copy(loadState = ShareState.ShareDataLoadState.Failed(error)) }
+  }
+
+  /**
+   * Whether the current failure has already been surfaced to the user. Prevents the share error
+   * toast from being shown again when the activity is recreated while the state is still failed.
+   */
+  fun hasShareErrorBeenShown(): Boolean = shareErrorShown
```
tuncoglu, line 90: `RxStore.stateFlowable` replays the latest state on subscription, so an activity recreated while the failure state is current (e.g. rotation) would re-show the toast. The flag lives on the ViewModel because that survives configuration changes — the activity instance does not.

tuncoglu:
Hi! The `Android CI` and `APK Diff` runs for this PR have been waiting for approval since Aug 31 (the runs created when the PR opened expired unapproved after 30 days). Could a maintainer approve the pending runs when someone has a moment? Happy to rebase on latest `main` or address anything CI surfaces. Thanks!

tuncoglu:
Quick update: I've rebased this branch onto the latest `main` (8.28.3). The diff is unchanged, but fresh `Android CI` and `APK Diff` runs were created for the new head `dc36ce94` and are awaiting approval — please approve those rather than the earlier pending runs, which point at the pre-rebase code. Thanks!

PR #14969, opened 2026-08-29
Fix message notification vibration toggle bug

Andrea-Bruno:
This change fixes a bug in the notification settings screen. Before this fix, toggling the vibration for message notifications would mistakenly toggle the entire notification system. I have updated the code to ensure that changing the vibration setting only affects vibration as expected.


mtang-signal (approved):
Thanks! This will go out in the next release.

Edit: Actually, you will need to sign the Contributor License Agreement first. Let me know here when that's done.

PR #14971, opened 2026-08-30
Fix stories sorting by type instead of by time

### LINKED ISSUE #14965: Stories are no longer ordered by their time of posting
### Guidelines

### Bug description

Ever since a few updates ago, seeing your own Signal stories showed that Signal stories were no longer ordered by the time they were posted, but instead, it seemed to follow an order of media first and then text. More recently, this is now affecting Signal stories as a whole.

### Screenshots

_No response_

### Device

moto g84

### Android version

Android 16 QPR2 (CalyxOS 7)

### Signal version

8.24.2

### Link to debug log

https://debuglogs.org/android/8.24.2/42aefc059f50157f6d760268220e97ec5be82eec1d797976e243dbd67f92d968

theabhishekchandra:

### Contributor checklist
  * Vivo V2036. Not reproduced manually through the UI (posting a story sends to real contacts), but verified on-device: the ordering cases run as instrumentation tests against real Android SQLite and fail pre-fix. Full `database` instrumentation package: 117 tests, 0 failures. Plus 7 unit tests, each confirmed failing before the fix.

----------

### Description

`getOutgoingStoriesTo`, `getAllOutgoingStories`, `getAllStoriesFor`, and `getUnreadStories` in `MessageTable.kt` query via `message_story_type_index` (an index on `story_type` alone) with no explicit `ORDER BY`. Without one, SQLite returns rows in the index's own key order, so media stories (`StoryType` codes 1-2) always come back before text stories (codes 3-4), regardless of when they were actually posted. This causes a story with mixed text/media posts to display and open out of chronological order (see #14965).

This is a regression from e54f3f501a ("Fix improper index usage on story queries"), which added the forced index to these four functions as a performance fix but didn't add a compensating `ORDER BY` the way other functions in this file already do (e.g. `getArchiveScreenStoriesPage`, `getOldestStorySendTimestamp`). Before that commit, these queries had no index constraint at all and fell back to insertion order, which happened to be chronological.

The fix adds an explicit `ORDER BY date_sent` to each of the four functions. `getAllOutgoingStories` takes it in both directions — `ASC` normally, `DESC` when `reverse = true`, since an empty `orderBy` would fall through to `ORDER BY _id DESC` (insertion order), the same class of bug. Its only reverse caller, `MyStoriesRepository`, wants newest-first and now gets it by send time. The `reverse` arg to `queryMessages` is dropped there as dead once `orderBy` is set.

Added 7 regression tests covering all four functions, both directions, and the `limit` interaction. Each is arranged so insertion order and send order disagree, and each fails against the pre-fix code.

Fixes #14965

### DIFF
--- a/app/src/main/java/org/thoughtcrime/securesms/database/MessageTable.kt
+++ b/app/src/main/java/org/thoughtcrime/securesms/database/MessageTable.kt
@@ -1523,12 +1523,13 @@ open class MessageTable(context: Context?, databaseHelper: SignalDatabase) : Dat
       whereArgs = buildArgs(threadId)
     }
 
-    return MmsReader(queryMessages(where, whereArgs, index = INDEX_STORY_TYPE))
+    return MmsReader(queryMessages(where, whereArgs, orderBy = "$DATE_SENT ASC", index = INDEX_STORY_TYPE))
   }
 
   fun getAllOutgoingStories(reverse: Boolean, limit: Int): Reader {
     val where = "$IS_STORY_CLAUSE AND ($outgoingTypeClause)"
-    return MmsReader(queryMessages(where, null, reverse, limit.toLong(), index = INDEX_STORY_TYPE))
+    val orderBy = if (reverse) "$DATE_SENT DESC" else "$DATE_SENT ASC"
+    return MmsReader(queryMessages(where, null, limit = limit.toLong(), orderBy = orderBy, index = INDEX_STORY_TYPE))
   }
 
   fun markAllIncomingStoriesRead(): List<MarkedMessageInfo> {
@@ -1553,7 +1554,7 @@ open class MessageTable(context: Context?, databaseHelper: SignalDatabase) : Dat
     val threadId = threads.getThreadIdIfExistsFor(recipientId)
     val where = "$IS_STORY_CLAUSE AND $THREAD_ID = ?"
     val whereArgs = buildArgs(threadId)
-    val cursor = queryMessages(where, whereArgs, false, limit.toLong(), index = INDEX_STORY_TYPE)
+    val cursor = queryMessages(where, whereArgs, limit = limit.toLong(), orderBy = "$DATE_SENT ASC", index = INDEX_STORY_TYPE)
     return MmsReader(cursor)
   }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/thoughtcrime/securesms/database/MessageTable.kt
```diff
@@ -1523,12 +1523,13 @@ open class MessageTable(context: Context?, databaseHelper: SignalDatabase) : Dat
...
     }
 
-    return MmsReader(queryMessages(where, whereArgs, index = INDEX_STORY_TYPE))
+    return MmsReader(queryMessages(where, whereArgs, orderBy = "$DATE_SENT ASC", index = INDEX_STORY_TYPE))
   }
 
   fun getAllOutgoingStories(reverse: Boolean, limit: Int): Reader {
     val where = "$IS_STORY_CLAUSE AND ($outgoingTypeClause)"
-    return MmsReader(queryMessages(where, null, reverse, limit.toLong(), index = INDEX_STORY_TYPE))
+    val orderBy = if (reverse) "" else "$DATE_SENT ASC"
```
mtang-signal: if reverse is true, we should sort by DATE_SENT DESC

app/src/test/java/org/thoughtcrime/securesms/database/MessageTableTest_stories.kt
mtang-signal, line 1: I think a lot of comments here are unnecessary, especially when the test name should explain it.

app/src/main/java/org/thoughtcrime/securesms/database/MessageTable.kt
theabhishekchandra: Agreed. An empty `orderBy` isn't unordered — it falls through to `ORDER BY _id DESC` in `queryMessages`, i.e. insertion order, which is the same bug this PR fixes. It only looked right because `_id` and `date_sent` usually agree.

Now `if (reverse) "$DATE_SENT DESC" else "$DATE_SENT ASC"`. Also dropped the `reverse` arg passed to `queryMessages` — it only fed that branch, so it's dead once `orderBy` is set. Added tests for the `limit` interaction too; all ordering tests fail against the pre-fix code.

app/src/test/java/org/thoughtcrime/securesms/database/MessageTableTest_stories.kt
theabhishekchandra: Done — removed, leaving just the `// GIVEN` / `// WHEN` / `// THEN` markers used elsewhere in the file.

app/src/main/java/org/thoughtcrime/securesms/database/MessageTable.kt
theabhishekchandra: One related note while you're in here: `StoryViewerPageRepository` calls `getAllOutgoingStories(false, 100)`, so with `ASC` + `LIMIT` it now takes the oldest 100 rather than an arbitrary 100. That follows from the `ORDER BY date_sent ASC` in the original commit rather than this round, and matches what `getAllStoriesFor(recipientId, 100)` already does — flagging in case you'd rather that call site take the newest 100.

mtang-signal (approved):
thanks, this will go in the next release

